PR #12, opened 2024-02-03
Fix for macOS in index.adoc

uhooi:
The download path for macOS was incorrect and has been corrected.


bioball:
Thanks for this! This was rolled into #21

PR #25, opened 2024-02-04
.idea folder removed

paolodenti:
The `.idea` folder has been pushed, probably by mistake.
It is also explicitly excluded in the gitignore here https://github.com/apple/pkl/blob/main/.gitignore#L10

This PR removes the `.idea` folder


bioball:
Thanks, but it's intentional. You can see the subdirectories that are intentionally kept here: https://github.com/apple/pkl/blob/main/.gitignore#L11-L15

PR #29, opened 2024-02-04
Fix macOS download URL mismatch

TinkoLiu:
The download URL mismatched with the architecture for macOS builds.


bioball:
Good catch!

PR #40, opened 2024-02-04
fix spell error in language-reference page

shufanhao:
thanks for opensource `pkl`, that's awesome.  It is very useful, especially to generate repeating k8s yaml files. It's time to get ride of helm. 

after went through https://pkl-lang.org/main/current/language-reference/index.html, found few spell errors. This PR to fix these errors.


codebendr-product-designer (approved):

shenson:
I've approved https://github.com/apple/pkl/pull/40/, which should cover this. Thanks for the submission though.

PR #54, opened 2024-02-05
Use the `KOTLIN_OFFICIAL` code style defaults for Kotlin

KushalP:
JetBrains both ships the language and an editor for Kotlin. We should conform to their house style where it makes sense to us.

### DIFF
--- a/pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/JavaCodeGenerator.kt
+++ b/pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/JavaCodeGenerator.kt
@@ -40,894 +40,931 @@ import org.pkl.core.util.CodeGeneratorUtils
 class JavaCodeGeneratorException(message: String) : RuntimeException(message)
 
 data class JavaCodegenOptions(
-  /** The characters to use for indenting generated Java code. */
-  val indent: String = "  ",
+    /** The characters to use for indenting generated Java code. */
+    val indent: String = "  ",
 
-  /**
-   * Whether to generate public getter methods and protected final fields instead of public final
-   * fields.
-   */
-  val generateGetters: Boolean = false,
+    /**
+     * Whether to generate public getter methods and protected final fields instead of public final
+     * fields.
+     */
+    val generateGetters: Boolean = false,
 
-  /** Whether to generate Javadoc based on doc comments for Pkl modules, classes, and properties. */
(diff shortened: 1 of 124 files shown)


holzensp (changes requested):
We could do this, but it requires that `spotless` is setup to use the same style (see [here](https://github.com/apple/pkl/blob/main/buildSrc/src/main/kotlin/pklKotlinLibrary.gradle.kts#L24))

KushalP:
Thanks. Updated that line to use `.kotlinlangStyle()`.

KushalP:
I've separated the configuration changes from the `spotlessApply` changes.

KushalP:
It will be easiest to review this PR with whitespace hidden:

Screenshot 2024-02-05 at 13 13 23

KushalP:
And the original (IntelliJ) configuration change came from the following editor suggestion:

Screenshot 2024-02-05 at 12 38 21

KushalP:
I'm going to close this for now as it's less important than other things. Should anyone come across this, it's possible to ignore the pop-up from IntelliJ and use the Pkl project's configuration.

pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/JavaCodeGenerator.kt
```diff
@@ -40,894 +40,931 @@ import org.pkl.core.util.CodeGeneratorUtils
...
+                                propertyType = propertyType.baseType
+                            }
+                        }
+                        propertyType is PType.Class && propertyType.pClass == pClass
+                    }
+                if (modulePropertiesWithMatchingType.size == 1) {
+                    // exactly one module property has this type -> make it available for direct
+                    // injection
+                    // (potential improvement: make type available for direct injection if it occurs
+                    // exactly
```
holzensp, line 542: That's... disappointing. I would have hoped spotless could line-wrap better [throughout]

holzensp (commented):
Leaving it open for other maintainers to chime in.

I don't mind it and here-and-there prefer the Kotlin Language layout. That said, it's a high volume PR. I understand the inclination to default into the suggestion from the IntelliJ pop-up, but it might be too much churn for what it delivers (I cancel those pop-ups by default).

The vast majority of changes in this PR seem to be cause by having a narrower line-width. If that's tweakable, friction might drop significantly.

PR #65, opened 2024-02-06
Fix typo "maintian" to "maintain"

Auspicus:


bioball:
There's some commits that don't belong here. Try `git rebase -i main`, and pick just the last commit and force-push.

holzensp:
This was actually covered by #44. Thanks for the help, though.

PR #70, opened 2024-02-06
Fix broken search on pkldoc

bioball:
Fixes two issues:

1. The search-index.js script doesn't get imported in the docsite
2. The regexes aren't accessible to the function, thanks to hoisting

### DIFF
--- a/pkl-doc/src/main/resources/org/pkl/doc/scripts/search-worker.js
+++ b/pkl-doc/src/main/resources/org/pkl/doc/scripts/search-worker.js
@@ -14,6 +14,7 @@ if (isWorker) {
   const searchIndexUrl = workerName === "main" ?
       '../search-index.js' :
       '../' + workerName + '/search-index.js';
+  importScripts(searchIndexUrl);
   initSearchIndex();
   addEventListener('message', e => {
     const {query, packageName, moduleName, className} = e.data;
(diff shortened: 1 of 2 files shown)


pkl-doc/src/main/resources/org/pkl/doc/scripts/search-worker.js
```diff
@@ -177,12 +178,11 @@ function toWordStarts(characters) {
   return result;
 }
 
-const regexIsUppercase = /\p{Lu}/u
-
-const regexIsNumericCharacter = /\p{N}/u
```
bioball, line 182: Declaring these outside of `toCharClass` was causing problems because `function toCharClass` gets hoisted to the top of the file, while these consts do not. So, the consts aren't actually available to the body of `function toCharClass`.

JS is silly sometimes..

stackoverflow (approved):
LGTM

pkl-doc/src/main/resources/org/pkl/doc/scripts/search-worker.js
holzensp: *Which* JS are you calling silly? :p

holzensp (approved):
Expediting shipping, because <3 search

Issue #61, opened 2024-02-05
Crash help text has mangled URL


Edward-Knight:
The help text output when a crash happens has a URL to help submit a bug report, however this URL is mangled due to a missing newline:

```
$ echo "a { b = a }" | pkl eval - 2>&1 | head
An unexpected error has occurred. Would you mind filing a bug report?
Cmd+Double-click the link below to open an issue.
Please copy and paste the entire error output into the issue's description, provided you can share it.

https://github.com/apple/pkl/issues/newjava.lang.StackOverflowError

–– Pkl Error ––
Stack overflow

Pkl 0.25.1 (macOS 14.1, native)
```

The URL is https://github.com/apple/pkl/issues/newjava.lang.StackOverflowError ~~instead of https://github.com/apple/pkl/issues/new~~.

~~This seems to be caused by a missing newline at the end of this paragraph:~~

https://github.com/apple/pkl/blob/6eb3d20b472f4ae53595edf13b3499520c075b76/pkl-core/src/main/java/org/pkl/core/runtime/VmExceptionRenderer.java#L57-L62

---

I can open a separate issue for the stack overflow if needed - seems sometimes this is caught, e.g.:
```
$ echo "output { value = output }" | pkl eval - 2>&1
–– Pkl Error ––
A stack overflow occurred.

┌─ 4001 repetitions of:
│ 106 | text = renderer.renderDocument(value)
│              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
│ at pkl.base#Module.output.text (https://github.com/apple/pkl/blob/0.25.1/stdlib/base.pkl#L106)
└─
```

Edward-Knight:
Or actually, looking at the code further, is the URL missing a `?body=`? So it should be https://github.com/apple/pkl/issues/new?body=java.lang.StackOverflowError?

holzensp:
Fixed by PR #73

Edward-Knight:
Also should I open a separate issue about the stack overflow or is the current behaviour expected?

holzensp:
The stack overflow is expected, although arguably it's missing a stack frame (cc @stackoverflow and not just for the ironically appropriate username). In Pkl, `a = a` is the simplest way to cause a stack overflow. That's essentially what you're doing here, because in `output { value = output }`, the inner `output` "ties a loop" with the outer `output`.

Re-opening to follow up about the query parts of the URL (which may have been rolled into the URL encoding fix)

PR #73, opened 2024-02-06
fix mangled url

shubhgng:
Crash help text has mangled URL #61

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmExceptionRenderer.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmExceptionRenderer.java
@@ -59,7 +59,7 @@ public class VmExceptionRenderer {
         .append("Cmd+Double-click the link below to open an issue.\n")
         .append(
             "Please copy and paste the entire error output into the issue's description, provided you can share it.\n\n")
-        .append("https://github.com/apple/pkl/issues/new");
+        .append("https://github.com/apple/pkl/issues/new\n\n");
 
     builder.append(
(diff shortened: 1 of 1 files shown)


pkl-core/src/main/java/org/pkl/core/runtime/VmExceptionRenderer.java
```diff
@@ -59,7 +59,8 @@ private void renderBugException(VmBugException exception, StringBuilder builder)
         .append("Cmd+Double-click the link below to open an issue.\n")
         .append(
             "Please copy and paste the entire error output into the issue's description, provided you can share it.\n\n")
-        .append("https://github.com/apple/pkl/issues/new");
+        .append("https://github.com/apple/pkl/issues/new")
+        .append("\n\n");
```
holzensp, line 63: Yeah, except maybe keep it brief
```suggestion
        .append("https://github.com/apple/pkl/issues/new\n\n");
```

holzensp (approved):
Awaiting a CI config snag, but LGTM (see nit)

pkl-core/src/main/java/org/pkl/core/runtime/VmExceptionRenderer.java
shubhgng: @holzensp sure I have made the changes.

stackoverflow:
Some tests are broken and should be fixed before we can merge this.

holzensp:
Shooting from the hip, but I'd imagine _all_ of these tests have become flakey since the Gradle bump. We should kick this off again after that's reverted.

shubhgng:
Much appreciated, thanks

PR #79, opened 2024-02-06
Create SECURITY.md

narmidm:
I think it would be a good practice to have one, as it can help us protect our main email account from spam, phishing, and other threats. It can also help us manage our security notifications and alerts more easily.


madrob:
Thank you for your contribution. We already have a security policy at https://github.com/apple/pkl/blob/main/SECURITY.adoc - please follow the instructions there to submit security reports.

PR #81, opened 2024-02-06
Fix documentation homepage

bioball:
This adds a trailing slash that was missing from the documentation homepage.


stackoverflow (approved):
Yeah, I was thinking this was the culprit but didn't get as far as testing it.

PR #82, opened 2024-02-07
fix concat operator and string typo

lanzafame:
Based on the Language Reference concatenation example, `To concatenate strings, use the + (plus) operator, as in "abc" + "def" + "ghi".`. I double checked the lang ref for a `+=` operator and it doesn't appear to be defined.


holzensp (approved):
You are correct; this was a typo. Thanks for the fix.

Issue #55, opened 2024-02-05
Improve development setup instructions


translatenix:
I'm trying to build Pkl on Ubuntu 22.04/WSL 2.0.14. (I'd prefer to build directly in Windows but ran into https://github.com/apple/pkl/issues/20 and failing tests.) After reading through `DEVELOPMENT.adoc`, I have the following questions:

What does "Setup Gradle on your system" entail? (`./gradlew build` seems to work fine without further setup.)

How do I make `./gradlew buildNative` work? Currently, task `:pkl-cli:alpineExecutableAmd64` fails with:
```
/usr/bin/ld: cannot find -lz: No such file or directory
```

Further up I see:
```
[7/7] Creating image...                                                                                  (0.0s @ 1.86GB)
Fatal error: java.lang.RuntimeException: There was an error linking the native image: Linker command exited with 1

Based on the linker command output, possible reasons for this include:
1. It appears as though libz:.a is missing. Please install it.
```

Installing package `zlib1g-dev`, which contains `/usr/lib/x86_64-linux-gnu/libz.a`, didn't solve the problem.

The good news is that after tinkering with the build scripts to exclude the Alpine build and test tasks, `./gradlew build buildNative` succeeded. So I think I'm pretty close...

bioball:
The Alpine executable depends on musl. You can see how we set it up in our CI here: https://github.com/apple/pkl/blob/main/.circleci/jobs/BuildNativeJob.pkl#L78-L90

But, if you just want to build it locally, you should be able to just do `./gradlew linuxExecutableAarch64` or `./gradlew linuxExecutableAmd64` depending on your architecture. But yeah, changing the build script to exclude the Alpine stuff works too!

For Gradle: I suppose that also implies setting up the JDK, but evidently enough you've already done that.

translatenix:
Thanks for the pointers! I found more info here: https://www.graalvm.org/latest/reference-manual/native-image/guides/build-static-executables/#prerequisites-and-preparation

I think most (WSL) Linux users won't bother to set up a musl toolchain. But we'd still like to run the native build as advertised (`./gradlew buildNative`). How about only running the Alpine tasks if a musl toolchain (i.e., `~/staticdeps/bin/x86_64-linux-musl-gcc`) is present? If this sounds good, I'll send a pull request.

> For Gradle: I suppose that also implies setting up the JDK, but evidently enough you've already done that.

[...]

PR #83, opened 2024-02-07
Only build static executable if musl toolchain is installed

translatenix:
This improves the development experience for (WSL) Linux users. They can now run "./gradlew buildNative" without having a musl toolchain installed. In this case, only the dynamically linked executable will be built.

### DIFF
--- a/buildSrc/src/main/kotlin/BuildInfo.kt
+++ b/buildSrc/src/main/kotlin/BuildInfo.kt
@@ -97,7 +97,7 @@ open class BuildInfo(project: Project) {
 
   val hasMuslToolchain: Boolean by lazy {
     // see "install musl" in .circleci/jobs/BuildNativeJob.pkl
-    File(System.getProperty("user.home"), "staticdeps/bin/musl-gcc").exists()
+    File(System.getProperty("user.home"), "staticdeps/bin/x86_64-linux-musl-gcc").exists()
   }
 
   val os: org.gradle.internal.os.OperatingSystem by lazy {
--- a/buildSrc/src/main/kotlin/BuildInfo.kt
+++ b/buildSrc/src/main/kotlin/BuildInfo.kt
@@ -95,6 +95,11 @@ open class BuildInfo(project: Project) {
     java.lang.Boolean.getBoolean("releaseBuild")
   }
 
+  val hasMuslToolchain: Boolean by lazy {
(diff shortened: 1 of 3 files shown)


holzensp:
Help me understand WSL a little better (I haven't used it for years and back then not for real); why would WSL users use the `Alpine` builds? I guess this comes with two sub-questions:
* Does `gw linuxExecutableAmd64` work on WSL?
* If so, shouldn't the fix be to improve platform detection (`buildNative` should end up building `linuxExecutableAmd64`)?

translatenix:
This PR achieves exactly what you're proposing: If no musl toolchain is installed, `buildNative` should only build `linuxExecutableAmd64` and should skip `alpineExecutableAmd64`. The problem isn't specific to WSL but applies to Linux in general. Prior discussion: https://github.com/apple/pkl/issues/55#issuecomment-1929075962

buildSrc/src/main/kotlin/BuildInfo.kt
```diff
@@ -95,6 +95,11 @@ open class BuildInfo(project: Project) {
     java.lang.Boolean.getBoolean("releaseBuild")
   }
 
+  val hasMuslToolchain: Boolean by lazy {
+    // see "install musl" in .circleci/jobs/BuildNativeJob.pkl
+    File(System.getProperty("user.home"), "staticdeps/bin/musl-gcc").exists()
```
bioball, line 100: Quick suggestion: this is slightly less brittle because this is ultimately what our `native-image` build looks for

```suggestion
    File(System.getProperty("user.home"), "staticdeps/bin/x86_64-linux-musl-gcc").exists()
```

bioball (approved):
Thank you!

buildSrc/src/main/kotlin/BuildInfo.kt
```diff
@@ -97,7 +97,7 @@ open class BuildInfo(project: Project) {
 
   val hasMuslToolchain: Boolean by lazy {
     // see "install musl" in .circleci/jobs/BuildNativeJob.pkl
-    File(System.getProperty("user.home"), "staticdeps/bin/musl-gcc").exists()
+    File(System.getProperty("user.home"), "staticdeps/bin/x86_64-linux-musl-gcc").exists()
```
translatenix, line 100: FYI I used “staticdeps/bin/musl-gcc” because according to a comment in BuildNativeJob.pkl, that’s the path expected by GraalVM.

PR #84, opened 2024-02-07
update 'Loading Modules' example to match config contents

igmrrf:
The example config content:
bird {
  species = "Pigeon"
  diet = "Seeds"
}
parrot {
  species = "Parrot"
  diet = "Berries"
}

But in the 'Loading Modules' example, 'bird.name' was used which in most cases will return an error for those following;
–– Pkl Error ––
Cannot find property name in object of type Dynamic.

1 | bird.name
    ^^^^^^^^^
at  (repl:pkl1)

Available properties:
default
diet
species


bioball:
Thank you!

PR #91, opened 2024-02-07
Update Hidden Properties documentation example

maiph:
These examples on the documentation where not clear and also didn't have valid syntax.


bioball:
Thank you!

PR #96, opened 2024-02-08
Fixes for pkldoc

bioball:
* Add URI encoding to paths and fragments
* Render quotes around identifiers when appropriate


pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
```diff
@@ -365,13 +365,13 @@ public static URI relativize(URI uri, URI base) {
...
-    var basePath = Path.of(base.getRawPath());
+    var basePath = Path.of(base.getPath());
     if (!base.getRawPath().endsWith("/")) basePath = basePath.getParent();
-    var resultPath = basePath.relativize(Path.of(uri.getRawPath()));
+    var resultPath = basePath.relativize(Path.of(uri.getPath()));
 
     try {
       return new URI(
-          null, null, null, -1, resultPath.toString(), uri.getRawQuery(), uri.getRawFragment());
+          null, null, null, -1, resultPath.toString(), uri.getQuery(), uri.getFragment());
```
bioball, line 374: This was incorrect; using the raw API means we end up double encoding things. So, for example, a path value `/foo bar` gets turned into URI `/foo%2520bar`, when it should actually be `/foo%20bar`

pkl-core/src/test/kotlin/org/pkl/core/runtime/DefaultModuleResolverTest.kt
```diff
@@ -163,7 +163,7 @@
 //            Arrays.asList("file:///pkl/.*", "https://.*\\.apple\\.com"));
 //
 //    resolver.resolve("file:///pkl/foo.pkl", fileUrlModule, sourceSection);
-//    resolver.resolve("https://pkl.io/foo.pkl", fileUrlModule, sourceSection);
+//    resolver.resolve("https://example.com/foo.pkl", fileUrlModule, sourceSection);
```
bioball, line 166: Unrelated: change all `pkl.io` to `example.com`.

pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
holzensp: Good catch

pkl-doc/src/main/kotlin/org/pkl/doc/DocPackageInfo.kt
```diff
@@ -159,7 +159,14 @@ data class DocPackageInfo(
   internal fun getModuleImportUri(moduleName: String): URI =
     when (importUri) {
       "pkl:/" -> "pkl:${moduleName.substring(4)}".toUri()
-      else -> "$importUri${getModulePath(moduleName, moduleNamePrefix)}.pkl".toUri()
+      else -> {
+        val path =
+          getModulePath(moduleName, moduleNamePrefix)
+            .split("/")
+            .map { it.uriEncoded }
+            .joinToString("/") { it } + ".pkl"
```
holzensp, line 167: `DocPackageInfo` only restricts `moduleNamePrefix` to `!contains("/")`. What are the odds someone will cheat a query component or fragment in there? Wondering also whether uses might spread and whether this logic of avoiding `/`s should live in what is now `uriEncoded` (although that might need renaming)

pkl-doc/src/test/files/DocGeneratorTest/output/com.package1/1.2.3/baseModule/BaseClass.html
```diff
@@ -39,8 +39,8 @@ <h1 id="declaration-title">BaseClass<span id="declaration-version">1.2.3</span><
         <h2 class="member-group-title">Methods<span class="toggle-inherited-members">(<span class="toggle-inherited-members-link button-link">show inherited</span>)</span></h2>
         <ul>
           <li>
-            <div id="getClass()" class="anchor"> </div>
-            <div class="member inherited expandable hidden collapsed"><a class="member-selflink material-icons" href="#getClass()">link</a>
+            <div id="getClass%28%29" class="anchor"> </div>
```
holzensp, line 42: `id`s are not URIs. Except when the attribute is itself a URI/URL should it be escaped as such. In other cases, this should be XML escaped.

pkl-doc/src/main/kotlin/org/pkl/doc/DocPackageInfo.kt
bioball: It should be okay if they put other stuff in there; it will all get URI encoded in our generated file. A doc package whose name is `foo?bar=baz` is treated as path segment `foo%3Fbar%3Dbaz`

pkl-doc/src/test/files/DocGeneratorTest/output/com.package1/1.2.3/baseModule/BaseClass.html
bioball: Good catch!

bioball: Actually, I just tested this, and the browser doesn't treat XML escapes the same way.

`https://example.com/foo#%3E` won't navigate to an anchor with ID `"&gt;"`, but it _will_ navigate to an anchor of ID `"%3E".

PR #102, opened 2024-02-08
logical and fix in mathUtils

shubhgng:
minor fix in the mathUtils.java file

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/util/MathUtils.java
+++ b/pkl-core/src/main/java/org/pkl/core/util/MathUtils.java
@@ -63,7 +63,7 @@ public final strictfp class MathUtils {
    * Long.bitCount(Long.MIN_VALUE) == 1}, but {@link Long#MIN_VALUE} is not a power of two.
    */
   public static boolean isPowerOfTwo(long x) {
-    return x > 0 & (x & (x - 1)) == 0;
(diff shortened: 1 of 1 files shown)


pkl-core/src/main/java/org/pkl/core/util/MathUtils.java
```diff
@@ -63,7 +63,7 @@ public static boolean isMathematicalInteger(double x) {
    * Long.bitCount(Long.MIN_VALUE) == 1}, but {@link Long#MIN_VALUE} is not a power of two.
    */
   public static boolean isPowerOfTwo(long x) {
-    return x > 0 & (x & (x - 1)) == 0;
+    return x > 0 && (x & (x - 1)) == 0;
```
bioball, line 66: This actually has the same logic; bitwise AND and logical AND have the same meaning for booleans.

But, it probably makes sense to change this for style reasons. Also, bitwise AND will not short-circuit so that's an added bonus.

bioball (approved):

shubhgng:
thanks.

pkl-core/src/main/java/org/pkl/core/util/MathUtils.java
shubhgng: yes

PR #120, opened 2024-02-11
Minor documentation fixes

hoxell:
At docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc:35, I guess it makes at least some sense to keep it as is (it was prettier before my change), but I believe the proposed change makes it ever so slightly more accurate, making it a tiny bit easier for first-time readers to understand the nomenclature.

(Absolutely love this tool btw! 😍)

### DIFF
--- a/docs/modules/language-reference/pages/index.adoc
+++ b/docs/modules/language-reference/pages/index.adoc
@@ -685,7 +685,7 @@ dodo {
 As you probably guessed, the nested property `class` can be accessed with `dodo.taxonomy.class`.
 
 Like all values, objects are _immutable_, which is just a fancy (and short!) way to say that their properties never change.
-So what happens when Pigeon moves to a different street? Do we have to construct a new object from scratch?
+So what happens when dodo moves to a different street? Do we have to construct a new object from scratch?
 
 [[amending-objects]]
 === Amending Objects
--- a/docs/modules/language-reference/pages/index.adoc
+++ b/docs/modules/language-reference/pages/index.adoc
@@ -685,7 +685,7 @@ dodo {
 As you probably guessed, the nested property `class` can be accessed with `dodo.taxonomy.class`.
 
 Like all values, objects are _immutable_, which is just a fancy (and short!) way to say that their properties never change.
-So what happens when dodo moves to a different street? Do we have to construct a new object from scratch?
+So what happens when Dodo moves to a different street? Do we have to construct a new object from scratch?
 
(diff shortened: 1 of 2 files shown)


docs/modules/language-reference/pages/index.adoc
```diff
@@ -685,7 +685,7 @@ dodo {
 As you probably guessed, the nested property `class` can be accessed with `dodo.taxonomy.class`.
 
 Like all values, objects are _immutable_, which is just a fancy (and short!) way to say that their properties never change.
-So what happens when Pigeon moves to a different street? Do we have to construct a new object from scratch?
+So what happens when dodo moves to a different street? Do we have to construct a new object from scratch?
```
holzensp, line 688:
```suggestion
So what happens when Dodo moves to a different street? Do we have to construct a new object from scratch?
```

(if referring to the _property_ it should be between backticks, but this previously spoke casually about Pigeon, referring to the given `name`).

docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
```diff
@@ -32,8 +32,8 @@ parrot = (bird) {
 }
 ----
 
-Parrot and Pigeon have nearly identical properties.
-They only differ in their name and taxonomy, so if you have already written out `bird`, you can say that `parrot` is just like `pigeon` except `name` is `"Parrot"`, diet is `"Berries"` the `taxonomy.order` is `"Psittaciformes"`.
+`parrot` and `bird` have nearly identical properties.
```
holzensp, line 35: Same as above (although not critical); this refers to the named birds, not their describing properties.
```suggestion
Parrot and Pigeon have nearly identical properties.
```

docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
```diff
@@ -32,8 +32,8 @@ parrot = (bird) {
...
+They only differ in their name and taxonomy, so if you have already written out `bird`, you can say that `parrot` is just like `bird` except `name` is `"Parrot"`, diet is `"Berries"` the `taxonomy.order` is `"Psittaciformes"`.
```
holzensp, line 36: Really, the `bird` property should probably be renamed to `pigeon`.

holzensp (approved):
Thanks for these fixes! I'll roll in my suggested changes to move it forward.

PR #121, opened 2024-02-11
Introduce Typed objects before referring to them

hoxell:
Currently, typed objects are only briefly mentioned in the sentence stating that (typed) objects cannot be amended with new properties, only existing ones. This is unnecessarily confusing since all examples up until that point have been concerned with dynamic objects. Since the tutorial goes on to amend a dynamic object with a new property in the next section, a reader who isn't yet aware there are two types of objects might be confused by what seems to be a contradiction.

### DIFF
--- a/docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
+++ b/docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
@@ -58,11 +58,11 @@ parrot {
 }
 ----
 
-[IMPORTANT]
+[NOTE]
 ====
-_Amending_ does not allow us to _add_ properties to the (typed) object we are amending.
+So far, we have only worked with link:{uri-stdlib-Dynamic}[Dynamic] objects. The other type available to us is link:{uri-stdlib-Typed}[Typed] objects. _Amending_ allows us to override, amend, and add new properties to a dynamic object, but a typed object will only let us amend or override existing properties, not add entirely new ones.
 The xref:03_writing_a_template.adoc[next part of the tutorial] discusses types in more detail.
-There, you see that amending _never changes the type_ of the object.
+There, you see that amending _never changes the type_ of a typed object.
 ====
 
 You can also amend nested objects.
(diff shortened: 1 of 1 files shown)


docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
```diff
@@ -58,11 +58,11 @@ parrot {
 }
 ----
 
-[IMPORTANT]
+[NOTE]
 ====
-_Amending_ does not allow us to _add_ properties to the (typed) object we are amending.
+So far, we have only worked with link:{uri-stdlib-Dynamic}[Dynamic] objects. The other type available to us is link:{uri-stdlib-Typed}[Typed] objects. _Amending_ allows us to override, amend, and add new properties to a dynamic object, but a typed object will only let us amend or override existing properties, not add entirely new ones.
```
holzensp, line 63: One code style note: We keep to one sentence per line (gives somewhat nicer diffing over time).

For presentation purposes, we prefer the _second_ person (it's more activating).

Although I agree with this fix, I feel some wordsmithing is called for, especially because `Listing` and `Mapping` live at the same level as `Dynamic` and `Typed`. Leaving it to you to accept/alter:
```suggestion
So far, you have only worked with link:{uri-stdlib-Dynamic}[Dynamic] objects.
Pkl also offers link:{uri-stdlib-Typed}[Typed] objects.
_Amending_ allows you to override, amend, and add new properties to a dynamic object.
Typed objects will only let you amend or override existing properties, not add entirely new ones.
```

docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
```diff
@@ -58,11 +58,11 @@ parrot {
...
 The xref:03_writing_a_template.adoc[next part of the tutorial] discusses types in more detail.
-There, you see that amending _never changes the type_ of the object.
+There, you see that amending _never changes the type_ of a typed object.
```
holzensp, line 65: This one, I'd revert, because `Dynamic` is also a type that doesn't change by amending.
```suggestion
There, you see that amending _never changes the type_ of the object.
```

holzensp (changes requested):
I like it and the call-out about confusion is entirely correct. I wanted to suggest a _little_ tweaking, though.

docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
hoxell: Ah, got it! I somehow was under the impression that there was some type associated with a `Dynamic` object, too, based on its properties, elements, and entries, but it seems to be more akin to the Lua table data type then.

docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
hoxell: Thanks, that's an improvement; applied 🙏🏻

holzensp (approved):

holzensp:
The failing tests are all `org.pkl.cli.CliDownloadPackageCommandTest` and known to be flakey. Merging past the failures.

PR #131, opened 2024-02-12
Don't run JUnit tests for non-existing native executables

translatenix:
This allows to "Run all Tests" in IntelliJ without getting lots of test errors because "pkl-alpine-linux-amd64" etc. doesn't exist.

### DIFF
--- a/pkl-core/src/test/kotlin/org/pkl/core/LanguageSnippetTestsEngine.kt
+++ b/pkl-core/src/test/kotlin/org/pkl/core/LanguageSnippetTestsEngine.kt
@@ -1,5 +1,9 @@
 package org.pkl.core
 
+import org.junit.platform.engine.EngineDiscoveryRequest
+import org.junit.platform.engine.TestDescriptor
+import org.junit.platform.engine.UniqueId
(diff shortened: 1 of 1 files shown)


pkl-core/src/test/kotlin/org/pkl/core/LanguageSnippetTestsEngine.kt
```diff
@@ -137,6 +142,15 @@ abstract class AbstractNativeLanguageSnippetTestsEngine : AbstractLanguageSnippe
     Regex(".*/import1b\\.pkl"),
   )
 
+  override fun discover(discoveryRequest: EngineDiscoveryRequest, uniqueId: UniqueId): TestDescriptor {
```
holzensp, line 145: Warrants a comment.
```suggestion
  /**
   * Avoid running tests for native binaries when those native binaries have not been built.
   */
  override fun discover(discoveryRequest: EngineDiscoveryRequest, uniqueId: UniqueId): TestDescriptor {
```

holzensp (approved):
Yes please ;)

PR #138, opened 2024-02-13
Addressing documentation discrepancies from pkl examples.

grantabbott:


docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
```diff
@@ -372,11 +372,11 @@ pipelines {
 <1> There is no pipeline object to amend. The `new` keyword gives you an object to amend.
 
 So far, you've defined objects the same way you amended them.
-When the name `foo` didn't occur before, `foo { ... }` _creates_ a property called `foo` and assigns to it the object specified on the `...`.
-If `foo` is an existing object, this notation is an _amend expression_; resulting in a new _object_ (value), but _not_ a new (named) property.
+When the `name` didn't occur before, `new { ... }` _creates_ a property called `name` and assigns to it the object specified on the `Listing`.
```
holzensp, line 375: This is incorrect. The point that is being made here, is that - without an `amends` clause - writing
```
foo {
  bar = "baz"
  qux = 42
}
```
_creates_ the property `foo`. `new` is the operator used to let Pkl auto-resolve the default to amend _in an expression position_. Similarly, the `...` at the end of that sentence refers to `foo { ... }` and `foo` is actually a `Dynamic` if it was defined by the first-use scenario described here.

That said, this sentence can certainly use a improvement.
```suggestion
If the name `foo` didn't occur before, `foo { ... }` _creates_ a property called `foo` and assigns to it the contents in place of the `...` (given that `...` is not valid Pkl, but an abbreviation for illustration purposes).
```

Is this a clearer formulation? Feel free to have another stab at it.

docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
```diff
@@ -372,11 +372,11 @@ pipelines {
...
+If `name` is an existing object, this notation is an _amend expression_; resulting in a new _object_ (value), but _not_ a new (named) property.
```
holzensp, line 376:
```suggestion
If `foo` is an existing object, this notation is an _amend expression_; resulting in a new _object_ (value), but _not_ a new (named) property.
```

^^ same

docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
```diff
@@ -372,11 +372,11 @@ pipelines {
...
 Since `pipelines` is a listing, you can _add_ elements by writing expressions in an amend expression.
 
-In this case, though, there is no object to amend. Writing `myNewPipeline { ... }` defines a _property_, but listings may only include _elements_.
+In this case, though, there is no object to amend. Writing `pipelines { ... }` defines a _property_, but listings may only include _elements_.
```
holzensp, line 379: The point being made here is about what happens _inside_ `pipelines`. At the top-level, where we said `pipelines { ... }`, there _is_ an object to amend (namely, the brand new one, just created, as per the above).

holzensp (changes requested):
The first two fixes are entirely correct. The rest needs a little tweaking.

docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
grantabbott: We can keep the current verbiage if that is desired.

holzensp (approved):

docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
holzensp:
```suggestion
In this case, though, there is no object to amend. Writing `myNewPipeline { ... }` defines a _property_, but listings may only include _elements_.
```

PR #140, opened 2024-02-13
Issue Template Forms

moritz-t-w:
There will likely be a lot of similar issues submitted to this repo, which is why I think it's important to organize them.
For the issue forms to work, the following labels need to exist:
- [x] enhancement
- [ ] bindings


moritz-t-w (commented):
I am having trouble running this... when manually executing the script it works but it doesn't seem to get run by git...? Maybe I made a mistake in the script or I need to configure something... According to the [git docs](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) the hooks are supposed to be in `.git/hooks`
(that's the reason I've force pushed twice :clown_face:)

.github/ISSUE_TEMPLATE/language_bindings_request.pkl
```diff
@@ -0,0 +1,69 @@
...
+    id = "additional-info"
+    attributes {
+      label = "Additional Information (Optional)"
+      description = "Any other information you'd like to add"
+    }
+    validations {
+      required = false
+    }
+  }
+}
```
holzensp, line 69: As a suggestion to make this more discoverable for anyone who comes at this from a GitHub angle; it's always good to mark generated files as generated. This suggestion uses `reflect.Module`, which requires that you `import "pkl:reflect"` at the top.

```suggestion
}

output {
  text = """
    # This file is GENERATED and should not be edited manually.  Edit \(reflect.Module(module).name).pkl instead.
    \(super.text)
    """
}
```

.github/ISSUE_TEMPLATE/language_bindings_request.pkl
```diff
@@ -0,0 +1,69 @@
+name = "Language Bindings Request"
```
holzensp, line 1: It's fine to postpone, because you might be able to iterate faster without, but adding a template for this (and start this with `amends "path/to/template/maybe/even/package/on/pkl-pantry"`) would give you the full Pkl experience ;)

You can also start by just putting type annotations in _this_ file.

holzensp (commented):
I _love_ what you're doing here. Don't let me slow you down, but I thought I'd drop a comment.

.github/ISSUE_TEMPLATE/language_bindings_request.pkl
moritz-t-w: I was thinking about doing something like that but then I decided to just write the `yml` assisted by the existing template and basing the `pkl` on that, since it's not gonna be changed a lot and I didn't wanna bury myself in a rabbithole. 
I'm not gonna go that deep atm but I bet as this gets more popular someone is gonna make a template that can then be used here.
Really seeing the use-cases of this language tho :D

moritz-t-w:
Is there a plan for proposals of changes to the syntax or should I also just make a template for that?

moritz-t-w:
I just saw that I didn't adhere to the commit message guidelines. I'm gonna edit some commit messages.

.github/ISSUE_TEMPLATE/language_bindings_request.pkl
moritz-t-w: I will write a template tho, where should I put that? I don't think it makes sense to put it in the ISSUE_TEMPLATE folder or one might confuse it with an actual issue template that didn't get rendered by accident. What do you think?

holzensp: In a very general sense; you can publish packages anywhere you like (that's just FYI for whatever else you might be looking to create templates for). All repositories maintained by the core Pkl team should really only be use packages in those repositories. This could be an excellent contribution to [pkl-pantry](https://github.com/apple/pkl-pantry). It's worth having a think about naming. I'm hesitant to suggest `com.github` as a package name, because that would suggest an authoritative source, but maybe `com.github.contrib`? (cc @bioball).

moritz-t-w: I was gonna write just a template for this specific usecase, not a completely generalized one. This would just include the "GENERATED" message and shared attributes between similar templates. Obviously that's not really necessary, but I wanted to try out some of the features of pkl, without writing a complete implementation of github's issue form standard.

holzensp: That’s entirely fair. In that case, I’d keep it local and remove the `amends` before commit or - probably better - just do that type work inline (I.e. not a template, but proper validation)

moritz-t-w: > I’d keep it local and remove the `amends` before commit

@holzensp what did you mean by this?

moritz-t-w:
I was gonna do bug report and feature request forms but then I thought, it should probably be separated between cli, standard library, etc.? Can someone give me a list of templates that I should make?
Currently there's:
- Language Bindings Request - Request for Pkl bindings for a programming language
- Output Format Request - Request for a output format to be added to Pkl

.github/ISSUE_TEMPLATE/language_bindings_request.pkl
moritz-t-w: > "this isn't a general implementation of the GitHub Issues API, it's just something used here."

Exactly. If there is enough demand, someone is gonna make a full suite of GitHub scoped templates, I'll leave that to them and use a simple template for this use case.

moritz-t-w:
> Is there a plan for proposals of changes to the syntax or should I also just make a template for that?

@holzensp

.github/ISSUE_TEMPLATE/language_bindings_request.pkl
moritz-t-w: So my question remains, where should I put the template file?

moritz-t-w:
@holzensp
https://github.com/apple/pkl/pull/140#issuecomment-1945795260

PR #144, opened 2024-02-13
Fix name resolution in `typealias` with constraint

KushalP:
The body of a `typealias` gets inlined into wherever the `typealias` is used. This fixes a bug where the name resolution would resolve an additional type constraint on the next scope of `this`.

Lack of a frame scope
---------------------

`typealias` doesn't currently have a frame scope, which is why we have to assign the `owner` and the `receiver` here. If it did, we could read both of these at the point where we're trying to resolve the `typealias` in question.


bioball (changes requested):
This looks pretty good! Left some comments but overall it seems like this fixes the bug.

KushalP:
Thanks. I have amended and pushed.

Issue #137, opened 2024-02-13
[feat] Render to INI-style formats


moritz-t-w:
I know your bandwidth is limited, just putting it out there. I would love to see support for INI-like formats, such as systemd configuration files or even just plain ini files.

holzensp:
It's a totally legit ask. For other output formats, especially "simpler" ones, we recommend implementing them in Pkl itself. We did this with the TOML renderer, for example: https://github.com/apple/pkl-pantry/tree/main/packages/pkl.toml

Since bandwidth is indeed limited, you might want to have a go at it that way yourself; we're more than happy to support / review / etc. and since this can be written in Pkl itself, it doesn't require expertise about Pkl internals.

Madmegsox1:
Hi, ill have a look at starting to create a `INIRenderer`.

moritz-t-w:
> since this can be written in Pkl itself, it doesn't require expertise about Pkl internals.

I love that
Unfortunately my bandwith is also limited :skull: but I might get into making some other renderers at a later point, sounds like a fun challenge

Madmegsox1:
Its looking good so far the ini renderer corretly renders ini files as in the correct structure just need to handle escape code and correctly formatting the ini file. Here is a look at what the test outputted.

```ini
int = 123
float = 1.23
bool = true
string = Pigeon
unicodeString = abc😀abc😎abc
multiLineString = have a
great
day
[map]
one = 123
two = 1.23
three = true
four = Pigeon
five = abc😀abc😎abc
six = have a
great
day
[map.seven]
name = Pigeon
[mapping]
one = 123
two = 1.23
three = true
four = Pigeon
five = abc😀abc😎abc
six = have a
great
day
[mapping.seven]
name = Pigeon
[typedObject]
name = Pigeon
age = 30
[typedObject.address]
street = Folsom St.
[container]
name = Pigeon
age = 30
[container.address]
street = Folsom St.
```

Madmegsox1:
Its now outputting correctly and handles escape chars. Is there anywhere else i need to add the `IniRenderer` to?

bioball:
Nice! Do you want to PR this? That will make it a little easier to iterate.

To add a new in-language renderer, you will need to create a new standard library module. To do that:

1. Create the stdlib module itself; `stdlib/ini.pkl`
[...]

PR #149, opened 2024-02-14
[Addition] Added IniRenderer and IniRenderer test cases #137

Madmegsox1:
Added ini renderer and ini renderer test cases. 
Changes that still need to be added are as follows:
- a new in-language renderer 
- a new standard library module, as mentioned in [this issue](https://github.com/apple/pkl/issues/137#issuecomment-1942946804).

Ill create another pr later with these changes.

In `IniUtils`, the list of charecters that need to be escaped within the .ini file are in accourdance with [INI file Wiki](https://en.wikipedia.org/wiki/INI_file#Escape_characters).

The `IniRenderer` is very similar to [ PropertiesRenderer](https://github.com/apple/pkl/blob/main/pkl-core/src/main/java/org/pkl/core/PropertiesRenderer.java) due to .properties file having a similar style to an .ini file

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/IniRenderer.java
+++ b/pkl-core/src/main/java/org/pkl/core/IniRenderer.java
@@ -1,0 +1,247 @@
+/**
+ * Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.core;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.io.Writer;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
+import org.pkl.core.util.Nullable;
+import org.pkl.core.util.ini.IniUtils;
+
+// To instantiate this class, use ValueRenderers.properties().
+final class IniRenderer implements ValueRenderer {
+
+  private final Writer writer;
+  private final boolean omitNullProperties;
+  private final boolean restrictCharset;
+
+  public IniRenderer(Writer writer, boolean omitNullProperties, boolean restrictCharset) {
+    this.writer = writer;
+    this.omitNullProperties = omitNullProperties;
+    this.restrictCharset = restrictCharset;
+  }
+
+  @Override
+  public void renderDocument(Object value) {
+    new Visitor().renderDocument(value);
+  }
+
+  @Override
+  public void renderValue(Object value) {
+    new Visitor().renderValue(value);
+  }
+
+  public class Visitor implements ValueConverter<String> {
+
+    public void renderValue(Object value) {
+      write(convert(value), false, restrictCharset);
+    }
+
+    public void renderDocument(Object value) {
+      if (value instanceof Composite) {
+        doVisitMap(null, ((Composite) value).getProperties());
+      } else if (value instanceof Map) {
+        doVisitMap(null, (Map<?, ?>) value);
+      } else if (value instanceof Pair) {
+        Pair<?, ?> pair = (Pair<?, ?>) value;
+        doVisitKeyAndValue(null, pair.getFirst(), pair.getSecond());
+      } else {
+        throw new RendererException(
+            String.format(
+                "The top-level value of a Java properties file must have type `Composite`, `Map`, or `Pair`, but got type `%s`.",
+                value.getClass().getTypeName()));
+      }
+    }
+
+    private void doVisitMap(@Nullable String keyPrefix, Map<?, ?> map) {
+      for (Map.Entry<?, ?> entry : map.entrySet()) {
+        doVisitKeyAndValue(keyPrefix, entry.getKey(), entry.getValue());
+      }
+    }
+
+    private void doVisitKeyAndValue(@Nullable String keyPrefix, Object key, Object value) {
+      if (omitNullProperties && value instanceof PNull) {
+        return;
+      }
+      var baseKey = convert(key);
+      // gets existing key and appends the existing keypreifx if it's not null
+      var keyString = keyPrefix == null ? baseKey : keyPrefix + "." + baseKey;
+      // discovered a new section and writes key then writes the child values
+      // e.g. [example.dog]
+      if (value instanceof Composite) {
+        writeIniKey(keyString);
+        doVisitMap(keyString, ((Composite) value).getProperties());
+      } else if (value instanceof Map) {
+        writeIniKey(keyString);
+        doVisitMap(keyString, (Map<?, ?>) value);
+      } else {
+        writeIniValue(baseKey, convert(value));
+      }
+    }
+
+    @Override
+    public String convertNull() {
+      return "";
+    }
+
+    @Override
+    public String convertString(String value) {
+      return value;
+    }
+
+    @Override
+    public String convertBoolean(Boolean value) {
+      return value.toString();
+    }
+
+    @Override
+    public String convertInt(Long value) {
+      return value.toString();
+    }
+
+    @Override
+    public String convertFloat(Double value) {
+      return value.toString();
+    }
+
+    @Override
+    public String convertDuration(Duration value) {
+      throw new RendererException(
+          String.format(
+              "Values of type `Duration` cannot be rendered as Properties. Value: %s", value));
+    }
+
+    @Override
+    public String convertDataSize(DataSize value) {
+      throw new RendererException(
+          String.format(
+              "Values of type `DateSize` cannot be rendered as Properties. Value: %s", value));
+    }
+
+    @Override
+    public String convertPair(Pair<?, ?> value) {
+      throw new RendererException(
+          String.format(
+              "Values of type `Pair` cannot be rendered as Properties. Value: %s", value));
+    }
+
+    @Override
+    public String convertList(List<?> value) {
+      throw new RendererException(
+          String.format(
+              "Values of type `List` cannot be rendered as Properties. Value: %s", value));
+    }
+
+    @Override
+    public String convertSet(Set<?> value) {
+      throw new RendererException(
+          String.format("Values of type `Set` cannot be rendered as Properties. Value: %s", value));
+    }
+
+    @Override
+    public String convertMap(Map<?, ?> value) {
+      throw new RendererException(
+          String.format("Values of type `Map` cannot be rendered as Properties. Value: %s", value));
+    }
+
+    @Override
+    public String convertObject(PObject value) {
+      throw new RendererException(
+          String.format(
+              "Values of type `Object` cannot be rendered as Properties. Value: %s", value));
+    }
+
+    @Override
+    public String convertModule(PModule value) {
+      throw new RendererException(
+          String.format(
+              "Values of type `Module` cannot be rendered as Properties. Value: %s", value));
+    }
+
+    @Override
+    public String convertClass(PClass value) {
+      throw new RendererException(
+          String.format(
+              "Values of type `Class` cannot be rendered as Properties. Value: %s",
+              value.getSimpleName()));
+    }
+
+    @Override
+    public String convertTypeAlias(TypeAlias value) {
+      throw new RendererException(
+          String.format(
+              "Values of type `TypeAlias` cannot be rendered as Properties. Value: %s",
+              value.getSimpleName()));
+    }
+
+    @Override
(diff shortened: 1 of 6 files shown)


ghyatzo:
great that this has already been made, any specific reason why lists\sets are not supported?
something like this:
```
[section]
mylist=[a, list, of, things]
```
I have encountered them plenty of time in the wild. Also, sets could be rendered exactly the same, since the uniqueness of all items in a set should be checked in the pkl side.

Madmegsox1:
One way of doing this for the time being is as follows. I will add support list/sets natively im just currently working on the ini std-lib:
```
output {
   renderer = new IniRenderer {
     converters {
       // render lists as comma-separated values
       [List] = (it) -> it.join(",")
    }
  }
}
```

bioball:
Thanks for the PR! Please also add `ini.pkl` to this PR too.

The Java-side renderer is actually the less important renderer; those renderers are meant for Java users that want to take an already evaluated Java representation of a Pkl value. We only offer Java-side renderers for some formats.

The more important renderer is the in-language renderer, via `ini.pkl`, and I don't think we want the Java-side in without the in-language renderer also being there.

Madmegsox1:
Ive been working on the in-language renderer over the past couple days. I have written the .pkl std lib. I have also added an `IniModule` class and the `RenderNodes` class. 

I just need to write the `LanguageSnippetTest`

bioball:
The converters mechanism is a way for users to override the intrinsic behavior of the renderer. The standard library should be defining the intrinsic behavior, so, it's not appropriate to specify converters within `stdlib/ini.pkl`.

For handling arrays: seems like there's no agreed upon format? Some implementations use [commas](https://cloanto.com/specs/ini/#:~:text=An%20INI%20file%20is%20an,contains%20zero%20or%20more%20values.&text=Section%20names%20are%20enclosed%20in,key%20names%20are%20case%2Dinsensitive), and others use [square brackets](https://github.com/npm/ini), and others don't support them at all as far as I can tell.

Maybe one way we can handle this is to add a renderer option for some commonly used styles.

```
class Renderer {
  arrayStyle: ("comma"|"bracket")?
}
```

Where `null` means: throw if any `List` or `Listing` are found.

Madmegsox1:
Thank you for your help. Test cases should be done today and i love the idea of adding the array style config to ini.pkl!

holzensp:
All of what @bioball said and just to make the point explicit; the out-of-the-box-default behaviour of a renderer should always be restricted to target format standard representations. If there is no such standard, the default behaviour must be to error out. This preserves the principle of "least surprise." (This is the reason why, for example `JsonRenderer` doesn't know how to deal with `Duration`s, even though there are perfectly reasonable default _choices_.)

pkl-core/src/test/resources/org/pkl/core/iniRendererTest.pkl
```diff
@@ -0,0 +1,66 @@
+class Person {
+  name: String
+  age: Int
+  address: Address
+  friend: Person?
+}
```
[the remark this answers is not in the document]
Madmegsox1: Okay, Thank you. Ill have a look and update these tests with some more up to date ones

Madmegsox1:
Thank you for confirming. Very helpful!

Madmegsox1:
Pushed changes for code review purposes. I know i still need to add some more in-lang renderer tests 👍

pkl-core/src/main/java/org/pkl/core/stdlib/ini/RendererNodes.java
```diff
@@ -0,0 +1,409 @@
...
+              visit(memberValue);
+              currPath.pop();
+            }
+            return true;
+          }));
+    }
+
+    // ignored for now
+    @Override
+    protected void endListing(VmListing value, boolean isEmpty) {}
```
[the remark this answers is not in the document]
Madmegsox1: Okay 👍 i think your right with what you said, users can always overwrite the List converter if they want to add lists

Madmegsox1:
Having an issue with the tests that overwrite the render functuons...

So all the values get rendered out correctly but due to the way the ini renderer works using the end functuins to group up Maps, Dynamics etc, it renderes them still at the bottom of the file
```
res1 = gnirts
res2 = false
res3 = 43
res4 = 2.33
res5 = 3 s
res6 = 4 mb
res8 = [\n  43,\n  false,\n  \"gnirts\"\n]
res9 = [\n  \"gnirts\",\n  false,\n  43,\n  5\n]
res10 = {\n  \"eno\"\: false,\n  \"owt\"\: 2.33,\n  \"eerht\"\: \"3 s\"\n}
res11 = [\n  \"gnirts\",\n  false,\n  43,\n  5\n]
res12 = {\n  \"eman\"\: \"noegip\",\n  \"ega\"\: 31,\n  \"eerht\"\: \"3 s\"\n}
res13 = {\n  \"name\"\: \"noegip\",\n  \"age\"\: 31,\n  \"other\"\: \"rehto\"\n}
res14 = {\n  \"name\"\: \"noegip\",\n  \"age\"\: 41\n}
res15 = String
```

pkl-core/src/main/java/org/pkl/core/IniRenderer.java
```diff
@@ -0,0 +1,247 @@
...
+
+    public void renderDocument(Object value) {
+      if (value instanceof Composite) {
+        doVisitMap(null, ((Composite) value).getProperties());
+      } else if (value instanceof Map) {
+        doVisitMap(null, (Map<?, ?>) value);
+      } else if (value instanceof Pair) {
+        Pair<?, ?> pair = (Pair<?, ?>) value;
+        doVisitKeyAndValue(null, pair.getFirst(), pair.getSecond());
+      } else {
```
bioball, line 65: To be consistent with other renderers, let's not support rendering pairs.

```suggestion
      } else {
```

bioball (commented):
General question: is there a specification that you are following here?

I'm having a hard time finding a specification for INI files. It seems like the closest is that different platforms provide their own spec for how they parse INI, and they're all slightly different.

Here's just some of the specs that I can find through my googling:

* https://cloanto.com/specs/ini/
* https://www.nongnu.org/chmspec/latest/INI.html
* https://moz-services-docs.readthedocs.io/en/latest/server-devguide/confspec.html
* https://spec.editorconfig.org/#file-format

The implementation here makes some choices that are incompatible with some of these specs. For example, non-string values are serialized, and how `.` in a section header means nesting, and using LF for line endings.

Is there a well known implementation that is being followed here?

Maybe the correct way to approach this is something like:

```
class IniRenderer {
  flavor: "cloanto"|"somethingElse"
}
```

And using that as a way to determine how to render to INI?

Another alternative would be to break these options into their own properties.

```
class IniRenderer {
  booleans: "truefalse"|"yesno"|"none" = "none"
  lineEndings: "\r\n"|"\n" = "\r\n"
  // etc
}
```

Either of these approaches have flaws, though. The first one is not flexible and wouldn't be able to capture some new INI format, and we'd have a hard time chasing down all of the variations and subtleties if taking the second approach.

Madmegsox1:
Have you had a chance to look at the issue i am having with the tests?

ghyatzo:
It is annoying that ini files have such a loose format. With there being so many variations, to what extent can we then customize the renderer through converters?
For example, some more variations that I have encountered in the wild:
- using `:` instead of `=` for the key-value separator
- no nesting, but instead everything is flattened into a single section

Is this something that can be done through converters or should it be a renderer setting like bioball suggested above (or straight up not supprted)?

My main concern is that exactly because the format is so loose, then there is a proliferation of dialects, but we should also be able to adapt this single renderer to any dialect (that doesn't take it too far) with ease.
Luckily an ini file doesn't have that many moving parts to generalize.

Madmegsox1:
Indeed the format that i am basing this renderer off can be found [here](https://en.wikipedia.org/wiki/INI_file).
Currently im trying to fix some of the tests but need some of the maintainers help on a issue i am having the tests.
Bioballs suggestion by letting the usere choose how to rendere the file will be implemented once i have fixed these tests

Madmegsox1:
> Having an issue with the tests that overwrite the render functuons...
> 
> So all the values get rendered out correctly but due to the way the ini renderer works using the end functuins to group up Maps, Dynamics etc, it renderes them still at the bottom of the file
> 
> ```
> res1 = gnirts
> res2 = false
> res3 = 43
> res4 = 2.33
> res5 = 3 s
> res6 = 4 mb
> res8 = [\n  43,\n  false,\n  \"gnirts\"\n]
> res9 = [\n  \"gnirts\",\n  false,\n  43,\n  5\n]
> res10 = {\n  \"eno\"\: false,\n  \"owt\"\: 2.33,\n  \"eerht\"\: \"3 s\"\n}
> res11 = [\n  \"gnirts\",\n  false,\n  43,\n  5\n]
> res12 = {\n  \"eman\"\: \"noegip\",\n  \"ega\"\: 31,\n  \"eerht\"\: \"3 s\"\n}
> res13 = {\n  \"name\"\: \"noegip\",\n  \"age\"\: 31,\n  \"other\"\: \"rehto\"\n}
> res14 = {\n  \"name\"\: \"noegip\",\n  \"age\"\: 41\n}
> res15 = String
> 
> [res10]
> eno = false
> owt = 2.33
> 
> [res12]
> eman = noegip
> ega = 31
> 
> [res13]
> name = noegip
> age = 31
> 
> [res14]
> name = noegip
> age = String
> ```
> 
> **Im wondering if there is a flag somewhere that could be accessed by the end functions to tell it that the main render function for that data type has been overwritten?**

@holzensp is there functionality there for the Renderer to tell if parts of it has been overwritten by pkl?

bioball:
> Im wondering if there is a flag somewhere that could be accessed by the end functions to tell it that the main render function for that data type has been overwritten?
> @holzensp is there functionality there for the Renderer to tell if parts of it has been overwritten by pkl?

@Madmegsox1: I'm not totally understanding the question. What do you mean by "main render function for that data type has been overwritten"?

Can you share a snippet of the Pkl input, and your INI output, and your expected INI output?

Madmegsox1:
Hi has anyone been able to take a look at this?

pkl-core/src/main/java/org/pkl/core/stdlib/ini/RendererNodes.java
```diff
@@ -0,0 +1,409 @@
...
+              if (memberKey instanceof Identifier) {
+                currPath.push(memberKey);
+              } else {
+                currPath.push(converter.convert(memberKey, List.of()));
+              }
+              visit(memberValue);
+              currPath.pop();
+            }
+            return true;
+          }));
```
holzensp, line 310: Should have caught this earlier; this is where your bug is. You already print the contents of an object by visiting each property _before_ this method gets called (that provides the expected output). This is the bit that produces the unexpected extra output. The `endDynamic` / `endTyped` / `endMap` / etc methods are there for other kinds of admin, such as decreasing indent (in JSON, YAML, Pcf) and printing a closing `}` (in JSON, Pcf). You shouldn't revisit the members.

holzensp (commented):
Very sorry for letting this sit; travel and other distractions. I believe I've found your issue, though.

pkl-core/src/main/java/org/pkl/core/stdlib/ini/RendererNodes.java
Madmegsox1: Ah, Thank you. I understand where you are coming from, at the time i though it would be a good idea to use them for complex Pkl files. See the issue comees when you have a pkl file like this for example:
```pkl
mapping = new Mapping{
  ["one"] = "value1"
  ["two"] = new Person {name = "Pigeon" age = 19 address {
    street = "fun Street"
  }}
  ["three"] = "value3"
  ["four"] = 1.2
  ["five"] = new Mapping {
    ["six"] = 123
  }
  ["seven"] = true
}
```
The expected output would have the correct sections such as `mapping.one` or `mapping.two.address` and then the values under said sections. For example:

```pkl
[mapping]
one = value1
three = value3
four = 1.2
seven = true

[mapping.two]
name = Pigeon
age = 19

[mapping.two.address]
street = fun Street

[mapping.five]
six = 123
```

So the use of these "end functions" is to make sure that the values are correctly under each section. If i didnt use the end functions the rendered ini file would look like this: 
```pkl
[mapping]
one = value1

[mapping.two]
name = Pigeon
age = 19

[mapping.two.address]
street = fun Street
three = value3
four = 1.2

[mapping.five]
six = 123
seven = true
```
And as you can see some values such as `seven` and `four` are under the incorrect section.

Im not sure if you or anyone else could think of another way of overcoming this problem if so it would be great to get this solved!

Madmegsox1: @holzensp have you had time to look at my issue?

holzensp: Again, sluggish; sorry. I get it now; you're having to compensate for `.ini` not having hierarchies, so things that would be nested, now come _after_ the object is entirely rendered. You should still not _revisit the object's members_, though. I'd keep a stack of `StringWriter`s in your renderer. Every `startFoo` you push a new one with the qualified name of the property already in it. Every `endFoo` you then pop it and write the top-writers contents.

Madmegsox1: Thank you so much! Ill have a look soon

PR #151, opened 2024-02-14
Add Gradle task "testNative"

translatenix:
Support running native tests with "gw testNative" for consistency with "gw test".


holzensp (approved):
LGTM

PR #162, opened 2024-02-15
Fix typo in constructor parameter name

translatenix:
This is a backwards-incompatible change. I assume this isn't a public API because it isn't documented.

PS: `CliDownloadPackageCommandTest` is very flaky for me. I believe the reason is that `PackageServer` binds to a hardcoded port (12110).


bioball (approved):
The CLI is actually a library, and this is meant to be a public API too despite the lack of documentation.

It's okay for this to be breaking, though, because 0.x versions allow breaking changes in semver terms. But I think this is fine regardless.

PR #163, opened 2024-02-15
Fix usage of wrong lock object

translatenix:


bioball (approved):
Ah, good catch, thanks!

PR #165, opened 2024-02-15
[pkl-doc] Sort stdlib first on the main page

HT154:


stackoverflow (approved):
Having the stdlib first makes sense. Thanks!

PR #169, opened 2024-02-16
Rename some CLI classes for consistency

translatenix:
CliDownloadPackageCommand -> CliPackageDownloader (consistent with CliProjectPackager/CliProjectResolver) CliAbstractProjectCommand -> CliProjectCommand (consistent with CliCommand)

As I learned yesterday, I'm touching a public API here. Let me know if this is acceptable.


holzensp:
Yeah, I see the value of consistency in the naming, but it is indeed a breaking change. Thoughts, @bioball / @stackoverflow?

stackoverflow:
I wouldn't worry too much about the breaking here. These classes are supposed to be used in the CLI only (most of them are final). It doesn't make much sense to depend on them.

holzensp (approved):

PR #175, opened 2024-02-17
Simplify code in ResourceReaders

translatenix:
Removing the methods fixes several IntelliJ warnings. They are overridden in all subclasses.


stackoverflow (approved):
LGTM

Issue #188, opened 2024-02-18
Clear task names in Gradle plugin


sgammon:
It is currently pretty hard to locate the task names for generating code from the Pkl Gradle plugin; it would help if these task names were given a `category` and `description`, so that they show up when a Pkl user runs `./gradlew tasks`.

In my specific use case, I'm looking for the tasks I can configure to run another task reliably after Pkl's code-gen step; to remedy #187 locally, I want to amend the files and move them to a package path, but it's hard to do that even with Gradle's extensive override facilities if I don't know where plugin authors prefer me to hook in and call `finalizedBy`.

PR #190, opened 2024-02-18
Add summaries and groups to gradle tasks

sgammon:
Fixes and closes apple/pkl#188 by adding task names and groups to major Pkl Gradle Plugin tasks. I would love advice from library authors that these are the right tasks to expose to users.


pkl-gradle/src/main/java/org/pkl/gradle/PklPlugin.java
```diff
@@ -171,6 +171,8 @@ private void configureJavaCodeGenTasks(NamedDomainObjectContainer<JavaCodeGenSpe
           createModulesTask(JavaCodeGenTask.class, spec)
               .configure(
                   task -> {
+                    task.setDescription(TaskConstants.GENERATE_JAVA_DESCRIPTION);
+                    task.setGroup(TaskConstants.TASK_GROUP_CODEGEN);
```
sgammon, line 175: In general, this is how I'm configuring `description` and `group` on these tasks.

pkl-gradle/src/main/java/org/pkl/gradle/TaskConstants.java
```diff
@@ -0,0 +1,38 @@
...
+  static final String TASK_GROUP_DOCS = "Documentation";
+
+  /** Gradle task description for generating Java code from Pkl modules. */
+  static final String GENERATE_JAVA_DESCRIPTION = "Generate Java code from Pkl modules";
+
+  /** Gradle task description for generating Kotlin code from Pkl modules. */
+  static final String GENERATE_KOTLIN_DESCRIPTION = "Generate Kotlin code from Pkl modules";
+
+  /** Gradle task description for generating Pkldoc from Pkl modules. */
+  static final String GENERATE_PKLDOC_DESCRIPTION = "Generate Pkldoc from Pkl modules";
```
sgammon, line 37: Task constants that will show in `./gradlew tasks` for a project with the Pkl Gradle Plugin added.

pkl-gradle/src/main/java/org/pkl/gradle/TaskConstants.java
```diff
@@ -0,0 +1,38 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.gradle;
+
+/** Constant values used by tasks */
+class TaskConstants {
```
netvl, line 19: I believe it is more conventional to put these constants directly to the plugin class.

sgammon: What about the category settings, which must be shared between plugins? I think there may be symbols for those somewhere in Gradle's API. I'll look for appropriate ones to use.

bioball: There's [LifecycleBasePlugin](https://github.com/gradle/gradle/blob/master/platforms/software/platform-base/src/main/java/org/gradle/language/base/plugins/LifecycleBasePlugin.java) which provides the base task names and groups. 

Then there's [JvmConstants](https://github.com/gradle/gradle/blob/master/platforms/jvm/platform-jvm/src/main/java/org/gradle/api/internal/tasks/JvmConstants.java), but this seems less appropriate for us.

I think it's fine to just have these as strings, though.

bioball (changes requested):
Maybe one letter and lowercase these group names?

Also; per @netvl's comments, would be good to follow convention and put these constants within `PklPlugin`.

pkl-gradle/src/main/java/org/pkl/gradle/TaskConstants.java
netvl: There is only one plugin in the project, so there is nothing to share, or an I missing something?

Anyway, if you look at the built-in Gradle plugins, they always contain constants related to tasks configured by these plugins within the plugin class. If we ever decide to make new plugins, we can move these constants to the appropriate location at that time.

pkl-gradle/src/main/java/org/pkl/gradle/TaskConstants.java
netvl: FWIW, I think that if these strings are not reused anywhere, it might make sense to inline them to the specific task definitions within the plugin code. But I don't have a strong opinion on this.

PR #194, opened 2024-02-18
Add codegen setting governing Kotlin package

sgammon:
## Summary

Adds a setting to the Kotlin code generator which acts as a prefix for the target Kotlin package during codegen.

This has been filed on top of #192 to avoid conflicts.

## Changelog

- feat(gradle): `kotlinPackage` property in gradle plugin
- feat(codegen): use `kotlinPackage` as prefix for kotlin codegen
- fix: pin version of `kotlinx.html` and `kotlinx.serialization`
- test(codegen): add tests for custom kotlin package
- test(gradle): add tests for generating with custom kotlin package

### DIFF
--- a/pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/CliKotlinCodeGeneratorOptions.kt
+++ b/pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/CliKotlinCodeGeneratorOptions.kt
@@ -29,6 +29,9 @@ data class CliKotlinCodeGeneratorOptions(
   /** The characters to use for indenting generated source code. */
   val indent: String = "  ",
 
+  /** Kotlin package to use for generated code; if none is provided, the root package is used. */
+  val kotlinPackage: String = "",
+
   /** Whether to generate Kdoc based on doc comments for Pkl modules, classes, and properties. */
   val generateKdoc: Boolean = false,
 
@@ -39,5 +42,11 @@ data class CliKotlinCodeGeneratorOptions(
   val implementSerializable: Boolean = false
 ) {
   fun toKotlinCodegenOptions(): KotlinCodegenOptions =
-    KotlinCodegenOptions(indent, generateKdoc, generateSpringBootConfig, implementSerializable)
+    KotlinCodegenOptions(
+      indent = indent,
+      generateKdoc = generateKdoc,
+      generateSpringBootConfig = generateSpringBootConfig,
+      implementSerializable = implementSerializable,
+      kotlinPackage = kotlinPackage,
+    )
 }
--- a/pkl-codegen-kotlin/src/test/kotlin/org/pkl/codegen/kotlin/KotlinCodeGeneratorTest.kt
+++ b/pkl-codegen-kotlin/src/test/kotlin/org/pkl/codegen/kotlin/KotlinCodeGeneratorTest.kt
@@ -133,7 +133,8 @@ class KotlinCodeGeneratorTest {
       pklCode: String,
       generateKdoc: Boolean = false,
       generateSpringBootConfig: Boolean = false,
-      implementSerializable: Boolean = false
+      implementSerializable: Boolean = false,
+      kotlinPackage: String? = null,
     ): String {
 
       val module = Evaluator.preconfigured().evaluateSchema(ModuleSource.text(pklCode))
@@ -144,7 +145,8 @@ class KotlinCodeGeneratorTest {
           KotlinCodegenOptions(
             generateKdoc = generateKdoc,
             generateSpringBootConfig = generateSpringBootConfig,
-            implementSerializable = implementSerializable
+            implementSerializable = implementSerializable,
+            kotlinPackage = kotlinPackage ?: "",
           )
         )
       return generator.kotlinFile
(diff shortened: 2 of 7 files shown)


bench/gradle.lockfile
```diff
@@ -26,17 +34,17 @@ org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.7.10=testCompileClasspath,testImplemen
...
+org.junit.jupiter:junit-jupiter-params:5.10.2=testCompileClasspath,testImplementationDependenciesMetadata,testRuntimeClasspath
+org.junit.platform:junit-platform-commons:1.10.2=testCompileClasspath,testImplementationDependenciesMetadata,testRuntimeClasspath,testRuntimeOnlyDependenciesMetadata
+org.junit.platform:junit-platform-engine:1.10.2=testRuntimeClasspath,testRuntimeOnlyDependenciesMetadata
+org.junit:junit-bom:5.10.2=testCompileClasspath,testImplementationDependenciesMetadata,testRuntimeClasspath,testRuntimeOnlyDependenciesMetadata
+org.openjdk.jmh:jmh-core:1.37=jmh,jmhCompileClasspath,jmhImplementationDependenciesMetadata,jmhRuntimeClasspath
+org.openjdk.jmh:jmh-generator-asm:1.37=jmh,jmhCompileClasspath,jmhImplementationDependenciesMetadata,jmhRuntimeClasspath
+org.openjdk.jmh:jmh-generator-bytecode:1.37=jmh,jmhCompileClasspath,jmhImplementationDependenciesMetadata,jmhRuntimeClasspath
+org.openjdk.jmh:jmh-generator-reflection:1.37=jmh,jmhCompileClasspath,jmhImplementationDependenciesMetadata,jmhRuntimeClasspath
+org.opentest4j:opentest4j:1.3.0=testCompileClasspath,testImplementationDependenciesMetadata,testRuntimeClasspath,testRuntimeOnlyDependenciesMetadata
```
stackoverflow, line 47: There are way too many changes on these lockfiles. If you update a library you should only update this library here.

sgammon: Thanks for catching this, but these lockfile changes occur just by running `./gradlew updateDependencyLocks`; is it possible this drift is caused by dynamic versions in the `libs.catalog.toml`?

During regular development, how do you guys update the lockfiles normally? I must be doing something wrong. Thanks for catching this, I'll push a fix to this PR.

stackoverflow: You should use `--update-locks` to specify which libraries you want to update. See https://docs.gradle.org/current/userguide/dependency_locking.html#selectively_updating_lock_state_entries

sgammon: Thank you @stackoverflow; I had tried the `--update-locks` flag but didn't realize it took an argument. This is going to improve my workflow in a big way. I will amend this PR to clean it up.

stackoverflow: Happy to help :)

sgammon: @stackoverflow I realized I could land this change without any dependency pins/changes at all, so I have updated this not to include any lockfile changes :) thanks again.

pkl-codegen-kotlin/src/test/kotlin/org/pkl/codegen/kotlin/KotlinCodeGeneratorTest.kt
```diff
@@ -553,6 +555,92 @@ class KotlinCodeGeneratorTest {
...
+      module my.mod
+
+      class Person {
+        name: String
+        age: Int
+        hobbies: List<String>
+        friends: Map<String, Person>
+        sibling: Person?
+      }
+    """,
```
holzensp, line 572: I'm surprised spotless doesn't flag this. Maybe auto-format (there are multiple sites)

pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/CliKotlinCodeGeneratorOptions.kt
```diff
@@ -39,5 +42,11 @@ data class CliKotlinCodeGeneratorOptions(
   val implementSerializable: Boolean = false
 ) {
   fun toKotlinCodegenOptions(): KotlinCodegenOptions =
-    KotlinCodegenOptions(indent, generateKdoc, generateSpringBootConfig)
+    KotlinCodegenOptions(
+      indent = indent,
+      generateKdoc = generateKdoc,
+      generateSpringBootConfig = generateSpringBootConfig,
+      implementSerializable = implementSerializable,
```
holzensp, line 49: sheesh... good find

pkl-gradle/src/test/kotlin/org/pkl/gradle/KotlinCodeGeneratorsTest.kt
```diff
@@ -125,6 +140,7 @@ class KotlinCodeGeneratorsTest : AbstractTest() {
             sourceModules = ["mod.pkl"]
             outputDir = file("build/generated")
             settingsModule = "pkl:settings"
+            ${if (kotlinPackage != null) "kotlinPackage = \"$kotlinPackage\"" else ""}
```
holzensp, line 143: and you used the more kotlinesque way above ;)
```suggestion
            ${kotlinPackage?.let { "kotlinPackage = \"$it\"" } ?: ""}
```

pkl-codegen-kotlin/src/test/kotlin/org/pkl/codegen/kotlin/KotlinCodeGeneratorTest.kt
```diff
@@ -553,6 +555,92 @@ class KotlinCodeGeneratorTest {
...
+        """
+      module my.mod
+
+      class Person {
+        name: String
+        age: Int
+        hobbies: List<String>
+        friends: Map<String, Person>
+        sibling: Person?
+      }
```
holzensp, line 571: Very low-priority, but for other/future cases as well; we default to using bird examples. The Person instances you'll find were missed in cleanup.

holzensp (dismissed):
Works for me. Did @stackoverflow have any not-yet covered issues?

stackoverflow (dismissed):
LGTM. You need to rebase main on your branch before we can merge.

sgammon:
@stackoverflow / @holzensp Thank you for the review; I've adopted all suggestions and made all changes requested. There are no lockfile updates in this change anymore so we should be good there as well.

stackoverflow (approved):

bioball:
@sgammon: did you try setting a Pkl module name instead?

The normal way to solve this is to give your Pkl module a fully qualified name. For example:

```
module com.example.sgammon.MyModule
```

And this will generate `package com.example.sgammon` in the generated Kotlin code.

sgammon:
@bioball Yes, I'd noticed that, but this applies an unconditional prefix only for Kotlin, which might be needed in some corner cases. For example, in cases where there is no module involved, Pkl uses the root module, but JPMS forbids this.

Issue #157, opened 2024-02-14
Use java.net.http.HttpClient instead of legacy java.net.Http(s)URLConnection


translatenix:
Moving to `java.net.http.HttpClient` brings many benefits, including:

* HTTP/2 support
* optional async mode
* no need to set a system-wide default `SSLSocketFactory`

Instead of directly interfacing with `java.net.http.HttpClient`, I'd recommend to introduce a lightweight `org.pkl.core.http.HttpClient` facade. This keeps some flexibility and allows to enforce things like setting `User-Agent`.

If this sounds good, I'll work on a draft PR.

bioball:
Hi, this sounds good! Go ahead and submit a PR for this.

PR #200, opened 2024-02-19
Upgrade `gradle` version in use to 7.6

KushalP:
Command that was used to produce the upgrade diff:

```
./gradlew wrapper --gradle-version=7.6
```


holzensp:
Digging into that flakey test.

translatenix:
@holzensp All tests that use PackageServer are flaky.
Higher Gradle versions just seem better at exposing the flakiness.
The proper fix is to bind PackageServer to a dynamic port.
(Currently the port is hardcoded to 12110.)

I'd be happy to take this on. It's a natural follow-up to #157.

holzensp:
> @holzensp All tests that use PackageServer are flaky. Higher Gradle versions just seem better at exposing the flakiness. The proper fix is to bind PackageServer to a dynamic port. (Currently the port is hardcoded to 12110.)

Yes, it appears to be increased concurrency with which the tests are run. We are no playing around with assigning different ports.

> I'd be happy to take this on. It's a natural follow-up to #157.

Thanks for the offer, but work is underway as we speak, so let's avoid the duplication of effort. Press on as if these are fixed!

translatenix:
@holzensp FYI every clean build of the main branch fails with connection refused errors on my machine. I always need to run the build a second time to get past the flaky tests. In my experience, binding test servers to dynamic ports (or cutting out the network layer) is the only good solution.

I'm ready to send a PR for #157 but have been holding back to give you guys a chance to catch up with other PRs.

holzensp:
Much appreciated @translatenix. @KushalP is also working on a PR to avoid port binding conflicts when tests are run in parallel. The tests really aren't rigid enough. Your #157 will likely help with remaining fickleness of these tests. Reviewing today (promise).

PR #206, opened 2024-02-20
Delete stale files in LanguageSnippetTests/output

translatenix:
When I delete pkl-core/src/test/files/LanguageSnippetTests/output and run "gw test", most output files are regenerated, but 46 of them aren't. I assume these are stale files that should be deleted.


holzensp:
Your reasoning is sound, but your conclusion is a leap... I'm thinking this is signal some of our tests aren't being executed as expected. This requires a little more deep-diving.

translatenix:
It seems that for every stale output file there's a similarly named generated output file. For example `jsonRenderer1/2/3/6.json.pcf` (stale) vs. `jsonRenderer1/2/3/6.json` (generated).

translatenix:
Another observation is that out of 46 stale output files, 44 have the extension `.pcf`.

bioball (approved):
This is correct; these are indeed stale expected output files that were never cleaned up.

PR #210, opened 2024-02-20
Update Gradle syntax in docs

translatenix:
Replace the legacy "compile" scope with "implementation". This matches Maven Central's recommended syntax:
https://central.sonatype.com/artifact/org.pkl-lang/pkl-core


bioball:
Thanks!

PR #204, opened 2024-02-19
Upstream: Build Upgrades

sgammon:
## Summary

**Not for merge**

This is a massive PR that upgrades the Gradle build in a big way. I did this sort of for fun, over the weekend, while working on other PRs which are reasonable and not enormous.

We are intending to use and extend Pkl so we could maintain a fork, and I have no expectation that this will get merged as-is. That being said, since we've done the leg work on some of this stuff and it could be useful, I wanted to file and **offer any of the below** as a menu of changes the Pkl team could get upstreamed from us if desired.

There are a handful of test failures and checksums have changed, which I am fixing. Total broken tests number less than 10, I believe, and the breakages seem fixable.

Feedback is of course welcome but since this PR isn't designed to be merged, such feedback would be incorporated in another PR, which is smaller and reviewable.

Thank you to the Pkl team for a very cool project!

## Update: Feb 21st

New batch of changes pushed with the following:

- chore: `buildSrc` refactored to `includeBuild("build-logic")` ([here's why](https://proandroiddev.com/stop-using-gradle-buildsrc-use-composite-builds-instead-3c38ac7a2ab3))
- feat: upgrade GraalVM → latest (`23.1.2`)
- feat: support for `--strict-image-heap`
- feat: support for GraalVM PGO in native mode
- feat: preinitialization of `pkl` truffle module
- feat: support for engine sharing/context re-use
- feat: native jpms support (`module-info`)
- feat: faster native builds with `-Ob`
- feat: light plugin system for embedded customization of `pkl-core` engine
- ~~feat: use `epsilon` gc in native binary~~ _note: rolling this back bc `pkl server`_

Initial native binary benchmarks show a modest improvement:

Screenshot 2024-02-21 at 11 24 08 AM

### Download URLs to try it out

- **macOS `aarch64`**: [zip](https://dl.elide.cloud/cli/pkl/labs/pkl-gvm-23.1.2.zip), [tar.gz](https://dl.elide.cloud/cli/pkl/labs/pkl-gvm-23.1.2.tar.gz)

## Update: Split-out PRs (Feb 20th)

- #213
- #212
- #211 

## Changes enclosed

**0) Fun developer ergonomics stuff.**
- Pkl icon for IDEA project
- Pkl syntax highlighting in some test fragments using IDEA [Language Injections](https://www.jetbrains.com/help/idea/using-language-injections.html)

**1) Build upgrades.**
- Support for Gradle toolchains
- Support for Gradle's `jvm-test-suite` plugins
[...]

PR #211, opened 2024-02-20
Move tool versions to Version Catalog

sgammon:
## Summary

Add a fix to make Version Catalog symbols visible within the `buildSrc` plugins, and then move tool versions (Google Java Format and ktfmt) into the catalog.

Originally previewed as part of apple/pkl#204.

## Changelog

- fix: make version catalog accessible from `buildSrc` plugins
- chore: declare `googleFormatVersion` in version catalog
- chore: declare `ktfmt` in version catalog


holzensp (approved):
Much appreciated. CI green after manual restart (session timed out for the last run)

PR #214, opened 2024-02-20
writing a template documentation fixes

grantabbott:
On line 170 I believe the filename that is meant to be referenced is`pklTutorialPart3.pkl` as there is not a `pklTutorial.pkl` file mentioned in this section. I also made some fun changes with the template using the date that pkl was publicly released.


bioball (approved):
LGTM, thanks!

Issue #215, opened 2024-02-20
`validatePom` expects missing task


sgammon:
Currently on `main`, there is a small bug introduced by the `validatePom` step within the `pklPublishLibrary` convention plugin; this step is added to validate published library POMs, but some projects (like `:stdlib`) do not configure publishing, and so a required task is missing.

This bug was easy to miss because it only surfaces with very generic project-wide tasks, like `./gradlew tasks`.

## Error

```
* What went wrong:
Execution failed for task ':tasks'.
> Could not create task ':stdlib:validatePom'.
   > Task with name 'generatePomFileForLibraryPublication' not found in project ':stdlib'.
```

**Relevant stacktrace**
```
Caused by: org.gradle.api.UnknownTaskException: Task with name 'generatePomFileForLibraryPublication' not found in project ':stdlib'.
        at org.gradle.api.internal.tasks.DefaultTaskCollection.createNotFoundException(DefaultTaskCollection.java:102)
        at org.gradle.api.internal.tasks.DefaultTaskCollection.createNotFoundException(DefaultTaskCollection.java:46)
        at org.gradle.api.internal.DefaultNamedDomainObjectCollection.named(DefaultNamedDomainObjectCollection.java:360)
        at org.gradle.api.internal.tasks.DefaultTaskCollection.named(DefaultTaskCollection.java:112)
        at org.gradle.api.internal.tasks.DefaultTaskCollection.named(DefaultTaskCollection.java:46)
        at org.gradle.api.internal.DefaultNamedDomainObjectCollection.named(DefaultNamedDomainObjectCollection.java:375)
        at org.gradle.api.internal.tasks.DefaultTaskCollection.named(DefaultTaskCollection.java:122)
        at org.gradle.kotlin.dsl.TaskContainerExtensionsKt.named(TaskContainerExtensions.kt:213)
        at org.gradle.kotlin.dsl.TaskContainerExtensionsKt.provideDelegate(TaskContainerExtensions.kt:85)
        at PklPublishLibrary_gradle$validatePom$2.invoke(pklPublishLibrary.gradle.kts:53)
        at PklPublishLibrary_gradle$validatePom$2.invoke(pklPublishLibrary.gradle.kts:3)
```

<details>
<summary>Full stacktrace</summary>
<pre>
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':tasks'.
[...]

PR #216, opened 2024-02-20
Skip POM validation for unpublished projects

sgammon:
## Summary

This small change fixes a bug introduced by the `validatePom` task within the `pklPublishLibrary` plugin. When running project-wide tasks like `./gradle tasks`, projects which don't provide a POM fail the build, because the `generatePomFileForLibraryPublication` cannot be found, but it is expected to be present.

With this change applied, `validatePom` only runs if the task `generatePomFileForLibraryPublication` is present in a given project.

Fixes and closes apple/pkl#215


holzensp (approved):
Nice catch

PR #219, opened 2024-02-20
Fix almost-bug in PackageResolvers

translatenix:
The only call site already resolved the path against tmpDir.


bioball (approved):
Ah, great catch! This was probably the result of some refactoring gone wrong.

PR #220, opened 2024-02-21
Fix concurrency bug in PackageResolvers

translatenix:
Access to field "isClosed" must be guarded.


bioball (approved):
Good catch, thanks!

PR #221, opened 2024-02-21
Tweak User-Agent header

translatenix:
It is customary to separate elements with a semicolon.


bioball (approved):
Works with me. @holzensp thoughts?

holzensp (approved):
No issue... it's more common with, but I see plenty without. No reason not to.

PR #227, opened 2024-02-22
Bind PackageServer to ephemeral port to avoid port conflicts

translatenix:
Here is my take on how to best solve the "flaky PackageServer tests" problem.
It is ready to be merged after #217 .
See commit message for details.
If you prefer a different solution, perhaps this can serve as inspiration.


translatenix:
@bioball Rebased and ready for review.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -135,6 +136,14 @@ class BaseOptions : OptionGroup() {
       .path()
       .multiple()
 
+  // hidden option used by native tests
+  private val testPort: Int by
+    option(names = arrayOf("--test-port"), help = "Internal test option", hidden = true)
+      .single()
+      .int()
+      .default(-1)
+      .validate { it in 1024..65535 }
```
bioball, line 145: Doesn't seem very useful to validate given it's a test flag, and it's also provisioned by the OS. Do we _know_ it's going to be in this range?

```suggestion
```

pkl-core/src/main/java/org/pkl/core/http/HttpClient.java
```diff
@@ -108,6 +108,15 @@ interface Builder {
      */
     Builder addBuiltInCertificates();
 
+    /**
+     * Sets a test server's listening port.
+     *
+     * <p>If set, requests that specify port 12110 are rewritten to use the given port. This enables
+     * users of this client to connect to a test server without having to know its (ephemeral)
+     * listening port.
```
bioball, line 116:
```suggestion
     * listening port.
     *
     * <p>This is an internal option used for testing purposes only.
```

pkl-core/src/main/java/org/pkl/core/util/HttpUtils.java
```diff
@@ -47,4 +48,22 @@ public static void checkHasStatusCode200(HttpResponse<?> response) throws IOExce
...
+      return new URI(
+          uri.getScheme(),
+          uri.getUserInfo(),
+          uri.getHost(),
+          port,
+          uri.getPath(),
+          uri.getQuery(),
+          uri.getFragment());
+    } catch (URISyntaxException e) {
+      throw new AssertionError(e); // only port changed
```
bioball, line 66:
```suggestion
      throw PklBugException.unreachableCode(); // only port changed
```

pkl-executor/src/main/java/org/pkl/executor/ExecutorOptions2.java
```diff
@@ -49,6 +50,7 @@ public class ExecutorOptions2 extends ExecutorOptions {
    * @param projectDir API equivalent of the {@code --project-dir} CLI option.
    * @param certificateFiles API equivalent of the {@code --ca-certificates} CLI option
    * @param certificateUris API equivalent of the {@code --ca-certificates} CLI option
+   * @param testPort API equivalent of the {@code --test-port} CLI option
```
bioball, line 53:
```suggestion
   * @param testPort Internal option used for testing purposes.
```

pkl-executor/src/main/java/org/pkl/executor/ExecutorOptions2.java
```diff
@@ -62,7 +64,8 @@ public ExecutorOptions2(
       /* @Nullable */ Path moduleCacheDir,
       /* @Nullable */ Path projectDir,
       List<Path> certificateFiles,
-      List<URI> certificateUris) {
+      List<URI> certificateUris,
+      int testPort) {
```
bioball, line 68: Since `testPort` is an internal test-only option, we should have another constructor that doesn't have it as an option, and sets `testPort` to `-1`.

bioball (changes requested):
Overall, this is looking great! I have some minor comments.

pkl-executor/src/main/java/org/pkl/executor/ExecutorOptions2.java
translatenix: Should probably address the root cause and provide a builder for `ExecutorOptions(2)`. For example, almost nobody will want to pass certificates either.

I thought about modernizing the `pkl-executor` API to look more like `Evaluator` and language bindings. But then I realized that my time is better spent on Java language bindings, which have the potential to make `pkl-executor` unnecessary.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
translatenix: Not terribly important to have this validated, but ephemeral ports are guaranteed to be high ports (1024-65535).
https://en.wikipedia.org/wiki/Ephemeral_port#Range

pkl-core/src/main/java/org/pkl/core/http/HttpClient.java
translatenix: I didn't see a need to "hide" this option at the `HttpClient` level. Anyone who codes against `HttpClient`, rather than just passing it around, may find this option useful.

Should switch from 12210 to 0 in a follow-up PR.

pkl-executor/src/main/java/org/pkl/executor/ExecutorOptions2.java
translatenix: Removed this param from the public API.

pkl-core/src/main/java/org/pkl/core/util/HttpUtils.java
translatenix: Done. `org.pkl.core.util` depending on `org.pkl.core` seems problematic. But since `PklBugException` is already used in this class, it's not making things worse.

pkl-executor/src/main/java/org/pkl/executor/ExecutorOptions2.java
translatenix: Added a builder. See commit message for details.

translatenix:
@bioball Ready from my side. See latest commit message for details. Quite happy with the cleanup.

pkl-executor/src/main/java/org/pkl/executor/ExecutorOptions.java
```diff
@@ -15,47 +15,226 @@
...
 
   /** Returns the module cache dir that the CLI uses by default. */
   public static Path defaultModuleCacheDir() {
     return Path.of(System.getProperty("user.home"), ".pkl", "cache");
   }
 
+  public static Builder builder() {
+    return new Builder();
+  }
+
```
bioball, line 69: Good idea with the builder!

We have an existing pattern of having `unconfigured()` and `preconfigured()` builders; see `EvaluatorBuilder`, `ConfigEvaluatorBuilder`, `ValueMapperBuilder`, etc. I think let's follow that.

```suggestion
  // keep in sync with SecurityManagers.defaultAllowedModules
  /// Returns the allowed module patterns that the CLI uses by default.
  public static final List<String> defaultAllowedModules =
    List.of(
      "repl:",
      "file:",
      "jar:file:",
      "modulepath:",
      "https:",
      "pkl:",
      "package:",
      "projectpackage:"
    );

  // keep in sync with SecurityManagers.defaultAllowedResources
  /// Returns the allowed resources patterns that the CLI uses by default.
  public static final List<String> defaultAllowedResources =
    List.of(
      "prop:",
      "env:",
      "file:",
      "modulepath:",
      "package:",
      "projectpackage:",
      "https:");

  /**
   * Creates a builder that has no options set.
   */
  public static Builder unconfigured() {
    return new Builder();
  }

  /**
   * Creates a builder configured with:
   *
   * <ul>
   *   <li>{@link #defaultAllowedModules}</li>
   *   <li>{@link #defaultAllowedResources}</li>
   *   <li>System environment variables</li>
   *   <li>System properties</li>
   * </ul>
   */
  public static Builder preconfigured() {
    //noinspection unchecked,rawtypes
    return unconfigured()
      .allowedModules(defaultAllowedModules)
      .allowedResources(defaultAllowedResources)
      .environmentVariables(System.getenv())
      .externalProperties((Map) System.getProperties());
  }
```

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
bioball: Gotcha, thanks for the link. Let's still remove this, though; it adds some unnecessary brittleness to our tests.

pkl-core/src/main/java/org/pkl/core/http/HttpClient.java
bioball: It's the same with `org.pkl.commons.cli.CliBaseOptions`; both are public APIs but expose flags that are only meant for test purposes (e.g. `testMode`, `testPort` added by this PR) and we call them "internal" in the doc comments.

bioball (changes requested):
Just a couple minor comments, but this is pretty close!

translatenix:
> We have an existing pattern of having unconfigured() and preconfigured() builders; see EvaluatorBuilder, ConfigEvaluatorBuilder, ValueMapperBuilder, etc. I think let's follow that.

According to the Javadoc, `Executor` "evaluates Pkl modules in a sandbox".
Does preconfiguring with `System.getenv()` and `System.getProperties()` make sense here?
And shouldn't granting file system and network access be an explicit decision?

Apart from adding `unconfigured()/preconfigured()`,
we could add `builder().grantFileSystemAccess(rootDir).grantNetworkAccess()`.
Or we could stick to a basic builder for now.
Let me know how to proceed.

> Let's still remove this, though; it adds some unnecessary brittleness to our tests.

Removed.
Just to clarify, this validation can't possibly cause brittleness. It *prevents* brittleness by catching the mistake of picking a well-known port as test port. Operating systems would never pick a well-known port as ephemeral port.

> both are public APIs but expose flags that are only meant for test purposes (e.g. testMode, testPort added by this PR) and we call them "internal" in the doc comments.

There is a big difference: `testMode` is only useful for testing Pkl itself, whereas `testPort` is equally useful for testing third-party code that uses `CliBaseOptions` or `HttpClient`. But I won't fight over this one.

bioball:
> According to the Javadoc, Executor "evaluates Pkl modules in a sandbox".
Does preconfiguring with System.getenv() and System.getProperties() make sense here?
And shouldn't granting file system and network access be an explicit decision?

I don't see any issues with the set of options provided by `preconfigured()`. This is an opt-in behavior, and users are expected to know what `preconfigured()` means. It's actually not that different from using our evaluator in pkl-core (which is also a sandbox).

translatenix:
I've made 2 out of 3 requested changes and have squashed my commits. I'll leave any builder changes to you.

bioball:
The executor API is typically meant for services that use Pkl as a "platform" that needs to evaluate 3rd party Pkl code. For instance, if you are a service that accepts Pkl files as configuration. And, like you said, in those cases, you wouldn't want to pass your own env vars or system properties to Pkl.

The core thing that the executor provides that `ConfigEvaluator` (or just `Evaluator`) don't is the ability to use multiple different Pkl versions for evaluation. The rest of the sandboxing is already part of `Evaluator`/`ConfigEvaluator` (e.g. allowed modules, allowed resources).

Thought about this a little more--yeah, there's probably not much of a use-case for `ExecutorOptions#preconfigured()`. I'm okay merging this as-is.

bioball (approved):

PR #231, opened 2024-02-22
Replace `foo` with `pigeon` in 02_filling_out_a_template

StefMa:
Title ☝️


bioball:
Thanks! This was already addressed in https://github.com/apple/pkl/pull/138

We still need to cherry-pick those changes back into release/0.25, which I'll do now.

PR #233, opened 2024-02-22
Update 03_writing_a_template

StefMa:
While reading the docs suddenly a `pklTutorial.pkl` appears which was never defined somewhere (Line 170).

To make it a bit more clear, I changed the name of the "initial" module to just `pklTutorial` (without parts) and changed the naming everywhere too.
I also made it more exclicit that we are talking about this one 🙃


bioball:
Thanks! This was already fixed in https://github.com/apple/pkl/pull/214. But, this reminds me that we need to update our docs still, which I'll do now.

PR #235, opened 2024-02-22
Fix IntelliJ nullability errors

translatenix:
IntelliJ doesn't recognize nullability annotations in code such as "java.time.@Nullable Duration". This causes erroneous nullability errors at call sites. Fix this by replacing the affected code with code such as "@Nullable java.time.Duration". This requires allowing "@Nullable" in more places.


translatenix:
An even better solution would be to switch to https://jspecify.dev, the future of Java nullness annotations.
IntelliJ correctly handles `java.time.@Nullable Duration` when using JSpecify.

bioball:
Hm? This warns just fine for me:

Screenshot 2024-02-22 at 3 08 04 PM

translatenix:
The problem may be Kotlin related. For example, `CliCommand.kt` line 100 was giving an error.

Switching to JSpecify would be my preferred solution, but I'll take almost anything over a red squiggle. I find false positive errors very disturbing.

bioball:
I don't see an error on that line, either. What error are you seeing? Can you provide a screenshot?

translatenix:
The error in line 100 is now gone, but here's another one from the same file:

Screenshot 2024-02-22 170851

bioball:
What is your IJ version? And, do you have the same null/nonnull settings as me?

Screenshot 2024-02-22 at 7 53 54 PM

Screenshot 2024-02-22 at 7 53 37 PM

translatenix:
IntelliJ IDEA 2023.3.4 (Community Edition)
Build #IC-233.14475.28, built on February 12, 2024

Same configuration except that I don't have the io.reactivex annotations.

PR #243, opened 2024-02-23
Delete unused image

translatenix:
Not sure what this is,
but it doesn't seem to be used anywhere.


bioball:
An old artifact from times gone by. Whoops, and thanks!

bioball (approved):

PR #253, opened 2024-02-26
Add CI check to test patch file

bioball:


stackoverflow (approved):
LGTM
We should merge your fix to the patch file first.
Once we drop jdk11 we can remove this job.

stackoverflow (approved):

Issue #166, opened 2024-02-15
NullPointerException on chained module import and glob import


othompson2:
Run into an issue when I chain an `@module` import which in turn contains an import with a glob pattern. If I replace either the `@module` or glob pattern with a direct import the problem doesn't occur, so both seem to work independently just not when combined.

---
error:
```
java.lang.NullPointerException

–– Pkl Error ––
None (cause has no message)

6 | styling = import*("styles/*.pkl")
    ^^^^^^^
at section#styling (projectpackage://example.com/sections@0.0.1#/example/section.pkl)

106 | text = renderer.renderDocument(value)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
at pkl.base#Module.output.text (https://github.com/apple/pkl/blob/0.25.2/stdlib/base.pkl#L106)

Pkl 0.25.2 (macOS 14.1, native)

java.lang.NullPointerException
	at java.base@17.0.9/java.util.regex.Matcher.getTextLength(Matcher.java:1769)
	at java.base@17.0.9/java.util.regex.Matcher.reset(Matcher.java:415)
	at java.base@17.0.9/java.util.regex.Matcher.<init>(Matcher.java:252)
	at java.base@17.0.9/java.util.regex.Pattern.matcher(Pattern.java:1134)
	at org.pkl.core.util.IoUtils.isUriLike(IoUtils.java:67)
	at org.pkl.core.util.IoUtils.toUri(IoUtils.java:75)
	at org.pkl.core.util.IoUtils.createUri(IoUtils.java:81)
	at org.pkl.core.util.IoUtils.resolve(IoUtils.java:273)
	at org.pkl.core.runtime.ReaderBase.resolveUri(ReaderBase.java:74)
	at org.pkl.core.module.ModuleKeys$DependencyAwareModuleKey.resolveUri(ModuleKeys.java:559)
	at org.pkl.core.util.GlobResolver.resolveGlob(GlobResolver.java:519)
	at org.pkl.core.ast.expression.unary.ImportGlobNode.executeGeneric(ImportGlobNode.java:104)
	at org.pkl.core.ast.PklRootNode.executeBody(PklRootNode.java:41)
	at org.pkl.core.ast.MemberNode.executeBody(MemberNode.java:71)
	at org.pkl.core.ast.member.TypeCheckedPropertyNode.evalTypedObjectCached(TypeCheckedPropertyNode.java:55)
[...]

PR #255, opened 2024-02-26
Support Projects loaded from arbitrary URIs

HT154:
Currently, Pkl assumes that a PklProject file and any declared local dependencies are accessible on the filesystem (i.e. via `file:` URIs. For usage of the `pkl` CLI this is sufficient, but for usage of language bindings this presents an obstacle for evaluating Pkl code using projects from sources that are not the filesystem. One compelling use case would be consuming Pkl code using project dependencies from a git repository; using `go-git` and its `go-billy` filesystem abstraction, it should be possible to do this completely in-memory. With this change and https://github.com/apple/pkl-go/pull/21 this now works end-to-end.

Brief summary of the changes here:
* Replace usage of `Path projectDir` with `URI projectBaseUri` throughout
* Switch reading of `PklProject.deps.json` to use resource readers
* Update `pkl:Project` to accept local dependencies when the scheme and authority are the same, rather than checking for `file:` URIs specifically
* Add a test in `pkl-core` for custom-defined module and resource readers
* Fix NPE in pkl-server's handling of empty/absent `pathElements` in `List*Response`s


pkl-core/src/test/files/LanguageSnippetTests/input/projects/project1/globbing.pkl
```diff
@@ -34,4 +34,9 @@ examples {
   ["glob-read absolute package uri"] {
     read*("package://localhost:12110/birds@0.5.0#/catalog/*.pkl")
   }
+  
+//   https://github.com/apple/pkl/issues/166
+//   ["glob-import behind local project import"] {
+//     import("@project6/children.pkl")
+//   }
```
HT154, line 41: This is a test for #166. I suspected these changes might resolve the issue, but they do not. This change did seem to surface a clearer error though. It looks like when the `ImportGlobNode` is executed, `currentModule` has a `file:` URI while `importUri` has a `projectpackage:` URI. This disagreement causes `ModuleKeys.ProjectPackage.listElements` to eventually bark since it doesn't know about `file:` URIs.

pkl-core/src/main/java/org/pkl/core/runtime/ResourceManager.java
```diff
@@ -165,8 +167,38 @@ public Optional<Object> read(URI resourceUri, Node readNode) {
...
+  public boolean hasElement(URI elementUri) throws IOException, SecurityManagerException {
+    var reader = resourceReaders.get(elementUri.getScheme());
+    if (reader == null) {
+      throw new VmExceptionBuilder()
+          .evalError("noResourceReaderRegistered", elementUri.getScheme())
+          .build();
+    }
+
+    return reader.hasElement(securityManager, elementUri);
+  }
```
HT154, line 203: I am not a huge fan of needing to pipe these through `ResourceManager` like this. I'm more than open to alternate suggestions on how to handle this

Issue #62, opened 2024-02-06
Document contributing a language library. 


spotandjake:
It would be beneficial if the process of writing a library for a given language was documented and semi standardized. So library support could be moved to the broader community. Especially for more niche languages.

bioball:
Yeah, I agree, we'll provide something here.

PR #257, opened 2024-02-27
Add specification for language binding API

bioball:
This adds documentation for how the language bindings works, for those that wish to create their own bindings.

Closes https://github.com/apple/pkl/issues/62


stackoverflow (dismissed):
Looks good! With feedback from the community we can keep improving it further.

stackoverflow (approved):

PR #262, opened 2024-02-28
Add Gradle toolchain support

### LINKED ISSUE #184: Use Gradle Java toolchain to avoid Java check
When you add gradle Java toolchain to this project it literally doesn't matter which jvm version users are using. As long as it can run gradle, the toolchain will make sure that the correct jvm version is downloaded and uses to build the project.

See https://docs.gradle.org/current/userguide/toolchains.html

There is no need for the "is Java 11 check" in the settings.gradle anymore. Additionally, also CI would use that exact same jdk version.

StefMa:
fixes #184 

With that we can get rid of the java 11 check and the information around it.
Also, we can "on the fly" update the jdk version the project will build on.
We only need "a" java version installed to start Gradle.

As discussed in the ticket, I also had to update Gradle to version 7.6  which (seems) brings flaky tests.
But we can't use toolchains before that version.
So maybe this PR wil hang here until this is fixed 🙃 

Questions / Todos
* I guess this is not going to work anymore (or has to be adjusted)
👉  https://github.com/apple/pkl/blob/2499e2c4937fc228e49e94e41fdaac065c203168/.circleci/config.pkl#L98C7-L107

* I guess this is not going to work anymore (or has to be adjusted)
👉  https://github.com/apple/pkl/blob/2499e2c4937fc228e49e94e41fdaac065c203168/patches/graalVm23.patch#L25-L30

Let me know if there is more like these 🙃

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -1,5 +1,3 @@
-import org.jetbrains.kotlin.config.JvmTarget
-
 plugins {
   `kotlin-dsl`
 }
@@ -17,17 +15,6 @@ dependencies {
   implementation(files(libs.javaClass.superclass.protectionDomain.codeSource.location))
 }
 
-java {
-  sourceCompatibility = JavaVersion.VERSION_11
-  targetCompatibility = JavaVersion.VERSION_11
-}
-
 kotlin {
-  target {
-    compilations.configureEach {
-      kotlinOptions {
-        jvmTarget = "11"
-      }
-    }
-  }
+  jvmToolchain(11)
 }
--- a/buildSrc/src/main/kotlin/JvmToolchain.kt
+++ b/buildSrc/src/main/kotlin/JvmToolchain.kt
@@ -1,12 +1,11 @@
 import org.gradle.jvm.toolchain.JavaLanguageVersion
 import org.gradle.jvm.toolchain.JvmVendorSpec
-import org.gradle.api.Project
 
 private const val JDK_VERSION = 11
 
 val jvmToolchainVersion: JavaLanguageVersion
     get() {
-        val jvmVersion = System.getProperty("PKL_JVM_VERSION", JDK_VERSION.toString()).toInt()
+        val jvmVersion = System.getProperty("PKL_JVM_VERSION")?.toInt() ?: JDK_VERSION
         return JavaLanguageVersion.of(jvmVersion)
     }
 
--- a/buildSrc/src/main/kotlin/JvmToolchain.kt
+++ b/buildSrc/src/main/kotlin/JvmToolchain.kt
@@ -1,0 +1,14 @@
+import org.gradle.jvm.toolchain.JavaLanguageVersion
+import org.gradle.jvm.toolchain.JvmVendorSpec
+import org.gradle.api.Project
+
+private const val JDK_VERSION = 11
+
+val jvmToolchainVersion: JavaLanguageVersion
+    get() {
+        val jvmVersion = System.getProperty("PKL_JVM_VERSION", JDK_VERSION.toString()).toInt()
+        return JavaLanguageVersion.of(jvmVersion)
+    }
+
+val jvmToolchainVendor: JvmVendorSpec 
+    get() = JvmVendorSpec.ADOPTIUM
--- a/buildSrc/src/main/kotlin/pklKotlinLibrary.gradle.kts
+++ b/buildSrc/src/main/kotlin/pklKotlinLibrary.gradle.kts
@@ -25,8 +25,8 @@ tasks.compileKotlin {
 }
 
 kotlin.jvmToolchain {
-  languageVersion.set(JavaLanguageVersion.of(11))
-  vendor.set(JvmVendorSpec.ADOPTIUM)
+  languageVersion.set(jvmToolchainVersion)
+  vendor.set(jvmToolchainVendor)
 }
 
 spotless {
--- a/buildSrc/src/main/kotlin/pklKotlinLibrary.gradle.kts
+++ b/buildSrc/src/main/kotlin/pklKotlinLibrary.gradle.kts
@@ -24,6 +24,11 @@ tasks.compileKotlin {
   enabled = true // disabled by pklJavaLibrary
 }
 
+kotlin.jvmToolchain {
+  languageVersion.set(JavaLanguageVersion.of(11))
+  vendor.set(JvmVendorSpec.ADOPTIUM)
+}
+
 spotless {
   kotlin {
     ktfmt(libs.versions.ktfmt.get()).googleStyle()
--- a/.circleci/config.pkl
+++ b/.circleci/config.pkl
@@ -141,11 +141,11 @@ local buildNativeJobs: Mapping<String, BuildNativeJob> = new {
 
 local gradleCheckJobs: Mapping<String, GradleCheckJob> = new {
   ["gradle-check-jdk11"] {
-    javaVersion = "11.0"
+    javaVersion = "11"
     isRelease = false
   }
   ["gradle-check-jdk17"] {
-    javaVersion = "17.0"
+    javaVersion = "17"
     isRelease = false
   }
 }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -583,7 +583,7 @@ jobs:
     steps:
     - checkout
     - run:
-        command: ./gradlew --info --stacktrace check
+        command: ./gradlew --info --stacktrace check -DPKL_JVM_VERSION=11
         name: gradle check
     - run:
         command: |-
@@ -601,7 +601,7 @@ jobs:
     steps:
     - checkout
     - run:
-        command: ./gradlew --info --stacktrace check
+        command: ./gradlew --info --stacktrace check -DPKL_JVM_VERSION=17
         name: gradle check
     - run:
         command: |-
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -614,7 +614,7 @@ jobs:
     environment:
       LANG: en_US.UTF-8
     docker:
-    - image: cimg/openjdk:17.0
+    - image: cimg/openjdk:11.0
   bench:
     steps:
     - checkout
--- a/.circleci/jobs/GradleCheckJob.pkl
+++ b/.circleci/jobs/GradleCheckJob.pkl
@@ -22,7 +22,7 @@ javaVersion: "11.0"|"17.0"
 steps {
   new Config.RunStep {
     name = "gradle check"
-    command = "./gradlew \(module.gradleArgs) check"
+    command = "./gradlew \(module.gradleArgs) check -DPKL_JVM_VERSION=\(javaVersion.dropLast(2))"
   }
 }
 
--- a/.circleci/jobs/GradleCheckJob.pkl
+++ b/.circleci/jobs/GradleCheckJob.pkl
@@ -17,19 +17,19 @@ extends "GradleJob.pkl"
 
 import "package://pkg.pkl-lang.org/pkl-pantry/com.circleci.v2@1.1.0#/Config.pkl"
 
-javaVersion: "11.0"|"17.0"
+javaVersion: "11"|"17"
 
 steps {
   new Config.RunStep {
     name = "gradle check"
-    command = "./gradlew \(module.gradleArgs) check -DPKL_JVM_VERSION=\(javaVersion.dropLast(2))"
+    command = "./gradlew \(module.gradleArgs) check -DPKL_JVM_VERSION=\(javaVersion)"
   }
 }
 
 job {
   docker {
     new {
-      image = "cimg/openjdk:\(javaVersion)"
+      image = "cimg/openjdk:11.0"
     }
   }
 }
--- a/DEVELOPMENT.adoc
+++ b/DEVELOPMENT.adoc
@@ -2,11 +2,9 @@
 :uri-gng: https://gng.dsun.org
 :uri-jenv: https://www.jenv.be
 :uri-intellij: https://www.jetbrains.com/idea/download/
-:uri-jdk: https://adoptopenjdk.net/releases.html
 
 == Setup
 
-. (mandatory) Install {uri-jdk}[OpenJDK 11 HotSpot] (as long as we support JDK11)
 . (recommended) Install {uri-intellij}[IntelliJ IDEA 2023.x] +
 To import the project into IntelliJ, go to File->Open and select the project's root directory.
 If the project is opened but not imported, look for a popup in the lower right corner
--- a/buildSrc/src/main/kotlin/pklAllProjects.gradle.kts
+++ b/buildSrc/src/main/kotlin/pklAllProjects.gradle.kts
@@ -22,15 +22,8 @@ configurations {
   }
 }
 
-plugins.withType(JavaPlugin::class).configureEach {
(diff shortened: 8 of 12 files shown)


translatenix:
In my opinion, it’s important that the Java version/vendor can be changed in a single place. For example, I often use this to reproduce a failed CI build.

StefMa:
Any suggestion/recommendation how to do that?
Maybe an optional gradle property? 🤔

translatenix:
Perhaps you could add `pklJvmProject.gradle.kts` and apply it in the other files.

translatenix:
IMHO the property is unnecessary. It's much easier to edit `private const val DEFAULT_JVM_VERSION = 11`. How about renaming it to `JDK_VERSION`? I'd also add a `val` for the vendor.

translatenix:
I think `JDK_VERSION` is correct, but mainly I'd like to get rid of `DEFAULT`.

`const` is only an optimization, but a vendor `val` isn't essential.

System property would be a small improvement because setting the project property in a build script will likely not work correctly.

StefMa:
Hmm.. unfourtnalty I'm unable to update he dependency lock files 🤔 
This is my change:
```diff
index 445bfc4..bdb5c2a 100644
--- a/buildSrc/src/main/kotlin/JvmToolchain.kt
+++ b/buildSrc/src/main/kotlin/JvmToolchain.kt
@@ -2,7 +2,7 @@ import org.gradle.jvm.toolchain.JavaLanguageVersion
 import org.gradle.jvm.toolchain.JvmVendorSpec
 import org.gradle.api.Project
 
-private const val JDK_VERSION = 11
+private const val JDK_VERSION = 17
 
 val jvmToolchainVersion: JavaLanguageVersion
     get() {
diff --git a/gradle/libs.versions.toml b/gradle/libs.versions.toml
index 62fc47f..cce9276 100644
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -8,11 +8,11 @@ downloadTaskPlugin = "4.1.2"
 geantyref = "1.+"
 googleJavaFormat = "1.15.0"
 # must not use `+` because used in download URL
-graalVm = "22.3.3"
-# intentionally empty; replaced by patch file when building pkl-cli macos/aarch64
-graalVM23JdkVersion = "replace-me"
+graalVm = "23.0.2"
+graalVM23JdkVersion = "17.0.9"
 # slightly hacky but convenient place so we remember to update the checksum
 graalVmSha256-darwin-amd64 = "cab6a1436626adc28ec0f72791772315678e7c758e2fbae2cb6758a38f27c56a"
+graalVmSha256-macos-aarch64 = "2214b6ecb32faacc84dffcbfae930450abe77c31730c4b6310e22d8f743959a5"
 graalVmSha256-linux-amd64 = "959154e1248108dcd6eb279032cecdd2a6076bdebb345de9532681185231d255"
 graalVmSha256-linux-aarch64 = "177d682f3455d00fa2491246e809d9c9b743187a82115ad3f578998b4935d5a1"
 ideaExtPlugin = "1.1"
```

When running `./gradlew updateDependencyLocks` I will get some erros like:
<details>

```
* What went wrong:
Execution failed for task ':pkl-tools:updateDependencyLocks'.
> Could not resolve all files for configuration ':pkl-tools:compileClasspath'.
   > Could not resolve project :pkl-codegen-java.
     Required by:
         project :pkl-tools
      > No matching variant of project :pkl-codegen-java was found. The consumer was configured to find an API of a library compatible with Java 11, preferably in the form of class files, preferably optimized for standard JVMs, and its dependencies declared externally but:
          - Variant 'apiElements' capability org.pkl-lang:pkl-codegen-java:0.26.0-SNAPSHOT declares an API of a library, packaged as a jar, preferably optimized for standard JVMs, and its dependencies declared externally:
              - Incompatible because this component declares a component compatible with Java 17 and the consumer needed a component compatible with Java 11
          - Variant 'javadocElements' capability org.pkl-lang:pkl-codegen-java:0.26.0-SNAPSHOT declares a runtime of a component, and its dependencies declared externally:
              - Incompatible because this component declares documentation and the consumer needed a library
              - Other compatible attributes:
                  - Doesn't say anything about its target Java environment (preferred optimized for standard JVMs)
                  - Doesn't say anything about its target Java version (required compatibility with Java 11)
                  - Doesn't say anything about its elements (required them preferably in the form of class files)
          - Variant 'mainSourceElements' capability org.pkl-lang:pkl-codegen-java:0.26.0-SNAPSHOT declares a component, and its dependencies declared externally:
              - Incompatible because this component declares a component of category 'verification' and the consumer needed a library
              - Other compatible attributes:
                  - Doesn't say anything about its target Java environment (preferred optimized for standard JVMs)
                  - Doesn't say anything about its target Java version (required compatibility with Java 11)
                  - Doesn't say anything about its elements (required them preferably in the form of class files)
                  - Doesn't say anything about its usage (required an API)
          - Variant 'runtimeElements' capability org.pkl-lang:pkl-codegen-java:0.26.0-SNAPSHOT declares a runtime of a library, packaged as a jar, preferably optimized for standard JVMs, and its dependencies declared externally:
              - Incompatible because this component declares a component compatible with Java 17 and the consumer needed a component compatible with Java 11
          - Variant 'sourcesElements' capability org.pkl-lang:pkl-codegen-java:0.26.0-SNAPSHOT declares a runtime of a component, and its dependencies declared externally:
              - Incompatible because this component declares documentation and the consumer needed a library
              - Other compatible attributes:
                  - Doesn't say anything about its target Java environment (preferred optimized for standard JVMs)
                  - Doesn't say anything about its target Java version (required compatibility with Java 11)
                  - Doesn't say anything about its elements (required them preferably in the form of class files)
          - Variant 'testResultsElementsForTest' capability org.pkl-lang:pkl-codegen-java:0.26.0-SNAPSHOT:
              - Incompatible because this component declares a component of category 'verification' and the consumer needed a library
              - Other compatible attributes:
                  - Doesn't say anything about how its dependencies are found (required its dependencies declared externally)
                  - Doesn't say anything about its target Java environment (preferred optimized for standard JVMs)
                  - Doesn't say anything about its target Java version (required compatibility with Java 11)
                  - Doesn't say anything about its elements (required them preferably in the form of class files)
                  - Doesn't say anything about its usage (required an API)
```

</details>

StefMa:
Seems we run into a few other Gradle 7x toolchain issues 🙈 
https://github.com/gradle/gradle/issues/22796
https://github.com/gradle/gradle/issues/19382

Lets wait for #245

StefMa:
Hey @bioball and/or @stackoverflow I might need your help here:

1. For clarification. The CI checks `gradle-check-jdk11` and `gradle-check-jdk17` are meant to be check if the code got compiled/works with Java 11 and Java 17, right? 

If this is the case, then my changes are correct.
Because *before* we rely on the *installed* Java version on the machine. 
Since Gradle handles  this now, we have to put the Java version into the build script (using the system property `PKL_JVM_VERSION`).
For now I left using the respective Docker images (11 and 17), however, they are **not required anymore**.
Gradle toolchains make sure that the correct JDK will be installed before compiling anything.
We only require an java image that can "Start" Gradle 🙃 

2. Can someone help me out to fix the `graalVm23.patch` file? 🙈 

I tried various things, but non of the results look like it should (or maybe? 🤷 )
I tried to run `updateDependencyLocks`, but this seems to update *too many* updates.
I tried to run `buildNative classes --update-locks  org.graalvm.compiler:\*,org.graalvm.sdk:\*,org.graalvm.js:\*,org.graalvm.nativeimage:\*,org.graalvm.truffle:\*`, but this seems not to update all of the dependencies even if I mentioning them 😭 
The new patch file contains only 266 lines (instead of the orignal 530 lines).
For example, the whole `stdlib/gradle.lockfile` is not included in the patch at all.

So what I'm doing wrong?
How can I update it?
(Maybe these changes because of the Gradle 7.6 update? 🤷 I don't know)
This is the patchfile I end up with: https://pastebin.com/xFVt1XRa

bioball:
Hey @StefMa! Just getting around to looking at this.

To fix the patch file: you can apply the patch on a file-by-file basis using `git apply --reject patches/graalVm23.patch`. This will apply changes to files where there are no conflicts, and create `.rej` files for those where conflicts are found.

In this case, I think what you need to do is:

1. `git apply --reject patches/graalVm23.patch`
1. Delete all `.rej` files (we don't need those changes anymore thanks to this PR)
1. Create the new patch file
    - `git diff > patches/graalVm23.patch`
1. Commit the new changed patch file
1. Cleanup; reset the rest of the un-staged changes

buildSrc/src/main/kotlin/JvmToolchain.kt
```diff
@@ -0,0 +1,14 @@
+import org.gradle.jvm.toolchain.JavaLanguageVersion
+import org.gradle.jvm.toolchain.JvmVendorSpec
+import org.gradle.api.Project
+
+private const val JDK_VERSION = 11
+
+val jvmToolchainVersion: JavaLanguageVersion
+    get() {
+        val jvmVersion = System.getProperty("PKL_JVM_VERSION", JDK_VERSION.toString()).toInt()
```
bioball, line 9: [nit] unnecessary conversion from int to string back to int for the default case

```suggestion
        val jvmVersion = System.getProperty("PKL_JVM_VERSION")?.toInt() ?: JDK_VERSION
```

bioball:
The CI tests are failing on task `:pkl-config-java:generateTestConfigClasses` for gradle-check-jdk17.

Cause:

```
Caused by: org.gradle.process.internal.ExecException: Process 'command '/usr/local/jdk-11.0.22/bin/java'' finished with non-zero exit value 1
        at org.gradle.process.internal.DefaultExecHandle$ExecResultImpl.assertNormalExitValue(DefaultExecHandle.java:431)
        at org.gradle.process.internal.DefaultJavaExecAction.execute(DefaultJavaExecAction.java:52)
        at org.gradle.api.tasks.JavaExec.exec(JavaExec.java:165)
```

I tried running this locally in docker using `openjdk:11`, and am seeing this:

```
> Task :pkl-config-java:generateTestConfigClasses FAILED
Error: LinkageError occurred while loading main class org.pkl.codegen.java.Main
        java.lang.UnsupportedClassVersionError: org/pkl/codegen/java/Main has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0
```

So, it seems like Gradle toolchains might affect built-in Java compile and run tasks, but for some reason it doesn't affect `JavaExec`. This looks related: https://github.com/gradle/gradle/issues/16791

buildSrc/src/main/kotlin/JvmToolchain.kt
netvl, line 9: I believe that in Gradle, it is more conventional to use project properties rather than system properties to specify this kind of build parameters. I'd write this as
```kotlin
private const val DEFAULT_JDK_VERSION = 11

val Project.jvmToolchainVersion: JavaLanguageVersion
    get() {
        val jvmVersion = if (extra.has("pklJvmVersion")) extra.get("pklJvmVersion")?.toInt() else null
        return JavaLanguageVersion.of(jvmVersion ?: DEFAULT_JDK_VERSION)
    }
```

Then you'd run the build as `./gradlew check -PpklJvmVersion=17` (`-P` vs `-D`).

buildSrc/build.gradle.kts
```diff
@@ -17,17 +15,6 @@ dependencies {
...
-
 kotlin {
-  target {
-    compilations.configureEach {
-      kotlinOptions {
-        jvmTarget = "11"
-      }
-    }
-  }
+  jvmToolchain(11)
```
netvl, line 19: I think it would it make sense to configure the global toolchain here as well, for consistency, via the `java { toolchain {  } }` block.

buildSrc/src/main/kotlin/JvmToolchain.kt
netvl: Sorry, which discussion it is? I'm not quite sure I understand the reasoning.

Unfortunately, Gradle docs on various kinds of properties do not provide an explicit prescription on where to use which kind of thing, but they are interchangeable in a way - you can define system properties via project properties (`-PsystemProp.a.b.c=value`) and vice versa (`-Dorg.gradle.project.a.b.c=value`).

So, from my experience, I suggest using project properties instead of system properties, because:
1. This mechanism is a part of the Gradle API, and thus project properties can be retrieved in a variety of useful ways (like `val someProp by project.extra` delegates)
2. As a consequence of 1, project properties are used more in idiomatic Gradle code than system properties, in my experience.

netvl: But overall I don't think I have a very strong stance on it, that was just a suggestion about what I think is a best practice. As I said, system and project properties are mostly interchangeable in Gradle, so the primary difference is the API which is used to access them, and arguably without delegates, system properties API is marginally nicer since it does not throw exceptions)

StefMa: The PR started with a Gradle property.
@translatenix suggested [here](https://github.com/apple/pkl/pull/262#issuecomment-1969963577) to use an System property:

> I think the project property is only useful for command-line invocations, so it could be a system property.

And followed up [here](https://github.com/apple/pkl/pull/262#issuecomment-1971318196) with:

> System property would be a small improvement because setting the project property in a build script will likely not work correctly.

Since I don't have a strong stance about it either, I changed it 🤷 

because you (@netvl) say the following, I'm going to resolve this conversation 🙃 
> But overall I don't think I have a very strong stance on it, that was just a suggestion about what I think is a best practice.

buildSrc/build.gradle.kts
StefMa: The `build.gradle.kts` is only responsible for the code within `buildSrc`, right?
Since we are using Kotlin only here, why should we set the `java.toolchain {}`? 🤔 

Or what do you mean by that?

netvl: Yes, you are correct that this build script is responsible for `buildSrc`.

In Gradle, the `java` plugin is _always_ applied to JVM-based projects, even if the project uses some other language as the core of its implementation. Therefore, for consistency, it is considered idiomatic to use the `java` plugin methods to configure something first, and use language-specific methods to adjust the configuration, if necessary.

In the toolchain case, it is specified explicitly in Kotlin docs (see [here](https://kotlinlang.org/docs/gradle-configure-project.html#gradle-java-toolchains-support)) that the toolchain version in the `java` block will be used by Kotlin compilation tasks, therefore, it would be conventional to use the `java` block to set it up.

(BTW, it's curious that the docs state that setting the toolchain in the `kotlin` extension will also configure it for `JavaCompile` tasks; this is another reason to set it in the `java` block, since it is the built-in way to configure the toolchain, and their effects would be similar anyway)

buildSrc/src/main/kotlin/pklKotlinLibrary.gradle.kts
```diff
@@ -24,6 +24,11 @@ tasks.compileKotlin {
   enabled = true // disabled by pklJavaLibrary
 }
 
+kotlin.jvmToolchain {
+  languageVersion.set(jvmToolchainVersion)
+  vendor.set(jvmToolchainVendor)
+}
```
netvl, line 30: This should not be necessary because this plugin applies `pklJavaLibrary`, and the latter already applies the toolchain in the `java {}` block.

StefMa:
Closed as this is outdated.
See also https://github.com/apple/pkl/pull/876#issuecomment-2573855464

PR #271, opened 2024-02-29
Revert "Add setting for Kotlin package to codegen (#194)"

bioball:
This reverts commit 7f404fff490e2dd883cf9683c21f6fd38fd84518.

The package is derived from the module name.
Having `module com.example.Foo` in Pkl
will create Kotlin `package com.example`.

### DIFF
--- a/pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/CliKotlinCodeGeneratorOptions.kt
+++ b/pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/CliKotlinCodeGeneratorOptions.kt
@@ -29,9 +29,6 @@ data class CliKotlinCodeGeneratorOptions(
   /** The characters to use for indenting generated source code. */
   val indent: String = "  ",
 
-  /** Kotlin package to use for generated code; if none is provided, the root package is used. */
-  val kotlinPackage: String = "",
-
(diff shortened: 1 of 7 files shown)


bioball:
Heads up to @sgammon: this reverts your commit, because there's already a way to control the package name.

sgammon:
@bioball No problem, understood. But did you see my comment about JPMS? This was added as an escape hatch for that specific issue

stackoverflow (approved):

bioball:
Ah, I see. How does a special package prefix flag change or fix this? To be clear, what's the difference between:

```
module my.Module
```

```
kotlinCodeGenerators {
  register("configClasses") {
    packagePrefix.set("some.cool.pkg")
  }
}
```

And this?

```
module some.coo.pkg.my.Module
```

bioball:
@sgammon gotcha. I think the concern that you're calling out is that you cannot modify 3d party Pkl code (code that's hosted outside of a Kotlin codebase). That's indeed a valid concern, but I don't think this is the right solution to that problem.

I think what we should have instead is a way to map Pkl module names to Kotlin module names. The setting should look something like:

```
javaCodeGenerators {
  register("configClasses") {
    mappings.set(mapOf(
      "package://example.com/foo@1.0.0" to "com.example.foo"
    ))
  }
}
```

Generally, first-party Pkl code should have their module name match the expected package name.

PR #294, opened 2024-03-06
Improve CircleCI builds

bioball:
* Prevent build_artifacts.txt files from being published as GitHub releases
* Bump version to 1.0.2; make PRBs not depend on flakey pr-approval job

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -698,6 +698,8 @@ jobs:
         at: '.'
     - run:
         command: |-
+          # exclude build_artifacts.txt from publish
+          rm pkl-cli/build/executable/*.build_artifacts.txt
           gh release create "${CIRCLE_TAG}" \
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -751,15 +753,12 @@ workflows:
     - gradle-check-jdk11:
         requires:
         - hold
-        - pr-approval/authenticate
```
StefMa, line 754: Does this mean PRs will run without authenticate from your side?

bioball: No; this is a job that auto-approves the `hold` job if the commit was authored by a maintainer.

This job has been a little flakey for us, though, and it's been preventing us from testing pull requests due to being a dependency for the actual test.

This gets rid of that dependency (it doesn't actually need to be a dependency)

stackoverflow (approved):

PR #295, opened 2024-03-06
Clean up http-client changes

bioball:
* pkl-excutor tests: copy 0.25.0 distribution into pkl-executable/build
* Use `IoUtils.getPklHomeDir` in HttpClientBuilder
* Simplify Exceptions.java
* Enable testing for older pkl-executor distribution

(CC @translatenix)

### DIFF
--- a/pkl-executor/pkl-executor.gradle.kts
+++ b/pkl-executor/pkl-executor.gradle.kts
@@ -6,7 +6,7 @@ plugins {
 }
 
 val pklDistributionCurrent: Configuration by configurations.creating
-val pklDistribution025: Configuration by configurations.creating
+val pklHistoricalDistributions: Configuration by configurations.creating
 
 // Because pkl-executor doesn't depend on other Pkl modules
 // (nor has overlapping dependencies that could cause a version conflict),
@@ -15,7 +15,7 @@ val pklDistribution025: Configuration by configurations.creating
 dependencies {
   pklDistributionCurrent(project(":pkl-config-java", "fatJar"))
   @Suppress("UnstableApiUsage")
-  pklDistribution025(libs.pklConfigJavaAll025)
+  pklHistoricalDistributions(libs.pklConfigJavaAll025)
 
   implementation(libs.slf4jApi)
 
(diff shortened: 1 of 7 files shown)


pkl-executor/src/test/kotlin/org/pkl/executor/EmbeddedExecutorTest.kt
```diff
@@ -121,7 +118,7 @@ class EmbeddedExecutorTest {
             "${Release.current().version().withBuild(null).toString().replaceFirst("dev", "SNAPSHOT")}.jar")
       path.apply {
         if (!exists()) throw AssertionError("Missing test fixture. " +
-          "To fix this problem, run `./gradlew :pkl-executor:prepareTest`.")
+          "To fix this problem, run `./gradlew :pkl-config-java:build`.")
```
translatenix, line 121: This doesn't look right. `:pkl-executor:prepareTest` should be exactly what's needed. In the current build, this can be abbreviated to `gw prepTest`, but that's not what I'd use in the message.

pkl-executor/src/test/kotlin/org/pkl/executor/EmbeddedExecutorTest.kt
```diff
@@ -98,18 +98,15 @@ class EmbeddedExecutorTest {
         if (executor.isInitialized()) executor.value.close()
       }
     }
-    
+
     // a Pkl distribution that supports ExecutorSpiOptions up to v1
     private val pklDistribution1: Path by lazy {
-      val path = System.getProperty("pklDistribution025")?.toPath() ?:
```
translatenix, line 104: I don't understand why you removed this system property but not the other. And in the build script you still set both of them.

pkl-executor/pkl-executor.gradle.kts
```diff
@@ -52,10 +52,15 @@ sourceSets {
   }
 }
 
+val copyHistoricalDistribution by tasks.registering(Copy::class) {
```
translatenix, line 55: Should match the configuration name (which can of course be renamed). What's needed is a distribution that doesn't have ExecutorSpiOptions2. That's why I used the overly specific name `025`.

bioball: I am leaning towards "historical" because when there's the inevitable ExecutorSpiOptions3, maybe we will need to download another distribution that has `ExecutorSpiOptions2` but is missing `ExecutorSpiOptions3`.

pkl-executor/src/test/kotlin/org/pkl/executor/EmbeddedExecutorTest.kt
bioball: Ah, thanks, should get rid of this too.

pkl-executor/src/test/kotlin/org/pkl/executor/EmbeddedExecutorTest.kt
bioball: Ah, I changed this because somehow `prepareTest` wasn't building a new `pkl-config-java-all` jar. But, I just tested again and it's working, so 🤷 

Will change this line back.

> I can't tell if you want every Gradle build to prepare for running EmbeddedExecutorTest with IntelliJ. If you do, there's no point in keeping the system properties around

Yep, that's the intention.

pkl-executor/pkl-executor.gradle.kts
translatenix: Oh I see. You want to put all distributions into a single configuration.

bioball: Yup, exactly!

> By the way, would be easy to create symlinks instead of copying (iterate over the configuration in doLast and use Files.createSymbolicLink).

Switched to making symlinks

stackoverflow (approved):
LGTM

PR #297, opened 2024-03-07
fix bug with for generator variables in mixin

stackoverflow:


pkl-core/src/main/java/org/pkl/core/ast/expression/literal/AmendFunctionNode.java
```diff
@@ -76,8 +79,17 @@ public AmendFunctionNode(
...
+        forGenVars.add(
+            new Slot(hostFrameDesecriptor.getSlotKind(i), hostFrameDesecriptor.getSlotName(i)));
+      } else {
+        forGenVars.add(new Slot(hostFrameDesecriptor.getSlotKind(i), Identifier.DUMMY));
+      }
+    }
+    // only add for generator vars if there's any
+    if (hasForGenVars) {
+      for (var slot : forGenVars) {
+        builder.addSlot(slot.kind, slot.name, null);
```
bioball, line 92: We can avoid the extra allocation here; can just start filling things in the first time we encounter a for generator variable.

```suggestion
        if (!hasForGenVars) {
          hasForGenVars = true;
          for (var j = 0; j < i; j++) {
            builder.addSlot(FrameSlotKind.Illegal, new Object(), null);
          }
        }
        builder.addSlot(hostFrameDesecriptor.getSlotKind(i), hostFrameDesecriptor.getSlotName(i), null);
      } else if (hasForGenVars) {
        builder.addSlot(FrameSlotKind.Illegal, new Object(), null);
```

pkl-core/src/main/java/org/pkl/core/runtime/Identifier.java
```diff
@@ -152,6 +152,9 @@ public final class Identifier implements Comparable<Identifier> {
   // common in lambdas etc
   public static final Identifier IT = get("it");
 
+  // dummy, unrepresentable identifier
+  public static final Identifier DUMMY = get("`#_");
```
bioball, line 156: Let's use this name too in `FunctionNode.java`:

```diff
diff --git a/pkl-core/src/main/java/org/pkl/core/ast/member/FunctionNode.java b/pkl-core/src/main/java/org/pkl/core/ast/member/FunctionNode.java
index 31c7d8791..bdedbda41 100644
--- a/pkl-core/src/main/java/org/pkl/core/ast/member/FunctionNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/member/FunctionNode.java
@@ -165,8 +165,8 @@ public final class FunctionNode extends MemberNode {
     var parameters = CollectionUtils.<String, PType>newLinkedHashMap(paramCount);
     for (var i = 0; i < paramCount; i++) {
       var slotName = getFrameDescriptor().getSlotName(i);
-      // Ignored parameters (`_`) have no name
-      var paramName = slotName == null ? "_#" + i : slotName.toString();
+      // This is a blank identifier; use dummy identifier name (impossible to represent in Pkl code)
+      var paramName = slotName == null ? Identifier.DUMMY.toString() + i : slotName.toString();
       parameters.put(paramName, TypeNode.export(parameterTypeNodes[i]));
     }
 

```

pkl-core/src/test/files/LanguageSnippetTests/input/generators/mixinForGeneratorVariable.pkl
```diff
@@ -0,0 +1,10 @@
+
+function addElements(keys: List<String>): Mixin<Mapping<String, String>> = new {
+  for (key in keys) {
+    [key] = key
+  }
+}
+
+res = new Mapping<String, String> {
+  ["base"] = "alreadyThere"
+} |> addElements(List("newElement"))
```
bioball, line 10: Move this test to file `forGeneratorInMixins`

bioball (changes requested):
Generally LGTM! Some comments.

pkl-core/src/main/java/org/pkl/core/runtime/Identifier.java
bioball: Ah, scratch that. These names need to be unique because this is part of the reflect API.

pkl-core/src/main/java/org/pkl/core/ast/expression/literal/AmendFunctionNode.java
```diff
@@ -76,8 +79,17 @@ public AmendFunctionNode(
       //
       // As a remedy, we simply copy outer for-generator variables into this frame.
```
bioball, line 80: Oh also, we should update this comment

pkl-core/src/main/java/org/pkl/core/ast/expression/literal/AmendFunctionNode.java
stackoverflow: Good one.

pkl-core/src/main/java/org/pkl/core/ast/expression/literal/AmendFunctionNode.java
```diff
@@ -77,19 +75,19 @@ public AmendFunctionNode(
       // frame (e.g. with `new Mixin { ... }` syntax), so it injects for-generator vars into the
       // wrong frame.
       //
-      // As a remedy, we simply copy outer for-generator variables into this frame.
+      // As a remedy, we simply copy outer variables into this frame if there are any for generator
+      // variables.
```
bioball, line 79:
```suggestion
      // variables.
      //
      // We need to preserve the frame slot index, so we insert dummy identifiers
      // for other slots that aren't for generator variables.
```

bioball (approved):
I have a suggestion for the comment, but overall LGTM

stackoverflow:
> I have a suggestion for the comment, but overall LGTM

Already added your suggestion, or do you have another one?

bioball:
Yeah, this one, which I don't think is part of a commit? https://github.com/apple/pkl/pull/297/files#r1516633570

stackoverflow:
It was added in my last commit.

bioball:
Oh I see it. My bad

PR #314, opened 2024-03-13
Add a build job for the `jpkl` executable

KushalP:
This is an artefact that we currently build and test for, but that we weren't releasing. Add a build job so that it can go out with a release.


.circleci/config.pkl
```diff
@@ -112,6 +113,7 @@ jobs {
   for (jobName, job in buildNativeJobs) {
     [jobName] = job.job
   }
+  ["pkl-cli-java"] = new BuildJavaExecutableJob {}.job
```
bioball, line 116: We already have a jpkl created as part of our build; it's built during our maven deploy jobs.

Actually I think the only change that's needed is a new `PersistToWorkspaceStep` in `DeployJob`; diff here:

```diff
diff --git a/.circleci/jobs/DeployJob.pkl b/.circleci/jobs/DeployJob.pkl
index a8281d6f1..bfc05b31c 100644
--- a/.circleci/jobs/DeployJob.pkl
+++ b/.circleci/jobs/DeployJob.pkl
@@ -32,4 +32,11 @@ steps {
   new Config.RunStep {
     command = "./gradlew \(self.gradleArgs) \(module.command)"
   }
+  // add jpkl to workspace so it gets published as a GitHub release
+  new Config.PersistToWorkspaceStep {
+    root = "."
+    paths {
+      "pkl-cli/build/executable/"
+    }
+  }
 }
```

bioball (approved):

PR #326, opened 2024-03-16
Use layout.buildDirectory

bioball:
This switches Gradle scripts to use `layout.buildDirectory` instead of hard-coded "build".

### DIFF
--- a/pkl-cli/pkl-cli.gradle.kts
+++ b/pkl-cli/pkl-cli.gradle.kts
@@ -56,13 +56,13 @@ dependencies {
 
   testImplementation(projects.pklCommonsTest)
 
-  val buildDir = layout.buildDirectory
-  stagedMacAmd64Executable(files(buildDir.dir("executable/pkl-macos-amd64")))
-  stagedMacAmd64Executable(files(buildDir.dir("executable/pkl-macos-amd64")))
-  stagedMacAarch64Executable(files(buildDir.dir("executable/pkl-macos-aarch64")))
-  stagedLinuxAmd64Executable(files(buildDir.dir("executable/pkl-linux-amd64")))
-  stagedLinuxAarch64Executable(files(buildDir.dir("executable/pkl-linux-aarch64")))
-  stagedAlpineLinuxAmd64Executable(files(buildDir.dir("executable/pkl-alpine-linux-amd64")))
+  fun executableDir(name: String) = files(layout.buildDirectory.dir("executable/$name"))
+  stagedMacAmd64Executable(executableDir("pkl-macos-amd64"))
+  stagedMacAmd64Executable(executableDir("pkl-macos-amd64"))
+  stagedMacAarch64Executable(executableDir("pkl-macos-aarch64"))
+  stagedLinuxAmd64Executable(executableDir("pkl-linux-amd64"))
+  stagedLinuxAarch64Executable(executableDir("pkl-linux-aarch64"))
+  stagedAlpineLinuxAmd64Executable(executableDir("pkl-alpine-linux-amd64"))
 }
 
 tasks.jar {
--- a/pkl-cli/pkl-cli.gradle.kts
+++ b/pkl-cli/pkl-cli.gradle.kts
@@ -56,11 +56,13 @@ dependencies {
 
   testImplementation(projects.pklCommonsTest)
 
-  stagedMacAmd64Executable(files("build/executable/pkl-macos-amd64"))
-  stagedMacAarch64Executable(files("build/executable/pkl-macos-aarch64"))
-  stagedLinuxAmd64Executable(files("build/executable/pkl-linux-amd64"))
-  stagedLinuxAarch64Executable(files("build/executable/pkl-linux-aarch64"))
-  stagedAlpineLinuxAmd64Executable(files("build/executable/pkl-alpine-linux-amd64"))
+  val buildDir = layout.buildDirectory
+  stagedMacAmd64Executable(files(buildDir.dir("executable/pkl-macos-amd64")))
+  stagedMacAmd64Executable(files(buildDir.dir("executable/pkl-macos-amd64")))
+  stagedMacAarch64Executable(files(buildDir.dir("executable/pkl-macos-aarch64")))
+  stagedLinuxAmd64Executable(files(buildDir.dir("executable/pkl-linux-amd64")))
+  stagedLinuxAarch64Executable(files(buildDir.dir("executable/pkl-linux-aarch64")))
+  stagedAlpineLinuxAmd64Executable(files(buildDir.dir("executable/pkl-alpine-linux-amd64")))
 }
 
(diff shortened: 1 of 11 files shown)


pkl-config-java/pkl-config-java.gradle.kts
```diff
@@ -10,14 +10,18 @@ val pklCodegenJava: Configuration by configurations.creating
...
+  val outputDir = layout.buildDirectory.dir("testConfigClasses")
+  outputs.dir(outputDir)
   inputs.dir("src/test/resources/codegenPkl")
 
   classpath = pklCodegenJava
   mainClass.set("org.pkl.codegen.java.Main")
-  args("--output-dir", "build/testConfigClasses")
-  args("--generate-javadoc")
-  args(fileTree("src/test/resources/codegenPkl"))
+  argumentProviders.add(CommandLineArgumentProvider {
```
translatenix, line 19: Using an argument provider has no benefits here.
This should work fine:
```
args("--output-dir", outputDir.get().asFile)
args("--generate-javadoc")
args(fileTree("src/test/resources/codegenPkl"))
```

pkl-config-kotlin/pkl-config-kotlin.gradle.kts
```diff
@@ -28,18 +28,21 @@ dependencies {
...
-  outputs.dir("build/testConfigClasses")
+  val outputDir = layout.buildDirectory.dir("testConfigClasses")
+  outputs.dir(outputDir)
   inputs.dir("src/test/resources/codegenPkl")
 
   classpath = pklCodegenKotlin
   mainClass.set("org.pkl.codegen.kotlin.Main")
-  args("--output-dir", "build/testConfigClasses")
-  args(fileTree("src/test/resources/codegenPkl"))
+  argumentProviders.add(CommandLineArgumentProvider {
```
translatenix, line 37: using an argument provider has no benefits here

pkl-cli/pkl-cli.gradle.kts
```diff
@@ -56,11 +56,13 @@ dependencies {
 
   testImplementation(projects.pklCommonsTest)
 
-  stagedMacAmd64Executable(files("build/executable/pkl-macos-amd64"))
-  stagedMacAarch64Executable(files("build/executable/pkl-macos-aarch64"))
-  stagedLinuxAmd64Executable(files("build/executable/pkl-linux-amd64"))
-  stagedLinuxAarch64Executable(files("build/executable/pkl-linux-aarch64"))
-  stagedAlpineLinuxAmd64Executable(files("build/executable/pkl-alpine-linux-amd64"))
+  val buildDir = layout.buildDirectory
```
translatenix, line 59: `fun executableDir(name: String) = files(layout.buildDirectory.dir("executable/$name"))`

pkl-config-java/pkl-config-java.gradle.kts
bioball: Calling `outputDir.get()` is eager. So, this just lets us configure the task while being able to keep the laziness of the `buildDirectory` provider.

As far as I can tell, this is the only way to configure an exec task while keeping it lazy.

translatenix: And what's the benefit of keeping the laziness of the buildDirectory provider? Either my Gradle knowledge is outdated, or you have some misconceptions about this.

bioball: According to their docs, the whole reason they switched from the eager `getBuildDir()` to this lazy API is to prevent ordering issues: https://docs.gradle.org/current/userguide/upgrading_version_8.html#deprecations_4

Seems like using this API eagerly kind of defeats the purpose; we'd still be vulnerable to ordering issues because we'd be configuring the task eagerly.

Note: I don't think we'd actually be vulnerable since we're not changing the build dir. But, in any case, it's good to try to be more idiomatic.

translatenix: Right, the `buildDir` ordering issue is irrelevant to Pkl and most other projects.

Gradle has been slowly migrating from `Foo` to `Property<Foo>`/`Provider<Foo>` for years.
For APIs that have been migrated, it makes sense to avoid `get()` at configuration time 
and use `property.set(provider)` instead of `property.set(foo)`.
This prevents ordering issues and enables task dependency tracking.
For the many APIs that haven't been migrated, using `get()` is normal and expected.

`JavaExec.args` hasn't been migrated to `ListProperty`, which would allow `args.add(outputDir)`.
Using an argument provider only to delay `outputDir.get()` isn't worth its weight.

If you'd like to dive deeper into this madness, I recommend reading through https://github.com/gradle/gradle/issues/4685.

stackoverflow (approved):

StefMa (approved):
I have the feeling this change based on my commit in the previous PR so I have the feeling I have to say something about it here as well 🙈

One one side, these changes might not bring any benefit in regards of build time. I agree that we probably use an lazy provider here and there that just adds a little bit of noise (cause we have to call `get()` or chain other lazy APIs that add no benefit but are required to use because we decided to go with the `layout.buildDirectory` property). 

I also agree that probably going with a hardcoded `build` string would sufficient and safe to use for the years to come. 

So I can fully understand @translatenix arguments here. 

However I am also a friend of using the newest stuff and don't hardcoded values that don't have to be hardcoded. So in case new APIs are available and can be used without **that much noise** I'm all in with the migration. Even though it doesn't bring any benefit and might look here and there a bit odd. 

So I can fully understand why @bioball opt in to make the changes 🙃. 

Just my 2 cents about this 🤝

pkl-config-java/pkl-config-java.gradle.kts
```diff
@@ -10,14 +10,18 @@ val pklCodegenJava: Configuration by configurations.creating
...
+    listOf(
+      "--output-dir", outputDir.get().asFile.absolutePath,
```
translatenix, line 21: It's always safe to use `.path` instead of `.absolutePath` in Gradle build scripts.
If a build script mistakenly creates a relative `File` by using `File("relative/path")` instead of `file("relative/path")`,
`.absolutePath` won't save the day because it will use the wrong base path (current working dir instead of `project.dir`).

pkl-config-java/pkl-config-java.gradle.kts
bioball: > For the many APIs that haven't been migrated, using `get()` is normal and expected.

I'd agree if there was no way to lazily configure the arguments. But since there is here, we'll just keep it this way. I don't think it's adding too much complexity to do it this way.

But yeah, it's pretty silly that they haven't provided a way to set `args` directly with lazy values yet.

pkl-config-java/pkl-config-java.gradle.kts
bioball: Will change!

PR #327, opened 2024-03-16
Fix build issue where test fixtures weren't generated

translatenix:
Make sure that declaring a dependency on project `pkl-commons-test` suffices to have its test fixtures generated.

This fix should work reliably.
However, there may be a more idiomatic way to achieve the same result.


bioball (approved):
Thanks! Yeah, this should be good enough, despite these no longer being resources.

PR #328, opened 2024-03-17
Follow HTTP redirects

translatenix:
- Change HttpClient to follow all redirects except HTTPS to HTTP.
- Run language snippet tests with --no-cache and real PackageServer instead of pre-seeded cache. This increases HTTP test coverage and enables testing of package redirects.
- Change PackageServer to return 301 for request paths starting with /HTTP301/ and 307 for request paths starting with /HTTP307/.
- Update some outdated test package checksums that apparently weren't verified.

### DIFF
--- a/pkl-core/src/test/kotlin/org/pkl/core/LanguageSnippetTestsEngine.kt
+++ b/pkl-core/src/test/kotlin/org/pkl/core/LanguageSnippetTestsEngine.kt
@@ -4,9 +4,12 @@ import org.junit.platform.engine.EngineDiscoveryRequest
 import org.junit.platform.engine.TestDescriptor
 import org.junit.platform.engine.UniqueId
 import org.junit.platform.engine.support.descriptor.EngineDescriptor
+import org.pkl.commons.test.FileTestUtils
 import org.pkl.commons.test.InputOutputTestEngine
 import org.pkl.commons.test.PackageServer
+import org.pkl.core.http.HttpClient
 import org.pkl.core.project.Project
+import org.pkl.core.util.IoUtils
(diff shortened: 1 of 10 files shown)


pkl-core/src/test/kotlin/org/pkl/core/LanguageSnippetTestsEngine.kt
```diff
@@ -58,6 +58,15 @@ abstract class AbstractLanguageSnippetTestsEngine : InputOutputTestEngine() {
     return expectedOutputDir.resolve(stdoutPath)
   }
 
+  override fun beforeAll() {
+    // disable SHA verification for packages
+    IoUtils.setTestMode()
```
translatenix, line 63: I'd love to get rid of this.

translatenix: The snippet tests now pass without setting `IoUtils.setTestMode()`.

I'm no longer sure what the "Qualify of life improvement" in `PackageResolvers.verifyPackageMetadataBytes` accomplishes. It checks for `IoUtils.isTestMode()`, which I think is always `false` unless when running native tests.

bioball (approved):
LGTM, thanks! You beat me to this PR.

pkl-core/src/test/kotlin/org/pkl/core/LanguageSnippetTestsEngine.kt
bioball: Our snippet tests used to use a real server, but we switched to a pre-seeded cache because our tests were really flakey.

PR #329, opened 2024-03-17
Polish HttpClient API

translatenix:
- Rename builder methods from setFoo() to foo(). As of Java 17, the getFoo()/setFoo(value) naming convention has been effectively replaced with foo()/foo(value). Especially for a pre-1.0 project, it makes sense to align with this development.
- Do not rename addFoo() builder methods to distinguish them from setters.
- Change HttpClientInitException to extend RuntimeException instead of PklException. Except for PklBugException, no other exception class extends PklException.


translatenix:
I just came across a new builder API in JDK 22 that uses `withFoo` for both setters and adders.
Need some time to think about what I'd like to propose.

PR #340, opened 2024-03-19
Update pkl-config-java.adoc

zihluwang:
There was an unpaired useless end bracket at _Installation_ > _Gradle_ > Codes in _Groovy_ which is removed in this PR.
Screenshot 2024-03-19 at 15 53 39


bioball (approved):
Thanks!

PR #341, opened 2024-03-19
Removed undeclared method reference in examples.

zihluwang:
There is no method `evaluateText` in `ConfigEvaluator` or `ConfigEvaluatorImpl`.


bioball (approved):
Good catch, thanks!

PR #344, opened 2024-03-20
Eliminate unnecessary use of GregorianCalendar

translatenix:
It's unnecessary to load and instantiate an expensive Calendar class to compute this constant.

jshell> new GregorianCalendar(1980, Calendar.FEBRUARY, 1, 0, 0, 0).getTimeInMillis();
$1 ==> 318240000000

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/project/ProjectPackager.java
+++ b/pkl-core/src/main/java/org/pkl/core/project/ProjectPackager.java
@@ -26,9 +26,9 @@ import java.nio.file.Path;
 import java.security.DigestOutputStream;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.time.LocalDateTime;
+import java.time.Month;
 import java.util.ArrayList;
-import java.util.Calendar;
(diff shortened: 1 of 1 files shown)


translatenix:
Turns out that the code comment is wrong: `new GregorianCalendar(1980, Calendar.FEBRUARY, 1, 0, 0, 0)` isn't 1980 February 1st CET but 1980 February 1st in the JVM's default time zone. This means that Pkl package ZIPs aren't deterministic. I'll investigate if there's a way to fix this.

translatenix:
I think I've found the correct solution. Unfortunately I can't trigger CI.

bioball (approved):
If you're curious about _why_ it's this value, take a read through the madness here: https://github.com/gradle/gradle/blob/9473938dd977cf8c76a1b7fe34cebaa041c4c633/subprojects/core/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java#L40-L55

translatenix:
It seems the madness is primarily caused by ZIP being a very old legacy format.

PR #345, opened 2024-03-20
Replace magic test port 12110 with 0

translatenix:
Also introduce constants for some test package SHAs to make them easier to update.


pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
```diff
@@ -105,9 +105,9 @@ private HttpRequest rewriteRequest(HttpRequest original) {
   }
 
   private URI rewriteUri(URI uri) {
-    // Would be nice to use port 0 instead of 12110,
+    // Would be nice to use port 0 instead of 0,
     // but this is best done in a separate commit.
```
bioball, line 109:
```suggestion
```

bioball (approved):
You didn't like the shoutout to [Latham, NY](https://www.unitedstateszipcodes.org/12110/)?

Jk; this port was arbitrarily picked. `0` is definitely better.

PR #349, opened 2024-03-20
Re-enable gradle compatibility tests

bioball:


stackoverflow (approved):
Oh. Totally forgot about that...

PR #350, opened 2024-03-21
Update dependencies

translatenix:
- Update dependencies without bumping versions in libs.versions.toml.
- Freeze kotlinx.html and kotlinx.serialization versions until Kotlin itself is updated.

Most version changes affect test dependencies.


holzensp:
The Graal23 patch file needed updating. Did a `git rebase-patch` and resolved the conflicts. It's not all there yet, because running `git apply --reject patches/graalVm23.patch && gw :pkl-config-java:shadowJar` fails for me on a `Unsupported class file major version 64`, so there's some Java 21 creeping into the build.

translatenix:
> so there's some Java 21 creeping into the build

That’s because some of the Graal JARs are multi-release JARs, i.e., they contain classes for multiple JDK versions (17-21) and the JVM picks the right ones at runtime. However, the shadowJar plugin needs to be able to process all of them. The solution is to update the shadowJar plugin. I did this in my JDK 21 PR, which also got stuck on the patch file.

translatenix:
I updated the shadowJar plugin, which should take care of the "unsupported class file version" problem. The patch still applies cleanly. I can't test the patch because it doesn't work on Linux x64. Can the team trigger another PR build?

translatenix:
Ready to squash and merge from my side.

translatenix:
Further dependency updates and JDK21/22 CI builds are blocked on this.

bioball (approved):
LGTM, thank you!

translatenix:
@bioball Some CI builds are failing. I believe jline should be locked to its previous version (3.23.0) in `libs.versions.toml` because newer versions have a problem with native image. I already filed an issue for jline.

bioball:
Yeah, sounds good. Where is the issue that you filed? They're using an option that got introduced for GraalVM for JDK 21: https://www.graalvm.org/release-notes/JDK_21/

translatenix:
just sent a PR with a link to my jline PR

PR #357, opened 2024-03-22
Avoid resolving paths on fragments

holzensp:
The more principled solution is to have `importUri` be a `URI` instead of a `String` and to test its scheme.

Generally, the problem is that `URI::resolve` can not assume fragments to be a path (and that it discards fragments altogether).

### DIFF
--- a/pkl-doc/src/main/kotlin/org/pkl/doc/DocPackageInfo.kt
+++ b/pkl-doc/src/main/kotlin/org/pkl/doc/DocPackageInfo.kt
@@ -161,7 +161,7 @@ data class DocPackageInfo(
       "pkl:/" -> "pkl:${moduleName.substring(4)}".toUri()
       else -> {
         val path = getModulePath(moduleName, moduleNamePrefix).uriEncoded + ".pkl"
-        URI(importUri).resolve(path)
+        URI(importUri + path)
       }
     }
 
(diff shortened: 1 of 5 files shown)


pkl-doc/src/main/kotlin/org/pkl/doc/DocPackageInfo.kt
```diff
@@ -165,7 +165,7 @@ data class DocPackageInfo(
             .split("/")
             .map { it.uriEncoded }
             .joinToString("/") { it } + ".pkl"
-        URI(importUri).resolve(path)
+        if (importUri.startsWith("package")) URI(importUri + path) else URI(importUri).resolve(path)
```
bioball, line 168: We can actually just use `URI(importUri + path)` in all cases.

Looks like I was the one that broke this logic in [an earlier PR](https://github.com/apple/pkl/commit/dddae1190febb148821a8560ffafbbba7dba413a#diff-08d9462f05ae0598bb1e90c634ccf7295edb7119c9bfc29d4a83fe93c9bd0ca9R162-R169)

```suggestion
        URI(importUri + path)
```

bioball (approved):
Approving to unblock, just one minor comment.

stackoverflow (approved):
LGTM

PR #359, opened 2024-03-22
Change HttpClientInitException to extend RuntimeException

translatenix:
PklException is mainly used to indicate errors during evaluation of Pkl code. It isn't a suitable superclass for HttpClientInitException.


bioball (approved):
Missed this in the review. Thanks!

PR #362, opened 2024-03-25
Fix source links in pkldoc

bioball:
Fixes an issue where source links are incorrectly URI encoded; i.e. `https%3A//github.com` instead of `https://github.com`.

This was causing the browser to resolve these as relative to the enclosing page.


stackoverflow (approved):
LGTM

PR #368, opened 2024-03-27
Fix wrong data type in kotlin example

TheFruxz:
Hey, by trying out the example, provided in the kotlin docs, i figured out, that the data type for seeds is not a viable list and errors are thrown if you're trying to use the example.

So no super large fix, but smth. that will improve the way to get into this


bioball:
Thanks! There's a little more to be done here. I submitted #372 that rolls in your changes.

PR #373, opened 2024-03-28
Ensure owner and receiver are reset after executing alias

bioball:
This fixes an issue where the frame's owner/receiver are not reset if a type test on a typealias fails.

### DIFF
--- a/pkl-core/src/test/files/LanguageSnippetTests/input/types/typeAlias2.pkl
+++ b/pkl-core/src/test/files/LanguageSnippetTests/input/types/typeAlias2.pkl
@@ -35,3 +35,10 @@ typealias Bar = Foo
 class Baz extends Bar {}
 
 res7 = (new Baz {}).value
+
+typealias MappingOrListing = Mapping|Listing
+
+res8 = new {
+  foo = 5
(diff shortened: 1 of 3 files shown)


pkl-core/src/test/files/LanguageSnippetTests/input/types/typeAlias2.pkl
```diff
@@ -35,3 +35,10 @@ typealias Bar = Foo
 class Baz extends Bar {}
 
 res7 = (new Baz {}).value
+
+typealias MappingOrListing = Mapping|Listing
+
+res8 = new {
+  foo = 5
+  bar = if (foo is MappingOrListing) "bar" else foo
```
bioball, line 43: This is failing in the main branch because the type check on `MappingOrListing` (`aliasedTypeNode.execute(...)` throws a (caught) type check exception, thus the owner/receiver don't get reset. Pkl then executes the `else` branch, and the owner/receiver are incorrectly set to the module.

holzensp: Good catch

holzensp (approved):

PR #383, opened 2024-04-02
Use Files.newInputStream()/newOutputStream() where appropriate

translatenix:
Use `Files.newInputStream(path)` instead of `new FileInputStream(path.toFile())` and `Files.newOutputStream(path)` instead of `new FileOutputStream(path.toFile())`.


bioball (approved):
LGTM, thanks!

PR #390, opened 2024-04-03
Turn EvalTask into a tracked Gradle task

translatenix:
Rationale: "Output file names are known only after execution" isn't a legitimate reason to make this task untracked because the output directory is known in this case. This is no different from (say) the `JavaCompile` task, where only the class file output directory is known upfront. Forum discussion didn't reveal another reason why this task should be untracked.

### DIFF
--- a/pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
+++ b/pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
@@ -19,26 +19,32 @@ import java.io.File;
 import org.gradle.api.file.DirectoryProperty;
 import org.gradle.api.file.RegularFileProperty;
 import org.gradle.api.provider.Property;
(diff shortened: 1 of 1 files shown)


StefMa (commented):
Do we want to have a test for this? 🤔

translatenix:
A test for what exactly?

StefMa:
Testing that if running it twice, the second run is up-to-date.

translatenix:
I think the only potential benefit of such a test is that it fails if someone adds back `@Untracked`. Otherwise it's just testing Gradle.

holzensp (approved):

bioball:
The reason this is untracked is because both `getOutputFile()` and `getMultipleFileOutputDir()` accepts [placeholder values](https://pkl-lang.org/main/current/pkl-cli/index.html#batch-evaluation), that get substituted out before Pkl writes any files.

Issue #393, opened 2024-04-04
[Bug] sorting


jw-y:
``` pkl
local com = (a, b) -> a >= b
local a = List(0, 0, 0, 0, 1, 1, 1, 1, 1, 1)
b = a.sortWith(com)
```

expected output:
```
b = List(1, 1, 1, 1, 1, 1, 0, 0, 0, 0)
```

got:
```
b = List(1, 1, 1, 1, 0, 0, 0, 0, 1, 1)
```

holzensp:
This is _wild_! I managed to reproduce. Thanks for the bug report.

translatenix:
I don't think this is a bug.
The documentation of `List.sortWith` says:
> comparator should return true if its first argument comes before its second argument in the sort order, and false otherwise.

So it should be `local com = (a, b) -> a > b` instead of `local com = (a, b) -> a >= b`.

I checked Swift's and Scala's `sortWith` functions and they have the same requirement. Swift additionally states that the comparator function must be *irreflexive*.

translatenix:
That indeed looks like a bug. I guess your incorrect comparator function should only give you an unstable/inefficient sort, not a wrong sort.

PR #394, opened 2024-04-04
Fix MergeSort.java

jw-y:
Fix for https://github.com/apple/pkl/issues/393#issue-2224982881


bioball:
Thanks!

Let's add a test for this in https://github.com/apple/pkl/blob/58ed8242af664ecf156ad1ea72a79214c794b4bc/pkl-core/src/test/files/LanguageSnippetTests/input/api/list.pkl#L242

Then run `OVERWRITE_SNIPPETS=1 ./gradlew pkl-core:check` to update the output file.

PR #397, opened 2024-04-04
Add quick link to standard library docs

translatenix:
Motivation:
The standard library docs are some of the most important docs but currently very hard to find. I've noticed that even advanced users have never heard of them and complain about "zero docs".


holzensp (approved):
Yeah, this is a good shout.

PR #403, opened 2024-04-05
Make EvalTask track resolved output paths

bioball:
This makes Gradle track `getOutputPaths()` and `getMultipleFileOutputPaths`. These represent the actual outputs of the EvalTask.

### DIFF
--- a/pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
+++ b/pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
@@ -69,8 +69,8 @@ public abstract class EvalTask extends ModulesTask {
   @OutputFiles
   @Optional
   public SetProperty<File> getOutputPaths() {
-    var ret = getObjectFactory().setProperty(File.class).value(getCliEvaluator().getOutputFiles());
-    ret.disallowChanges();
+    var ret = getObjectFactory().setProperty(File.class);
+    ret.value(getCliEvaluator().getOutputFiles()).disallowChanges();
     return ret;
   }
 
--- a/pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
+++ b/pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
@@ -33,7 +33,8 @@ public abstract class EvalTask extends ModulesTask {
   @Internal
   public abstract RegularFileProperty getOutputFile();
 
-  @Internal
+  @Input
+  @Optional
   public abstract Property<String> getOutputFormat();
 
   @Input
--- a/pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
+++ b/pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
@@ -16,12 +16,14 @@
 package org.pkl.gradle.task;
 
 import java.io.File;
-import javax.inject.Inject;
+import java.util.Collections;
+import java.util.Set;
+import javax.annotation.Nullable;
 import org.gradle.api.file.DirectoryProperty;
+import org.gradle.api.file.FileCollection;
 import org.gradle.api.file.RegularFileProperty;
-import org.gradle.api.model.ObjectFactory;
 import org.gradle.api.provider.Property;
-import org.gradle.api.provider.SetProperty;
+import org.gradle.api.provider.Provider;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.Optional;
@@ -33,59 +35,62 @@ import org.pkl.cli.CliEvaluatorOptions;
 public abstract class EvalTask extends ModulesTask {
 
   // not tracked because it might contain placeholders
+  // required
   @Internal
   public abstract RegularFileProperty getOutputFile();
 
   // not tracked because it might contain placeholders
+  // optional
   @Internal
   public abstract DirectoryProperty getMultipleFileOutputDir();
 
   @Input
-  @Optional
   public abstract Property<String> getOutputFormat();
 
   @Input
-  @Optional
   public abstract Property<String> getModuleOutputSeparator();
 
   @Input
-  @Optional
   public abstract Property<String> getExpression();
 
-  @Inject
-  public abstract ObjectFactory getObjectFactory();
-
-  private CliEvaluator getCliEvaluator() {
-    return new CliEvaluator(
-        new CliEvaluatorOptions(
-            getCliBaseOptions(),
-            getOutputFile().get().getAsFile().getAbsolutePath(),
-            getOutputFormat().get(),
-            getModuleOutputSeparator().get(),
-            mapAndGetOrNull(getMultipleFileOutputDir(), (it) -> it.getAsFile().getAbsolutePath()),
-            getExpression().get()));
-  }
+  private final Provider<CliEvaluator> cliEvaluator =
+      getProviders()
+          .provider(
+              () ->
+                  new CliEvaluator(
+                      new CliEvaluatorOptions(
+                          getCliBaseOptions(),
+                          getOutputFile().get().getAsFile().getAbsolutePath(),
+                          getOutputFormat().get(),
+                          getModuleOutputSeparator().get(),
+                          mapAndGetOrNull(
+                              getMultipleFileOutputDir(), it -> it.getAsFile().getAbsolutePath()),
+                          getExpression().get())));
 
   @OutputFiles
   @Optional
-  public SetProperty<File> getOutputPaths() {
-    var ret = getObjectFactory().setProperty(File.class);
-    ret.value(getCliEvaluator().getOutputFiles()).disallowChanges();
-    return ret;
+  public FileCollection getEffectiveOutputFilePaths() {
+    return getObjects()
+        .fileCollection()
+        .from(cliEvaluator.map(e -> nullToEmpty(e.getOutputFiles())));
   }
 
   @OutputDirectories
   @Optional
-  public SetProperty<File> getMultipleFileOutputPaths() {
-    var ret = getObjectFactory().setProperty(File.class);
-    ret.value(getCliEvaluator().getOutputDirectories()).disallowChanges();
-    return ret;
+  public FileCollection getOutputDirectories() {
+    return getObjects()
+        .fileCollection()
+        .from(cliEvaluator.map(e -> nullToEmpty(e.getOutputDirectories())));
+  }
+
+  private static <T> Set<T> nullToEmpty(@Nullable Set<T> set) {
+    return set == null ? Collections.emptySet() : set;
   }
 
   @Override
   protected void doRunTask() {
     //noinspection ResultOfMethodCallIgnored
     getOutputs().getPreviousOutputFiles().forEach(File::delete);
-    getCliEvaluator().run();
+    cliEvaluator.get().run();
   }
 }
--- a/pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
+++ b/pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
@@ -16,17 +16,19 @@
 package org.pkl.gradle.task;
 
 import java.io.File;
+import java.util.Set;
 import org.gradle.api.file.DirectoryProperty;
 import org.gradle.api.file.RegularFileProperty;
 import org.gradle.api.provider.Property;
+import org.gradle.api.provider.Provider;
+import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.Internal;
-import org.gradle.api.tasks.UntrackedTask;
+import org.gradle.api.tasks.Optional;
+import org.gradle.api.tasks.OutputDirectories;
+import org.gradle.api.tasks.OutputFiles;
 import org.pkl.cli.CliEvaluator;
 import org.pkl.cli.CliEvaluatorOptions;
 
-@UntrackedTask(
-    because =
-        "Output file names accept placeholder values, and actual file names and directories can't be known ahead of time")
 public abstract class EvalTask extends ModulesTask {
   @Internal
   public abstract RegularFileProperty getOutputFile();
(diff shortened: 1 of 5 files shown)


pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
```diff
@@ -19,32 +19,26 @@
...
-import org.gradle.api.tasks.Input;
-import org.gradle.api.tasks.Optional;
-import org.gradle.api.tasks.OutputDirectory;
-import org.gradle.api.tasks.OutputFile;
+import org.gradle.api.tasks.Internal;
+import org.gradle.api.tasks.UntrackedTask;
 import org.pkl.cli.CliEvaluator;
 import org.pkl.cli.CliEvaluatorOptions;
 
+@UntrackedTask(because = "Output file names are known only after execution")
```
StefMa, line 27: Maybe explain it a bit better here as well (that the output files contain placeholders)? 🤔

bioball: Done!

StefMa (approved):
LGTM 👌

translatenix:
I see no indication that `EvalTask` supports placeholders. It only accepts `RegularFileProperty getOutputFile()` and `DirectoryProperty getMultipleFileOutputDir()`, which clearly shouldn't contain placeholders.

StefMa:
You can read it up here:
https://pkl-lang.org/main/current/pkl-gradle/index.html#configuration-options



So yes it's a bit confusing that you can put a file into it that can contain a placeholder that doesn't work with Gradle because the file location is unknown at that point.

translatenix:
This is a design mistake in `EvalTask`. Should fix that instead of making the task untracked.

StefMa:
> This is a design mistake in `EvalTask`. Should try to fix that instead of making the task untracked.

If we would use a normal String as input it would work right? 🤔

translatenix:
I think `EvalTask` could have separate `Property<String>` properties and invalidate outputs if those are set.
But I see that `CliEvaluator` has `outputFiles` and `outputDirectories` properties, which might be all that's needed for accurate tracking.

bioball:
> But I see that CliEvaluator has outputFiles and outputDirectories properties, which might be all that's needed for accurate tracking.

Great idea! Updated the PR to add new properties that are tracked.

pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
```diff
@@ -16,50 +16,69 @@
...
 
 public abstract class EvalTask extends ModulesTask {
-  @OutputFile
-  @Optional
+  @Internal
   public abstract RegularFileProperty getOutputFile();
 
-  @Input
-  @Optional
+  @Internal
```
translatenix, line 36: Why is this not an input?

pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
```diff
@@ -16,50 +16,69 @@
...
   public abstract DirectoryProperty getMultipleFileOutputDir();
 
   @Input
   @Optional
   public abstract Property<String> getExpression();
 
+  @Internal
+  public Provider<CliEvaluator> getCliEvaluator() {
```
translatenix, line 51: I don't think `Provider` helps for this and the following methods, as they are only called if and when the information is needed.

pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
bioball: Missed this, updated!

pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
bioball: > Now the only question is if this works for a provider obtained with getProject().provider(). Another option is to inject a ProviderFactory into the task.

Hm... probably not, in that case. I'm guessing we need to reference the inputs.

StefMa: IIRC `getProject` would also fail if using configuration cache.
See https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution

translatenix: Good to know. I think that `@Inject`ing a `ProviderFactory` might be the right solution. Could check task dependency tracking with an automated or manual test.

translatenix: Related: https://discuss.gradle.org/t/auto-create-task-dependecies-in-case-of-computed-outputfile-provider-regularfile/45381

pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
```diff
@@ -39,27 +48,44 @@ public abstract class EvalTask extends ModulesTask {
...
+            getOutputFile().get().getAsFile().getAbsolutePath(),
+            getOutputFormat().get(),
+            getModuleOutputSeparator().get(),
+            mapAndGetOrNull(getMultipleFileOutputDir(), (it) -> it.getAsFile().getAbsolutePath()),
+            getExpression().get()));
+  }
+
+  @OutputFiles
+  @Optional
+  public SetProperty<File> getOutputPaths() {
```
translatenix, line 71: Is `getEffectiveOutputFiles()` a better name?

pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
```diff
@@ -39,27 +48,44 @@ public abstract class EvalTask extends ModulesTask {
...
+    var ret = getObjectFactory().setProperty(File.class);
+    ret.value(getCliEvaluator().getOutputFiles()).disallowChanges();
+    return ret;
+  }
+
+  @OutputDirectories
+  @Optional
+  public SetProperty<File> getMultipleFileOutputPaths() {
```
translatenix, line 79: Is `getEffectiveMultipleFileOutputDirs()` a better name?

translatenix (commented):
LGTM. Two more naming suggestions for your consideration.

pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
```diff
@@ -39,27 +48,44 @@ public abstract class EvalTask extends ModulesTask {
...
-
   @Input
   @Optional
   public abstract Property<String> getExpression();
 
+  @Inject
+  public abstract ObjectFactory getObjectFactory();
+
+  private CliEvaluator getCliEvaluator() {
+    return new CliEvaluator(
```
translatenix, line 59: Note that this results in multiple `CliEvaluator` instances, each of which recomputes `CliEvaluator.outputFiles` and `CliEvaluator.outputDirectories`.

netvl (commented):
There was a reason why we originally made `EvalTask` untracked.

Unfortunately, Gradle's model of input and output properties and their caching expects that all properties have _statically known values_ after configuration time. In other words, it is expected that Gradle knows exact values of all input and output properties of a task, in order to do its caching and invalidation.

This, in turn, implies that it is not possible (under the model constraints) to change or otherwise dynamically compute values of properties at the _task execution time_. As an example, this is not feasible:
```kotlin
class ExampleTask : DefaultTask() {
    @get:Output
    val someProperty: Property<String>

    @TaskAction
    fun doWork() {
        someProperty.set("xyz")
    }
}
```

To be absolutely precise, this _is_ possible to write, and it will kind of work as expected, however, it will not play nicely with caching and invalidation, because Gradle will read input and output property values before it executes the task and build its task state and cache based on those values, and it will not track any changes happening to properties during the task execution.

There are tasks in Gradle which may produce files whose names are unknown before execution, like compilation or code generation. However, in these cases these tasks' output properties are not individual files but directories, whose names are statically known.

In the `EvalTask` specifically, due to the nature of patterns, I originally assumed that the specific output file and directory names will not be known until evaluation. Additionally, because often you need to produce a file into a specific directory (for example, evaluate a `config.pkl` file into a `config.yml` file which must be present in the same directory), you cannot really declare the parent directory of the output pattern to be an "output directory", because it may interfere with other tasks and in general is not "controlled" by the eval task.

Because of this, I originally made this task untracked, because it ensures that the cache consistency is not violated, and that the task is invalidated at appropriate times.

Given the above, it is only realistically possible to track the outputs of this task in the following cases:
1. Computing output paths from a pattern does not, in fact, require Pkl evaluation. Looking at how `CliEvaluator` is used here, it seems like it might be the case. If this is true, then idiomatic description of this behavior would be different - I will describe it below.
2. We are okay with evaluating Pkl files at configuration time. I personally don't think this is a correct approach, and would vote for the task to remain untracked, because the whole point of splitting task configuration and evaluation is to avoid costly computations at configuration time and to have consistent inputs which do not depend on build script evaluation order. Still, if this is the case, I would still recommend to reimplement the logic as described below.

So, if we assume scenario 1, the core idea would be to build `CliEvaluator` as `Provider<CliEvaluator>` derived from other properties in the class, and then derive `Provider<File>` and `FileCollection` from that:
```java
private Provider<CliEvaluator> getCliEvaluator() {
    return getOutputFile().flatMap(outputFile ->
      getOutputFormat().flatMap(outputFormat ->
        ...new CliEvaluator(...)));
}

@OutputFile
public Property<File> getOutputFilePath() {
    // This logic should be changed to return only one file properly,
    // because there can be only one file
    return getCliEvaluator().map(evaluator ->
        evaluator.getOutputFiles().iterator().next()
    );
}

@OutputDirectories
public FileCollection getOutputDirectories() {
    return outputDirectoriesCollection;
}

private final ConfigurableFileCollection outputDirectoriesCollection = getProject().files();

{
    outputDirectoriesCollection.from(
        getCliEvaluator().map(evaluator ->
            evaluator.getOutputDirectories()
        )
    );
}
```

We can also cache the `CliEvaluator` instance, but this will require some extra code. Following an example in [here](https://github.com/gradle/gradle/issues/20820), we can make a utility function:

```java
public static <T> Provider<T> cache(ObjectFactory objects, Class<T> elementType, Provider<T> provider) {
    var property = objects.property(elementType);
    property.value(provider);
    property.disallowChanges();
    property.finalizeValueOnRead();
    return property;
}
```

Then, define the `CliEvaluator` property like this:

```java
private final Property<CliEvaluator> cliEvaluator = cache(
    getObjects(),
    CliEvaluator.class,
    getOutputFile().flatMap(outputFile ->
      getOutputFormat().flatMap(outputFormat ->
        ...new CliEvaluator(...)));
);
```
and use that as the source of output properties:

```java
@OutputFile
public Property<File> getOutputFilePath() {
    // This logic should be changed to return only one file properly,
    // because there can be only one file
    return cliEvaluator.map(evaluator ->
        evaluator.getOutputFiles().iterator().next()
    );
}

@OutputDirectories
public FileCollection getOutputDirectories() {
    return outputDirectoriesCollection;
}

private final ConfigurableFileCollection outputDirectoriesCollection = getProject().files();

{
    outputDirectoriesCollection.from(
        cliEvaluator.map(evaluator ->
            evaluator.getOutputDirectories()
        )
    );
}
```

I haven’t tested yet that this works fully correctly (I’m going to do it right now), and I do see that all of this is kind of clunky and more complicated than the original approach, but unfortunately that’s what is required here if we are to follow recommended Gradle practices.

<hr>

Actually I have just remembered another reason why it makes sense to keep the task untracked — Pkl evaluation can depend on external, non-file inputs, like environment variables. Thus it is possible for an evaluation task to become cached when it should not be, for example: the user changed an env var and re-run the task => Pkl script depends on the env var but there is no way for Gradle to track it => Pkl is not re-evaluated, old values are still used => the user is confused.

<hr>

As a side note — I would recommend not making every property `@Optional` by default. For example, properties like `outputFormat` always contain some default value from the spec, thus they should not really be optional.

pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
```diff
@@ -39,27 +48,44 @@ public abstract class EvalTask extends ModulesTask {
...
+            getExpression().get()));
+  }
+
+  @OutputFiles
+  @Optional
+  public SetProperty<File> getOutputPaths() {
+    var ret = getObjectFactory().setProperty(File.class);
+    ret.value(getCliEvaluator().getOutputFiles()).disallowChanges();
+    return ret;
+  }
```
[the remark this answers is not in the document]
translatenix: > Unfortunately, Gradle's model of input and output properties and their caching expects that all properties have statically known values after configuration time

This can't be true. Surely a task can take a `FileTree` as input whose files don't exist at configuration time?

> Pkl evaluation can depend on external, non-file inputs, like environment variables. 

I don't think that's an issue here because all such inputs are configured through the task (see `BasePklTask`, etc.).

Evaluating Pkl files at configuration time is not OK. But looking at `CliEvaluator`, I think that "computing output paths from a pattern does not, in fact, require Pkl evaluation" is correct.

netvl:
Btw, let me correct myself - it seems I forgot the code base a bit - of course there can be multiple output files with single output pattern; then the logic would look like this:
```kotlin
  @OutputFiles
  @Optional
  public Provider<Set<File>> getEffectiveOutputFilePaths() {
      return cliEvaluator.map(CliEvaluator::getOutputFiles);
  }
  
  @OutputDirectories
  @Optional
  public Provider<Set<File>> getOutputDirectories() {
      return cliEvaluator.map(CliEvaluator::getOutputDirectories);
  }
```

I also switched to `Provider<Set<File>>` instead of `FileCollection`, because unfortunately file collections are more cumbersome to create than providers, especially if empty providers is a possibility.

pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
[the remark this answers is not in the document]
netvl: > Surely a task can take a FileTree as input whose files don't exist at configuration time?

Yes, it most definitely can, but it does not contradict what I'm saying. I might not have expressed it fully clearly, but essentially what I wanted to say is that in Gradle model, file collections must be fully configured before the task execution time. It does not mean that their contents can't be dynamic - they most certainly can, like class files produced by compilation tasks, and they can be absent before some task is executed - but that their configuration can't be dynamic, i.e. you cannot add new sources to a file collection or a file tree at task execution time.

When I was implementing the first version of the evaluation task, I assumed that it is not possible to determine exact file names and directory names from patterns without evaluation, i.e. without executing the task logic first. Because patterns are flexible, it would've been impossible to statically configure a `FileCollection` representing output files/directories, and therefore I had to make the tasks untracked.

Since apparently it _is_ possible to resolve patterns and determine outputs without running evaluation first, therefore, it should be possible to determine the full set of output files/directories at configuration time, and therefore we can make the task tracked.

> I don't think that's an issue here because all such inputs are configured through the task (see superclasses of EvalTask).

Fair point, forgot that these items were there. Yes, in this case evaluation should be self-contained and thus cached.

netvl (approved):
LGTM!

pkl-gradle/src/test/kotlin/org/pkl/gradle/EvaluatorsTest.kt
```diff
@@ -501,7 +501,7 @@ class EvaluatorsTest : AbstractTest() {
       }
 
       val printEvalFiles by tasks.registering {
-        inputs.files(doEval)
+        inputs.files(doEval.map { it.effectiveOutputFiles })
```
netvl, line 501: This should not be necessary - it is idiomatic to depend on a task rather than on a specific output property, and also it is not really possible for these properties to be non-empty at the same time, so there is no ambiguity.

pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
```diff
@@ -67,17 +67,18 @@ public abstract class EvalTask extends ModulesTask {
...
+  public FileCollection getEffectiveOutputFiles() {
     return getObjects()
         .fileCollection()
         .from(cliEvaluator.map(e -> nullToEmpty(e.getOutputFiles())));
   }
 
   @OutputDirectories
   @Optional
-  public FileCollection getOutputDirectories() {
+  public FileCollection getEffectiveOutputDirectories() {
```
netvl, line 81: Minor - I believe in all of the plugins we use `getSomeDir()` for directory properties, not `getSomeDirectory()`, so it might make sense to name this `getEffectiveOutputDirs()`. WDYT?

stackoverflow (approved):

pkl-gradle/src/main/java/org/pkl/gradle/task/EvalTask.java
holzensp: Non-blocking, but I agree with @netvl.

holzensp (approved):

pkl-gradle/src/test/kotlin/org/pkl/gradle/EvaluatorsTest.kt
bioball: Fair; will undo this change

PR #409, opened 2024-04-08
Enable caching of native image building and test tasks

stefanobaghino:
This pull requests enables caching for the native image building task and improves the caching effectiveness of some of the test tasks in the project.
1. the native image building is a generic `Exec` tasks for which caching needs to be enabled explicitly (see [docs](https://docs.gradle.org/current/userguide/build_cache.html#using_the_runtime_api)) as the build tool doesn't have natively have enough information to cache it effectively -- furthermore some steps need to be taken to normalize the classpath (JARs come with an embedded timestamp in their manifest and this needs to be ignored), as well as not rely on absolute paths
2. while tests are usually cacheable, path relativity needs to be added as a property to ensure the results can be retrieved from a cache without relying on the environment of the machine running the task

All touched properties got a name to make sure it was easy to find them with [Build Scan®](https://scans.gradle.com/).

These issues have been diagnosed using [the build validation scripts by Gradle](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/bf70ec7e5e57ab0996d68a970745512b79462134/Gradle.md).

### DIFF
--- a/pkl-cli/pkl-cli.gradle.kts
+++ b/pkl-cli/pkl-cli.gradle.kts
@@ -138,9 +138,10 @@ fun Exec.configureExecutable(isEnabled: Boolean, outputFile: Provider<RegularFil
   enabled = isEnabled
   dependsOn(":installGraalVm")
 
-  inputs.files(sourceSets.main.map { it.output })
-  inputs.files(configurations.runtimeClasspath)
+  inputs.files(sourceSets.main.map { it.output }).withPropertyName("mainSourceSets").withPathSensitivity(PathSensitivity.RELATIVE)
+  inputs.files(configurations.runtimeClasspath).withPropertyName("runtimeClasspath").withNormalizer(ClasspathNormalizer::class)
   outputs.file(outputFile)
+  outputs.cacheIf { true }
 
   workingDir(outputFile.map { it.asFile.parentFile })
   executable = "${buildInfo.graalVm.baseDir}/bin/native-image"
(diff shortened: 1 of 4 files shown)


holzensp (commented):
LGTM; wanting a second pair of eyes; @bioball 

No idea why CircleCI errored out on approval, but appears stuck for me.

ribafish (approved):

stefanobaghino:
@holzensp Thanks for taking the time to review, I'll wait for @bioball's second look. 🙇🏻

stefanobaghino:
👋🏻 Hi @bioball! Can this tiny PR get your attention in the coming days? 🙂

pkl-cli/pkl-cli.gradle.kts
```diff
@@ -138,9 +138,10 @@ fun Exec.configureExecutable(isEnabled: Boolean, outputFile: Provider<RegularFil
   enabled = isEnabled
   dependsOn(":installGraalVm")
 
-  inputs.files(sourceSets.main.map { it.output })
-  inputs.files(configurations.runtimeClasspath)
+  inputs.files(sourceSets.main.map { it.output }).withPropertyName("mainSourceSets").withPathSensitivity(PathSensitivity.RELATIVE)
+  inputs.files(configurations.runtimeClasspath).withPropertyName("runtimeClasspath").withNormalizer(ClasspathNormalizer::class)
```
bioball, line 142: We should provide `bin/native-image` as an input here as well, right?

```suggestion
  inputs.files(configurations.runtimeClasspath).withPropertyName("runtimeClasspath").withNormalizer(ClasspathNormalizer::class)
inputs.files(file(buildInfo.graalVm.baseDir).resolve("bin/native-image")).withPropertyName("graalVmNativeImage").withPathSensitivity(PathSensitivity.ABSOLUTE)
```

bioball (changes requested):

stefanobaghino:
Thank you for taking the time to review! 🙂

PR #413, opened 2024-04-09
Fix the sequence length in the YAML parser

luuvish:
The initialSize in EconomicsMaps.create() is not the actual stored size, hence the accurate size is not determined until ObjectMembers are added after addMembers() called.

Since the size set during the creation of VmListing cannot be changed afterwards, the size of node.getValue() is used instead of members' size.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/stdlib/yaml/ParserNodes.java
+++ b/pkl-core/src/main/java/org/pkl/core/stdlib/yaml/ParserNodes.java
@@ -345,14 +345,15 @@ public final class ParserNodes {
       @Override
       public VmListing construct(Node node) {
         var sequenceNode = (SequenceNode) node;
-        var members = EconomicMaps.<Object, ObjectMember>create(sequenceNode.getValue().size());
+        var size = sequenceNode.getValue().size();
+        var members = EconomicMaps.<Object, ObjectMember>create(size);
 
         var result =
             new VmListing(
                 VmUtils.createEmptyMaterializedFrame(),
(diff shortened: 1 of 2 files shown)


holzensp (changes requested):
The test `pkl-core/src/test/files/LanguageSnippetTests/input/api/yamlParser3.pkl` fails with this change. Run `gw build` to reproduce.

luuvish:
in yamlParser3.pkl,

```yaml
product:
    - sku: BL3940
      quantify: 4
      description: Basketball
      price: 450.00 
    - sku: BL4438H
      quantify: 1
      description: Super Hoop
      price: 2392.00
```
it's result is
```yaml
product {
    "other"
    new {
      sku = "BL4438H"
      quantify = 1
      description = "Super Hoop"
      price = "2392.00"
    }
}
```

but, correct result is
```yaml
product {
    new {
      sku = "BL3940"
      quantify = 4
      description = "Basketball"
      price = 450.00
    }
    new {
      sku = "BL4438H"
      quantify = 1
      description = "Super Hoop"
      price = "2392.00"
    }
    "other"
}
```

The length of the sequence was incorrectly set to 0, causing "other" to be inserted at index 0 and overwriting the contents of index 0.

holzensp:
I meant the file describing the expected result (`pkl-core/src/test/files/LanguageSnippetTests/output/api/yamlParser3.pcf`) was not updated yet. I've done that for you now.

Btw, totally skipped over saying "nice catch," but it is, so... nice catch! ;)

holzensp (approved):

PR #416, opened 2024-04-10
Enable useCustomStringDelimiters for pkl test example output

lilyball:
This should make the `<file>-expected.pcf` file a bit easier to read when examples contain quotes or backslashes.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/runtime/TestRunner.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/TestRunner.java
@@ -263,7 +263,7 @@ public class TestRunner {
                 VmUtils.createSyntheticObjectProperty(Identifier.EXAMPLES, "examples", examples)),
             0);
     var builder = new StringBuilder();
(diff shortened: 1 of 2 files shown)


lilyball:
There are no tests for this right now as I wasn't sure where to find any existing test coverage for this, and I don't actually know Java or Kotlin myself. If someone wants to point me to where existing tests might be located that would be helpful.

bioball:
You'll want to add tests in pkl-core/src/test/kotlin/org/pkl/core/EvaluateTestsTest.kt.

If it's all a little too unfamiliar, we can take it over and bring it to the finish line!

lilyball:
Ok, I took a look at that file with the thought that maybe I could copy and modify an existing test. With that approach I could perhaps write a test that puts custom string delimiters in the expected file and ensures that parses fine, but since I don't know Kotlin I don't know how I'd do the more useful test, of ensuring that writing out the example uses custom string delimiters where expected. So given that, I would very much appreciate if you could take it over and add the tests yourself (This PR is already set to allow edits by maintainers). Thank you!

bioball (approved):

PR #426, opened 2024-04-12
Fix pkl.reflect#typedType to mirror the right type

lilyball:


bioball (approved):
Yikes, good catch. Thanks!

PR #428, opened 2024-04-13
Cannot assign property with fixed/const in spread

luuvish:
When using spread syntax with typed object, properties that are fixed or const cannot be assigned or amended.
This should be checked in the checkTypedProperty() function.


bioball (approved):
Great catch, thanks!

PR #432, opened 2024-04-16
Suppress warnings for `System.getProperty("line.separator")`

KushalP:
`class System` gets initialized at build time through `native-image` and added to the heap. We initialize everything at build time via the `--initialize-at-build-time=` flag.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
+++ b/pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
@@ -175,6 +175,7 @@ public final class IoUtils {
   }
 
   // not stored to avoid build-time initialization by native-image
+  @SuppressWarnings("SystemGetProperty")
   public static String getLineSeparator() {
     return System.getProperty("line.separator");
   }


pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
```diff
@@ -176,7 +176,7 @@ public static Path getDefaultModuleCacheDir() {
 
   // not stored to avoid build-time initialization by native-image
   public static String getLineSeparator() {
-    return System.getProperty("line.separator");
+    return System.lineSeparator();
```
bioball, line 179: Per comment on line 177, this is likely problematic because `System` gets initialized at build time and added to the heap (we initialize everything at build time via the `--initialize-at-build-time=` flag).

bioball (changes requested):
Per comment on line 177, this is likely problematic because class System gets initialized at build time and added to the heap (we initialize everything at build time via the --initialize-at-build-time= flag).

There's a chance I'm wrong here, but we need a Windows executable first so we can test it out.

Alternatively, I'd be happy to accept a PR that adds @SuppressWarnings("SystemGetProperty") to this method.

KushalP:
> Alternatively, I'd be happy to accept a PR that adds @SuppressWarnings("SystemGetProperty") to this method.

Updated the PR to do this. Have updated the PR title and description as well.

bioball (approved):
LGTM

PR #434, opened 2024-04-17
Fix IntelliJ scope definition

translatenix:


bioball (approved):
Ah, an artifact from times gone by. Thanks!

PR #449, opened 2024-04-24
Fix globbed read/import bugs

translatenix:
This PR fixes several bugs related to globbed reads/imports and makes code improvements along the way.

Bug fixes:
* Fix caching of globbed reads
  * Glob pattern isn't guaranteed to be a constant
  * Both glob resolution and `ResolvedGlobElement.getPath()` depend on the current module URI
* Fix `ClassCastException` when reflecting on globbed import
* Avoid creating an unbounded number of Truffle root nodes per globbed read/import

See commit messages for further details.


pkl-core/src/main/java/org/pkl/core/runtime/VmMap.java
```diff
@@ -219,12 +225,39 @@ public void force(boolean allowUndefinedValues) {
...
+      assert value != null;
+      if (key instanceof String) {
+        builder.addProperty(Identifier.get((String) key), value);
+      } else {
+        builder.addEntry(key, value);
+      }
+    }
+    return builder.toDynamic();
+  }
+
```
bioball, line 253: This is a replacement for code that was visible to Truffle inlining, so it shouldn't be then moved behind a `TruffleBoundary`.

This works:

```suggestion
  public VmMapping toMapping() {
    var builder = new VmObjectBuilder(getLength());
    for (var entry : this) {
      builder.addEntry(VmUtils.getKey(entry), VmUtils.getValue(entry));
    }
    return builder.toMapping();
  }

  public VmDynamic toDynamic() {
    var builder = new VmObjectBuilder(getLength());
    for (var entry : this) {
      var key = VmUtils.getKey(entry);
      var value = VmUtils.getValue(entry);
      if (key instanceof String) {
        builder.addProperty(Identifier.get((String) key), value);
      } else {
        builder.addEntry(key, value);
      }
    }
    return builder.toDynamic();
  }

```

pkl-core/src/main/java/org/pkl/core/stdlib/base/MapNodes.java
```diff
@@ -287,19 +287,7 @@ protected VmTyped eval(VmMap self, VmClass clazz) {
...
-            members,
-            VmUtils.getKey(entry),
-            VmUtils.createSyntheticObjectEntry("", VmUtils.getValue(entry)));
-      }
-
-      return new VmMapping(
-          VmUtils.createEmptyMaterializedFrame(),
-          BaseModule.getMappingClass().getPrototype(),
-          members);
+      return self.toMapping();
```
bioball, line 290: Should change `class toDynamic` to also call out to `self.toDynamic()` yeah?

pkl-core/src/main/java/org/pkl/core/util/GlobResolver.java
```diff
@@ -463,15 +465,15 @@ private static Pair<String, String[]> splitGlobPatternIntoBaseAndWildcards(
    * <p>Each pair is the expanded form of the glob pattern, paired with its resolved absolute URI.
    */
   @TruffleBoundary
-  public static List<ResolvedGlobElement> resolveGlob(
+  public static Map<String, ResolvedGlobElement> resolveGlob(
```
bioball, line 468: Why did this change from `List` to `Map`?

bioball (changes requested):
Overall, great work!!

I have some minor blocking comments, but overall this is a great improvement. Thanks so much for your PR.

bioball:
Hi @translatenix, are you still planning on working on this PR, and the other ones? If not, we'd be happy to take it past the finish line (and credit you, of course)

pkl-core/src/main/java/org/pkl/core/util/GlobResolver.java
bioball: Never mind, I see; this gets stored directly in extra storage.

bioball:
FYI @translatenix given the lack of activity here, I'll go ahead and update this PR with my comments, and do the same for #455, #378 and #197. Thanks again for your contributions!

bioball (approved):

PR #452, opened 2024-04-25
Don't install GraalVM for an architecture that can't be built 

### LINKED ISSUE #450: `gw buildNative` runs aarch64 tasks on amd64
On amd64 WSL/Ubuntu, `gw buildNative` now additionally runs `:downloadGraalVmAarch64` (which succeeds) and `:installGraalVmAarch64` (which fails). Neither of these tasks should be run.

bioball:
Fixes an issue where native builds will try to install GraalVM for the wrong architecture.

Closes #450

### DIFF
--- a/pkl-cli/pkl-cli.gradle.kts
+++ b/pkl-cli/pkl-cli.gradle.kts
@@ -135,13 +135,10 @@ tasks.check {
 }
 
 fun Exec.configureExecutable(
-  isEnabled: Boolean,
   graalVm: BuildInfo.GraalVm,
   outputFile: Provider<RegularFile>,
   extraArgs: List<String> = listOf()
 ) {
-  enabled = isEnabled
-
   inputs.files(sourceSets.main.map { it.output }).withPropertyName("mainSourceSets").withPathSensitivity(PathSensitivity.RELATIVE)
   inputs.files(configurations.runtimeClasspath).withPropertyName("runtimeClasspath").withNormalizer(ClasspathNormalizer::class)
   inputs.files(file(graalVm.baseDir).resolve("bin/native-image")).withPropertyName("graalVmNativeImage").withPathSensitivity(PathSensitivity.ABSOLUTE)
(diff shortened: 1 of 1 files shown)


pkl-cli/pkl-cli.gradle.kts
```diff
@@ -218,39 +215,48 @@ fun Exec.configureExecutable(
...
  */
 val macExecutableAmd64: TaskProvider<Exec> by tasks.registering(Exec::class) {
-  dependsOn(":installGraalVmAmd64")
-  configureExecutable(
-    buildInfo.os.isMacOsX,
-    buildInfo.graalVmAmd64,
-    layout.buildDirectory.file("executable/pkl-macos-amd64")
-  )
+  if (buildInfo.os.isMacOsX) {
+    dependsOn(":installGraalVmAmd64")
```
bioball, line 219: Curious if there's a better way to describe a conditional task dependency than this?

translatenix: Why is this dependency conditional?

pkl-cli/pkl-cli.gradle.kts
```diff
@@ -218,39 +215,48 @@ fun Exec.configureExecutable(
...
+    configureExecutable(
+      buildInfo.graalVmAmd64,
+      layout.buildDirectory.file("executable/pkl-macos-amd64")
+    )
+  } else {
+    enabled = false
+  }
```
holzensp, line 226: Shouldn't this just be
```suggestion
  enabled = buildInfo.os.isMacOsX
  dependsOn(":installGraalVmAmd64")
  configureExecutable(
    buildInfo.graalVmAmd64,
    layout.buildDirectory.file("executable/pkl-macos-amd64")
  )
```

? [throughout]

Surely, if `!enabled`, tasks this one `dependsOn` are not executed.

holzensp (approved):
Non-blocking suggestion for simplification.

pkl-cli/pkl-cli.gradle.kts
bioball: Without the `if`, I am seeing that `installGraalVmAmd64` gets executed on every OS/arch.

translatenix: Sounds like something else is wrong with the build script. The `if` doesn’t make sense to me.

bioball: Task `buildNative` depends on `macExecutableAmd64`, `macExecutableAarch64`, etc, etc.

I am seeing that these tasks themselves don't run if `enabled = false`, but its dependent tasks still run, unless I _also_ don't call `dependsOn` at all.

PR #453, opened 2024-04-25
Switch to adapter pattern

bioball:
This switches the LSP implementation to use the adapter pattern, deferring as much as possible to the ANTLR-generated classes.

It also gets rid of `CstBuilder`, instead moving diagnostic handling to the analyzers.

It's not complete but provides a skeleton to be filled in.


translatenix:
Have you considered having a separate repository for the language server, similar to pkl-Intellij? A language server is a large project, quite different from the rest of pkl/pkl, should be releasable on its own, and pkl/pkl already has some modularity issues. I’m also worried that this will slow down the build and especially the IntelliJ experience, which is already barely tolerable on my (not so old) laptop.

bioball:
We've thought about it. We're still a ways off from releasing the LSP. At the same time, putting it here is useful for us to share things with the rest of Pkl. We'll see.

> I’m also worried that this will slow down the build and especially the IntelliJ experience, which is already barely tolerable on my (not so old) laptop

Is this because of WSL? Might be better once we have Windows support available for you. I haven't noticed any perf issues on macOS yet.

PR #455, opened 2024-04-25
Improve configuration of native image compilation

translatenix:
- Solve msgpack issue with `--initialize-at-run-time`.
- Use quick build mode for 40% faster compilation and 20% smaller executable (tested on Linux amd64). If necessary, this can be disabled for release builds. Note that with the move from GraalVM CE 22 to Oracle GraalVM 23, native image compilation time has doubled and binary size has increased to 120 MiB on Linux amd64.
- Use Intel Skylake (2015) CPU features for amd64 executables.
- Remove options that are commented out.


pkl-cli/pkl-cli.gradle.kts
```diff
@@ -160,47 +160,29 @@ fun Exec.configureExecutable(
         // that the "initialize everything at build time" *CLI* option is likely here to stay
         "--initialize-at-build-time="
         // needed for messagepack-java (see https://github.com/msgpack/msgpack-java/issues/600)
-        ,"--add-opens=java.base/java.nio=ALL-UNNAMED"
-        ,"--add-opens=java.base/sun.nio.ch=ALL-UNNAMED"
+        ,"--initialize-at-run-time=org.msgpack.core.buffer.DirectBufferAccess"
```
bioball, line 163: Have you tested this? I think this would just move our problem from build time to run time.

pkl-cli/pkl-cli.gradle.kts
```diff
@@ -160,47 +160,29 @@ fun Exec.configureExecutable(
...
-        //,"-H:DashboardDump=dashboard.dump", "-H:+DashboardAll"
+        // disable automatic support for JVM CLI options (puts our main class in full control of argument parsing)
+        ,"-H:-ParseRuntimeOptions"
+        // quick build mode: 40% faster compilation, 20% smaller (but presumably also slower) executable
+        ,"-Ob"
+        // Use Intel Skylake (2015) CPU features, which should also be supported by AMD ZEN (2017) CPUs.
+        // Tested on ZEN 2.
+        // Feature list: CX8, CMOV, FXSR, MMX, SSE, SSE2, SSE3, SSSE3, SSE4_1, SSE4_2, POPCNT,
+        // LZCNT, AVX, AVX2, AES, CLMUL, BMI1, BMI2, FMA, AMD_3DNOW_PREFETCH, ADX, FLUSHOPT
+        , if (buildInfo.arch == "amd64") "-march=skylake" else ""
```
bioball, line 185: What does this do? Does this mean native executables are not compatible with Intel CPUs older than 2015, and AMD CPUs older than 2017?

bioball (changes requested):

pkl-cli/pkl-cli.gradle.kts
translatenix: The only way to instantiate a `DirectByteBuffer` is `ByteBuffer.allocateDirect()`, which isn't used in the Pkl codebase or its dependencies. That's why moving this to run time is a cleaner workaround than opening JDK internals.
Is msgpack exercised in native tests? If so, this change should be very safe.

> If DirectByteBuffer is not used, msgpack-java 0.9.3 will work even if these two options are not set.

pkl-cli/pkl-cli.gradle.kts
translatenix: Yes, this raises the baseline for supported CPUs from "x64-v3" (default) to "skylake". In return, the native image compiler can use many additional CPU features/instructions, resulting in faster and more efficient assembly code. 
I arrived at this by checking the output of `-march=list`. "skylake" seemed to be a good middle ground between supporting a wide range of CPUs and leveraging many CPU features.

https://www.graalvm.org/latest/reference-manual/native-image/overview/BuildOutput/#recommendation-cpu

pkl-cli/pkl-cli.gradle.kts
```diff
@@ -160,47 +160,29 @@ fun Exec.configureExecutable(
...
-        //,"-J-Dtruffle.TruffleRuntime=com.oracle.truffle.api.impl.DefaultTruffleRuntime"
-        //,"-J-Dcom.oracle.truffle.aot=true"
-        //,"-J:-ea"
-        //,"-J:-esa"
-        // for use with https://www.graalvm.org/docs/tools/dashboard/
-        //,"-H:DashboardDump=dashboard.dump", "-H:+DashboardAll"
+        // disable automatic support for JVM CLI options (puts our main class in full control of argument parsing)
+        ,"-H:-ParseRuntimeOptions"
+        // quick build mode: 40% faster compilation, 20% smaller (but presumably also slower) executable
+        ,"-Ob"
```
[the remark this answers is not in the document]
translatenix: Done. Note that binary size keeps increasing (now at 120 MiB for amd64). I think it's worth trying to switch to `--initialize-at-runtime` for most things other than the Truffle interpreter. I also tried switching to the G1 garbage collector, but this added another 20 MiB.

pkl-cli/pkl-cli.gradle.kts
bioball: These CLIs are libraries, and have a broad set of use-cases. For that reason, we should probably be conservative here. I'm thinking we should hold off on this change.

CC @holzensp @stackoverflow for comments

pkl-cli/pkl-cli.gradle.kts
bioball: Not right now. Our native tests just use our regular evaluator, and don't exercise the message passing stuff.

bioball: Looking through msgpack-java, I'm relatively certain that this would end up throwing at runtime: https://github.com/msgpack/msgpack-java/blob/141f66dd615e797c96113169371d1767b251a0ae/msgpack-core/src/main/java/org/msgpack/core/buffer/DirectBufferAccess.java#L286-L289

pkl-cli/pkl-cli.gradle.kts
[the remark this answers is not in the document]
bioball: In addition to Truffle, we also want to initialize the standard library at build time, to cut down the overhead of Pkl execution.

There's certainly some tuning that we can do here, though.

translatenix:
> How much perf do we gain from targeting skylake? Can you enable the native CLI builds to CI so we can compare it to the default architecture? Or, if you're on an x86 machine, feel free to just run some tests locally and report the results.

Does Pkl have a comprehensive benchmark suite? If you want proof, that’s probably the only way.

> These CLIs are libraries, and have a broad set of use-cases. For that reason, I'm thinking that we should probably be conservative here. We probably shouldn't just bump the version, but if there's a significant perf gain, we can consider adding this as an additional architecture.

Your call. I feel that “skylake” is fairly conservative yet enables most CPU features. “x64-v3” is an arbitrary GraalVM default—we should figure out what’s right for Pkl. Also, this is easy to undo depending on feedback.

bioball:
> Does Pkl have a comprehensive benchmark suite? If you want proof, that’s probably the only way.

Not yet. We have some very lightweight benchmark tests, but we don't have an environment (yet) that can run workloads with bare metal isolation. Until then, the best we can do is run benchmarks locally the compare.

translatenix:
> Until then, the best we can do is run benchmarks locally then compare

But what benchmarks do you run? `./gradlew pkl-core:testLinuxExecutableAmd64` doesn’t strike me as a meaningful benchmark.

bioball:
There is an in-language [stdlib module](https://pkl-lang.org/package-docs/pkl/current/Benchmark/index.html) that we use to write on-off benchmarks when we want to test specific things when iterating locally.

There's also the `bench:jmh` task, although we haven't been using that nearly as much to get any signal.

The suggestion of running `testLinuxExecutableAmd64` is just a coarse-grained way to get some signal, but definitely not a real benchmark.

bioball:
Here's a quick benchmark that might be a good starting point for comparison:

```groovy
amends "pkl:Benchmark"

import "package://pkg.pkl-lang.org/pkl-pantry/org.json_schema.contrib@1.0.5#/internal/ModulesGenerator.pkl"
import "package://pkg.pkl-lang.org/pkl-pantry/org.json_schema@1.0.2#/Parser.pkl"
import "package://pkg.pkl-lang.org/pkl-pantry/pkl.experimental.uri@1.0.1#/URI.pkl"

local githubActionJson = read("https://json.schemastore.org/github-action.json")

microbenchmarks {
  ["json schema generator"] {
    expression = 
      let (parsed = Parser.parse(githubActionJson))
        new ModulesGenerator {
          rootSchema = parsed
          baseUri = URI.parse("file:///foo")!!
        }
          .modules
          .map((it) -> it.moduleNode.output.text)
  }
}
```

pkl-cli/pkl-cli.gradle.kts
stackoverflow: Agree with Dan that we should be conservative here, even though skylake is almost 9 years old. Specially without any performance data to compare. If the performance bump is noticeable we can create a new architecture release or just drop old CPUs in the new version.

.circleci/config.yml
```diff
@@ -753,6 +753,9 @@ workflows:
     - gradle-check-jdk21:
         requires:
         - hold
+    - pkl-cli-linux-alpine-amd64-snapshot:
+        requires:
+        - hold
```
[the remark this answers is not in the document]
translatenix: Good catch. This commit belongs to a different PR, and I don't know how it ended up here. Removed.

translatenix:
> Here's a quick benchmark that might be a good starting point for comparison.

On my laptop, the variance of this benchmark is too high to draw any conclusions about`-march` settings.
However, I can tell that quick build mode significantly degrades performance, which isn't surprising.

`-march=x86-64-v3` requires Intel Haswell (2013) or AMD Excavator (2015), which isn't far off from `-march=skylake`.
To move forward with this PR, I'll remove `-march-skylake`.

pkl-cli/pkl-cli.gradle.kts
holzensp: I agree with @bioball and @stackoverflow. We're also considering spending a version-cycle on GraalVM/Truffle/native-image updates and perf (including binary size). For now, let's be quite conservative.

bioball (changes requested):
Sounds good, thanks for running that benchmark for us!

The one remaining request is to undo the change to `--add-opens` here.

translatenix:
> The one remaining request is to undo the change to --add-opens here.

I’m still working on that. I’m not yet convinced that --add-opens is required.

translatenix:
Ready to review from my side.

The "Improve BinaryEvaluatorSnippetTests(Engine)" refactoring isn't essential because I ultimately added `NativeServerTest` instead of `NativeBinaryEvaluatorSnippetTests`. But I think it's a worthwhile improvement regardless.

bioball:
Look like you're right; `--add-opens` is not needed :D

pkl-commons-test/src/main/kotlin/org/pkl/commons/test/AbstractSnippetTestsEngine.kt
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  protected fun String.stripFilePaths() = replace(snippetsDir.toString(), "/\$snippetsDir")
+
+  protected fun String.stripLineNumbers() =
+    replace(lineNumberRegex) { result ->
+      // replace line number with equivalent number of 'x' characters to keep formatting intact
+      (result.groups[1]!!.value) + "x".repeat(result.groups[3]!!.value.length) + " |"
+    }
+}
```
bioball, line 69: I don't know if this change improves our code; this moves some fields that are only relevant to the language snippet test engine (e.g. the package server, and also `selection`, which is a quality of life thing when working on snippet tests for pkl-core.

I think let's keep it the way it was. Also, we might move the binary evaluator into pkl-core, which means the binary evaluator snippet tests will change then, making this abstraction for naught.

pkl-commons-test/src/main/kotlin/org/pkl/commons/test/PklExecutablePaths.kt
```diff
@@ -0,0 +1,40 @@
...
+  val alpineAmd64: Path = executablePath("pkl-alpine-linux-amd64")
+
+  // order (aarch64 before amd64, linux before alpine) affects [firstExisting]
+  val all: List<Path> = listOf(macAarch64, macAmd64, linuxAarch64, linuxAmd64, alpineAmd64)
+
+  val existing: List<Path>
+    get() = all.filter(Files::exists)
+
+  val firstExisting: Path
+    get() = existing.first()
```
bioball, line 36:
```suggestion
  val firstExisting: Path
    get() {
      require(existing.isNotEmpty()) {
        "Native executable not found on system. Create one with `./gradlew buildNative`."
      }
      return existing.first()
    }
```

pkl-server/src/test/kotlin/org/pkl/server/NativeServerTest.kt
```diff
@@ -0,0 +1,39 @@
...
+    client = TestTransport(MessageTransports.stream(server.inputStream, server.outputStream))
+    executor.execute { client.start() }
+  }
+
+  @AfterEach
+  fun afterEach() {
+    client.close()
+    server.destroy()
+  }
+}
```
bioball, line 39: Very nice! Great addition!

bioball (changes requested):
Great work!

I think let's keep the language snippet stuff the way it is right now (see comment). But the other changes here look good to me.

pkl-commons-test/src/main/kotlin/org/pkl/commons/test/AbstractSnippetTestsEngine.kt
translatenix: It fixes dozens of IntelliJ warnings and eliminates code duplication. PackageServer is defined in commons-test and only started if needed. And yes, binary evaluator belongs into pkl-core.

bioball: Let's just suppress those warnings for now, then; once we move it into pkl-core, this will all change again, and we'd probably end up removing this abstraction layer if it existed.

translatenix: So you’re saying that I should keep the changes that suppress the warnings and discard the remaining changes, even though they remove significant code duplication and will make it easier to move the tests into pkl-core?

pkl-commons-test/src/main/kotlin/org/pkl/commons/test/PklExecutablePaths.kt
translatenix: added a slightly modified version

pkl-server/src/test/kotlin/org/pkl/server/NativeServerTest.kt
translatenix: I just hope that `nativeTest.dependsOn(":pkl-cli:assembleNative")` is OK from a build cache perspective. (`assembleNative` is a lifecycle task that doesn't declare outputs.)

pkl-commons-test/src/main/kotlin/org/pkl/commons/test/AbstractSnippetTestsEngine.kt
bioball: The abstraction might create more work for us later on if we eliminate `BinaryEvaluatorSnippetTests` later on if we move the binary evaluator into pkl-core.

Also, this would give us: `LanguageSnippetTestsEngine` -> `AbstractLanguageSnippetTestsEngine` -> `AbstractSnippetTestsEngine` -> `InputOutputTestEngine` -> `HierarchicalTestEngine`. That's just a lot of complexity that adds cognitive burden when trying to understand our test code.

It also moves properties that either shouldn't be moved, or don't have any meaning for the binary evaluator snippet test (`val selection: String = ""` is a property that exists for the purpose of working on the snippet tests in pkl-core).

To reduce the friction, I'd be willing to accept this PR if we moved these members into `AbstractLanguageSnippetTestsEngine`:

* `packageServer`
* `selection`
* `includedTests`
* `excludedTests`
* `Path.getProjectDir()`

stackoverflow (approved):
Looks good. Just some minor comments.

PR #458, opened 2024-04-27
Fix IntelliJ Java warnings, make classes final if possible

translatenix:
Fix all IntelliJ warnings in pkl-core production code except for bogus spelling warnings and warnings about unused public methods. I didn't delete any unused public method because it's hard to tell for sure which packages are public APIs.

Changes made:
- use text block instead of string concatenation
- extract method to avoid code duplication
- use switch expression
- fix Javadoc syntax and spelling
- fix spelling in comment
- increase class visibility to match visibility of use site
- delete overriding method with same implementation
- use String.isEmpty() and StringBuilder.isEmpty()
- add `@Serial` annotation
- make field final
- remove unused field
- remove unused private method
- remove exceptions that aren't thrown from throws clause
- insert non-null assertion
- annotate overriding method with `@Nonnull`
- suppress warning
- delete unused class (WriteAuxiliarySlotNode)


pkl-core/src/main/java/org/pkl/core/RendererException.java
```diff
@@ -15,7 +15,7 @@
  */
 package org.pkl.core;
 
-public class RendererException extends RuntimeException {
+public final class RendererException extends RuntimeException {
```
stackoverflow, line 18: Is IntelliJ complaining about this, or is it an unrelated change?

holzensp:
Overall, LGTM... Why does _my_ IntelliJ not warn about lacking / suggest adding `final` modifiers to classes?

translatenix:
> Why does my IntelliJ not warn about lacking / suggest adding final modifiers to classes?

That commit doesn’t fix IntelliJ warnings (nor does it claim to). It fixes gaps in making classes final, a policy that is generally adhered to in the codebase.

pkl-gradle/src/main/java/org/pkl/gradle/PklPlugin.java
```diff
@@ -398,6 +398,7 @@ private void configureIdeaModule(CodeGenSpec spec) {
               var outputDir = spec.getOutputDir().get().getAsFile();
               module.getGeneratedSourceDirs().add(outputDir);
               if (spec.getSourceSet().get().getName().toLowerCase().contains("test")) {
+                //noinspection deprecation
```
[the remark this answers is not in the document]
translatenix: The comment says “This part of logic can be removed once we stop supporting Kotlin plugin with version less than 1.8.x.”. Do we really need to leave warnings in the code?

pkl-core/src/main/java/org/pkl/core/RendererException.java
translatenix: Unlike the other commits, this one doesn’t address IntelliJ warnings (see commit messages). But it also falls into the “cleanup” category: Generally, classes are final if possible, but some aren’t.

stackoverflow (approved):
LGTM. You need to fix some formatting problems to get a green build, though.

translatenix:
Squashed the "fix IntelliJ warnings" commits.

pkl-gradle/src/main/java/org/pkl/gradle/PklPlugin.java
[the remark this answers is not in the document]
bioball: I am on board with the `noinspection` added on lines 480 and 482.

I agree with Phil that we should fix this specific warning, though. This is just about using a newer API that got introduced in Gradle 7.4.

Should be as simple as:

```
module.getTestSources().from(append(module.getTestSources().getFiles(), outputDir));
```

pkl-gradle/src/main/java/org/pkl/gradle/PklPlugin.java
```diff
@@ -398,6 +398,7 @@ private void configureIdeaModule(CodeGenSpec spec) {
...
                 module.setTestSourceDirs(append(module.getTestSourceDirs(), outputDir));
```
bioball, line 402:
```suggestion
                module.getTestSources().from(append(module.getTestSources().getFiles(), outputDir));
```

translatenix: fixed

pkl-gradle/src/main/java/org/pkl/gradle/PklPlugin.java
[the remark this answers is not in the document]
translatenix: Fixed. I thought @holzensp suggested to keep the warning to address this later.

bioball (approved):
LGTM; this looks great. Thanks!

PR #465, opened 2024-04-30
Improve conditional configuration of native build tasks

translatenix:
- Move all conditional configuration to the `assembleNative` and `testNative` "root" tasks
- Also pick default macOS tasks according to CPU architecture (wasn't the case before, but I assume it makes sense)

### DIFF
--- a/pkl-cli/pkl-cli.gradle.kts
+++ b/pkl-cli/pkl-cli.gradle.kts
@@ -215,48 +215,36 @@ fun Exec.configureExecutable(
  * Builds the pkl CLI for macOS/amd64.
  */
 val macExecutableAmd64: TaskProvider<Exec> by tasks.registering(Exec::class) {
-  if (buildInfo.os.isMacOsX) {
-    dependsOn(":installGraalVmAmd64")
-    configureExecutable(
-      buildInfo.graalVmAmd64,
-      layout.buildDirectory.file("executable/pkl-macos-amd64")
-    )
-  } else {
-    enabled = false
-  }
+  dependsOn(":installGraalVmAmd64")
+  configureExecutable(
+    buildInfo.graalVmAmd64,
+    layout.buildDirectory.file("executable/pkl-macos-amd64")
+  )
 }
 
 /**
  * Builds the pkl CLI for macOS/aarch64.
  */
 val macExecutableAarch64: TaskProvider<Exec> by tasks.registering(Exec::class) {
-  if (buildInfo.os.isMacOsX) {
-    dependsOn(":installGraalVmAarch64")
-    configureExecutable(
-      buildInfo.graalVmAarch64,
-      layout.buildDirectory.file("executable/pkl-macos-aarch64"),
-      listOf(
-        "-H:+AllowDeprecatedBuilderClassesOnImageClasspath"
-      )
+  dependsOn(":installGraalVmAarch64")
+  configureExecutable(
+    buildInfo.graalVmAarch64,
+    layout.buildDirectory.file("executable/pkl-macos-aarch64"),
+    listOf(
+      "-H:+AllowDeprecatedBuilderClassesOnImageClasspath"
     )
-  } else {
-    enabled = false
-  }
+  )
 }
 
 /**
  * Builds the pkl CLI for linux/amd64.
  */
 val linuxExecutableAmd64: TaskProvider<Exec> by tasks.registering(Exec::class) {
-  if (buildInfo.os.isLinux && buildInfo.arch == "amd64") {
-    dependsOn(":installGraalVmAmd64")
-    configureExecutable(
-      buildInfo.graalVmAmd64,
-      layout.buildDirectory.file("executable/pkl-linux-amd64")
-    )
-  } else {
-    enabled = false
-  }
+  dependsOn(":installGraalVmAmd64")
+  configureExecutable(
+    buildInfo.graalVmAmd64,
+    layout.buildDirectory.file("executable/pkl-linux-amd64")
+  )
 }
 
 /**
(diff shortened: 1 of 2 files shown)


StefMa (approved):
Oh yes! This looks way nicer!

pkl-cli/pkl-cli.gradle.kts
```diff
@@ -284,25 +268,37 @@ val linuxExecutableAarch64: TaskProvider<Exec> by tasks.registering(Exec::class)
...
 
 tasks.assembleNative {
-  dependsOn(macExecutableAmd64, macExecutableAarch64, linuxExecutableAmd64, linuxExecutableAarch64, alpineExecutableAmd64)
+  when {
+    buildInfo.os.isMacOsX && buildInfo.arch == "aarch64" -> {
+      dependsOn(macExecutableAarch64)
+    }
+    buildInfo.os.isMacOsX && buildInfo.arch == "amd64" -> {
+      dependsOn(macExecutableAmd64)
+    }
```
bioball, line 291:
```suggestion
    buildInfo.os.isMacOsX -> {
      dependsOn(macExecutableAarch64)
      dependsOn(macExecutableAmd64)
    }
```

pkl-core/pkl-core.gradle.kts
```diff
@@ -273,11 +268,23 @@ val testAlpineExecutableAmd64 by tasks.registering(Test::class) {
...
-  dependsOn(testMacExecutableAmd64)
-  dependsOn(testMacExecutableAarch64)
-  dependsOn(testAlpineExecutableAmd64)
+  when {
+    buildInfo.os.isMacOsX && buildInfo.arch == "aarch64" -> {
+      dependsOn(testMacExecutableAarch64)
+    }
+    buildInfo.os.isMacOsX && buildInfo.arch == "amd64" -> {
+      dependsOn(testMacExecutableAmd64)
+    }
```
bioball, line 277:
```suggestion
    buildInfo.os.isMacOsX -> {
      dependsOn(testMacExecutableAarch64)
      dependsOn(testMacExecutableAmd64)
    }
```

bioball (changes requested):
Generally looks good! Except, we actually cross-compile macOS; both amd64 and aarch64 binaries are built on aarch64 machines. It was intentional to let `buildNative` build both architectures.

pkl-core/pkl-core.gradle.kts
translatenix: Do you really want this by default, for every developer build? (You can still run the individual tasks if you need to.) Does it also work for Intel Macs?

bioball: These tasks (`buildNative`, `testNative`) aren't part of the base Gradle lifecycle tasks (build, test); they don't get run unless explicitly run. I think it's fine to let `buildNative` do a cross-build on macOS; you can also selectively build an architecture by running one of the subtasks.

But yeah, good point on Intel; we should skip out on the aarch64 stuff there.

translatenix: I'd optimize for the most common dev need. But OK, made the requested change.

bioball: Well, it's also consistent with how the linux/amd64 build works; on there, `buildNative` will build both libc and musl if it can.

pkl-cli/pkl-cli.gradle.kts
```diff
@@ -284,25 +268,38 @@ val linuxExecutableAarch64: TaskProvider<Exec> by tasks.registering(Exec::class)
...
 tasks.assembleNative {
-  dependsOn(macExecutableAmd64, macExecutableAarch64, linuxExecutableAmd64, linuxExecutableAarch64, alpineExecutableAmd64)
+  when {
+    buildInfo.os.isMacOsX && buildInfo.arch == "aarch64" -> {
+      dependsOn(macExecutableAarch64)
+      dependsOn(macExecutableAmd64)
+    }
+    buildInfo.os.isMacOsX && buildInfo.arch == "amd64" -> {
+      dependsOn(macExecutableAmd64)
+    }
```
bioball, line 292:
```suggestion
    buildInfo.os.isMacOsX -> {
      dependsOn(macExecutableAmd64)
      if (buildInfo.arch == "aarch64") {
        dependsOn(macExecutableAarch64)
      }
    }
```

pkl-core/pkl-core.gradle.kts
```diff
@@ -273,11 +268,24 @@ val testAlpineExecutableAmd64 by tasks.registering(Test::class) {
...
-  dependsOn(testMacExecutableAarch64)
-  dependsOn(testAlpineExecutableAmd64)
+  when {
+    buildInfo.os.isMacOsX && buildInfo.arch == "aarch64" -> {
+      dependsOn(testMacExecutableAarch64)
+      dependsOn(testMacExecutableAmd64)
+    }
+    buildInfo.os.isMacOsX && buildInfo.arch == "amd64" -> {
+      dependsOn(testMacExecutableAmd64)
+    }
```
bioball, line 278:
```suggestion
    buildInfo.os.isMacOsX -> {
      dependsOn(testMacExecutableAmd64)
      if (buildInfo.arch == "aarch64") {    
        dependsOn(testMacExecutableAarch64)
      }
    }
```

bioball (approved):
One stylistic nit.

Also; there's a merge conflict that needs addressing. Will merge once that's been addressed.

pkl-core/pkl-core.gradle.kts
translatenix: I'd definitely ask for a change if I could get the musl build to work. Until then I'm OK. :-)

PR #470, opened 2024-05-02
Remove hidden modifier from many reflect properties for better equality/rendering

stackoverflow:


pkl-core/src/test/kotlin/org/pkl/core/LanguageSnippetTestsEngine.kt
```diff
@@ -79,6 +79,9 @@ abstract class AbstractLanguageSnippetTestsEngine : InputOutputTestEngine() {
   // can't think of a better solution right now
   protected fun String.stripVersionCheckErrorMessage() =
     replace("Pkl version is ${Release.current().version()}", "Pkl version is xxx")
+  
+  protected fun String.stripStdlibLocationSha(): String =
+    replace(Regex("https://github.com/apple/pkl/blob/[0-9a-fA-F]+/stdlib/"), "https://github.com/apple/pkl/blob/stdlib/")
```
holzensp, line 84: It is generally wise to leave some marker that something was removed (compare, for example, `xxx` in `stripVersionCheckErrorMessage`).
```suggestion
    replace(Regex("https://github.com/apple/pkl/blob/[0-9a-fA-F]+/stdlib/"), "https://github.com/apple/pkl/blob/__hash__/stdlib/")
```

holzensp (commented):
This is a double-edged sword. Quite a few of these (in other classes in this module) were made `hidden` to make the rendered output more readable and/or to prevent stack overflows. I'm not entirely sure whether this makes for _better_ rendering and it only makes equality "come closer to correct." The `hidden`s you remove in this PR clearly don't introduce stack overflows, but it does seem a little arbitrary to just unhide what we happen to can without issues and leave other things hidden.

I think the deeper problem is that equality is somewhat broken and I'm wondering whether we shouldn't "just" fix that. I think we can't _generally_ fix equality because one can always define recursive, non-converging infinite amends chains, which make the problem undecidable. One of the reasons for the choice to only observe visible properties for equality was due to cycles (any instance of which would otherwise be a stack overflow). Keeping a stack of `seen` objects (using native JVM object identity) should help and then we can always give it a maximum recursion depth.

WDYT?

pkl-core/src/test/kotlin/org/pkl/core/LanguageSnippetTestsEngine.kt
stackoverflow: Done

stackoverflow:
> This is a double-edged sword. Quite a few of these (in other classes in this module) were made `hidden` to make the rendered output more readable and/or to prevent stack overflows. I'm not entirely sure whether this makes for _better_ rendering and it only makes equality "come closer to correct." The `hidden`s you remove in this PR clearly don't introduce stack overflows, but it does seem a little arbitrary to just unhide what we happen to can without issues and leave other things hidden.
> 
> I think the deeper problem is that equality is somewhat broken and I'm wondering whether we shouldn't "just" fix that. I think we can't _generally_ fix equality because one can always define recursive, non-converging infinite amends chains, which make the problem undecidable. One of the reasons for the choice to only observe visible properties for equality was due to cycles (any instance of which would otherwise be a stack overflow). Keeping a stack of `seen` objects (using native JVM object identity) should help and then we can always give it a maximum recursion depth.
> 
> WDYT?

I agree that conflating equality with rendering is not ideal, but the solution to that is beyond the scope of this PR.
This is just a bugfix to an issue I created when I fixed the stack overflow rendering reflected values.
As reflected values are not really configuration and mostly rendered for debug purposes I think it's fine to make everything we can not hidden. If/when we implemented a smart equality calculation we can unhide even more things.
The important part here is making `SourceLocation` not hidden and thus, part of equality, as locations are always unique.

pkl-core/src/test/kotlin/org/pkl/core/LanguageSnippetTestsEngine.kt
```diff
@@ -79,6 +79,9 @@ abstract class AbstractLanguageSnippetTestsEngine : InputOutputTestEngine() {
   // can't think of a better solution right now
   protected fun String.stripVersionCheckErrorMessage() =
     replace("Pkl version is ${Release.current().version()}", "Pkl version is xxx")
+  
+  protected fun String.stripStdlibLocationSha(): String =
+    replace(Regex("https://github.com/apple/pkl/blob/[0-9a-fA-F]+/stdlib/"), "https://github.com/apple/pkl/blob/_hash_/stdlib/")
```
bioball, line 84: I think we can match the commit ID directly

```suggestion
    replace("https://github.com/apple/pkl/blob/${Release.current().commitId()}/stdlib/", "https://github.com/apple/pkl/blob/_hash_/stdlib/")
```

stackoverflow: Indeed. This way we don't need a regex.

pkl-core/src/test/kotlin/org/pkl/core/LanguageSnippetTestsEngine.kt
```diff
@@ -79,6 +79,9 @@ abstract class AbstractLanguageSnippetTestsEngine : InputOutputTestEngine() {
   // can't think of a better solution right now
   protected fun String.stripVersionCheckErrorMessage() =
     replace("Pkl version is ${Release.current().version()}", "Pkl version is xxx")
+  
+  protected fun String.stripStdlibLocationSha(): String =
+    replace("https://github.com/apple/pkl/blob/${Release.current().commitId()}/stdlib/", "https://github.com/apple/pkl/blob/_hash_/stdlib/")
```
bioball, line 84: One more suggestion here: let's use a dollar-name like we do `$snippetsDir` and `$pklWebsite`.

```suggestion
    replace("https://github.com/apple/pkl/blob/${Release.current().commitId()}/stdlib/", "https://github.com/apple/pkl/blob/\$commitId/stdlib/")
```

bioball (approved):
One suggestion, but approving to unblock

PR #471, opened 2024-05-03
Declare members of http package NonnullByDefault

bioball:
* Add a package-info.java with NonnullByDefault annotation
* Make RequestCapturingClient return a response to satisfy the type checker

CC @translatenix


stackoverflow (approved):
LGTM

PR #484, opened 2024-05-09
Adjust filename according to style guide

garysassano:
Adjusted template's filename in tutorial's example to reflect [Pkl Style Guide](https://pkl-lang.org/main/current/style-guide/index.html#filename) recommendations.


.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
bioball, line 1: Can you delete this file?

bioball (changes requested):
Thanks for the typo fixes! There's a `.vscode/settings.json` file that shouldn't be checked in, though.

bioball (approved):
LGTM, thank you!

PR #486, opened 2024-05-10
Support scheme-agnostic projects

bioball:
Supercedes and closes #255


stdlib/Project.pkl
```diff
@@ -91,18 +91,15 @@ package: Package?
...
-/// A local dependency is another [Project] that is local to the file system.
-///
-/// To declare, use `import("path/to/PklProject")`
-typealias LocalDependency = Project(
-  isLocalPklProject,
-  this != module,
-  this.package != null
-)
+const local function isUriLocal(uri1: String, uri2: String): Boolean =
+  uri1.substring(0, uri1.indexOf(":")) == uri2.substring(0, uri2.indexOf(":"))
```
bioball, line 102: This isn't the perfect check. This should also check that the URIs have the same authority. This is hard to implement without a URI API within the stdlib though.

stackoverflow (approved):
Looks good. I'd add some note to our language reference explaining that projects can be loaded from different module keys, not only files. Probably in the `Resolving Dependencies` or `Local Dependencies` section.

HT154:
Thanks for picking this up! I've looked through the changes made for SPICE-0005 conformance and everything makes sense to me.

This also closes #166, which I'm delighted to see!

pkl-core/src/main/java/org/pkl/core/module/ProjectDependenciesManager.java
```diff
@@ -194,28 +206,42 @@ public Dependency getResolvedDependency(PackageUri packageUri) {
...
+          projectDeps = ProjectDeps.parse(depsJson);
+        } catch (IOException e) {
+          var exceptionBuilder =
+              new VmExceptionBuilder().evalError("cannotLoadProjectDepsJson", depsUri).withCause(e);
+          if (e.getMessage() != null) {
+            exceptionBuilder.withHint(e.getMessage());
+          } else {
+            exceptionBuilder.withHint("Encountered error: " + e);
+          }
+          throw exceptionBuilder.build();
```
holzensp, line 238:
```suggestion
          throw new VmExceptionBuilder()
            .evalError("cannotLoadProjectDepsJson", depsUri)
            .withCause(e)
            .withHint(e.getMessage() != null ? e.getMessage() : ("Encountered error: " + e))
            .build();
```

Some day, null-coalescence will come to Java ;)

pkl-core/src/main/java/org/pkl/core/util/GlobResolver.java
```diff
@@ -488,7 +489,14 @@ public static List<ResolvedGlobElement> resolveGlob(
         }
         return result;
       }
-      var baseUri = enclosingModuleKey.resolveUri(enclosingUri, URI.create(basePath));
+      URI baseUri;
+      try {
+        baseUri = IoUtils.resolve(securityManager, enclosingModuleKey, URI.create(basePath));
+      } catch (URISyntaxException e) {
+        // assertion: this is only thrown if the pattern starts with a triple-dot import.
+        // the language will throw an error if glob imports is combined with triple-dots.
```
holzensp, line 497: In that case, isn't this unreachable code?

holzensp (approved):
No blockers; some nits.

pkl-core/src/main/java/org/pkl/core/util/GlobResolver.java
bioball: I think I'll leave this as-is; not 100% sure this is unreachable.

pkl-core/src/test/files/LanguageSnippetTests/input/projects/project6/PklProject
```diff
@@ -0,0 +1,8 @@
+amends "pkl:Project"
+
+package {
+  name = "project6"
+  baseUri = "package://localhost:12110/project6"
+  version = "1.0.0"
+  packageZipUrl = "https://localhost:12110/project6/project6-\(version).zip"
```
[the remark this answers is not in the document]
bioball: This is just a snippet test, so, I'm going to leave it as-is.

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -680,6 +680,11 @@ Cannot resolve a triple-dot import from module URI `{0}`.\n\
 \n\
 Triple-dot imports may only be resolved by module schemes that are considered local, and have hierarchical URIs.
 
+moduleDoesNotSupportDependencies=\
+Module `{0}` does not support importing dependencies.\n\
+\n\
+Dependencies can only be imported in modules that belong to a project, or within a package.
```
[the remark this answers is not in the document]
bioball: Packages aren't projects, so, I think this is useful to call out.

pkl-core/src/test/files/LanguageSnippetTests/input/projects/project1/globbing.pkl
```diff
@@ -34,4 +34,9 @@ examples {
   ["glob-read absolute package uri"] {
     read*("package://localhost:0/birds@0.5.0#/catalog/*.pkl")
   }
+  
+//   https://github.com/apple/pkl/issues/166
```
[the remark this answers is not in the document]
bioball: Ended up getting rid of this comment; it's not really useful since the test already tells you exactly what it's checking for.

bioball:
For docs: I couldn't think of a natural place to add details about this behavior. Looking through our docs, I think we need to beef up our docs a little bit, especially when it relates to the APIs we provide when using Pkl as a library. I think let's consider that as out-of-scope for this PR, and something we should do in the future.

PR #489, opened 2024-05-10
Encode filepaths to be safe for Windows

bioball:
Also: don't URI-encode parentheses in pkldoc.

Implementation for [SPICE-0003](https://github.com/apple/pkl-evolution/pull/3)


pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
```diff
@@ -663,6 +663,69 @@ public static URI fixTripleSlashUri(URI baseUri, URI newUri) {
...
+          sb.append(")");
+        }
+        case '(' -> sb.append("((");
+        default -> sb.append(path.charAt(i));
+      }
+    }
+    return sb.toString();
+  }
+
+  /** The inverse of {@link #encodePath(String)} */
```
stackoverflow, line 690: Technically, this function is not really the inverse of `encodePath`, as it decodes anything, not only the chars that `encodePath` encodes. So one can decode `(66)(6f)(6f).text` as `foo.txt` which cannot be generated by `encodePath`. But I'm being pedantic and I think it's fine that it decodes anything.

pkl-doc/src/main/kotlin/org/pkl/doc/Util.kt
```diff
@@ -119,18 +118,11 @@ internal fun String.replaceSourceCodePlaceholders(
...
+/** Encodes a URI string, encoding characters that are part of URI syntax. */
 internal val String.uriEncodedComponent
   get(): String {
-    val ret = URLEncoder.encode(this, StandardCharsets.UTF_8)
-    // Replace `+` with `%20` to be safe
-    // (see
-    // https://stackoverflow.com/questions/2678551/when-should-space-be-encoded-to-plus-or-20#:~:text=%20%20is%20a%20valid%20way,encodeURIComponent()%20does%20in%20JavaScript.)
-    return ret.replace("+", "%20")
+    val ret = URI(null, null, this, null)
+    return ret.toString().replace("/", "%2F")
```
stackoverflow, line 125: Why the change here?

stackoverflow (approved):
Looks good.

pkl-doc/src/main/kotlin/org/pkl/doc/Util.kt
bioball: The previous implementation was encoding parentheses in percent encoding.

pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
bioball: Good point. Although, I'm actually getting rid of this altogether; this method isn't used anywhere.

PR #492, opened 2024-05-13
Add support for Windows

bioball:
Closes https://github.com/apple/pkl/issues/20

This adds support for Windows.
The in-language path separator is still `/`, to ensure Pkl programs are cross-platform.
The specifics behind this design are detailed in https://github.com/apple/pkl-evolution/pull/7.

Log lines are written using CRLF endings on Windows.
Modules that are combined with `--module-output-separator` uses LF endings to ensure
consistent rendering across platforms.

`jpkl` does not work on Windows as a direct executable.
However, it can work with `java -jar jpkl`.

Additional details:

* Adjust git settings for Windows
* Add native executable for pkl cli
* Add jdk17 windows Gradle check in CI
* Adjust CI test reports to be staged within Gradle rather than by shell script.
* Fix: encode more characters that are not safe Windows paths
* Skip running tests involving symbolic links on Windows (these require administrator privileges to run).
* Introduce custom implementation of `IoUtils.relativize`, because the current implementation delegates to `Path`, which will fail if the URI contains invalid Windows characters.
* Allow Gradle to initialize ExecutableJar `Property` values
* Add a way to enable remote debugging with `./gradlew test` commands
* Standard library `pkl:platform`'s `current.operatingSystem.name` is `"Windows"` for all Windows OSes (match treatment of macOS/Linux).

### DIFF
--- a/.circleci/config.pkl
+++ b/.circleci/config.pkl
@@ -14,26 +14,28 @@
 // limitations under the License.
 //===----------------------------------------------------------------------===// 
 // File gets rendered to .circleci/config.yml via git hook.
-amends "package://pkg.pkl-lang.org/pkl-project-commons/pkl.impl.circleci@1.1.0#/PklCI.pkl"
+amends "package://pkg.pkl-lang.org/pkl-project-commons/pkl.impl.circleci@1.1.1#/PklCI.pkl"
 
 import "jobs/BuildNativeJob.pkl"
 import "jobs/GradleCheckJob.pkl"
 import "jobs/DeployJob.pkl"
 import "jobs/SimpleGradleJob.pkl"
 
-local prbJobs: Listing<String> = gradleCheckJobs.keys.toListing()
+local prbJobs: Listing<String> = (gradleCheckJobs.keys.toListing()) {
+  "pkl-cli-windows-amd64-snapshot"
+}
 
 local buildAndTestJobs = (prbJobs) {
   "bench"
   "gradle-compatibility"
   ...buildNativeJobs.keys.filter((it) -> it.endsWith("snapshot"))
-}
+}.distinct
 
 local releaseJobs = (prbJobs) {
   "bench"
   "gradle-compatibility"
   ...buildNativeJobs.keys.filter((it) -> it.endsWith("release"))
-}
+}.distinct
 
 prb {
   jobs = prbJobs
@@ -96,6 +98,11 @@ local buildNativeJobs: Mapping<String, BuildNativeJob> = new {
       musl = true
       isRelease = _dist == "release"
     }
+    ["pkl-cli-windows-amd64-\(_dist)"] {
+      arch = "amd64"
+      os = "windows"
+      isRelease = _dist == "release"
+    }
   }
 }
 
@@ -107,6 +114,11 @@ local gradleCheckJobs: Mapping<String, GradleCheckJob> = new {
   ["gradle-check-jdk21"] {
     javaVersion = "21.0"
     isRelease = false
+  }
+  ["gradle-check-jdk17-windows"] {
+    javaVersion = "17.0"
+    isRelease = false
+    os = "windows"
   }
 }
 
--- a/.circleci/config.pkl
+++ b/.circleci/config.pkl
@@ -21,21 +21,19 @@ import "jobs/GradleCheckJob.pkl"
 import "jobs/DeployJob.pkl"
 import "jobs/SimpleGradleJob.pkl"
 
-local prbJobs: Listing<String> = (gradleCheckJobs.keys.toListing()) {
-  "pkl-cli-windows-amd64-snapshot"
-}
+local prbJobs: Listing<String> = gradleCheckJobs.keys.toListing()
 
 local buildAndTestJobs = (prbJobs) {
   "bench"
   "gradle-compatibility"
   ...buildNativeJobs.keys.filter((it) -> it.endsWith("snapshot"))
-}.distinct
+}
 
 local releaseJobs = (prbJobs) {
   "bench"
   "gradle-compatibility"
   ...buildNativeJobs.keys.filter((it) -> it.endsWith("release"))
-}.distinct
+}
 
 prb {
   jobs = prbJobs
--- a/.circleci/jobs/GradleJob.pkl
+++ b/.circleci/jobs/GradleJob.pkl
@@ -15,14 +15,18 @@
 //===----------------------------------------------------------------------===//
 abstract module GradleJob
 
-import "package://pkg.pkl-lang.org/pkl-pantry/com.circleci.v2@1.1.0#/Config.pkl"
+import "package://pkg.pkl-lang.org/pkl-pantry/com.circleci.v2@1.1.2#/Config.pkl"
 
 /// Whether this is a release build or not.
 isRelease: Boolean = false
 
+/// The OS to run on
+os: "macOS"|"linux"|"windows"
+
 fixed gradleArgs = new Listing {
   "--info"
   "--stacktrace"
+  "-DtestReportsDir=${HOME}/test-results"
   when (isRelease) {
     "-DreleaseBuild=true"
   }
@@ -37,15 +41,6 @@ job: Config.Job = new {
   steps {
     "checkout"
     ...module.steps
-    new Config.RunStep {
-      // find all test results and write them to the home dir
-      name = "Gather test results"
-      command = """
-        mkdir ~/test-results/
-        find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} ~/test-results/ \\;
-        """
-      `when` = "always"
-    }
     new Config.StoreTestResults {
       path = "~/test-results"
     }
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,4 @@
 /docs/** linguist-documentation
 
 *.pkl linguist-language=Groovy
+* text eol=lf
--- a/buildSrc/src/main/kotlin/BuildInfo.kt
+++ b/buildSrc/src/main/kotlin/BuildInfo.kt
@@ -26,6 +26,7 @@ open class BuildInfo(project: Project) {
       when {
         os.isMacOsX -> "macos"
         os.isLinux -> "linux"
+        os.isWindows -> "windows"
         else -> throw RuntimeException("${os.familyName} is not supported.")
       }
     }
@@ -36,7 +37,8 @@ open class BuildInfo(project: Project) {
 
     val downloadUrl: String by lazy {
       val jdkMajor = graalVmJdkVersion.takeWhile { it != '.' }
-      "https://download.oracle.com/graalvm/$jdkMajor/archive/$baseName.tar.gz"
+      val extension = if (os.isWindows) "zip" else "tar.gz"
+      "https://download.oracle.com/graalvm/$jdkMajor/archive/$baseName.$extension"
     }
 
     val installDir: File by lazy {
@@ -85,9 +87,14 @@ open class BuildInfo(project: Project) {
   val commitId: String by lazy {
     // only run command once per build invocation
     if (project === project.rootProject) {
-      Runtime.getRuntime()
-        .exec(arrayOf("git", "rev-parse", "--short", "HEAD"), arrayOf(), project.rootDir)
-        .inputStream.reader().readText().trim()
+      val process = ProcessBuilder()
+        .command("git", "rev-parse", "--short", "HEAD")
+        .directory(project.rootDir)
+        .start()
+      process.waitFor().also { exitCode ->
+        if (exitCode == -1) throw RuntimeException(process.errorStream.reader().readText())
+      }
+      process.inputStream.reader().readText().trim()
     } else {
       project.rootProject.extensions.getByType(BuildInfo::class.java).commitId
     }
--- a/buildSrc/src/main/kotlin/ExecutableJar.kt
+++ b/buildSrc/src/main/kotlin/ExecutableJar.kt
@@ -1,11 +1,11 @@
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.provider.ListProperty
+import org.gradle.api.provider.Property
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFile
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
-import org.gradle.kotlin.dsl.listProperty
 
 /**
  * Builds a self-contained Pkl CLI Jar that is directly executable on *nix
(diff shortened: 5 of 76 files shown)


.circleci/jobs/GradleJob.pkl
```diff
@@ -37,15 +41,6 @@ job: Config.Job = new {
...
     ...module.steps
-    new Config.RunStep {
-      // find all test results and write them to the home dir
-      name = "Gather test results"
-      command = """
-        mkdir ~/test-results/
-        find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} ~/test-results/ \\;
-        """
-      `when` = "always"
-    }
```
bioball, line 48: This step breaks on Windows.

Instead of trying to figure out how to write a Windows version, I added some logic to specify the test reports dir (see line 29).

.circleci/jobs/GradleJob.pkl
```diff
@@ -15,14 +15,18 @@
...
 abstract module GradleJob
 
-import "package://pkg.pkl-lang.org/pkl-pantry/com.circleci.v2@1.1.0#/Config.pkl"
+import "package://pkg.pkl-lang.org/pkl-pantry/com.circleci.v2@1.1.2#/Config.pkl"
 
 /// Whether this is a release build or not.
 isRelease: Boolean = false
 
+/// The OS to run on
+os: "macOS"|"linux"|"windows"
```
bioball, line 24: Moved this from `BuildNativeJob`, because this adds a `./gradlew check` job for Windows to PRBs.

buildSrc/src/main/kotlin/BuildInfo.kt
```diff
@@ -85,9 +87,14 @@ open class BuildInfo(project: Project) {
...
-      Runtime.getRuntime()
-        .exec(arrayOf("git", "rev-parse", "--short", "HEAD"), arrayOf(), project.rootDir)
-        .inputStream.reader().readText().trim()
+      val process = ProcessBuilder()
+        .command("git", "rev-parse", "--short", "HEAD")
+        .directory(project.rootDir)
+        .start()
+      process.waitFor().also { exitCode ->
+        if (exitCode == -1) throw RuntimeException(process.errorStream.reader().readText())
+      }
```
bioball, line 96: Tangential improvement: changed this because the failure was getting swallowed and I wasn't sure why the commit ID was not being populated correctly on Windows.

buildSrc/src/main/kotlin/ExecutableJar.kt
```diff
@@ -15,27 +15,25 @@ import org.gradle.kotlin.dsl.listProperty
...
-  val inJar: RegularFileProperty = project.objects.fileProperty()
+  abstract val inJar: RegularFileProperty
 
   @get:OutputFile
-  val outJar: RegularFileProperty = project.objects.fileProperty()
+  abstract val outJar: RegularFileProperty
 
   @get:Input
-  val jvmArgs: ListProperty<String> = project.objects.listProperty()
+  abstract val jvmArgs: ListProperty<String>
```
bioball, line 26: This change is unrelated to Windows. Happy to move this to a separate PR if you'd prefer.

pkl-cli/src/main/kotlin/org/pkl/cli/CliEvaluator.kt
```diff
@@ -158,7 +160,7 @@ constructor(
           } else {
             if (output.isNotEmpty()) {
               outputFile.writeString(
-                options.moduleOutputSeparator + IoUtils.getLineSeparator(),
+                options.moduleOutputSeparator + '\n',
```
bioball, line 163: Changed this to be `LF` and not platform dependent. Otherwise, `pkl eval foo.pkl bar.pkl` will give you different results on different OSes, which seems quite undesirable.

pkl-cli/src/test/kotlin/org/pkl/cli/CliProjectPackagerTest.kt
```diff
@@ -858,17 +867,18 @@ class CliProjectPackagerTest {
...
-      .out/project2@2.0.0/project2@2.0.0
-      .out/project2@2.0.0/project2@2.0.0.sha256
+      .out${sep}project1@1.0.0${sep}project1@1.0.0.zip
+      .out${sep}project1@1.0.0${sep}project1@1.0.0.zip.sha256
+      .out${sep}project1@1.0.0${sep}project1@1.0.0
+      .out${sep}project1@1.0.0${sep}project1@1.0.0.sha256
+      .out${sep}project2@2.0.0${sep}project2@2.0.0.zip
+      .out${sep}project2@2.0.0${sep}project2@2.0.0.zip.sha256
+      .out${sep}project2@2.0.0${sep}project2@2.0.0
+      .out${sep}project2@2.0.0${sep}project2@2.0.0.sha256
```
bioball, line 881: Log lines on Windows will output CRLF endings, and also output `\` as the path separator.

The rationale is that these are log lines, which are meant to be useful for the host machine specifically.

pkl-core/src/main/java/org/pkl/core/module/ModuleKeyFactories.java
```diff
@@ -141,15 +141,20 @@ public Optional<ModuleKey> create(URI uri) {
   private static class File implements ModuleKeyFactory {
     @Override
     public Optional<ModuleKey> create(URI uri) {
-      Path path;
-      try {
-        path = Path.of(uri);
```
bioball, line 146: Needed to adjust this because `Path.of("foo:bar")` throws an IllegalArgumentException, but yet should be treated as a file module key.

Changed this to initialize `Path` within the module key logic rather than here.

.gitattributes
```diff
@@ -4,3 +4,4 @@
 /docs/** linguist-documentation
 
 *.pkl linguist-language=Groovy
+* text eol=lf
```
bioball, line 7: Needed for git, and for spotless apply.

stackoverflow (approved):
Looks good.

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -175,45 +178,44 @@ Pkl {pkl-version} (Linux, native)
...
+=== Windows Executable
 
 [source,shell]
 [subs="+attributes"]
 ----
-curl -L -o jpkl '{uri-pkl-java-download}'
-chmod +x jpkl
-./jpkl --version
+Invoke-WebRequest '{uri-pkl-windows-download}' -OutFile pkl.exe
+chmod +x pkl.exe
```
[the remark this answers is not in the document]
bioball: Oh, good catch.

pkl-core/src/main/java/org/pkl/core/project/ProjectPackager.java
```diff
@@ -363,7 +370,7 @@ private List<Path> collectPackageElements(Project project, Package pkg) {
   }
 
   private boolean isAbsoluteImport(String importStr) {
-    return importStr.matches("\\w:.*") || importStr.startsWith("@");
+    return importStr.matches("\\w+:.*") || importStr.startsWith("@");
```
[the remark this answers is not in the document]
bioball: Yup

pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
```diff
@@ -180,6 +192,10 @@ public static String getLineSeparator() {
     return System.getProperty("line.separator");
   }
 
+  public static Boolean isWindows() {
+    return System.getProperty("os.name").contains("Windows");
```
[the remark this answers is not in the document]
bioball: I'll switch this to use `Platform.current().operatingSystem()`, which avoids the `contains` call every time.

pkl-core/src/main/java/org/pkl/core/module/ResolvedModuleKeys.java
```diff
@@ -75,7 +77,16 @@ public URI getUri() {
...
     @Override
     public String loadSource() throws IOException {
-      return Files.readString(path, StandardCharsets.UTF_8);
+      try {
+        return Files.readString(path, StandardCharsets.UTF_8);
+      } catch (AccessDeniedException e) {
+        // Windows throws `AccessDeniedException` when reading directories.
+        // Sync error between different OSes.
+        if (Files.isDirectory(path)) {
+          throw new IOException("Is a directory");
```
holzensp, line 86: Should we not preserve the trace? I'd, at least, add `e` as a cause.
```suggestion
          throw new IOException("Is a directory", e);
```

pkl-doc/src/test/kotlin/org/pkl/doc/CliDocGeneratorTest.kt
```diff
@@ -219,6 +226,11 @@ class CliDocGeneratorTest {
       .withFailMessage("Test bug: $actualFile should exist but does not.")
       .exists()
 
+    // symlinks on Git and Windows is rather finnicky; they create shortcuts by default unless
+    // a core Git option is set. Also, by default, symlinks require administrator privileges to run.
```
holzensp, line 230: ... but a `.lnk` is a text file with the redirected path in, no? Do you also need admin to read? If not; Is it an option to read `actualFile` instead and replace it with the contents? (Would possibly require a `while`, because of link-to-link cases.)

pkl-executor/src/main/java/org/pkl/executor/EmbeddedExecutor.java
```diff
@@ -157,7 +157,15 @@ private PklDistribution findCompatibleDistribution(
 
   private static Path toDisplayPath(Path modulePath, ExecutorOptions options) {
     var rootDir = options.getRootDir();
-    return rootDir == null ? modulePath : rootDir.relativize(modulePath);
+    return rootDir == null ? modulePath : relativize(modulePath, rootDir);
+  }
+
+  // Windows relativize will fail if the two paths have different roots.
+  private static Path relativize(Path path, Path base) {
+    if (path.isAbsolute() && base.isAbsolute() && !path.getRoot().equals(base.getRoot())) {
```
holzensp, line 165: I'd be inclined to push it in
```suggestion
    return relativize(modulePath, rootDir);
  }

  // Windows relativize will fail if the two paths have different roots.
  private static Path relativize(Path path, Path base) {
    if (base == null || path.isAbsolute() && base.isAbsolute() && !path.getRoot().equals(base.getRoot())) {
```

.circleci/config.pkl
```diff
@@ -134,6 +136,11 @@ local buildNativeJobs: Mapping<String, BuildNativeJob> = new {
       musl = true
       isRelease = _dist == "release"
     }
+    ["pkl-cli-windows-amd64-\(_dist)"] {
```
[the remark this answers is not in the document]
bioball: GraalVM doesn't have Windows ARM support. I'm actually not sure if they plan on supporting it; my question about it in Slack has not received any responses: https://graalvm.slack.com/archives/CNBFR78F9/p1715635155814669

pkl-cli/src/test/kotlin/org/pkl/cli/CliEvaluatorTest.kt
```diff
@@ -978,6 +984,56 @@ result = someLib.x
...
+    val moduleUri =
+      writePklFile(
+        "test.pkl",
+        """
+      output {
+        files {
+          ["foo:bar"] { text = "bar" }
+        }
+      }
+    """
```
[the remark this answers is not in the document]
bioball: Yes

pkl-core/src/main/java/org/pkl/core/module/ModuleKeyFactories.java
bioball: Yup; this is what we get on Windows:

```
jshell> Path.of(URI.create("file:///foo:bar"))
|  Exception java.nio.file.InvalidPathException: Illegal char <:> at index 4: /foo:bar
|        at WindowsPathParser.normalize (WindowsPathParser.java:182)
|        at WindowsPathParser.parse (WindowsPathParser.java:153)
|        at WindowsPathParser.parse (WindowsPathParser.java:77)
|        at WindowsPath.parse (WindowsPath.java:92)
|        at WindowsUriSupport.fromUri (WindowsUriSupport.java:166)
|        at WindowsFileSystemProvider.getPath (WindowsFileSystemProvider.java:98)
|        at Path.of (Path.java:203)
|        at (#6:1)
```

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -175,45 +178,44 @@ Pkl {pkl-version} (Linux, native)
...
+.\pkl.exe --version
```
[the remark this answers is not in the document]
bioball: `.\` is needed. Looks like `.exe` can be omitted, though.

pkl-doc/src/test/kotlin/org/pkl/doc/CliDocGeneratorTest.kt
bioball: As far as I can tell, there's no easy way to parse a `lnk` file in Java and figure out where it's pointed to.

pkl-cli/src/main/kotlin/org/pkl/cli/CliEvaluator.kt
bioball: This is configuration output meant to be saved to a file, or applied to a system.

The standard use-case here is to generate a YAML document stream. For example, `pkl eval podA.pkl podB.pkl > output.yaml`.

Note that if the `-o` option is set, it already uses `LF` endings. So, without this change, these two would have had different results:

```
pkl eval podA.pkl podB.pkl > output.yaml
```

```
pkl eval podA.pkl podB.pkl -o output.yaml
```

pkl-cli/src/test/kotlin/org/pkl/cli/CliProjectPackagerTest.kt
bioball: Can you elaborate on what is not consistent here?

pkl-core/src/main/java/org/pkl/core/module/ResolvedModuleKeys.java
bioball: While this is generally a good idea, it works against us here because the cause gets rendered in the `-- Pkl Error --` message. The idea here is that we want to make this error message identical on the two platforms.

pkl-core/src/main/java/org/pkl/core/module/ModuleKeys.java
```diff
@@ -316,7 +316,13 @@ public List<PathElement> listElements(SecurityManager securityManager, URI baseU
     public ResolvedModuleKey resolve(SecurityManager securityManager)
         throws IOException, SecurityManagerException {
       securityManager.checkResolveModule(uri);
-      var realPath = path.toRealPath();
+      // Disallow paths that contain `\\` characters if on Windows
+      // (require `/` as the path separator on all OSes)
+      var uriPath = uri.getPath();
+      if (java.io.File.separatorChar == '\\' && uriPath != null && uriPath.contains("\\")) {
+        throw new FileNotFoundException();
```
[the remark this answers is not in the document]
bioball: Fair point; will add an extra hint here about it. But I'm thinking let's just show this on every OS. It may or may not be useful to users on *nix platforms that are used to Windows style paths. Also, having this be Windows only greatly complicates our testing.

pkl-executor/src/main/java/org/pkl/executor/EmbeddedExecutor.java
bioball: I actually find the former a little easier to read, because of fewer binary operators grouped in the `if`

PR #493, opened 2024-05-15
Split out base settings in stdlib for reuse

holzensp:
This PR defines `pkl:AbstractSettings` as a common parent module for `pkl:settings` and `pkl:Project`.

Since user-level evaluator settings should be limited to user/machine specific ones (to avoid unnecessary "works on my machine"), whereas `PklProject`s might want to set more detailed settings.

There isn't an overwhelming win from making `pkl:AbstractSettings` a super-module, other than `evaluatorSettings` (and possibly other properties in the future) being defined in all children by default. Alternatively, `pkl:AbstractSettings` could define all `class`es that might be used somewhere, and it just being `import`ed, i.e.

`settings.pkl`:
```
module pkl.settings

import "pkl:AbstractSettings"

proxy: AbstractSettings.Proxy

//...
```

`Project.pkl`:
```
module pkl.Project

import "pkl:AbstractSettings"

evaluatorSettings: AbstractSettings.Evaluator

//...
```

### DIFF
--- a/pkl-core/src/test/files/LanguageSnippetTests/input/stdlib/settings.pkl
+++ b/pkl-core/src/test/files/LanguageSnippetTests/input/stdlib/settings.pkl
@@ -1,0 +1,25 @@
+amends ".../snippetTest.pkl"
+
+import "pkl:AbstractSettings"
+import "pkl:Project"
+import "pkl:settings"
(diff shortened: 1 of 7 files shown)


holzensp:
Rolling this into Proxy PR

PR #495, opened 2024-05-15
Introduce specialized ModuleKey(Factory) for `http` uris

bioball:
GenericUrl is a catch-all that uses URL.openConnection(). Since we now have special handling of HTTP urls, it makes more sense to put it in its own module key.

Those that don't use the Http module key specifically but use `GenericUrl` will still be able to fetch `http` urls, but it would use the standard `URL.openConnection` implementation. This works the same as `file:` URIs; if only the `GenericUrl` module key factory is used, Pkl can still load `file:` uris.

(FYI @translatenix)


stackoverflow (approved):
Looks good.

PR #472, opened 2024-05-03
Add support for HTTP proxying

bioball:
Following design of SPICE-0004: https://github.com/apple/pkl-evolution/pull/5

PR #506, opened 2024-05-27
Add support for HTTP proxying

holzensp:
Following design of [SPICE-0004](https://github.com/apple/pkl-evolution/pull/5).

Compared to #472; Re-organizes Pkl config, cleans up commits, and adds ability to use proxy settings from the system (macOS, Windows, GNOME).

Closes #472


stdlib/EvaluatorSettings.pkl
```diff
@@ -0,0 +1,125 @@
...
+  /// address = "http://myuser@\(proxyPassword):my.proxy.example.com:5080"
+  /// ```
+  address: Uri(startsWith("http://"))
+
+  /// Hosts to which all connections should bypass a proxy.
+  ///
+  /// Values can be either hostnames, or CIDR blocks.
+  ///
+  /// Follows the same rules as cURL's [--noproxy](https://curl.se/docs/manpage.html#--noproxy)
+  /// flag.
```
bioball, line 123: I'm not totally sure that we do follow cURL's rules. Better to spell this all out:

```suggestion
  /// Values can be either hostnames, or IP addresses.
  /// IP addresses can optionally be provided using [CIDR notation](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#CIDR_notation).
  ///
  /// The only wildcard is `"*"`, which disables all proxying.
  ///
  /// A hostname will match all subdomains.
  /// For example, `example.com` will match `foo.example.com`, but not `fooexample.com`.
  /// A hostname that is prefixed with a dot will match the hostname itself.
  /// Effectively, `.example.com` is the same as `example.com`.
  ///
  /// Optionally, a port can be specified.
  /// If a port is omitted, all ports are matched.
  ///
  /// Example:
  ///
  /// ```
  /// noProxy {
  ///   "127.0.0.1"
  ///   "169.254/16"
  ///   "example.com"
  ///   "localhost:5050"
  /// }
  /// ```
```

pkl-core/src/main/java/org/pkl/core/project/Project.java
```diff
@@ -325,7 +278,7 @@ private Project(
     return pkg;
   }
 
-  public EvaluatorSettings getSettings() {
+  public PklEvaluatorSettings getSettings() {
     return evaluatorSettings;
   }
```
bioball, line 283: Per my comment below to have a deprecation cycle, we can do this:

```suggestion
  /** Use {@link org.pkl.core.project.Project#getEvaluatorSettings()} instead. */
  @Deprecated(forRemoval = true)
  public EvaluatorSettings getSettings() {
    return new EvaluatorSettings(evaluatorSettings);
  }

  public PklEvaluatorSettings getEvaluatorSettings() {
    return evaluatorSettings;
  }
```

This works out nicely because `getSettings()` should probably be renamed anyway.

pkl-core/src/test/kotlin/org/pkl/core/project/ProjectTest.kt
```diff
@@ -116,7 +117,10 @@ class ProjectTest {
     """.trimIndent())
     val project = Project.loadFromPath(projectPath)
     assertThat(project.`package`).isEqualTo(expectedPackage)
-    assertThat(project.settings).isEqualTo(expectedSettings)
+    assertThat(project.settings)
+      .usingRecursiveComparison()
+      .withComparatorForType(compareBy(Pattern::toString), Pattern::class.java)
+      .isEqualTo(expectedSettings)
```
bioball, line 123: We should implement a custom `equals` and `hashCode` method in `PklEvaluatorSettings`; we _should_ be able to use `equals` to compare these two.

stdlib/settings.pkl
```diff
@@ -22,41 +22,45 @@
...
 
 /// The [Atom](https://atom.io) editor.
-hidden Atom: Editor = new {
+hidden const Atom: Editor = new {
   urlScheme = "atom://open?url=%{url}&line=%{line}&column=%{column}"
 }
 
 /// The [Visual Studio Code](https://code.visualstudio.com) editor.
-hidden VsCode: Editor = new {
+hidden const VsCode: Editor = new {
```
bioball, line 63: Not sure why these turned into `const`? Did I do this?

stdlib/settings.pkl
```diff
@@ -22,41 +22,45 @@
 @ModuleInfo { minPklVersion = "0.26.0" }
 module pkl.settings
 
+import "pkl:EvaluatorSettings"
+
 /// The editor for viewing and editing Pkl files.
 editor: Editor = System
 
+http: EvaluatorSettings.Http?
```
bioball, line 30:
```suggestion
/// Settings for controlling how Pkl makes HTTP(S) requests.
http: EvaluatorSettings.Http?
```

stdlib/EvaluatorSettings.pkl
```diff
@@ -0,0 +1,125 @@
...
+///
+/// Relative paths are resolved against PklProject's enclosing directory.
+rootDir: String?
+
+/// Configuration of outgoing HTTP requests.
+http: Http?
+
+/// Settings that control how Pkl talks to HTTP(S) servers.
+class Http {
+  /// Configuration of the HTTP proxy to use, `null` for none.
```
bioball, line 101: Not totally true; it will use the OS proxy settings if null.

```suggestion
  /// Configuration of the HTTP proxy to use.
  ///
  /// If [null], uses the operating system's proxy configuration.
```

pkl-cli/src/test/kotlin/org/pkl/cli/CliEvaluatorTest.kt
```diff
@@ -1206,6 +1214,196 @@ result = someLib.x
...
+          noProxy = listOf("*"),
+          allowedModules = SecurityManagers.defaultAllowedModules + Pattern.compile("http:"),
+        )
+      )
+    assertThatCode { evalToConsole(options) }
+      .hasMessageContaining("I/O error loading module `$targetAddress/foo.pkl`")
+  }
+
+  @Test
+  @Disabled // TODO: Re-enable after certification logic is fixed.
```
bioball, line 1255: [nit] this is a little clearer.

```suggestion
  @Disabled // TODO: figure out why this is failing.
```

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -136,6 +136,24 @@ class BaseOptions : OptionGroup() {
...
+      .convert { URI(it) }
+
+  val noProxy: List<String>? by
+    option(
+        names = arrayOf("--no-proxy"),
+        metavar = "<pattern1,pattern2>",
+        help = "Hostnames that should not be connected to via a proxy."
+      )
+      .single()
+      .split(",")
```
bioball, line 155: I think it'd be good to only allow this flag if `--proxy` is also set. Otherwise, it can be confusing why `pkl eval --no-proxy "*"` isn't doing anything.

```suggestion
      .split(",")
      .validate { 
        require(proxy != null) {
          "--no-proxy can only be specified alongside a --proxy"
        }
      }
```

bioball (changes requested):
Generally looks good! Some minor details, see comments.

Also, still missing:

* Docs
* Server mode API changes
* Gradle API changes

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
holzensp: Rolling this back after offline conversation; proxy config can come from other sources than CLI

stdlib/settings.pkl
holzensp: From a previous iteration where I used a default value somewhere. Will revert (although they might as well be const, no? Minor ergonomics point; if so, other PR).

pkl-core/src/main/java/org/pkl/core/http/NoProxyRule.java
```diff
@@ -0,0 +1,177 @@
...
+
+  private boolean ipV4Matches(String hostname) {
+    if (ipv4 == null) {
+      return false;
+    }
+    if (!ipv4Address.matcher(hostname).matches()) {
+      return false;
+    }
+    var address = parseIpv4(hostname);
+    if (ipv4 == address) {
```
stackoverflow, line 125: Any reason to use `==` here and not `equals`? `==` only works for small integers.

pkl-core/src/main/java/org/pkl/core/settings/PklSettings.java
```diff
@@ -72,47 +70,40 @@ public static PklSettings load(ModuleSource moduleSource) throws VmEvalException
...
-
-  @Override
-  public int hashCode() {
-    return editor.hashCode();
-  }
-
   @Override
   public String toString() {
-    return "PklSettings{" + "editor=" + editor + '}';
+    return "PklSettings{" + "editor=" + editor + ";http=" + http + '}';
```
stackoverflow, line 102: nit: Java will generate a `toString` for a record, similar to this. So not really needed.

pkl-core/src/main/java/org/pkl/core/settings/PklSettings.java
```diff
@@ -134,34 +125,17 @@ public static final class Editor {
...
-    }
-
-    @Override
-    public int hashCode() {
-      return urlScheme.hashCode();
-    }
-
     @Override
     public String toString() {
       return "Editor{" + "urlScheme='" + urlScheme + '\'' + '}';
```
stackoverflow, line 141: Same here. Can be removed.

stackoverflow (changes requested):
Looks pretty good!
As Dan mentioned, we still need documentation.

holzensp (commented):
* Addressed comments
* Documentation
* Gradle API

Not included;
* Server API changes. We do not currently allow setting any of the HTTP client settings through the Server API (certs, test port). Whether/how the HTTP client settings should be set through the Server API is a topic for further discussion.
* NetUtils parser for hosts and URIs (to follow)

pkl-core/src/test/kotlin/org/pkl/core/http/RequestCapturingClient.kt
```diff
@@ -1,18 +1,41 @@
...
+    }
+  }
+
+  override fun <T : Any?> send(
     request: HttpRequest,
     responseBodyHandler: HttpResponse.BodyHandler<T>
   ): HttpResponse<T> {
     this.request = request
-    return FakeHttpResponse()
+    return dummyResponse(request)
```
bioball, line 38: Bad rebase? Don't think this diff belongs in here.

stdlib/EvaluatorSettings.pkl
```diff
@@ -0,0 +1,145 @@
...
+/// Disables the file system cache for `package:` modules.
+///
+/// When caching is disabled, packages are loaded over the network and stored in memory.
+noCache: Boolean?
+
+/// A collection of jars, zips, or directories to be placed into the module path.
+///
+/// Module path modules and resources may be read and imported using the `modulepath:` scheme.
+///
+/// Relative paths are resolved against PklProject's enclosing directory.
```
bioball, line 78: This documentation doesn't make that much sense here.

I think let's add these details to the doc comment of property `evaluatorSettings` in `pkl:Project`.

Same goes for the comments for `moduleCacheDir` and `rootDir`.

stdlib/EvaluatorSettings.pkl
```diff
@@ -0,0 +1,145 @@
...
+class Proxy {
+  /// The proxy to use for HTTP(S) connections.
+  ///
+  /// At the moment, only HTTP proxies are supported.
+  ///
+  /// Example:
+  /// ```
+  /// address = "http://my.proxy.example.com:5080"
+  /// ```
+  address: Uri(startsWith("http://"))
```
bioball, line 117: I think both this and `noProxy` should be nullable.

stdlib/Project.pkl
```diff
@@ -403,83 +404,6 @@ class Package {
...
-  ///
-  /// Relative paths are resolved against PklProject's enclosing directory.
-  moduleCacheDir: String?
-
-  /// Restricts access to file-based modules and resources to those located under this directory.
-  ///
-  /// Relative paths are resolved against PklProject's enclosing directory.
-  rootDir: String?
-}
-
```
bioball, line 482: To preserve existing code, let's add:

```suggestion
/// class EvaluatorSettings has moved to its own module.
@Deprecated { since = "0.26.0"; replaceWith = "EvaluatorSettingsModule" }
typealias EvaluatorSettings = EvaluatorSettingsModule
```

And also change the import to:

```
import "pkl:EvaluatorSettings" as EvaluatorSettingsModule
```

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/CliCommand.kt
```diff
@@ -152,9 +155,33 @@ abstract class CliCommand(protected val cliOptions: CliBaseOptions) {
...
+      } else {
+        for (file in cliOptions.normalizedCaCertificates) addCertificates(file)
+      }
+      if (cliOptions.proxyAddress != null) {
+        setProxy(cliOptions.proxyAddress, cliOptions.noProxy)
+      } else {
+        project?.evaluatorSettings?.http?.proxy?.let {
+          setProxy(it.address, it.noProxy ?: listOf())
+        }
+          ?: settings.http?.proxy?.let { setProxy(it.address, it.noProxy ?: listOf()) }
```
bioball, line 177: I think this isn't what we want, right? I think we want the `address` and `noProxy` values each to be individually composed.

This way, this command means: use proxy settings from some lower precedence (~/.pkl/settings.pkl or system or whatever), except don't proxy localhost:

```
pkl eval --no-proxy localhost myModule.pkl
```

With the way it's written currently, it will throw away the `--no-proxy` flag.

bioball (changes requested):
Leftover blocking issues:

* The treatment of `--proxy` and `--no-proxy` flags; they should be individually composed so you can set `--no-proxy` at the CLI level and grab the proxy address from system settings for example
* `System.setProperty` is called in the wrong place
* Need documentation changes for:
  - Message passing API
  - Gradle API

It would also be good to have a new section in the CLI that details how configuring the HTTP proxy works (like how we have a section that details configuring CA certificates)

pkl-core/src/test/kotlin/org/pkl/core/http/NoProxyRuleTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    assertTrue(noProxyRule1.matches(URI("https://[1000::]")))
+    assertTrue(noProxyRule1.matches(URI("https://[1000:0:ffff:ffff:ffff:ffff:ffff:ffff]")))
+    assertFalse(noProxyRule1.matches(URI("https://[999::]")))
+    assertFalse(noProxyRule1.matches(URI("https://[1000:1::]")))
+
+    val noProxyRule2 = NoProxyRule("1000::ff/128")
+    assertTrue(noProxyRule2.matches(URI("https://[1000::ff]")))
+    assertFalse(noProxyRule2.matches(URI("https://[999::]")))
+    assertFalse(noProxyRule2.matches(URI("https://[1001::]")))
+  }
```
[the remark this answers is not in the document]
holzensp: Added more test cases than the `foo.com:5000` already in there, including with ports in the rule for IPv4, IPv6, and hostnames. Included `http` (80) and `https` (443) protocol-based deduction.

Offline discussion had concluded we shouldn't support CIDR + port, because it doesn't appear to exist anywhere else (and [we-need-to-talk-no-proxy](https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/#no_proxy-1) explicitly advices against CIDR use altogether: "Avoid using CIDR matching since only Go and Ruby support that.")

stackoverflow (approved):

pkl-core/src/main/java/org/pkl/core/http/ProxySelector.java
```diff
@@ -0,0 +1,70 @@
...
+final class ProxySelector extends java.net.ProxySelector {
+
+  public static final List<Proxy> NO_PROXY = List.of(Proxy.NO_PROXY);
+
+  private final @Nullable List<Proxy> myProxy;
+  private final List<NoProxyRule> noProxyRules;
+  private final @Nullable java.net.ProxySelector delegate;
+
+  ProxySelector(@Nullable URI proxyAddress, List<String> noProxyRules) {
+    this.noProxyRules = noProxyRules.stream().map(NoProxyRule::new).toList();
```
bioball, line 36: I think let's add some validation of the proxy address here:

```suggestion
    if (proxyAddress != null
        && (!proxyAddress.getScheme().equalsIgnoreCase("http")
            || proxyAddress.getHost() == null
            || !proxyAddress.getPath().isEmpty()
            || proxyAddress.getUserInfo() != null)) {
      throw new IllegalArgumentException(
          ErrorMessages.create("malformedProxyAddress", proxyAddress));
    }
```

pkl-core/src/main/java/org/pkl/core/http/NoProxyRule.java
```diff
@@ -0,0 +1,215 @@
...
+  private static final Pattern ipv4Address = Pattern.compile("^" + ipv4AddressString + "$");
+  private static final Pattern ipv4AddressOrCidr =
+      Pattern.compile("^" + ipv4AddressString + cidrString + portString + "$");
+  private static final String ipv6AddressString =
+      "(?<host>(?:[0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}|(?:[0-9a-fA-F]{1,4}:){1,7}:|(?:[0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|(?:[0-9a-fA-F]{1,4}:){1,5}(?::[0-9a-fA-F]{1,4}){1,2}|(?:[0-9a-fA-F]{1,4}:){1,4}(?::[0-9a-fA-F]{1,4}){1,3}|(?:[0-9a-fA-F]{1,4}:){1,3}(?::[0-9a-fA-F]{1,4}){1,4}|(?:[0-9a-fA-F]{1,4}:){1,2}(?::[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:(?::[0-9a-fA-F]{1,4}){1,6}|:(?:(?::[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(?::[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]+|::(?:ffff(:0{1,4})?:)?(?:(?:25[0-5]|(2[0-4]|1?[0-9])?[0-9])\\.){3}(?:25[0-5]|(?:2[0-4]|1?[0-9])?[0-9])|(?:[0-9a-fA-F]{1,4}:){1,4}:(?:(?:25[0-5]|(?:2[0-4]|1?[0-9])?[0-9])\\.){3}(?:25[0-5]|(?:2[0-4]|1?[0-9])?[0-9]))";
+  private static final Pattern ipv6AddressOrCidr =
+      Pattern.compile(
+          "^(?<open>\\[)?" + ipv6AddressString + cidrString + "(?<close>])?" + portString + "$");
```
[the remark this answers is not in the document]
holzensp: They are, but this is the cheapest way of implementing that. Depending on pedantry levels desired, we can verify that either one or the other is present.

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -827,3 +827,18 @@ These certificates can be overridden via either of the two options:
...
+By default, no proxy is configured.
+A proxy can be configured as follows:
+
+- Using `--proxy <uri>`.
+  The URI must (currently) have scheme `http`, and may not contain anything other than a host and port.
+- Using `--no-proxy <hosts>`.
+  The given (comma separated list of) hosts bypass the proxy.
+  Hosts can be specified by domain name (in which case all subdomains also bypass the proxy), IP addresses, or IP ranges (using link:https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#CIDR_notation[CIDR notation]).
+  For individual hosts (not CIDRs), ports can be specified.
+  When no port is specified for a given host, connections to all ports bypass the proxy.
```
bioball, line 844: By default, Pkl actually uses the system proxy settings.

I think something that's missing here is some documentation about how Pkl resolves proxy settings. E.g. the stuff about looking in the CLI flags, then PklProject, then the settings file.

BTW: the language here describes the settings as CLI flags, which is a little misleading (no proxy within `settings.pkl` isn't a comma-separated list).

bioball (changes requested):
Getting close!

stdlib/settings.pkl
```diff
@@ -22,9 +22,14 @@
 @ModuleInfo { minPklVersion = "0.26.0" }
 module pkl.settings
 
+import "pkl:EvaluatorSettings"
+
 /// The editor for viewing and editing Pkl files.
 editor: Editor = System
 
+/// Settings for controlling how Pkl makes HTTP(S) requests.
+http: EvaluatorSettings.Http?
```
bioball, line 31:
```suggestion
@Since { version = "0.26.0" }
http: EvaluatorSettings.Http?
```

bioball (approved):
LGTM!

PR #507, opened 2024-05-28
Fix `pathElements`; do not default to empty list

holzensp:
Extracted from #472, because it is unrelated.


bioball:
This change was separately made in https://github.com/apple/pkl/pull/480. Was there a particular reason that we wanted to revert it?

holzensp:
Nope, this was my bad. I was behind and got the direction of change wrong.

PR #509, opened 2024-05-29
Improve configuration of native image compilation

bioball:
This is supercedes and closes #455

Thanks to @translatenix for doing 99% of the work here

* Remove unnecessary `--add-opens` by deferring initialization of MessagePack library at build time
* Use quick build mode for 40% faster compilation and 20% smaller executable when not creating release builds
* Add test for native executables when running in server mode

### DIFF
--- a/pkl-core/pkl-core.gradle.kts
+++ b/pkl-core/pkl-core.gradle.kts
@@ -208,17 +208,7 @@ val testAlpineExecutableAmd64 by tasks.registering(Test::class) {
 
 val testWindowsExecutableAmd64 by tasks.registering(Test::class) {
   dependsOn(":pkl-cli:windowsExecutableAmd64")
-
-  inputs.dir("src/test/files/LanguageSnippetTests/input")
-  inputs.dir("src/test/files/LanguageSnippetTests/input-helper")
-  inputs.dir("src/test/files/LanguageSnippetTests/output")
-
-  testClassesDirs = files(tasks.test.get().testClassesDirs)
-  classpath = tasks.test.get().classpath
-
-  useJUnitPlatform {
-    includeEngines("WindowsLanguageSnippetTestsEngine")
-  }
+  configureExecutableTest("WindowsLanguageSnippetTestsEngine")
 }
 
 tasks.testNative {
--- a/pkl-core/pkl-core.gradle.kts
+++ b/pkl-core/pkl-core.gradle.kts
@@ -154,10 +154,7 @@ tasks.compileKotlin {
 }
 
 tasks.test {
-  inputs.dir("src/test/files/LanguageSnippetTests/input").withPropertyName("languageSnippetTestsInput").withPathSensitivity(PathSensitivity.RELATIVE)
-  inputs.dir("src/test/files/LanguageSnippetTests/input-helper").withPropertyName("languageSnippetTestsInputHelper").withPathSensitivity(PathSensitivity.RELATIVE)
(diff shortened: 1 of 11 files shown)


pkl-core/pkl-core.gradle.kts
```diff
@@ -316,3 +255,24 @@ spotless {
...
+    .withPathSensitivity(PathSensitivity.RELATIVE)
+  inputs.dir("src/test/files/LanguageSnippetTests/input-helper")
+    .withPropertyName("languageSnippetTestsInputHelper")
+    .withPathSensitivity(PathSensitivity.RELATIVE)
+  inputs.dir("src/test/files/LanguageSnippetTests/output")
+    .withPropertyName("languageSnippetTestsOutput")
+    .withPathSensitivity(PathSensitivity.RELATIVE)
+}
+
+private fun Test.configureExecutableTest(engineName: String) {
```
stackoverflow, line 271: Any reason why `testWindowsExecutableAmd64` doesn't call this function, just mac and linux?

pkl-server/src/test/kotlin/org/pkl/server/JvmServerTest.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.junit.jupiter.api.BeforeEach
+import org.pkl.core.http.HttpClient
+
+class JvmServerTest : AbstractServerTest() {
+  companion object {
+    private const val USE_DIRECT_TRANSPORT = false
+  }
+
+  private val transports: Pair<MessageTransport, MessageTransport> = run {
+    if (USE_DIRECT_TRANSPORT) {
```
stackoverflow, line 30: This is always `false`, why have an `if`?

stackoverflow (approved):
Looks good. Just some minor comments.

pkl-core/pkl-core.gradle.kts
bioball: Good catch!

pkl-server/src/test/kotlin/org/pkl/server/JvmServerTest.kt
bioball: This is just here to allow us to test the same logic if we wanted to run the server tests with an encoder/decoder. It's for development purposes. I'll add a note here.

pkl-server/src/test/kotlin/org/pkl/server/JvmServerTest.kt
```diff
@@ -0,0 +1,55 @@
...
+import java.io.PipedInputStream
+import java.io.PipedOutputStream
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.BeforeEach
+import org.pkl.core.http.HttpClient
+
+class JvmServerTest : AbstractServerTest() {
+  companion object {
+    private const val USE_DIRECT_TRANSPORT = false
+  }
```
StefMa, line 27: What are the reasons of using a companion object instead of a normal constant in that kt file?

StefMa: Yes exactly, that was my question 🙃.
But actually you end up with a `Foo.Companion` 🤷‍♂️ 😂

bioball: Yeah, true. We already need the companion class, and we _could_ mark them `@JvmStatic` to get around it. But, I think this would be splitting hairs a little bit.

PR #512, opened 2024-06-03
Remove documentation references to `packageInfos`

KushalP:
It references something that isn't present and should be removed.

```
❯ rg packageInfos
pkl-doc/src/main/kotlin/org/pkl/doc/DocGenerator.kt
48:   * A function to resolve imports in [modules], [packageInfos], and [docsiteInfo].
```


bioball (approved):
LGTM

PR #514, opened 2024-06-04
Fix native tests in CircleCI

bioball:
Fix an issue where the CI job attempts to install the wrong executable when on macOS/aarch64.

Also, adjusts native test definitions in pkl-server to match pkl-core.

### DIFF
--- a/pkl-server/pkl-server.gradle.kts
+++ b/pkl-server/pkl-server.gradle.kts
@@ -24,15 +24,62 @@ tasks.test {
   exclude("**/NativeServerTest.*")
 }
 
-val nativeTest by tasks.registering(Test::class) {
-  dependsOn(":pkl-cli:assembleNative")
+private fun Test.configureNativeTest() {
   testClassesDirs = files(tasks.test.get().testClassesDirs)
   classpath = tasks.test.get().classpath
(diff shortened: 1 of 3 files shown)


pkl-server/pkl-server.gradle.kts
```diff
@@ -24,15 +24,62 @@ tasks.test {
...
+}
+
 val testNative by tasks.existing
 testNative {
-  dependsOn(nativeTest)
+  when {
+    buildInfo.os.isMacOsX -> {
+      dependsOn(testMacExecutableAarch64)
+      if (buildInfo.arch == "aarch64") {
+        dependsOn(testMacExecutableAmd64)
```
stackoverflow, line 69: We depend on `amd64` if the arch is `aarch64`? I'm not following the logic here. Shouldn't this be done like linux with separate bodies for each architecture?

stackoverflow (approved):

pkl-server/pkl-server.gradle.kts
bioball: Ah, this logic was flipped around. Will fix.

We cross-build when on macOS, thanks to Rosetta. We can't do this on Linux, though.

PR #516, opened 2024-06-05
Require references from typealiases to be const

bioball:
This adds a language change that requires references from typealiases to the enclosing module to be `const`.

This is required because typealiases are not late-bound.

Rationale is laid out in [SPICE-0007](https://github.com/apple/pkl-evolution/pull/8).

Also:
* Update documentation to reflect new rules.
* Fix `Project.pkl`; mark method `const` to not break said rule.


stackoverflow (approved):
Looks good!

holzensp (approved):

PR #518, opened 2024-06-11
Update: Improve handling of CA certificates

bioball:
These are changes made on top of https://github.com/apple/pkl/pull/378. Thanks to @translatenix for doing most of the work here!

Changes on top of the original PR:

* Add ability to set CA certificates as a byte array
* Add CA certificates option to message passing API
* Revert logic to treat `--ca-certificate` flag as either directory or file

Closes #378


docs/modules/bindings-specification/pages/message-passing-api.adoc
```diff
@@ -189,7 +192,17 @@ class Checksums {
   /// The sha-256 checksum of this dependency's metadata.
   sha256: String
 }
+
+class Http {
+  /// Certificates to trust when making HTTP requests.
+  ///
+  /// If [null], Pkl will trust its own built-in certificates.
+  caCertificates: Binary?
```
bioball, line 200: At first, I considered accepting a path to a cert file here. However, this doesn't work so well if a client wants to trust system certs on macOS, because macOS's system certs are managed by Security.framework, and not stored as files on disk. Accepting certs directly as binary seems like a better choice for the message passing API.

pkl-core/src/main/java/org/pkl/core/http/HttpClientBuilder.java
```diff
@@ -67,39 +57,14 @@ public HttpClient.Builder setRequestTimeout(Duration timeout) {
...
-    if (certificateFiles.size() == fileCount) {
-      addBuiltInCertificates();
-    }
-    return this;
-  }
-
-  @Override
-  public HttpClient.Builder addBuiltInCertificates() {
-    certificateUris.add(getBuiltInCertificates());
+  public Builder addCertificates(byte[] certificateBytes) {
```
stackoverflow, line 66: Calling this `addCertificates` is misleading, as calling this method more than once will just reset the bytes, not add them like `addCertificates(Path)`. Either have a `List<ByteBuffer>` or rename it to `setCertificates`.

pkl-server/src/test/kotlin/org/pkl/server/AbstractServerTest.kt
```diff
@@ -1004,6 +1004,8 @@ abstract class AbstractServerTest {
       )
   }
 
+  @Test fun `custom ca certs`(@TempDir tempDir: Path) {}
```
stackoverflow, line 1007: This is not testing anything.

stackoverflow (approved):
Looks good. Some minor comments.

docs/modules/bindings-specification/pages/message-passing-api.adoc
HT154, line 200: If I'm reading this right, this is expected to be binary data containing PEM data. Am I right? I think it might make sense for that to be in the specification.

bioball: Yup, good call, will update

pkl-core/src/main/java/org/pkl/core/http/HttpClientBuilder.java
bioball: Changed this to have append semantics.

PR #519, opened 2024-06-11
Use Pkl syntax highlighting instead of Groovy

bioball:
Now that Pkl syntax highlighting is available on GitHub, we can remove the Groovy association with these files.

This also adds changes to mark `*.pcf` and `PklProject` files as Pkl.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,7 +3,9 @@
 
 /docs/** linguist-documentation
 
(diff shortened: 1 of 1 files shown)


stackoverflow (approved):

holzensp (approved):
Legogogo!

PR #523, opened 2024-06-12
Adjust docs for proxying

bioball:


docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -831,16 +831,31 @@ The CLI option takes precedence over the certificates in `~/.pkl/cacerts/`. +
...
-- Using `--no-proxy <hosts>`.
-  The given (comma separated list of) hosts bypass the proxy.
-  Hosts can be specified by domain name (in which case all subdomains also bypass the proxy), IP addresses, or IP ranges (using link:https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#CIDR_notation[CIDR notation]).
-  For individual hosts (not CIDRs), ports can be specified.
-  When no port is specified for a given host, connections to all ports bypass the proxy.
+== HTTP Proxying
+
+When making HTTP(s) requests, Pkl can possibly make use of an HTTP proxy.
+
+When determining proxy settings, Pkl will look at the following locations, in order of precedence (lowest to highest):
```
holzensp, line 838: This suggests (to me, at least) that if any higher priority defines any part of the proxy, definitions at lower priorities are ignored. It's worth making the point that this holds separately for the proxy address and no-proxy definitions.

holzensp (approved):
Approving to unblock, but I think a sentence or two about the different parts of the proxy definition would be useful.

docs/modules/pkl-cli/pages/index.adoc
bioball: Added more details here

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -831,16 +831,46 @@ The CLI option takes precedence over the certificates in `~/.pkl/cacerts/`. +
...
+
+[NOTE]
+====
+The proxy and noproxy values are individually set.
+For example, using the `--no-proxy` flag but not the `--proxy` flag will cause Pkl to look at the PklProject file, then the settings file, then system settings for the proxy address.
+
+One exception to this rule is that setting a proxy address will cause Pkl to ignore any noproxy values set at the OS level.
+====
+
+Pkl only supports HTTP proxies, and does not support HTTPS nor SOCKS proxies.
```
holzensp, line 854: My non-native speakerness may invalidate this point, but isn't it "neither... nor" _or_ "not ... or"?
```suggestion
Pkl only supports HTTP proxies, so neither HTTPS nor SOCKS proxies are supported.
```

holzensp (approved):

PR #524, opened 2024-06-13
Fix links to dependencies

bioball:
This fixes an issue where package dependencies are not rendered correctly.


pkl-doc/src/main/kotlin/org/pkl/doc/PageGenerator.kt
```diff
@@ -525,15 +525,30 @@ internal abstract class PageGenerator<out S>(
...
+              pageScope.siteScope?.packageScopes?.values?.find {
+                it.docPackageInfo.name == dep.name
+              }
+            if (localSitePackage != null) {
+              a {
+                href =
+                  pageScope.relativeSiteUrl
+                    .resolve(
+                      "${localSitePackage.docPackageInfo.name.pathEncoded}/current/index.html"
+                    )
```
bioball, line 545: This will only show a local link if the package is known to Pkldoc, and it will link to "current" rather than the depended upon version. If the package is not known, _and_ we don't know what its documentation URL is, this will just display plain text.

Note: I considered making this link directly to the version, e.g. `"${localSitePackage.docPackageInfo.name.pathEncoded}/${localSitePackage.docPackageInfo.version.pathEncoded}/index.htm`

However, we don't know if that version exists in the doc site. The page generator doesn't know about all the versions of each package, and we don't build this index until _after_ the static HTML has been built.

holzensp (approved):
No blockers; one possible smell

bioball:
This actually introduces a new problem--links to modules of a package dependency is now broken. I'm going to hold off on this for now.

PR #528, opened 2024-06-13
Change Gradle property from "packageMapping" to "renames"

bioball:
This changes the property to match the name of the CLI flag.

FYI @netvl


holzensp (approved):

pkl-gradle/src/main/java/org/pkl/gradle/spec/CodeGenSpec.java
```diff
@@ -32,5 +32,5 @@ public interface CodeGenSpec extends ModulesSpec {
 
   Property<Boolean> getImplementSerializable();
 
-  MapProperty<String, String> getPackageMapping();
+  MapProperty<String, String> getRenames();
```
netvl, line 35: Totally forgot about the Gradle plugin, sorry :(

PR #542, opened 2024-06-18
Pass `HOMEBREW_*` environment variables on through `native_image`

holzensp:
We currently have to use `NATIVE_IMAGE_DEPRECATED_BUILDER_SANITATION` to let builds pass in Homebrew. This is because `native-image` runs its child processes (such as the platform-dependent C compiler) in a "clean" environment. In Homebrew, such compilers (`clang`, `gcc`, etc) are wrapper scripts that first check the Homebrew environment is intact.

GraalVM folks have warned that `NATIVE_IMAGE_DEPRECATED_BUILDER_SANITATION` will go away.

This PR preserves all environment variables starting with `HOMEBREW_` for the inner `native-image` environment.

### DIFF
--- a/pkl-cli/pkl-cli.gradle.kts
+++ b/pkl-cli/pkl-cli.gradle.kts
@@ -241,6 +241,8 @@ fun Exec.configureExecutable(
       val processors = Runtime.getRuntime().availableProcessors() /
         if (buildInfo.os.isMacOsX && !buildInfo.isCiBuild) 4 else 1
       add("-J-XX:ActiveProcessorCount=${processors}")
+      // Pass through all `HOMEBREW_` prefixed environment variables to allow build with shimmed tools.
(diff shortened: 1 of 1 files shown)


bioball (approved):
Generally LGTM.

Question: Is this enough for Homebrew? Have we tested that this works?

The old behavior can be found here:
https://github.com/oracle/graal/blob/5e9a8f75fabf5ffddf14bb39bed806b4d0151450/substratevm/src/com.oracle.svm.driver/src/com/oracle/svm/driver/NativeImage.java#L1773-L1778

It passes through everything except for "JAVA_COMPILER", "_JAVA_OPTIONS", "JAVA_TOOL_OPTIONS", "JDK_JAVA_OPTIONS", "CLASSPATH".

holzensp:
It should. Will follow-up after more testing if necessary.

PR #552, opened 2024-06-26
Add colours to Pkl errors in Cli output

thomaspurchas:
To make error messages from Pkl eval easier to read, this change uses the Jansi library to colour the output, making it quicker and easier to scan error messages and understand what's happened.

The Jansi library also detects if the CLI output is a terminal capable of handling colours, and will automatically strip out escape codes if the output won't support them (e.g. piping the output somewhere else).

The implementation is a little crude, but think acceptable given a better approach would probably require rewriting how errors are handled completely, so that the output composition happens near the frontend of the CLI, rather than down in the bowls of Pkl. The CLI frontend is in a much better position to determine terminal capabilities, and adjust outputs as needed, and would allow us to use a much more expressive library like [Mordant](https://ajalt.github.io/mordant/) to really supercharge the output formatting.

image
image
image


thomaspurchas:
@holzensp that makes sense. I'll try and find a better way to disabling the colour output during tests, ideally it would be disabled by default (and without the need for the `removeColors` helper), except for tests looking to test the colour output, but I haven't managed to nail down a good approach yet.

thomaspurchas:
@holzensp I've rewritten how tests are handled, and managed to configure the test environment to complete remove the need for any special handling of colours in normal tests, and I've added a some tests that specifically test the colour outputs.

pkl-core/src/main/java/org/pkl/core/runtime/StackTraceRenderer.java
```diff
@@ -15,15 +15,23 @@
...
+import org.fusesource.jansi.Ansi.Color;
 import org.pkl.core.StackFrame;
 import org.pkl.core.util.Nullable;
 
 public final class StackTraceRenderer {
   private final Function<StackFrame, StackFrame> frameTransformer;
 
+  private static final Ansi.Color frameColor = Color.YELLOW;
+  private static final Ansi.Color lineNumColor = Color.BLUE;
+  private static final Ansi.Color repColor = Color.MAGENTA;
```
holzensp, line 33: Happy to see you're choosing colours that don't wobble to us deuteranomalous folks; that said... should these be configurable? More generally, I feel it's probably wise to put all the colour definitions in one place.

Besides that: I'm not _entirely_ certain what `rep` is the abbreviation of... guessing here;
```suggestion
  private static final Ansi.Color frameColor = Color.YELLOW;
  private static final Ansi.Color lineNumColor = Color.BLUE;
  private static final Ansi.Color repetitionColor = Color.MAGENTA;
```

pkl-core/src/main/java/org/pkl/core/runtime/StackTraceRenderer.java
```diff
@@ -40,27 +48,36 @@ void doRender(
...
-          builder.append(leftMargin).append("┌─ ").append(loop.count).append(" repetitions of:\n");
+          out.fgBright(frameColor)
+              .a(leftMargin)
+              .a("┌─ ")
+              .reset()
+              .bold()
+              .fg(repColor)
+              .a(Integer.toString(loop.count))
+              .reset()
+              .a(" repetitions of:\n");
```
holzensp, line 69: I'm quite happy with how (relatively) well the formatting inserts scale here.

pkl-core/src/main/java/org/pkl/core/runtime/VmExceptionRenderer.java
```diff
@@ -73,7 +75,8 @@ private void renderBugException(VmBugException exception, StringBuilder builder)
   }
 
   private void renderException(VmException exception, StringBuilder builder) {
-    var header = "–– Pkl Error ––";
+    var out = ansi(builder);
+    out.render("@|red –– Pkl Error ––|@");
```
holzensp, line 79: Re: defining all colours in one place; this (and below) hard-coded redness definitely included in that.

pkl-core/src/test/kotlin/org/pkl/core/ErrorColoringTest.kt
```diff
@@ -0,0 +1,51 @@
...
+    @JvmStatic
+    fun afterAll() {
+      evaluator.close()
+    }
+  }
+
+  private fun evaluate(program: String, expression: String): Any {
+    return evaluator.evaluateExpression(ModuleSource.text(program), expression)
+  }
+  @BeforeEach
```
holzensp, line 21:
```suggestion
  }

  @BeforeEach
```

nit (surprised Spotless doesn't balk at this)

pkl-core/src/test/kotlin/org/pkl/core/ErrorColoringTest.kt
```diff
@@ -0,0 +1,51 @@
...
+  }
+  
+  @Test
+  fun `repeated error`() {
+    val error = assertThrows<PklException> { evaluate("""self: String = "Strings; if they were lazy, you could tie the knot on \(self.take(7))"""", "self") }
+    assertThat(error).hasMessageContaining("[91mA stack overflow occurred.[m")
+    assertThat(error).hasMessageContaining("[93m┌─ [0;1;35m")
+    assertThat(error).hasMessageContaining("[m repetitions of:")
+    assertThat(error).hasMessageContaining("[93m│ [94m1 | [m")
+    assertThat(error).hasMessageContaining("[0;31m^^^^[m")
```
holzensp, line 49: let's make these slightly nicer
```suggestion
    assertThat(error).message()
      .contains("\u001B[31m–– Pkl Error ––\u001B[m")
      .contains("\u001B[94m1 | \u001B[m")
      .contains("\u001B[0;31m^")
  }
  
  @Test
  fun `repeated error`() {
    val error = assertThrows<PklException> { evaluate("""self: String = "Strings; if they were lazy, you could tie the knot on \(self.take(7))"""", "self") }
    assertThat(error).message()
      .contains("[91mA stack overflow occurred.[m")
      .contains("[93m┌─ [0;1;35m")
      .contains("[m repetitions of:")
      .contains("[93m│ [94m1 | [m")
      .contains("[0;31m^^^^[m")
```

holzensp (changes requested):
_Love_ it! A few nits to figure out. Also, we currently only do native-image builds on CI in prerelease; have you tested building native images locally?

pkl-core/src/main/java/org/pkl/core/runtime/StackTraceRenderer.java
thomaspurchas: Yeah `rep` is `repetition`. The colour choice being deuteranomalous friendly was more by accident than design. Was trying to stick to the basic colours that have very wide support, rather than trying to get fancy. Also means that colours will appear slightly different in different terminals, and I imagine that most terminals support customising the precise RGB values of each colour.

Wouldn't recommend making these configurable right now. The formatting API doesn't really lend itself to making that easy as the final colour is determined by the base colour here, but also the exact method used to apply the colour e.g. `fgBright()` vs `fg`. If we were serious about allowing colour configuration, then I think we should also look at reworking how these outputs are generated, and moving this formatting out of `pkl-core` and into `pkl-cli-common`, and adding extra info to the `PklException` so that those layers can access the underlying failure and render a completely different output based on terminal capability and user config.

thomaspurchas: Yeah `rep` is `repetition`. The colour choice being deuteranomalous friendly was more by accident than design. Was trying to stick to the basic colours that have very wide support, rather than trying to get fancy. Also means that colours will appear slightly different in different terminals, and I imagine that most terminals support customising the precise RGB values of each colour.

Wouldn't recommend making these configurable right now. The formatting API doesn't really lend itself to making that easy as the final colour is determined by the base colour here, but also the exact method used to apply the colour e.g. `fgBright()` vs `fg`. If we were serious about allowing colour configuration, then I think we should also look at reworking how these outputs are generated, and moving this formatting out of `pkl-core` and into `pkl-cli-common`, and adding extra info to the `PklException` so that those layers can access the underlying failure and render a completely different output based on terminal capability and user config.

thomaspurchas: Yeah `rep` is `repetition`. The colour choice being deuteranomalous friendly was more by accident than design. Was trying to stick to the basic colours that have very wide support, rather than trying to get fancy. Also means that colours will appear slightly different in different terminals, and I imagine that most terminals support customising the precise RGB values of each colour.

Wouldn't recommend making these configurable right now. The formatting API doesn't really lend itself to making that easy as the final colour is determined by the base colour here, but also the exact method used to apply the colour e.g. `fgBright()` vs `fg`. If we were serious about allowing colour configuration, then I think we should also look at reworking how these outputs are generated, and moving this formatting out of `pkl-core` and into `pkl-cli-common`, and adding extra info to the `PklException` so that those layers can access the underlying failure and render a completely different output based on terminal capability and user config.

thomaspurchas: Yeah `rep` is `repetition`. The colour choice being deuteranomalous friendly was more by accident than design. Was trying to stick to the basic colours that have very wide support, rather than trying to get fancy. Also means that colours will appear slightly different in different terminals, and I imagine that most terminals support customising the precise RGB values of each colour.

Wouldn't recommend making these configurable right now. The formatting API doesn't really lend itself to making that easy as the final colour is determined by the base colour here, but also the exact method used to apply the colour e.g. `fgBright()` vs `fg`. If we were serious about allowing colour configuration, then I think we should also look at reworking how these outputs are generated, and moving this formatting out of `pkl-core` and into `pkl-cli-common`, and adding extra info to the `PklException` so that those layers can access the underlying failure and render a completely different output based on terminal capability and user config.

bioball:
This looks great! Thanks for the PR!

With outputting colors in our console, I think we will also want a `--color` flag, where it accepts "never", "always", and "auto" (auto means: output colors if stdout/stderr are TTY).

> a better approach would probably require rewriting how errors are handled completely, so that the output composition happens near the frontend of the CLI, rather than down in the bowls of Pkl. The CLI frontend is in a much better position to determine terminal capabilities, and adjust outputs as needed, and would allow us to use a much more expressive library like [Mordant](https://ajalt.github.io/mordant/) to really supercharge the output formatting.

I don't think we really need a whole revamp of error messages here. Rather, the CLI can pass a text formatter into the APIs within pkl-core. This is already a pattern that we use (see `org.pkl.core.StackFrameTransformer`).

Also: it would be good to avoid adding jansi as a dependency in pkl-core. This would be another benefit of providing a "text formatter" interface, of which we can have an implementation that outputs Ansi colors, and an implementation that strips colors.

These comments don't necessarily need to block the PR, but we'll probably want to address this before we ship 0.27.

thomaspurchas:
> I don't think we really need a whole revamp of error messages here. Rather, the CLI can pass a text formatter into the APIs within pkl-core. This is already a pattern that we use (see org.pkl.core.StackFrameTransformer).

Happy to explore those options once we've got this merged. Agree it would be good to avoid the Jansi dependence in Pkl-core and find a better way to cleanly separate these concerns. I do also want to look at whole error generation and formatting process, allowing errors to propagate more context that the formatting process can take advantage, so colour hints etc can be embedded into the error messages themselves.

> Color suggestion: the red error message seems a little harsh to me. How about making it bold? Something like this (here, both "Pkl Error" and the message are bold)

We can adjust this, but it's worth remembering that the exact colours used here are determined by your terminal not by Pkl. So the "harshness" will depend quite a bit on the terminal config, so there not a clearcut good set of colour choices. I think the red matches peoples expectations of what error messages look like, so I don't think it worth changing, and it's aways easy to change it in the future based on feedback we get.

image

image

image

image

bioball:
Thanks for the additional screenshots. Albeit, I still feel that it's a little harsh in each of those screen grabs.

PR #553, opened 2024-06-26
Do not enable TLS certificate revocation checks by default

bioball:
This addresses an issue where network requests may fail if cert revocation checks error, which may occur due to availability issues, or due to lack of internet access.

This changes the HTTP client to not check revocation on TLS certificates by default.
Users of the Java client can enable this by setting Java property `com.sun.net.ssl.checkRevocation` to `true`.

Users of the native executables will not perform revocation checks.


stackoverflow (approved):

pkl-core/src/main/java/org/pkl/core/http/JdkHttpClient.java
```diff
@@ -130,43 +125,36 @@ private static SSLContext createSslContext(
...
-      var revocationChecker = (PKIXRevocationChecker) certPathBuilder.getRevocationChecker();
-      revocationChecker.setOptions(Set.of()); // prefer OCSP, fall back to CRLs
-
       var certFactory = CertificateFactory.getInstance("X.509");
-      Set<TrustAnchor> trustAnchors =
-          createTrustAnchors(certFactory, certificateFiles, certificateBytes);
-      var pkixParameters = new PKIXBuilderParameters(trustAnchors, new X509CertSelector());
-      // equivalent of "com.sun.net.ssl.checkRevocation=true"
-      pkixParameters.setRevocationEnabled(true);
-      pkixParameters.addCertPathChecker(revocationChecker);
```
bioball, line 149: Using `PKIXBuilderParameters` results in revocation checking being enabled by default when testing this locally.

So, I changed this to using `KeyStore` instead, which results in the Java client looking for property `com.sun.net.ssl.checkRevocation` before it does any revocation checks.

PR #570, opened 2024-07-07
Add CSS to make top doc box have a fade, rather than hard cutoff

thomaspurchas:
The current hard cutoff in the docs often results in people not realising that the doc box can be expanded, and often resulting in confusion because the most helpful examples are often in the module doc box.

This change uses some simple CSS tweaks to replace the hard cut-off with a visualfade, so it's obvious that there's content hidden out of view. Doing this required removing the CSS transition, as it hard to correctly transition the height property of CSS element of unknown target height. But the improved discoverablility of the doc content seems like a worthwhile tradeoff.

Before:
image

After:
image

### DIFF
--- a/pkl-doc/src/main/resources/org/pkl/doc/styles/pkldoc.css
+++ b/pkl-doc/src/main/resources/org/pkl/doc/styles/pkldoc.css
@@ -518,6 +518,20 @@ because the entire .member.with-page-link is effectively a link (via JS).
   display: none;
 }
 
+#_declaration .expandable {
+  transform: none;
+  transition: none;
+}
+
+#_declaration .expandable.collapsed {
+  mask: linear-gradient(rgb(0 0 0), transparent) content-box;
+  height: 100px;
+}
+
+#_declaration .expandable.hidden {
+  display: block;
+}
+
 /* show an otherwise hidden inherited member if it's a link target */
 .anchor:target + .expandable.collapsed.hidden {
   display: inherit;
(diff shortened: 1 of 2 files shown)


StefMa (commented):
Nice one!
I also stumbled over this in the past.

However, I am not sure if there is an easier way to make this more visible to users... 🤔

My idea would be to move the arrow from top left to bottom center. I guess this is where the most people would expect an "expand/collapse indicator", didn't they?

I am not an ux/ui expert. I also have only limited knowledge about css. Just wanted to share my thoughts on this. ☺

thomaspurchas:
@StefMa agree there better ways to do this. But this change only requires a few lines of CSS, so it’s quick and easy todo. I don’t consider this “good”, just better than what we have, and a solution that I could implement quickly in my spare time.

If people have suggestions for improvements, please submit a PR!

holzensp (approved):
I agree with @StefMa; moving the carrot might be an even bigger improvement. That said, this is undoubtedly already an improvement for discoverability, so... if anyone wants to work on further improvements, yes please. Until then; let's ship this.

Thanks @thomaspurchas!

StefMa:
Just wanted to refer to [this lines](https://github.com/apple/pkl/blob/main/pkl-doc%2Fsrc%2Fmain%2Fkotlin%2Forg%2Fpkl%2Fdoc%2FPageGenerator.kt#L679-L687) as I guess this is relevant in case we want to move the arrow 🙃

PR #571, opened 2024-07-07
Add a --color flag to the CLI and disable colors for Pkl test

thomaspurchas:
The --colors flag makes it possible to forcefully disable colour output, and also forcefully enable it. We make use of this in the native Pkl test engines by passing --colors=never into the pkl commands, which ensures we get a predictable output.

Additionally the pkl test command also sets --colors=never to ensure that pkl test output isn't accidentially contaminated with ansi escape chars, which will break example based tests.

### DIFF
--- a/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
+++ b/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
@@ -17,6 +17,7 @@ package org.pkl.commons.cli.commands
 
 import com.github.ajalt.clikt.parameters.groups.OptionGroup
 import com.github.ajalt.clikt.parameters.options.*
+import com.github.ajalt.clikt.parameters.types.choice
 import com.github.ajalt.clikt.parameters.types.int
 import com.github.ajalt.clikt.parameters.types.long
 import com.github.ajalt.clikt.parameters.types.path
@@ -198,6 +199,11 @@ class BaseOptions : OptionGroup() {
       .single()
       .split(",")
 
+  val colors: String by
+    option(names = arrayOf("--colors"), help = "Enable or disable colour output in the terminal")
+      .choice("auto", "never", "always")
+      .default("auto")
+
   // hidden option used by native tests
   private val testPort: Int by
     option(names = arrayOf("--test-port"), help = "Internal test option", hidden = true)
@@ -208,7 +214,8 @@ class BaseOptions : OptionGroup() {
   fun baseOptions(
     modules: List<URI>,
     projectOptions: ProjectOptions? = null,
-    testMode: Boolean = false
+    testMode: Boolean = false,
+    disableColors: Boolean = false,
   ): CliBaseOptions {
     return CliBaseOptions(
       sourceModules = modules,
@@ -230,7 +237,8 @@ class BaseOptions : OptionGroup() {
       noProject = projectOptions?.noProject ?: false,
       caCertificates = caCertificates,
       httpProxy = proxy,
-      httpNoProxy = noProxy ?: emptyList()
+      httpNoProxy = noProxy ?: emptyList(),
+      colors = if (disableColors) "never" else colors,
     )
   }
 }
--- a/pkl-gradle/src/main/java/org/pkl/gradle/task/BasePklTask.java
+++ b/pkl-gradle/src/main/java/org/pkl/gradle/task/BasePklTask.java
@@ -166,7 +166,8 @@ public abstract class BasePklTask extends DefaultTask {
               getTestPort().getOrElse(-1),
               Collections.emptyList(),
               getHttpProxy().getOrNull(),
-              getHttpNoProxy().getOrElse(List.of()));
+              getHttpNoProxy().getOrElse(List.of()),
+              "auto");
     }
     return cachedOptions;
   }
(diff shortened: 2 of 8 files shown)


pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -230,7 +237,8 @@ class BaseOptions : OptionGroup() {
       noProject = projectOptions?.noProject ?: false,
       caCertificates = caCertificates,
       httpProxy = proxy,
-      httpNoProxy = noProxy ?: emptyList()
+      httpNoProxy = noProxy ?: emptyList(),
+      colors = if (disableColors) "never" else colors,
```
bioball, line 241:
```suggestion
      color = color
```

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -208,7 +214,8 @@ class BaseOptions : OptionGroup() {
   fun baseOptions(
     modules: List<URI>,
     projectOptions: ProjectOptions? = null,
-    testMode: Boolean = false
+    testMode: Boolean = false,
+    disableColors: Boolean = false,
```
bioball, line 218:
```suggestion
```

pkl-gradle/src/main/java/org/pkl/gradle/task/BasePklTask.java
```diff
@@ -166,7 +166,8 @@ protected CliBaseOptions getCliBaseOptions() {
               getTestPort().getOrElse(-1),
               Collections.emptyList(),
               getHttpProxy().getOrNull(),
-              getHttpNoProxy().getOrElse(List.of()));
+              getHttpNoProxy().getOrElse(List.of()),
+              "auto");
```
bioball, line 170: I'm pretty sure this means that Gradle will not output any colors, because this is executed as a child process.

Anyway, we should provide the same options in Gradle (auto, never, always). Take a look at how the other options are configured to see what to do here.

Or, leave a TODO comment here so we can follow up in a future PR.

bioball (changes requested):
Some minor comments!

Also, this is missing documentation (cli-common-options.adoc, gradle-common-properties.adoc). Do you mind adding them?

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
thomaspurchas: This exists to provide a mechanism for the `test` subcommand to forcefully disable not just colours, but also the injection if ANSI codes completely. I've been having issues because the `pkl test` expected output because they don't travel along the standard output rails. As a consequence the ANSI codes don't get stripped if they're injected.

By providing this escape hatch, I can ensure that colour are disabled _before_ there's any opportunity for the Jansi formatting object to init, because once it's created you can't disable the injection of ANSI codes, only strip them out later. That mixed with the fact that caught errors in `pkl test` are truncated creates a bit of nightmare as the outputs get truncated to slightly different lengths depending on if colours were enabled, regardless of if those ANSI code actually made it to the output.

I tried to figure out where the error truncation etc was happening, so I could modify it to handle ANSI/non-ANSI outputs in an equivalent way, but I haven't been able to track down where it happens.

bioball: I think if (when?) we switch to a text formatter, we can control this ourselves by passing a no-op text formatter when `--color=never`. Then, we're in full control of how formatting happens rather than mutating global state.

Anyways, I don't know if that's relevant to my suggestion here. `CliBaseOptions` is a data class, and you can always modify it however you want because you get a derived `copy` method.

```kt
val options = baseOptions(myModules).copy(color = Color.NEVER)
```

thomaspurchas: It's only relevant due to how Jansi handles global state. It's caught me out a couple of times because Jansi doesn't handle changes to global state the way you might expect. The final result is what you expect, but the mechanism by which Jansi achieves that result depends on the exact order of operations (mutating global state before or after doing the initial Jansi setup). In most cases it doesn't make any practical difference, but for the specific case of Pkl test it does (which I learned the hard way).

Agree we can just get rid of this hack by implementing a proper formatting API, and having direct control over of control codes are injected.

holzensp:
This may be more complicated than what it's worth, but really, `pkl test` should color the "outside" exceptions and not the "inside" ones, i.e. if `pkl test` _itself_ throws, it should be colored, but anything inside a `module.catch(...)` should not.

I don't think this is something that should hold you back, especially considering we want to try and avoid the Jansi dependency in `pkl-core`. @bioball's suggestion of passing in a formatter does seem the most sensible way to go.

thomaspurchas:
@holzensp yeah my plan to stop working on this approach, and look at passing in a formatter. As I mentioned [here](https://github.com/apple/pkl/pull/571#discussion_r1671293034), the number of hacks and whatnot that are piling up to make this approach work doesn't feel good. I wouldn't be comfortable merging such a fragile pile of hacks and workarounds.

bioball:
Closing this; superseded by #746

Issue #573, opened 2024-07-08
🐛`fixed` properties not respected by `toTyped()`


thomaspurchas:
The `toTyped` API doesn't respect `fixed` properties on classes, resulting in the following unexpected outcome:

```pkl
class Bird {
  fixed type = "Parrot"
  name = "Polly"
}

untypedBird {
  type = "Cocker spaniel"
  name = "Morris"
}

result = untypedBird.toTyped(Bird)
```

[produces](https://pkl-playground.vercel.app/?share=truck-species-screen)

```
untypedBird {
  type = "Cocker spaniel"
  name = "Morris"
}
result {
  type = "Cocker spaniel"
  name = "Morris"
}
```

hmonfleur:
I don't think It's particular to fixed properties. I have the same kind of problem but with "normal" properties : 

```pkl
class A{
  x:Int
  y = x + 1
}

a = Map("x", 1).toTyped(A)
```
[produces](https://pkl-playground.vercel.app/?share=stretch-away-speed) the following error :

```Tried to read property `x` but its value is undefined.```

Which is understandable since :

class A{
  x:Int = 0
  y = x + 1
}

a = Map("x", 1).toTyped(A)

[produces](https://pkl-playground.vercel.app/?share=earth-me-go) :

```
{
  "a": {
    "x": 1,
    "y": 1
  }
}
```

If I'm missing something or if there is a way around it I'm interested :)

holzensp:
The point @thomaspurchas is making - I believe - is that he expected `type` to be `"Parrot"`. What isn't being respected here, is that `toTyped` ignores the `fixed`-ness of the property.

hmonfleur:
Ok, maybe I wrongly made a  link between my problem and the one that @thomaspurchas raised.
Still there's a behavior that is not the one I would have expected when using  `toTyped` , i.e., it does not produce the same object as normal instantiation would have. I have found some partial workaround using `let` to force evaluation at some points (I mean I think that's what it does) but it does not entirely solve the problem (in my case for instanciation of objects  contained in attributes and produced by the instanciation of an object that are not instanciated when using `toTyped` resulting in empty attributes).
Maybe I can sum that up and open a new issue about it.

Issue #202, opened 2024-02-19
Consistency with `toTyped` and class instantiation


Senjai:
Consider

```pkl
class Bird {
  name: String
  lifespan: Int
}

// Allowed
pigeon = new Dynamic {
  name = "Pigeon"
  lifespan = 8
  notInBird = "not in bird"
}.toTyped(Bird)

// not allowed
eagle = new Bird {
  name = "Pigeon"
  lifespan = 8
  notInBird = "not in bird"
}
```

`toTyped` is more lenient than class instantiation, and I'm not sure that should be the case. It might be a good default, since dynamic data can be well, dynamic and I can understand why we wouldn't want people to have to remove a bunch of non relevant data if they are working with truly dynamic data. That said, in cases like this, or when validating JSON which you want to be strongly typed with no leaked data, we are missing a `toStrictTyped` or `toTyped` with a boolean flag to ensure that it behaves _as if_ it was defined in Pkl.

Something like this is needed in my opinion both to ensure the normal course of converting from dynamic to typed can behave without surprises (e.g. toMap().remove(x).toTyped) and for validating schemas

holzensp:
You could define (using `import "pkl:reflect"`):
```
function expectedKeys(clazz: reflect.Class?): Set<String> = if (clazz == null) Set() else
  clazz.properties.keys + expectedKeys(clazz.superclass)

function hasAdditionalFields(object: Dynamic, clazz: Class): Boolean =
  let (props = expectedKeys(reflect.Class(clazz)))
    object.toMap().keys.every((it) -> props.contains(it) || throw("Unexpected property \(it)"))
```
and constrain it through that, but there is no built-in mechanism for it.

Senjai:
> You could define (using import "pkl:reflect"):

Thank you. I saw some of this in the deepToTyped package too, I'll note it down to dig into this.

> there is no built-in mechanism for it

Should there be? Curious to hear thoughts there.

holzensp:
Layers of “built-in.” For sure there is a place for this in `pkl-pantry`. I was thinking about being built-in in the language, and the problem there is that there are a few more constraints.

For example, Pkl is (quite) lazy. To do a properly deep type cast, requires forcing everything. This is also why `toTyped` is _shallow_. Thinking about properties existing that are not in the type, in terms of (lazy) evaluation order; when should that error surface? These are tricky questions that need some proper laying out, before we can roll anything into the language.

bioball:
[...]

PR #575, opened 2024-07-12
Fix default value evaluation in toTyped()

luuvish:
Fix the issue where default values defined in the class are not properly
generated when converting Dynamic or Map objects to Typed objects.
Also, fix the issue where values are not properly generated for amended
properties when making amends to the converted Typed objects.


Jimmydraper317 (approved):

holzensp (commented):
Very sorry for letting this sit so long and thanks for the contribution. I've created an issue for this (#768).

The approach in this PR is not quite right. `toTyped` should really produce an object with the class default in its prototype chain (i.e. the resulting object should have the same parent as if you'd evaluated `new Person { name = "NewPigeon" }`).

Aside from that, the solution of introducing three `readMember`-like functions is costly (maintenance-wise). There is a much longer-standing issue with `toTyped` and that is that it is "shallow" (see issue #769). We intend to rework `toTyped` to cover all known issues (also issues #202 and #573).

luuvish:
Thank you for your response.

I have also been contemplating various issues related to toTyped and have been solving them in my own way.
I'm very excited to hear that you are planning to work on covering all related issues.

Regarding issue #573, `const` properties have the same problem. I hope all issues will be resolved cleanly.

PR #580, opened 2024-07-16
Support `_` in String[toInt|toIntOrNull|toFloat|toFloatOrNull]

stackoverflow:
The logic is the same as the Pkl parser


bioball (approved):
LGTM!

Issue #583, opened 2024-07-17
Possible race condition when eval'ing packages in concurrently


bioball:
If two competing processes both attempt to download a package, Pkl can possible throw `FileAlreadyExistsException`

bioball:
Fixed in 0.26.2

PR #584, opened 2024-07-17
Fix race condition when concurrently downloading packages

bioball:
This fixes a possible race condition where multiple processes download the same package into the same temp dir.

The issue here is that we are using the _same_ filename when creating these temp files. Because of this, two concurrent processes can possibly run into `FileAlreadyExistsException` when attempting to create the temporary file.

This fixes the problem by using the `Files#createTempFile` API to ensure that these filenames are unique.

Closes https://github.com/apple/pkl/issues/583

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/packages/PackageResolvers.java
+++ b/pkl-core/src/main/java/org/pkl/core/packages/PackageResolvers.java
@@ -455,7 +455,6 @@ final class PackageResolvers {
 
     private byte[] downloadUriToPathAndComputeChecksum(URI downloadUri, Path path)
         throws IOException, SecurityManagerException {
-      Files.createDirectories(path.getParent());
       var inputStream = openExternalUri(downloadUri);
       try (var digestInputStream = newDigestInputStream(inputStream)) {
         Files.copy(digestInputStream, path, StandardCopyOption.REPLACE_EXISTING);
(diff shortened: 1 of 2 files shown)


stackoverflow (approved):
LGTM

pkl-core/src/main/java/org/pkl/core/packages/PackageResolvers.java
```diff
@@ -490,7 +490,8 @@ private Path getMetadataPath(
       if (Files.exists(cachePath)) {
         return cachePath;
       }
-      var tmpPath = tmpDir.resolve(metadataRelativePath);
+      Files.createDirectories(tmpDir);
+      var tmpPath = Files.createTempFile(tmpDir, null, ".json");
```
holzensp, line 494: Not blocking, but for `/tmp` hygiene, an informative prefix would be good. I'd just use the package name & version.

pkl-core/src/main/java/org/pkl/core/packages/PackageResolvers.java
```diff
@@ -490,7 +490,8 @@ private Path getMetadataPath(
       if (Files.exists(cachePath)) {
         return cachePath;
       }
-      var tmpPath = tmpDir.resolve(metadataRelativePath);
+      Files.createDirectories(tmpDir);
```
holzensp, line 493: Why wasn't this necessary before, even though `tmpDir` hasn't changed?

holzensp (approved):

pkl-core/src/main/java/org/pkl/core/packages/PackageResolvers.java
bioball: `Files.createTempFile` throws if the temp dir doesn't exist.

PR #596, opened 2024-07-24
Fix property parsing bug in the cli

stackoverflow:
This fixes a bug where a property set as `-p foo=` would be parsed as `"true"` instead of `""`.

### DIFF
--- a/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
+++ b/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
@@ -65,16 +65,14 @@ class BaseOptions : OptionGroup() {
           }
       }
 
-    fun RawOption.splitProps(): NullableOption<Pair<String, String>, Pair<String, String>> {
-      return convert {
-        val parts = it.split("=")
-        if (parts.size <= 1) parts[0] to "true" else parts[0] to parts[1]
-      }
-    }
-
     fun RawOption.associateProps():
       OptionWithValues<Map<String, String>, Pair<String, String>, Pair<String, String>> {
-      return splitProps().multiple().toMap()
+      return convert {
+          val parts = it.split("=")
+          if (parts.size <= 1) parts[0] to "true" else parts[0] to parts[1]
+        }
+        .multiple()
+        .toMap()
     }
   }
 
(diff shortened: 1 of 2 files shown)


StefMa:
Why is that a bug? 🤔
I wouldn't expect to have set a property to true if only `prop=` is set.
Maybe it should report an error instead of using a blank string? 🤔

stackoverflow:
> Why is that a bug?

Because before this fix you couldn't set a property to empty string.

> Maybe it should report an error instead of using a blank string?

Why would this be an error? Empty string is a valid property value.

StefMa:
God it. I misunderstood the implementation 😂
This is fine 👌

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -64,6 +64,18 @@ class BaseOptions : OptionGroup() {
...
+      return convert {
+        val parts = it.split("=")
+        if (parts.size <= 1) parts[0] to "true" else parts[0] to parts[1]
+      }
+    }
+
+    fun RawOption.associateProps():
+      OptionWithValues<Map<String, String>, Pair<String, String>, Pair<String, String>> {
+      return splitProps().multiple().toMap()
+    }
```
holzensp, line 78: I'd inline at least once (if not entirely on the `associateProps` call site:
```suggestion
    fun RawOption.associateProps():
      OptionWithValues<Map<String, String>, Pair<String, String>, Pair<String, String>> {
      return convert {
        val parts = it.split("=")
        parts[0] to (parts.getOrNull(1) ?: "true")
      }.multiple().toMap()
    }
```

pkl-commons-cli/src/test/kotlin/org/pkl/commons/cli/BaseCommandTest.kt
```diff
@@ -45,10 +45,10 @@ class BaseCommandTest {
 
   @Test
   fun `external properties without value default to 'true'`() {
-    cmd.parse(arrayOf("-p", "flag1", "-p", "flag2", "-p", "FOO=bar"))
+    cmd.parse(arrayOf("-p", "flag1", "-p", "flag2=", "-p", "FOO=bar"))
```
holzensp, line 48: The `flag1` and `flag2` test cases becoming overlapped should have been a hint before ;)

holzensp (approved):

bioball (approved):
LGTM

PR #599, opened 2024-07-25
Separate the cache check and member lookup into different methods

KushalP:
This allows for better inlining and optimization of the cache check, which is a common path.

Original results
----------------

```
Benchmark                                    Mode  Cnt   Score   Error  Units
VmUtilsBenchmarks.readMemberOrNullCacheHit   avgt    5  18.906 ± 0.726  ns/op
VmUtilsBenchmarks.readMemberOrNullCacheMiss  avgt    5  15.274 ± 0.149  ns/op
```

This change results
-------------------

```
Benchmark                                    Mode  Cnt   Score   Error  Units
VmUtilsBenchmarks.readMemberOrNullCacheHit   avgt    5  16.819 ± 0.837  ns/op
VmUtilsBenchmarks.readMemberOrNullCacheMiss  avgt    5  15.710 ± 0.906  ns/op
```

### DIFF
--- a/bench/bench.gradle.kts
+++ b/bench/bench.gradle.kts
@@ -30,6 +30,10 @@ dependencies {
   jmh(libs.antlrRuntime)
   truffle(libs.truffleApi)
   graal(libs.graalCompiler)
+
+  implementation(libs.truffleApi)
+  implementation(libs.graalSdk)
+  compileOnly(libs.jsr305)
 }
 
 jmh {
--- a/bench/gradle.lockfile
+++ b/bench/gradle.lockfile
@@ -1,6 +1,7 @@
 # This is a Gradle generated file for dependency locking.
 # Manual edits can break the build and are not advised.
 # This file is expected to be part of source control.
+com.google.code.findbugs:jsr305:3.0.2=compileClasspath,compileOnlyDependenciesMetadata,jmhCompileClasspath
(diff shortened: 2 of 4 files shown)


holzensp:
In the worst case (assuming variance is correlated), the cache miss here is ~8% slower. I don't trust anything on the `ns` scale, to be honest. Let's try and get some larger numbers to calibrate what these benchmarks mean.

holzensp:
@stackoverflow is trying this on a larger codebase. The intuition is quite counter as well. When the JIT decides this is *hot*, it would _inline_ (effectively _undo_ your change). Considering that there are no new/other uses of the small functions you split off in this PR, there should be no speed-up at all (in the worst-case, there should be minute degradation of warm-up).

holzensp:
Those results;
Before: `111.19s user 5.82s system 159% cpu 1:13.47 total`
After: `114.28s user 5.05s system 162% cpu 1:13.23 total`

I still don't quite understand how _user_ time increases (although it seem within variance bounds; rerun took `112`), while wall-clock time marginally decreases, but heyo...

On the whole, it doesn't seem to make a noticeable difference and it does add names for things. Were you looking for uses of _only_ `getCachedValue` or `lookupMember`? If so, this might still be useful; otherwise, fewer names are more better ;)

KushalP:
Thanks all. This was a speculative change to begin with. I'll close this.

PR #613, opened 2024-08-06
[docs] Add mention of "optional" to nullable types section

HT154:
I've experience a few users missing this section of the documentation because they searched "optional". This terminology is common in other languages (eg. Swift, Python). This adds a sentence to the documentation to ease search-ability.


bioball (approved):
LGTM, thanks!

PR #614, opened 2024-08-06
[docs] Document the class-as-a-function pattern

HT154:


docs/modules/language-reference/pages/index.adoc
```diff
@@ -5146,8 +5152,8 @@ It is defined by the presence of a `PklProject` file that amends the standard li
 Defining a project serves the following purposes:
 
 1. It allows defining common evaluator settings for Pkl modules within a logical project.
-2. It helps with managing <<packages,package>> dependencies for Pkl modules within a logical project.
-3. It enables packaging and sharing the contents of the project as a <<packages,package>>.
+2. It helps with managing <<package-asset-uri,package>> dependencies for Pkl modules within a logical project.
+3. It enables packaging and sharing the contents of the project as a <<package-asset-uri,package>>.
```
HT154, line 5156: These anchors are broken currently. I took an educated guess on where they should point, but I can revert or modify as needed.

bioball:
Thanks for the PR!

I'm on the fence about including this in the language reference. I think this might be better suited as a blog post instead. Or, perhaps in the style guide (although I'm less sure about that one because this is not a clear-cut rule).

WDYT?

HT154:
I opted for the style guide here, but I'm also torn. A blog post seems less discoverable in comparison. I think there's definite utility in linking this—wherever it ends up—from the language reference.

stackoverflow:
Yeah, the style guide doesn't seem to be the best place for this. Perhaps a blog post with a note in the language reference pointing to the post is better.

HT154:
Sounds good @stackoverflow! I opened https://github.com/apple/pkl-lang.org/pull/27 creating the blog post and have updated this branch accordingly.

bioball (approved):
LGTM, thanks!

PR #618, opened 2024-08-07
[docs] Document `hidden` equality/hashing behavior

HT154:


docs/modules/language-reference/pages/index.adoc
```diff
@@ -933,12 +934,20 @@ pigeon = new Bird { // <3>
...
+
+samePigeon = pigeon == favoritePigeon // <7>
 ----
 <1> Properties defined as `hidden` are accessible on any `Bird` instance, but not output by default.
 <2> Non-hidden properties can refer to hidden properties as usual.
 <3> `pigeon` is an object with _four_ properties, but is rendered with _three_ properties.
 <4> Accessing a `hidden` property from outside the class and object is like any other property.
 <5> Object conversions omit hidden properties, so the resulting `Dynamic` has three properties.
+<6> Our favorite pigeon should be indexed with her name!
+<7> Objects that differ only in `hidden` property values are considered equal
```
bioball, line 950: Point 6 seemed a little out of place?

```suggestion

favoritePigeon = (pigeon) {
  nameAndLifespanInIndex = "Bettie, \(lifespan)"
}

samePigeon = pigeon == favoritePigeon // <6>
----
<1> Properties defined as `hidden` are accessible on any `Bird` instance, but not output by default.
<2> Non-hidden properties can refer to hidden properties as usual.
<3> `pigeon` is an object with _four_ properties, but is rendered with _three_ properties.
<4> Accessing a `hidden` property from outside the class and object is like any other property.
<5> Object conversions omit hidden properties, so the resulting `Dynamic` has three properties.
<6> Objects that differ only in `hidden` property values are considered equal
```

bioball (approved):
LGTM, thanks!

stackoverflow (approved):
Looks good!

PR #624, opened 2024-08-13
[docs] Document object instantiation

HT154:
This content is "reverse engineered" entirely from using the language and is not an interpretation of the code, so it may be missing or misconstruing things. Please let me know if anything needs adjusting!

### DIFF
--- a/docs/modules/language-reference/pages/index.adoc
+++ b/docs/modules/language-reference/pages/index.adoc
@@ -3263,10 +3263,10 @@ In these cases, Pkl infers the amend operation's parent value based on context:
 * When assigning to a declared property, the property's default value is amended (<<amend-null, including `null`>>).
 If there is no type associated with the property, an empty `Dynamic` object is amended.
 * When assigning to an entry (e.g. a `Mapping` member) or element (e.g. a `Listing` member), the enclosing object's `default` property is applied to the corresponding index or key, respectively, to produce the value to be amended.
-* In other cases, type inference fails with the error message "Cannot tell which parent to amend".
+* In other cases, evaluation fails with the error message "Cannot tell which parent to amend".
 
-The type information of a <<methods,method>> parameter is not used for type inference.
-In this case, argument type information should be specified explicitly.
+The type annotation of a <<methods,method>> parameter is not used for inference.
+In this case, the argument's type should be specified explicitly.
 
 [source,{pkl}]
 ----
--- a/docs/modules/language-reference/pages/index.adoc
+++ b/docs/modules/language-reference/pages/index.adoc
@@ -3156,7 +3156,7 @@ pkl: TRACE: num1 * num2 = 672 (at file:///some/module.pkl, line 42)
 
 This section discusses language features that are generally more relevant to template and library authors than template consumers.
 
-<<object-instantiation,Object Instantiation>> +
+<<meaning-of-new,Meaning of `new`>> +
 <<let-expressions,Let Expressions>> +
 <<type-tests,Type Tests>> +
 <<type-casts,Type Casts>> +
@@ -3180,16 +3180,17 @@ This section discusses language features that are generally more relevant to tem
 <<blank-identifiers,Blank Identifiers>> +
 <<projects,Projects>>
 
-[[object-instantiation]]
-=== Object Instantiation
+[[meaning-of-new]]
+=== Meaning of `new`
 
-Object instantiation in Pkl always <<amending-objects, amends>> _some_ value.
-The `new` keyword is a special case of amending where a contextual default value is amended. 
-In Pkl, objects can be instantiated in three major ways:
+Objects in Pkl always <<amending-objects, amends>> _some_ value.
+The `new` keyword is a special case of amending where a contextual value is amended.
+In Pkl, there are two forms of `new` objects:
 
-* Amending some parent property or value directly.
-* Invoking `new` with explicit type information, which always amends the default value of the specified type.
-* Invoking `new` without type information, which amends a contextually inferred value.
+* `new` with explicit type information, for example, `new Foo {}`.
+* `new` without type information, for example, `new {}`.
+
+==== Type defaults
 
 To understand instantiation cases without explicit parent or type information, it's important to first understand implicit default values.
 When a property is declared in a module or class but is not provided an explicit default value, the property's default value becomes the type's default value.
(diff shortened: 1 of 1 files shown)


docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3179,33 @@ This section discusses language features that are generally more relevant to tem
 <<blank-identifiers,Blank Identifiers>> +
 <<projects,Projects>>
 
+[[the-meaning-of-new]]
+=== The Meaning of `new`
+
+Pkl has no "true" object instantiation: the `new` keyword is a special case of <<amending-objects, amending>>.
```
stackoverflow, line 3185: nit: you could rephrase this sentence to be a bit less negative. Something like "Object instantiation in Pkl is actually just amending" or something of the sort.

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3179,33 @@ This section discusses language features that are generally more relevant to tem
...
+In Pkl, objects can be instantiated in three major ways:
+
+* Amending a property/entry `someProperty {}` directly, which may specify properties that have not previously been declared or defined
```
stackoverflow, line 3188:
```suggestion
* Amending a property/entry `someProperty {}` directly, which may specify properties that have not previously been declared or defined.
```

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3179,33 @@ This section discusses language features that are generally more relevant to tem
...
+* Invoking `new` without type information `new {}`, which amends a contextually inferred value
```
stackoverflow, line 3189:
```suggestion
* Invoking `new` without type information `new {}`, which amends a contextually inferred value.
```

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3179,33 @@ This section discusses language features that are generally more relevant to tem
...
+* Invoking `new` with explicit type information: `new Dynamic {}`, which always amends the default value of the specified type
```
stackoverflow, line 3190:
```suggestion
* Invoking `new` with explicit type information: `new Dynamic {}`, which always amends the default value of the specified type.
```

stackoverflow, line 3190: nit: I think readability would be improved if you created a code block showing all 3 types of instantiation and then used notes to explain them, pointing back to the code.

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3179,33 @@ This section discusses language features that are generally more relevant to tem
...
+When amending previously undeclared properties and using the untyped `new` invocation, there is no explicit parent value to amend.
+In these cases, Pkl infers the amend operation's parent value based on context.
+In order of precedence:
+
+1. When amending an existing value, the parent value's type is used.
+2. When amending a declared but undefined property, the property's explicit type's default value is used (<<amend-null, including `null`>>).
+3. Within a `Listing` or `Mapping`, the container's `default` value is used.
+4. Within a `Listing` or `Mapping`, the default value of the container's explicit element or entry value type argument is used.
+5. In module or object body context, when no type information is otherwise available, an empty `Dynamic` object is used.
+6. In other cases, type inference fails with the error message "Cannot tell which parent to amend".
```
stackoverflow, line 3201: nit: again, having a code block showing all these cases would make it easier for newcomers to understand these concepts.

stackoverflow (approved):
Left some nits, but it already looks good!

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3179,33 @@ This section discusses language features that are generally more relevant to tem
...
+
+* Amending a property/entry `someProperty {}` directly, which may specify properties that have not previously been declared or defined
+* Invoking `new` without type information `new {}`, which amends a contextually inferred value
+* Invoking `new` with explicit type information: `new Dynamic {}`, which always amends the default value of the specified type
+
+When amending previously undeclared properties and using the untyped `new` invocation, there is no explicit parent value to amend.
+In these cases, Pkl infers the amend operation's parent value based on context.
+In order of precedence:
+
+1. When amending an existing value, the parent value's type is used.
```
holzensp, line 3196: Nit picking, but the parent value's type is never used. It's just the value itself that gets amended.

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3179,33 @@ This section discusses language features that are generally more relevant to tem
...
+2. When amending a declared but undefined property, the property's explicit type's default value is used (<<amend-null, including `null`>>).
```
holzensp, line 3197: Pardon the pedantry, but _technically_ this isn't accurate; declaring a property will also give it a default value (i.e. `foo: Foo` will give `foo` the default value as declared in `Foo`, if any).

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3179,33 @@ This section discusses language features that are generally more relevant to tem
...
+3. Within a `Listing` or `Mapping`, the container's `default` value is used.
```
holzensp, line 3198:
```suggestion
3. Within a `Listing` or `Mapping`, the container's (function-typed) `default` property is applied to the corresponding index, respectively key, to produce the value to be amended.
```

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3179,33 @@ This section discusses language features that are generally more relevant to tem
...
+4. Within a `Listing` or `Mapping`, the default value of the container's explicit element or entry value type argument is used.
```
holzensp, line 3199: Similar to `2.`; if I declare `foos: Listing<Foo>`, then `foos.default` is initialised to `(_) -> new Foo {}`

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3179,33 @@ This section discusses language features that are generally more relevant to tem
...
+5. In module or object body context, when no type information is otherwise available, an empty `Dynamic` object is used.
+6. In other cases, type inference fails with the error message "Cannot tell which parent to amend".
+
+These rules yield some corollaries that may not be obvious:
+ 
+* Instantiating with explicit type information ignores a collection's `default` value.
```
holzensp, line 3205:
```suggestion
* Instantiating with explicit type information ignores a collection's `default` property.
```

holzensp (changes requested):
It's certainly a good idea to document this behaviour explicitly, but it's hard to get exactly right. Some nits to iterate on.

docs/modules/language-reference/pages/index.adoc
bioball: Ditto--don't know what true instantiation is!

HT154:
I've almost entirely rewritten this to incorporate the great feedback @stackoverflow @holzensp

Notable callouts:
* I've added a separate section in the introduction on default values
* I broke the three cases out into their own sections and reorganized them
* I've added code blocks with examples for each instantiation type—I was a little conservative in the name of brevity, so if you see a case I didn't cover but think should be included, please let me know

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3180,157 @@ This section discusses language features that are generally more relevant to tem
...
+  }
+}
+----
+<1> Assigning a `new` explicitly-typed value to a property.
+<2> Adding an `new` explicitly-typed `Listing` element.
+The value will not have property `isDefault = true` as the `default` property of the `Listing` is not used.
+<3> Assigning a `new` explicitly-typed value to a `Mapping` entry.
+The value will not have property `isDefault = true` as the `default` property of the `Mapping` is not used.
+
+==== Untyped `new`
```
stackoverflow, line 3271: Even though untyped is technically correct, it may give the idea that the resulting value has no type. I think "Implicitly typed `new`", or perhaps, "Raw `new`" is a better name here.

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3180,157 @@ This section discusses language features that are generally more relevant to tem
...
+class Gizmo {
+  name: String = "doodad"
+}
+
+gizmo: Gizmo // <1>
+gizmoListing: Listing<Gizmo> // <2>
+gizmoMapping: Mapping<String, Gizmo> // <3> 
+----
+<1> Without an explicit default value, this property has default value `new Gizmo { name = "doodad" }`
+<2> With an explicit element type argument, this property's default value is equivalent to `new Listing<Gizmo> { default = new Gizmo { name = "doodad" } }`
```
bioball, line 3211:
```suggestion
<2> With an explicit element type argument, this property's default value is equivalent to `new Listing<Gizmo> { default = (_) -> new Gizmo { name = "doodad" } }`
```

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3180,157 @@ This section discusses language features that are generally more relevant to tem
...
+[[object-instantiation]]
+=== Object Instantiation
+
+Object instantiation in Pkl always <<amending-objects, amends>> _some_ value.
+The `new` keyword is a special case of amending where a contextual default value is amended. 
+In Pkl, objects can be instantiated in three major ways:
+
+* Amending some parent property or value directly.
+* Invoking `new` with explicit type information, which always amends the default value of the specified type.
+* Invoking `new` without type information, which amends a contextually inferred value.
```
bioball, line 3192: I think that the concept of amending a parent is pretty different to what the rest of this section covers. I don't know if we should group that together here; I'm thinking it might be better to just talk about `new` specifically.

```suggestion
=== Meaning of `new`

Objects in Pkl always <<amending-objects, amends>> _some_ value.
The `new` keyword is a special case of amending where a contextual value is amended. 
In Pkl, there are two forms of `new` objects:

* `new` with explicit type information, for example, `new Foo {}`.
* `new` without type information, for example, `new {}`.
```

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3180,157 @@ This section discusses language features that are generally more relevant to tem
...
+someMapping = new Mapping {
+  ["greetings"] { // <3>
+    hello = "everyone"
+  }
+}
+----
+<1> Amending a property.
+<2> Amending a value.
+<3> Amending a new `Mapping` entry.
+
```
bioball, line 3237: Per my previous comment; I don't know if this section belongs here; it's kind of a different concept.

But, FYI, `newProperty {}` means: `newProperty = (super.newProperty) {}`. And only if there is no supermodule, `super.newProperty` will give you `new Dynamic {}`. Otherwise, an error is thrown.

In a similar vein, `["someEntry"] {}` means `["someEntry"] = (super["someEntry"]) {}`

```suggestion
```

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3180,157 @@ This section discusses language features that are generally more relevant to tem
...
+==== Explicitly Typed `new`
+
+Instantiating an object with `new <type>` results in a value that amends the specified type's default value.
+Notably, creating a `Listing` element or assigning a `Mapping` value with an explicitly typed `new` ignores a collection's `default` value.
```
bioball, line 3241: `Collection` is the base type of `List` and `Set`. We avoid this term unless referring to those two types specifically.

```suggestion
Notably, creating a `Listing` element or assigning a `Mapping` entry value with an explicitly typed `new` ignores the object's `default` value.
```

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3180,157 @@ This section discusses language features that are generally more relevant to tem
...
+
+Object instantiation in Pkl always <<amending-objects, amends>> _some_ value.
+The `new` keyword is a special case of amending where a contextual default value is amended. 
+In Pkl, objects can be instantiated in three major ways:
+
+* Amending some parent property or value directly.
+* Invoking `new` with explicit type information, which always amends the default value of the specified type.
+* Invoking `new` without type information, which amends a contextually inferred value.
+
+To understand instantiation cases without explicit parent or type information, it's important to first understand implicit default values.
```
bioball, line 3194:
```suggestion
==== Type defaults

To understand instantiation cases without explicit parent or type information, it's important to first understand default values of types.
```

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3180,157 @@ This section discusses language features that are generally more relevant to tem
...
+  ["greetings"] = new Dynamic { // <3>
+    hello = "everyone"
+  }
+}
+----
+<1> Assigning a `new` explicitly-typed value to a property.
+<2> Adding an `new` explicitly-typed `Listing` element.
+The value will not have property `isDefault = true` as the `default` property of the `Listing` is not used.
+<3> Assigning a `new` explicitly-typed value to a `Mapping` entry.
+The value will not have property `isDefault = true` as the `default` property of the `Mapping` is not used.
```
bioball, line 3269: Changed this a bit to be a bit more canonical Pkl (avoid Dynamic), and use our Bird theme.

```suggestion
class Bird {
  /// The name of the bird
  name: String

  /// Whether this is a bird of prey or not.
  isPredatory: Boolean?
}

newProperty = new Bird { // <1>
  name = "Warbler"
}

someListing = new Listing<Bird> {
  default {
    isPredatory = true
  }
  new Bird { // <2>
    name = "Sand Piper"
  }
}

someMapping = new Mapping<String, Bird> {
  default {
    isPredatory = true
  }
  ["Penguin"] = new Bird { // <3>
    name = "Penguin"
  }
}
----
<1> Assigning a `new` explicitly-typed value to a property.
<2> Adding an `new` explicitly-typed `Listing` element.
The value will not have property `isPredatory = true` as the `default` property of the `Listing` is not used.
<3> Assigning a `new` explicitly-typed value to a `Mapping` entry.
The value will not have property `isPredatory = true ` as the `default` property of the `Mapping` is not used.
```

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3180,157 @@ This section discusses language features that are generally more relevant to tem
...
+==== Implicitly Typed `new`
+
+When using the implicitly typed `new` invocation, there is no explicit parent value to amend.
+In these cases, Pkl infers the amend operation's parent value based on context.
+In order of precedence:
+
+1. When assigning to a declared property, the property's default value is amended (<<amend-null, including `null`>>).
+2. Within a `Listing` or `Mapping`, the container's function-typed `default` property is applied to the corresponding index or key, respectively, to produce the value to be amended.
+3. In module or object body context, when no type information is otherwise available, an empty `Dynamic` object is amended.
+4. In other cases, type inference fails with the error message "Cannot tell which parent to amend".
```
bioball, line 3280:
```suggestion
1. When assigning to a property, the property's default value is amended (<<amend-null, including `null`>>). If there is no type associated with the property, an empty `Dynamic` object is amended.
2. When assigning to an entry (e.g. a `Mapping` member) or element (e.g. a `Listing` member), the enclosing object's `default` property is applied to the corresponding index or key, respectively, to produce the value to be amended.
3. Otherwise, evaluation fails with the error message "Cannot tell which parent to amend".
```

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3180,157 @@ This section discusses language features that are generally more relevant to tem
...
+
+Currently, the type information of a <<methods,method>> parameter is not used for type inference, which may lead to surprising type inference errors.
+In this case, argument type information should be specified explicitly.
```
bioball, line 3283: "Currently" suggests that this will change in the future, but I'm not so sure that we can ever change this.

```suggestion
The type information of a <<methods,method>> parameter is not used for type inference.
In this case, argument type information should be specified explicitly.
```

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3180,157 @@ This section discusses language features that are generally more relevant to tem
...
+  ["doohicky"] = new { // <5>
+    name = "thingamabob"
+  }
+}
+amendedMapping = (typedMapping) {
+  ["doohicky"] = new {} // <6>
+}
+
+doodadGizmos = typedProperty.gizmofy(new { "widget"; "whatchamacallit" }) // <7>
+----
```
bioball, line 3325: Can we make this bird themed?

Also, I think we should include the case where `new {}` amends a class property's default value. Something like:

```pkl
class Aviary {
  birds: Listing<Bird> = new {
    new { name = "Osprey" }
  }
}

aviary: Aviary = new {
  birds = new { // <-- this amends `new Listing { new Bird { name = "Osprey" } } and you end up with two birds here
    new { name = "Kiwi" }
  }
}
```

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3180,157 @@ This section discusses language features that are generally more relevant to tem
...
+[source,{pkl}]
+----
+class Gizmo {
+  name: String = "doodad"
+}
+
+gizmo: Gizmo // <1>
+gizmoListing: Listing<Gizmo> // <2>
+gizmoMapping: Mapping<String, Gizmo> // <3> 
+----
```
bioball, line 3209: Can we make this bird themed?

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3180,157 @@ This section discusses language features that are generally more relevant to tem
...
+<2> Adding an `new` explicitly-typed `Listing` element.
+The value will not have property `isDefault = true` as the `default` property of the `Listing` is not used.
+<3> Assigning a `new` explicitly-typed value to a `Mapping` entry.
+The value will not have property `isDefault = true` as the `default` property of the `Mapping` is not used.
+
+==== Implicitly Typed `new`
+
+When using the implicitly typed `new` invocation, there is no explicit parent value to amend.
+In these cases, Pkl infers the amend operation's parent value based on context.
+In order of precedence:
```
bioball, line 3275: There's no precedence here--what Pkl does depends on what the object member is.

Property: amend the class default value.
Entry/Element: call the default function of the enclosing object with my key/index.

```suggestion
```

bioball (changes requested):
Thanks for your PR!!

This overall looks really good, and definitely is a topic that is sorely missing. I have some comments--take a look.

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3180,162 @@ This section discusses language features that are generally more relevant to tem
...
+When using the implicitly typed `new` invocation, there is no explicit parent value to amend.
+In these cases, Pkl infers the amend operation's parent value based on context:
+
+* When assigning to a declared property, the property's default value is amended (<<amend-null, including `null`>>).
+If there is no type associated with the property, an empty `Dynamic` object is amended.
+* When assigning to an entry (e.g. a `Mapping` member) or element (e.g. a `Listing` member), the enclosing object's `default` property is applied to the corresponding index or key, respectively, to produce the value to be amended.
+* In other cases, type inference fails with the error message "Cannot tell which parent to amend".
+
+The type information of a <<methods,method>> parameter is not used for type inference.
+In this case, argument type information should be specified explicitly.
```
bioball, line 3269:
```suggestion
The type annotation of a <<methods,method>> parameter is not used for inference.
In this case, the argument's type should be specified explicitly.
```

docs/modules/language-reference/pages/index.adoc
```diff
@@ -3179,6 +3180,162 @@ This section discusses language features that are generally more relevant to tem
...
+
+==== Implicitly Typed `new`
+
+When using the implicitly typed `new` invocation, there is no explicit parent value to amend.
+In these cases, Pkl infers the amend operation's parent value based on context:
+
+* When assigning to a declared property, the property's default value is amended (<<amend-null, including `null`>>).
+If there is no type associated with the property, an empty `Dynamic` object is amended.
+* When assigning to an entry (e.g. a `Mapping` member) or element (e.g. a `Listing` member), the enclosing object's `default` property is applied to the corresponding index or key, respectively, to produce the value to be amended.
+* In other cases, type inference fails with the error message "Cannot tell which parent to amend".
```
bioball, line 3266:
```suggestion
* In other cases, evaluation fails with the error message "Cannot tell which parent to amend".
```

bioball (approved):

PR #627, opened 2024-08-15
Might be null

manuelsblanco:
**This PR refactors the VmMap class to enhance immutability handling and improve overall performance.** 

**Key changes include:**

- Refactored the VmMap class to ensure consistent use of TruffleBoundary annotations.
- Improved the Builder class for more efficient map construction.
- Added null checks and robustness improvements for handling empty collections.
- Optimized methods related to map and key order immutability.
- These changes aim to increase the stability and reliability of the VM runtime. Please review the updates, and let me know if any adjustments are needed.

### DIFF
--- a/bench/src/jmh/java/org/pkl/core/Fibonacci.java
+++ b/bench/src/jmh/java/org/pkl/core/Fibonacci.java
@@ -19,6 +19,8 @@ import static org.pkl.core.ModuleSource.modulePath;
 
 import java.util.concurrent.TimeUnit;
 import org.openjdk.jmh.annotations.*;
+import java.util.HashMap;
+import java.util.Map;
 
 @SuppressWarnings("unused")
 @Warmup(iterations = 5, time = 2)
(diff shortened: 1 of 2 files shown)


bench/src/jmh/java/org/pkl/core/Fibonacci.java
```diff
@@ -98,7 +100,23 @@ public long fib_lambda() {
...
+
+    // If the result for n is already calculated, return it from memo
+    if (memo.containsKey(n)) {
+      return memo.get(n);
+    }
+
+    // Otherwise, calculate Fibonacci recursively and store the result in memo
+    long result = fib(n - 1) + fib(n - 2);
+    memo.put(n, result);
+    return result;
```
bioball, line 120: This class is intentionally written this way--we use this to compare Pkl's fibonacci performance to native Java (see comment on line 99)

pkl-core/src/main/java/org/pkl/core/runtime/VmMap.java
```diff
@@ -98,7 +99,7 @@ public boolean hasNext() {
       @TruffleBoundary
       public Map.Entry<Object, Object> next() {
         var key = keyIterator.next();
-        return Map.entry(key, map.get(key));
+        return Map.entry(key, Objects.requireNonNull(map.get(key)));
```
bioball, line 102: Mmm I'm not sure why GitHub's diff is showing this, but this is out of date with what's on our main branch.

We already have a non-null assertion here, see:

https://github.com/apple/pkl/blob/a9c4963a4a60e06af683f4a700f290b75ca99d22/pkl-core/src/main/java/org/pkl/core/runtime/VmMap.java#L101-L103

bioball (commented):
Thanks for the PR! Unfortunately, I don't think there's anything here that we can accept for contribution (see my comments)

manuelsblanco:
Thanks for the reply.

Issue #406, opened 2024-04-05
Type checking a mapping/listing should be lazy


bioball:
Currently, accessing a property of type `Mapping<TypeA, TypeB>` or `Listing<Type>` will shallow-force the whole mapping/listing. The shallow-force gets skipped if the type is `unknown` or `Any` (e.g. `Listing`, or `Listing<unknown>`, or `Listing<Any>`).

This should not throw, but does currently:

```
hidden nums: Listing<Int> = new {
  throw("uh oh")
  1
}

result = nums[1]
```

This is related to #405

harryjackson:
This is just another example. The addition happens and we get 33 but at what point is the type checking meant to happen. Using the IntelliJ plugin I get warnings telling me I'm doing something stupid. 

```bash
foo = new Listing<YamlRenderer> {
  11
}

bar = new Listing<String> {
  22
}

output {
  value = foo[0] + bar[0]
  renderer = new YamlRenderer{}
}
```
output
```bash
33
```

bioball:
@harryjackson: what you're seeing is the #405 bug, rather than this one. To fix, you should add a type annotation to your properties:

```groovy
foo: Listing<YamlRenderer> = new {
  11
}

bar: Listing<String> = new {
  22
}

output {
  value = foo[0] + bar[0]
  renderer = new YamlRenderer{}
}
```

Issue #405, opened 2024-04-05
Type parameters in `new Mapping<TypeA, TypeB>`/ `new Listing<Type>` are not checked


bioball:
This does not throw, but should:

```
foo = new Listing<Int> {
  "hello"
}
```

In the below, both types should be checked:

```
foo: Listing<Int(isEven)> = new Listing<Int(this > 5)> { 1; 2; 3 }
```

The check should not follow when amended.

This is fine:

```
foo: new Listing<Int> { 1; 2; 3 }

bar = (foo) { "hello" }
```

This should probably be fine too:

```
foo = new Listing<Int> { 1 } { "hello" }
```

harryjackson:
This is just another example. The addition happens and we get 23 but at what point is the type checking meant to happen. Using the IntelliJ plugin I get warnings telling me I'm doing something stupid. 

```bash
foo = new Listing<YamlRenderer> {
  11
}

bar = new Listing<String> {
  22
}

output {
  value = foo[0] + bar[0]
  renderer = new YamlRenderer{}
}
```
output
```bash
33
```

translatenix:
A pragmatic solution would be to allow `x: Listing<A> = new Listing {...}`(*) but not `x = new Listing<A> {...}`. This would reflect the (current) reality that generics aren't reified. It would also encourage users to do the right thing and put type annotations on their properties.

(*) or `x: Listing<A> = (Listing) {...}`

holzensp:
That's less simple than you might think. The generic informs the default / IDE type checking. It's a common pattern to define `Listing`s inline to produce something of a very different type, such as `new Listing<String> { ... }.join(", ")` to produce a `String`.

translatenix:
Dealing with generics is never simple. Regarding your concrete arguments, `Listing.join` is defined for all `Listing`s, not just `Listing<String>`. And perhaps IDEs could infer the element type from the given elements. The default sounds like a problem though.

PR #628, opened 2024-08-15
Overhaul typechecks for Mapping/Listing

bioball:
This follows this SPICE: https://github.com/apple/pkl-evolution/pull/11

This changes how the language performs typechecks for mappings and
listings.

Currently, Pkl will shallow-force any Mapping and Listing to check it
the type parameter (e.g. `Listing<Person>` means each element is checked
to be an instance of `Person`).

This changes the language to check each member's type when the member
is accessed.

This PR is designed to be reviewed as two separate commits.
The first one contains business logic that changes how mapping/listing
typechecks work.
The second one adjusts the test runner to handle errors thrown within
tests.

Closes https://github.com/apple/pkl/issues/405
Closes https://github.com/apple/pkl/issues/406


pkl-core/src/test/files/LanguageSnippetTests/output/errors/listingTypeCheckError1.err
```diff
@@ -0,0 +1,11 @@
+–– Pkl Error ––
+Expected value of type `String`, but got type `Int`.
+Value: 1
+
+x | res = new Listing<String> {
+                      ^^^^^^
```
bioball, line 6: TODO: we should ideally show the source section of the member that failed this check here.

pkl-core/src/test/files/LanguageSnippetTests/input/listings/typeCheck.pkl
```diff
@@ -1,41 +1,367 @@
 import "pkl:test"
```
bioball, line 1: Note: these don't amend `pkl:test` because `facts` and `examples` are declared as `Mapping<String, Listing<Boolean>>` and `Mapping<String, Listing<unknown>>`. These are typed properties that themselves use Mapping/Listing type checks, and problems with typechecking can quickly be conflated between the tests themselves and the test framework.

pkl-core/src/test/files/LanguageSnippetTests/output/errors/listingTypeCheckError1.err
bioball: Figured it out!

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -67,16 +72,28 @@ public boolean isNoopTypeCheck() {
   public abstract TypeNode initWriteSlotNode(int slot);
 
   /**
-   * Checks if `value` conforms to this type. If so, returns normally. Otherwise, throws a
-   * `VmTypeMismatchException`.
+   * Checks if {@code value} conforms to this type, and possibly casts it in the case of {@link
+   * MappingTypeNode} or {@link ListingTypeNode}.
    */
-  public abstract void execute(VirtualFrame frame, Object value);
+  public abstract Object execute(VirtualFrame frame, Object value);
```
bioball, line 78: This is one of the biggest changes. Executing a typecheck will return a type-casted value.

For most types, this simply returns `value` if the value conforms to the type (e.g. `Int`). For `Listing` and `Mapping`, this returns a surrogate listing/mapping object. Other parameterized types may also return a new object if the parameterized type is `Listing` or `Mapping` (e.g. `List<Listing<Int>>`).

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -745,58 +942,86 @@ public VmClass getVmClass() {
...
+        elementTypeNode.executeEagerly(frame, elem);
+      }
+
+      LoopNode.reportLoopCount(this, value.getLength());
+      return value;
     }
   }
 
-  public abstract static class ListTypeNode extends ObjectSlotTypeNode {
+  public static final class ListTypeNode extends ObjectSlotTypeNode {
```
bioball, line 1001: Truffle's DSL was overriding `executeEagerly` when making this an abstract class and using specializations. I'm not sure if I'm doing this wrong?

But, regardless, having specializations here doesn't seem that useful anyway? Reading the same member value skips the typecheck the second time around because it gets cached. And reading a new value needs an `instanceof` call and will de-specialize.

pkl-core/src/main/java/org/pkl/core/ast/type/VmTypeMismatchException.java
```diff
@@ -35,18 +38,35 @@
...
     this.actualValue = actualValue;
   }
 
+  @TruffleBoundary
+  public void putInsertedStackFrame(CallTarget callTarget, StackFrame stackFrame) {
+    if (this.insertedStackFrames == null) {
+      this.insertedStackFrames = new HashMap<>();
+    }
+    this.insertedStackFrames.put(callTarget, stackFrame);
+  }
```
bioball, line 54: Added this so we can allow a `VmTypeMismatchException` to blame a source section when a typecheck fails (see `ListingOrMappingTypeNode#doEagerCheck`)

pkl-core/src/main/java/org/pkl/core/stdlib/base/MappingNodes.java
```diff
@@ -50,9 +51,14 @@ public abstract static class length extends ExternalPropertyNode {
...
-      self.iterateMemberValues(
-          (key, member, value) -> {
+      var count = new MutableLong(0);
+      var visited = new HashSet<>();
+      self.iterateMembers(
+          (key, member) -> {
+            var alreadyVisited = !visited.add(key);
+            // important to record hidden member as visited before skipping it
+            // because any overriding member won't carry a `hidden` identifier
+            if (alreadyVisited || member.isLocalOrExternalOrHidden()) return true;
```
bioball, line 61: This implementation is changed because `iterateMemberValues` might cause a `VmMapping`/`VmListing` to throw, because `VmListingOrMapping#getCachedValue()` can possibly perform an additional typecheck. `Mapping#length` should not need to execute any members, so, it should not need to run typechecks.

pkl-core/src/test/files/LanguageSnippetTests/input/listings/listings7.pkl
```diff
@@ -0,0 +1,4 @@
+// ensure that this member is only evaluated once (should result in only one trace)
```
stackoverflow, line 1: nit: all other tests are in the singular (`listing4/5/6.pkl`), only this one is plural.

stackoverflow (approved):
Nice solution! Less code than I expected. Just have some small things.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -805,29 +1033,49 @@ public TypeNode getElementTypeNode() {
...
+    public Object executeEagerly(VirtualFrame frame, Object value) {
+      if (!(value instanceof VmList vmList)) {
+        throw typeMismatch(value, BaseModule.getListClass());
+      }
+      if (skipElementTypeChecks) return vmList;
 
-      for (var elem : value) {
-        elementTypeNode.execute(frame, elem);
+      for (var elem : vmList) {
+        elementTypeNode.executeEagerly(frame, elem);
```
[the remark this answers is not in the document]
bioball: `executeEagerly` is a type test rather than a type cast. It will never return a different value; instead, it will throw if the type test fails.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -805,29 +1033,49 @@ public TypeNode getElementTypeNode() {
...
+      LoopNode.reportLoopCount(this, vmList.getLength());
+      return value;
     }
 
-    @Fallback
-    protected void fallback(Object value) {
-      throw typeMismatch(value, BaseModule.getListClass());
+    @Override
+    public Object execute(VirtualFrame frame, Object value) {
+      if (!(value instanceof VmList vmList)) {
```
holzensp, line 1062: Is this a newly introduced scenario, or a previously missed one?

pkl-core/src/main/java/org/pkl/core/runtime/VmListingOrMapping.java
```diff
@@ -0,0 +1,131 @@
...
+import org.pkl.core.PklBugException;
+import org.pkl.core.ast.member.ListingOrMappingTypeCheckNode;
+import org.pkl.core.ast.member.ObjectMember;
+import org.pkl.core.util.EconomicMaps;
+import org.pkl.core.util.Nullable;
+
+public abstract class VmListingOrMapping<SELF extends VmListingOrMapping<SELF>> extends VmObject {
+
+  /** The original that this listing/mapping is a surrogate of. It might have its own surrogatee. */
+  private final @Nullable SELF surrogatee;
```
holzensp, line 32: On the surrogate(e) naming; what about s/surrogate/typeCheckedThis and s/surrogatee/uncheckedThis?

pkl-core/src/main/java/org/pkl/core/runtime/VmListingOrMapping.java
```diff
@@ -0,0 +1,131 @@
...
+
+  private final ListingOrMappingTypeCheckNode typeCheckNode;
+  private final MaterializedFrame typeNodeFrame;
+  private final EconomicMap<Object, ObjectMember> cachedMembers = EconomicMaps.create();
```
holzensp, line 36: I'm a little worried about the memory load this imposes (2x this for all `Listing`s and `Mapping`s; especially the `cachedMembers`). Have you measured the cost of this on a significant corpus?

holzensp (approved):
Approving to unblock, but some significant questions; especially memory pressure!

pkl-core/src/main/java/org/pkl/core/runtime/VmListingOrMapping.java
bioball: This only introduces new cached members on _some_ mappings and listings.

For example, this still contains 2 cached members, because `foo` is hidden and never executed:

```pkl
hidden foo: Mapping = new {
  ["first"] = 1
  ["second"] = 2
}

bar: Mapping<String, Int> = foo
```

Same here:

```pkl
hidden foo: Mapping<String, Int> = new {
  ["first"] = 1
}
```

However, this example means we get four cached members, because `foo` is not hidden:

```pkl
foo: Mapping = new {
  ["first"] = 1
  ["second"] = 2
}

bar: Mapping<String, Int> = foo
```

We can certainly profile this on some real-world projects to see what this change looks like.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -259,6 +283,11 @@ public VmTyped getMirror() {
     protected PType doExport() {
       return PType.UNKNOWN;
     }
+
+    @Override
+    protected void acceptTypeNode(Consumer<TypeNode> visitor) {
+      visitor.accept(this);
+    }
```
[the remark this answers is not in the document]
bioball: Writing it this way is less error prone--having a default implementation made it too easy to miss if something did need to change for any one TypeNode.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -646,21 +733,106 @@ protected PType doExport() {
...
+    private boolean shouldEagerCheck() {
+      var seenListings = new MutableLong(0);
+      var seenMappings = new MutableLong(0);
+      var seenLists = new MutableLong(0);
+      var seenMaps = new MutableLong(0);
+      var seenSets = new MutableLong(0);
+      var seenPairs = new MutableLong(0);
+      var seenCollections = new MutableLong(0);
+      this.acceptTypeNode(
+          (typeNode) -> {
```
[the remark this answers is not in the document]
bioball: There is a truffle boundary here, so there's no impact on partial evaluation/inlining.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -646,21 +733,106 @@ protected PType doExport() {
 
     @Override
     @ExplodeLoop
-    public void execute(VirtualFrame frame, Object value) {
-      if (skipElementTypeChecks) return;
+    protected void acceptTypeNode(Consumer<TypeNode> visitor) {
+      visitor.accept(this);
+      //noinspection ForLoopReplaceableByForEach
```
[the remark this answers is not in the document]
bioball: Yeah; writing it this way allows Truffle to inline this loop into a series of statements. See [ExplodeLoop](https://www.graalvm.org/truffle/javadoc/com/oracle/truffle/api/nodes/ExplodeLoop.html) for details.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -646,21 +733,106 @@ protected PType doExport() {
...
+      var seenListings = new MutableLong(0);
+      var seenMappings = new MutableLong(0);
+      var seenLists = new MutableLong(0);
+      var seenMaps = new MutableLong(0);
+      var seenSets = new MutableLong(0);
+      var seenPairs = new MutableLong(0);
+      var seenCollections = new MutableLong(0);
+      this.acceptTypeNode(
+          (typeNode) -> {
+            if (typeNode instanceof ListingTypeNode) {
```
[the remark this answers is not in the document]
bioball: How does that work if, say, one of the members themselves is a union type?

pkl-cli/src/test/kotlin/org/pkl/cli/CliTestRunnerTest.kt
```diff
@@ -107,6 +107,138 @@ class CliTestRunnerTest {
     assertThat(err.toString()).isEqualTo("")
   }
 
+  @Test
+  fun `CliTestRunner thrown error in facts`(@TempDir tempDir: Path) {
```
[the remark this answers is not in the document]
bioball: Well, it's not `CliTestRunner` that throws the error. I'll change this to `CliTestRunner with thrown error in facts`

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -442,13 +442,21 @@ public Object visitNewExpr(NewExprContext ctx) {
...
-            createSourceSection(ctx),
-            visitType(typeCtx),
-            symbolTable.getCurrentScope().getQualifiedName()));
+    var parentType = visitType(typeCtx);
+    var expr =
+        doVisitObjectBody(
+            ctx.objectBody(),
+            new GetParentForTypeNode(
+                createSourceSection(ctx),
+                visitType(typeCtx),
```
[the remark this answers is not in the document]
bioball: Good catch!

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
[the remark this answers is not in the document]
bioball: Good idea!

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
bioball: I don't think so; I'm referring to the generated code that Truffle produces when you have an abstract class with specializations, which is a different concern than partial evaluation?

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
bioball: This is neither; I rewrote this class to remove the specialization (see comment here: https://github.com/apple/pkl/pull/628/files#r1720334213)

pkl-core/src/test/files/LanguageSnippetTests/input/listings/listing5.pkl
```diff
@@ -0,0 +1,168 @@
...
+
+  ["type check: propagate from List"] {
+    local l: List<Listing<String(length.isOdd)>> = List(
+      new Listing<String(this == capitalize())> {
+        "Ba"
+        "bar"
+      }
+    )
+    module.catch(() -> l[0][1])
+    module.catch(() -> l[0][2])
```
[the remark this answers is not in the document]
bioball: Whoops, good catch!

pkl-core/src/test/files/LanguageSnippetTests/input/listings/listings7.pkl
bioball: Good catch!

pkl-core/src/main/java/org/pkl/core/runtime/VmListingOrMapping.java
bioball: I added an additional optimization here: now, we will only store cached members on the delegated object _if_ the typecast results in a new object.

This results in two cached members:

```
foo: Mapping = new {
  ["first"] = 1
  ["second"] = 2
}

bar: Mapping<String, Int> = foo
```

This results in four cached members, due to the `Listing` to `Listing<Int>` typecast:

```pkl
foo: Mapping<Listing> = new {
  ["first"] = new { 1 }
  ["second"] = new { 2 }
}

bar: Mapping<String, Listing<Int>> = foo
```

bioball: I also added another optimization: A typecast will only return a new delegating object _if_ the types are different.

This results in no delegation:

```pkl
foo: Listing<String>

bar = foo as Listing<String> // no delegation here because `foo` is already `Listing<String>`
```

PR #629, opened 2024-08-18
Add support for handling remote file URIs.

### LINKED ISSUE #616: PklBugException when given a `file:` URI with a host
Pkl currently hard-fails if given a `file:` URI with a host set.

For example:

```
pkl eval file://example.com/bar/baz.pkl
```

Produces:

```
```

We should either support remote files, or throw a meaningful error here.

Carpe-Wang:
### Summary
This PR adds support for handling remote file URIs within the project. It addresses the issue where the system could not correctly process URIs that point to remote files.

### Changes
- Implemented functionality to download remote files specified by `file:` URIs.
- Added unit tests to verify the correctness of this new feature.

### Issue Link
Fixes #616

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
+++ b/pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
@@ -16,10 +16,13 @@
 package org.pkl.core.util;
 
 import com.oracle.truffle.api.TruffleOptions;
+import java.io.File;
 import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
+import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
@@ -55,6 +58,13 @@ public final class IoUtils {
 
   public static URL toUrl(URI uri) throws IOException {
     try {
+      if ("file".equalsIgnoreCase(uri.getScheme()) && uri.getHost() != null) {
+        // Dynamically determine the protocol, defaulting to http
+        String protocol = "http"; // Default protocol
+        URL remoteFileUrl = new URL(protocol, uri.getHost(), uri.getPort(), uri.getPath());
+        File localFile = downloadFile(remoteFileUrl);
+        return localFile.toURI().toURL();
+      }
       return uri.toURL();
     } catch (Error e) {
       // best we can do for now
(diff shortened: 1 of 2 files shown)


bioball:
Hey Carpe, thanks for the PR!

I don't know if this is the right solution to the problem. As far as I can tell, there is no agreed upon protocol when opening files with a remote hostname.

* Windows machines will use [SMB](https://en.wikipedia.org/wiki/Server_Message_Block) to open the file
* Safari on macOS will strip the hostname entirely and try to open the file from the local machine
* Java's `URL#openConnection()` will use FTP
* I can't find any examples of using HTTP to fetch the file

Without a protocol here, I think some valid choices are:
1. Explicitly don't support, and instead throw an error
2. Support on Windows only
3. Follow Java's lead and use FTP

I'm kind of leaning towards option 1 because it's the least amount of commitment and allows us to address this if the need arises later.

Carpe-Wang:
@bioball Hi bioball, I got your idea, If both of them also suggest modifying it to directly throw an exception, I will submit a new commit to implement this.

holzensp:
I would have expected NFS on Linux/Solaris/FreeBSD/etc. Agreed; randomly guessing protocols to start hammering seems risky, at best. Exception would be the way to go.

Carpe-Wang:
@bioball hi bioball, I change to throw the exception directly. Can you help me to ensure satisfying the requirement?

bioball:
Hi @Carpe-Wang: can you add this check to `org.pkl.core.module.ModuleKeys#file`?

That method should throw a `URISyntaxException` if there is a hostname set. And we should remove the check from `IoUtils#toUrl`; that's not quite the right place for it.

Thanks for the help here!

bioball (changes requested):
Hi @Carpe-Wang: can you add this check to `org.pkl.core.module.ModuleKeys#file`?

That method should throw a `URISyntaxException` if there is a hostname set. And we should remove the check from `IoUtils#toUrl`; that's not quite the right place for it.

Thanks for the help here!

Carpe-Wang:
Hi @bioball , I want move check URL logic to `ModuleKeys` inner `ProjectPackage`.class's `ResolvedModuleKey`method like below:
```java
    @Override
    public ResolvedModuleKey resolve(SecurityManager securityManager)
        throws IOException, SecurityManagerException {
      var uri = packageAssetUri.getUri();
      try {
        if (uri.getHost() != null) {
          throw new URISyntaxException(uri.toString(), "Host not allowed");
        }
      }catch (URISyntaxException e) {
        throw new IOException(e.getMessage());
      }
      securityManager.checkResolveModule(uri);
      var dependency =
          getProjectDependenciesManager().getResolvedDependency(packageAssetUri.getPackageUri());
      var local = getLocalUri(dependency);
      if (local != null) {
        var resolved =
            VmContext.get(null).getModuleResolver().resolve(local).resolve(securityManager);
        return ResolvedModuleKeys.delegated(resolved, this);
      }
      var dep = (Dependency.RemoteDependency) dependency;
      assert dep.getChecksums() != null;
      var bytes = getPackageResolver().getBytes(packageAssetUri, false, dep.getChecksums());
      return ResolvedModuleKeys.virtual(this, uri, new String(bytes, StandardCharsets.UTF_8), true);
    }
```

Do you think is okay? If I directly throw a URISyntaxException without catching it, I would need to modify the method signature to public `ResolvedModuleKey resolve(SecurityManager securityManager) throws IOException, SecurityManagerException, URISyntaxException`. Additionally, the interface and corresponding Override methods would also need to include URISyntaxException. Does this match your expectations?

bioball:
I think we should add the check to the `org.pkl.core.module.ModuleKeys#file` factory method; this is the entrypoint for all file-based modules.

This is a pattern that we already have! Take a look at `org.pkl.core.module.ModuleKeys#pkg` for another example of a module key that throws a `URISyntaxException` upon initialization.

Rochmarcos38:
Ok.


Sent from Yahoo Mail for iPhone


On Thursday, October 24, 2024, 6:59 PM, Carpe-Wang ***@***.***> wrote:




Hi @bioball , I want move check URL logic to ModuleKeys inner ProjectPackage.class's ResolvedModuleKeymethod like below:
    @Override
    public ResolvedModuleKey resolve(SecurityManager securityManager)
        throws IOException, SecurityManagerException {
      var uri = packageAssetUri.getUri();
      try {
        if (uri.getHost() != null) {
          throw new URISyntaxException(uri.toString(), "Host not allowed");
        }
      }catch (URISyntaxException e) {
        throw new IOException(e.getMessage());
      }
      securityManager.checkResolveModule(uri);
      var dependency =
          getProjectDependenciesManager().getResolvedDependency(packageAssetUri.getPackageUri());
      var local = getLocalUri(dependency);
      if (local != null) {
        var resolved =
            VmContext.get(null).getModuleResolver().resolve(local).resolve(securityManager);
        return ResolvedModuleKeys.delegated(resolved, this);
      }
      var dep = (Dependency.RemoteDependency) dependency;
      assert dep.getChecksums() != null;
      var bytes = getPackageResolver().getBytes(packageAssetUri, false, dep.getChecksums());
      return ResolvedModuleKeys.virtual(this, uri, new String(bytes, StandardCharsets.UTF_8), true);
    }
Do you think is okay? If I directly throw a URISyntaxException without catching it, I would need to modify the method signature to public ResolvedModuleKey resolve(SecurityManager securityManager) throws IOException, SecurityManagerException, URISyntaxException. Additionally, the interface and corresponding Override methods would also need to include URISyntaxException. Does this match your expectations?

PR #631, opened 2024-08-20
Fix parsing of external property values containing `=`

### LINKED ISSUE #630: Regression between 0.26.1 and 0.26.3: properties provided on the commandline with `=` in them are parsed differently
This appears to be a regression between `0.26.1` and `0.26.3`. Properties that have an `=` symbol in them are parsed differently.

## Reproduction

```
# 0.26.1
❯ ./pkl-0.26.1 eval -x 'read("prop:foo")' -p foo=bar=baz pkl:base
bar=baz

# 0.26.3
❯ ./pkl-0.26.3 eval -x 'read("prop:foo")' -p foo=bar=baz pkl:base
bar
```

holzensp:
Fixes #630


stackoverflow (approved):
My bad. Missed that case...

PR #640, opened 2024-09-09
Refactor message-passing logic into pkl-core

HT154:
This is preparatory work for [SPICE-0009](https://github.com/apple/pkl-evolution/pull/10). It is being contributed in a separate pull request to ease review.

The Message, Message(En|De)coder, and MessageTransport types have been ported to Java and moved to a new `org.pkl.core.messaging` package.


pkl-server/src/test/kotlin/org/pkl/server/JvmServerTest.kt
```diff
@@ -19,17 +19,23 @@ import java.io.PipedInputStream
...
     if (USE_DIRECT_TRANSPORT) {
-      MessageTransports.direct()
+      MessageTransports.direct(::log)
     } else {
       val in1 = PipedInputStream()
       val out1 = PipedOutputStream(in1)
       val in2 = PipedInputStream()
       val out2 = PipedOutputStream(in2)
-      MessageTransports.stream(in1, out2) to MessageTransports.stream(in2, out1)
+      Pair.of(
```
HT154, line 35: This is using `org.pkl.core.util.Pair` instead of Kotlin's `Pair` now that it's implemented in Java since the API's similar enough to work as a drop-in. Let me know if there's something different I should do!

Issue #641, opened 2024-09-09
PklBugException when reading assets from a local project dependency


bioball:
Using `read()` on an asset in a local project dependency causes an exception like:

```
–– Pkl Error ––
Resource reader `org.pkl.core.resource.ResourceReaders$ProjectPackageResource` returned unsupported resource type `class org.pkl.core.runtime.VmTyped`.
```

PR #642, opened 2024-09-09
Fix bug when reading local package assets

bioball:
Closes https://github.com/apple/pkl/issues/641


pkl-core/src/main/java/org/pkl/core/resource/ResourceReaders.java
```diff
@@ -488,7 +488,16 @@ public Optional<Object> read(URI uri)
...
+        var resourceManager = VmContext.get(null).getResourceManager();
+        var securityManager = VmContext.get(null).getSecurityManager();
+        securityManager.checkReadResource(local);
+        var reader = resourceManager.getResourceReader(local);
+        if (reader == null) {
+          throw new VmExceptionBuilder()
+              .evalError("noResourceReaderRegistered", uri.getScheme())
+              .build();
+        }
+        return resourceManager.doRead(reader, local, null);
```
bioball, line 500: The core problem is around using `ResourceManager#read`, which returns a `org.pkl.core.runtime.VmTyped` rather than a `org.pkl.core.resource.Resource`.

Additionally, we shouldn't use the ResourceManager here because that method statefully alters its cache. Since this method is already called within the same cache's `computeIfAbsent` call, mutating that cache results in a ConcurrentModificationException.

stackoverflow (approved):
Looks good.

PR #647, opened 2024-09-15
Add ResourceReaderFactory abstraction

HT154:
This is preparatory work for [SPICE-0009](https://github.com/apple/pkl-evolution/pull/10). It is being contributed in a separate pull request to ease review.

This is required to allow deferred launch of external reader processes without requiring up-front configuration of scheme->process mappings. More details on why this is required here: https://github.com/apple/pkl-evolution/pull/10#discussion_r1724145677

One concern here (which applies before this change as well) is that the semantics of module key factories and resource reader factories is reversed. For module key factories, the first factory that answers for a URI scheme "wins", while for resource reader factories, the _last_ reader factory that answers for a URI scheme "wins". This PR preserves that behavior, but it may be worth reconsidering this design at this time.

This behavior can be observed in practice in <code>ResourceReadersEvaluatorTest.\`module path\`</code> where `ResourceReaders.modulePath` is added after the pre-configured `ResourceReaders.classPath` and takes precedence.

This is a fairly large breaking API change, but in practice most clients will only need to replace a few lines in their calls to `EvaluatorBuilder`, eg.
```java
.addResourceReader(ResourceReaders.environmentVariable())
// becomes
.addResourceReaderFactory(ResourceReaderFactories.environmentVariable())
```

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/resource/ResourceReaderFactory.java
+++ b/pkl-core/src/main/java/org/pkl/core/resource/ResourceReaderFactory.java
@@ -1,0 +1,38 @@
+/**
+ * Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.core.resource;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Optional;
+
+/** A factory for {@link ResourceReader}s. */
+public interface ResourceReaderFactory {
+
+  /**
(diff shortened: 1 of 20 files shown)


pkl-core/src/main/java/org/pkl/core/resource/ResourceReaderFactory.java
```diff
@@ -0,0 +1,38 @@
...
+   *
+   * <p>Implementations must not perform any I/O related to the given URI. For example, they must
+   * not check if the module represented by the given URI exists.
+   *
+   * <p>Throws {@link URISyntaxException} if the given URI has invalid syntax.
+   *
+   * @param uri an absolute normalized URI
+   * @return a resource for the given URI
+   */
+  Optional<ResourceReader> create(URI uri);
```
HT154, line 37: For implementing external readers, this method will be the point at which the child process is spawned (if it is not already running) and the resource/module readers it implements will be discovered. When the process needs to be spawned, this call will incur that one-time delay.

bioball:
> One concern here (which applies before this change as well) is that the semantics of module key factories and resource reader factories is reversed. For module key factories, the first factory that answers for a URI scheme "wins", while for resource reader factories, the last reader factory that answers for a URI scheme "wins". This PR preserves that behavior, but it may be worth reconsidering this design at this time.

Ah, that's pretty unfortunate... that would be a pretty surprising change of behavior, for interfaces that otherwise look so similar. It'd probably be better to flip the order for resource readers, at the cost of a bigger breaking change here.

I'm am wondering if we do want the factory approach here. Do we need external module/resource readers to have unknown schemes ahead of time? Let's chat more about it [on the spice](https://github.com/apple/pkl-evolution/pull/10/files#r1762311549).

PR #655, opened 2024-09-20
add jbang catalog

maxandersen:
This makes it possible to use [jbang](https://jbang.dev) to easily run and install `pkl` jar based cli.

Once merged you will be able to run it like this:

`jbang pkl@apple/pkl`

you can also install it and then just use it as a script from wherever:

```
jbang app install pkl@apple/pkl
pkl --version
```

I include in this pr a `renovate.json` that enables automatic PR updates if you are able to install https://github.com/apps/forking-renovate on this repo.

if not, you will have to manually update the version - or we change the version used in jbang-catalog to `RELEASE` and it will fetch latest release always.

the `build.yml` is used to validate prs opened by renovate.

Thus if renovate is not an option I would remove those - for now just include it. Let me know and I can adjust PR.


maxandersen:
pushed update with alias for java and kotlin code generator.

`jbang app install pkl-codegen-java@apple/pkl`

and

``jbang app install pkl-codegen-kotlin@apple/pkl`

jbang-catalog.json
```diff
@@ -0,0 +1,18 @@
...
+    "pkl": {
+      "script-ref": "org.pkl-lang:pkl-cli-java:0.26.3",
+      "java-agents": []
+    },
+    "pkl-codegen-java": {
+      "script-ref": "org.pkl-lang:pkl-codegen-java:0.26.3",
+      "java-agents": []
+    },
+    "pkl-codegen-kotlin": {
+      "script-ref": "org.pkl-lang:pkl-codegen-kotlin:0.26.3",
```
bioball, line 13: The current version is 0.27.0-SNAPSHOT

stackoverflow:
We don't use GitHub actions, so you'll have to remove that part, but otherwise, looks fine! We'll update the version in the catalog ourselves whenever a new Pkl version is released.

StefMa:
> We don't use GitHub actions, so you'll have to remove that part [..]

It's free to use for open source projects. Why not (starting) using it? 🤔 
Doesn't mean you should switch everything to it, but maybe smaller checks/updates like this make sense?!

bioball:
We may explore it at some point! But, for now, we're sticking with CircleCI.

.github/workflows/build.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  push:
+    branches:
+    - main
+  pull_request:
+
+jobs:
+  build:
+    uses: jbanghub/.github/.github/workflows/shared-build.yml@main
+    secrets: inherit
```
StefMa, line 12: BTW, this has no effect as you can share only secrets in reusable workflows within the same org or between org in the same enterprise. 

Beside if that it looks also pretty dangerous 

See https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idsecretsinherit

maxandersen: good point - not actually sure why that is there by default. but i'll remove the .github part anyway

jbang-catalog.json
maxandersen: 0.27.0-snapshot is not availabe in maven central so would be bad to point to it.

maxandersen:
> We don't use GitHub actions, so you'll have to remove that part, but otherwise, looks fine! We'll update the version in the catalog ourselves whenever a new Pkl version is released.

cool - pushed update where no .github actions enabled.

stackoverflow (approved):
Thanks for your contribution.

Issue #656, opened 2024-09-22
Custom ResourceReader implementation in Golang fails to decode read data


barahona42:
# context

Hi all, first time submitting a bug report ever. Here's the info:

- What are you trying to achieve?
  - Implement a way for pkl to retrieve configuration values from AWS SSM Parameters when `read("ssm:/foo/SomeParameterPath")` is assigned to a pkl resource.
  - Working language: `go1.23.0`
- What's the impact of this bug?
  - Unmarshaling/resolving values from a module is failing.

## outputs

- `pkl --version: Pkl 0.26.3 (macOS 14.3.1, native)`
- complete error message:
    ```
    java.lang.InterruptedException

    –– Pkl Error ––
    None (cause has no message)

    1 | SomeParameter = read("ssm:ParameterName")
        ^^^^^^^^^^^^^^^^
    at config#SomeParameter (file:///etc/server/config.pkl)

    Pkl 0.26.3 (Linux 5.15.0-1057-aws, native)

    java.lang.InterruptedException
            at java.base@17.0.10/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:386)
            at java.base@17.0.10/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2073)
            at org.pkl.server.UtilsKt.getUnderlying(Utils.kt:64)
            at org.pkl.server.ClientResourceReader.doRead(ClientResourceReader.kt:104)
            at org.pkl.server.ClientResourceReader.read(ClientResourceReader.kt:46)
            at org.pkl.core.runtime.ResourceManager.lambda$read$1(ResourceManager.java:91)
            at java.base@17.0.10/java.util.HashMap.computeIfAbsent(HashMap.java:1220)
            at org.pkl.core.runtime.ResourceManager.read(ResourceManager.java:72)
            at org.pkl.core.ast.expression.unary.AbstractReadNode.doRead(AbstractReadNode.java:55)
            at org.pkl.core.ast.expression.unary.ReadNode.read(ReadNode.java:33)
            at org.pkl.core.ast.expression.unary.ReadNodeGen.executeAndSpecialize(ReadNodeGen.java:75)
            at org.pkl.core.ast.expression.unary.ReadNodeGen.executeGeneric(ReadNodeGen.java:66)
            at org.pkl.core.ast.PklRootNode.executeBody(PklRootNode.java:41)
            at org.pkl.core.ast.MemberNode.executeBody(MemberNode.java:53)
[...]

PR #657, opened 2024-09-22
Handle `null` contents gracefully in message passing API

HT154:
In messages "Read Resource Response" and "Read Module Response", if `contents` and `error` are both null, default to an empty byte array/string.

This resolves one of the issues reported in #656


bioball (approved):
Awesome, thanks for the fix!

PR #666, opened 2024-10-01
Add `String.splitLimit` API

HT154:
Open to naming input here!
* Java, Kotlin, and Swift use an overload of `String.split` for this
* Golang calls this [`SplitN`](https://pkg.go.dev/strings#SplitN)
* Python uses a defaulted argument on `str.split`

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/stdlib/base/StringNodes.java
+++ b/pkl-core/src/main/java/org/pkl/core/stdlib/base/StringNodes.java
@@ -739,6 +739,22 @@ public final class StringNodes {
     }
   }
 
+  public abstract static class splitLimit extends ExternalMethod2Node {
+    @TruffleBoundary
+    @Specialization
+    protected VmList eval(String self, String separator, long limit) {
+      var parts = self.split(Pattern.quote(separator), (int) limit);
+      return VmList.create(parts);
+    }
+
+    @TruffleBoundary
+    @Specialization
+    protected VmList eval(String self, VmRegex separator, long limit) {
+      var parts = separator.getPattern().split(self, (int) limit);
+      return VmList.create(parts);
+    }
+  }
+
   public abstract static class capitalize extends ExternalMethod0Node {
     @TruffleBoundary
     @Specialization
--- a/pkl-core/src/test/files/LanguageSnippetTests/output/api/string.pcf
+++ b/pkl-core/src/test/files/LanguageSnippetTests/output/api/string.pcf
@@ -109,6 +109,16 @@ examples {
     List("aa", "caa", "c")
     List("aabb", "aabb")
   }
+  ["splitLimit()"] {
+    List("aabbccaabbccaabbccaabbcc")
+    List("aa", "caabbccaabbccaabbcc")
+    List("aa", "caa", "caa", "caabbcc")
+    List("aabbccaabbccaabbccaabbcc")
+    List("aabb", "aabbccaabbccaabbcc")
+    List("aabb", "aabb", "aabb", "aabbcc")
+    "Type constraint `this > 0` violated. Value: 0"
+    "Type constraint `this > 0` violated. Value: -1"
+  }
   ["replaceAll()"] {
     "xxbbccxxbbcc"
     "aa(bbc|bb)caa(bbc|bb)c"
--- a/stdlib/base.pkl
+++ b/stdlib/base.pkl
@@ -1370,6 +1370,15 @@ external class String extends Any {
   /// Splits this string matches of [pattern], up to [limit] substrings.
   /// 
   /// Returns a [List] with at most [limit] elements.
+  /// If the limit has been reached, the last entry will contain the un-split remainder of this string.
+  ///
+  /// Facts:
+  /// ```
+  /// "a.b.c".splitLimit(".", 2) == List("a", "b.c")
+  /// "a.b.c".splitLimit(".", 1) == List("a.b.c")
+  /// "a.b.c".splitLimit(".", 50) == List("a", "b", "c")
+  /// "a.b:c".splitLimit(Regex("[.:]"), 3) == List("a", "b", "c")
+  /// ```
   @Since { version = "0.27.0" } 
   external function splitLimit(pattern: String|Regex, limit: Int(this > 0)): List<String>
 
--- a/stdlib/base.pkl
+++ b/stdlib/base.pkl
@@ -1367,6 +1367,12 @@ external class String extends Any {
   /// Splits this string around matches of [pattern].
   external function split(pattern: String|Regex): List<String>
 
+  /// Splits this string matches of [pattern], up to [limit] substrings.
+  /// 
+  /// Returns a [List] with at most [limit] elements.
+  @Since { version = "0.27.0" } 
+  external function splitLimit(pattern: String|Regex, limit: Int(this > 0)): List<String>
+
   /// Converts the first character of this string to title case.
   ///
   /// Facts:
--- a/pkl-core/src/test/files/LanguageSnippetTests/input/api/string.pkl
+++ b/pkl-core/src/test/files/LanguageSnippetTests/input/api/string.pkl
@@ -138,6 +138,19 @@ examples {
     str4.split("cc")
   }
 
+  local str7 = "aabbccaabbccaabbccaabbcc"
+
+  ["splitLimit()"] {
+    str7.splitLimit(Regex("b+c"), 1)
+    str7.splitLimit(Regex("b+c"), 2)
+    str7.splitLimit(Regex("b+c"), 4)
+    str7.splitLimit("cc", 1)
+    str7.splitLimit("cc", 2)
+    str7.splitLimit("cc", 4)
+    module.catch(() -> str7.splitLimit(Regex("b+c"), 0))
+    module.catch(() -> str7.splitLimit(Regex("b+c"), -1))
+  }
+
   ["replaceAll()"] {
     str4.replaceAll("aa", "xx")
     str4.replaceAll(Regex("(b+)c"), "($0|$1)")
--- a/pkl-core/src/test/files/LanguageSnippetTests/output/api/reflectedDeclaration.pcf
+++ b/pkl-core/src/test/files/LanguageSnippetTests/output/api/reflectedDeclaration.pcf
@@ -1333,11 +1333,33 @@ alias {
           parameters = Map("pattern", new {
             name = "pattern"
           })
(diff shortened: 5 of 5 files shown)


stdlib/base.pkl
```diff
@@ -1367,6 +1367,12 @@ external class String extends Any {
   /// Splits this string around matches of [pattern].
   external function split(pattern: String|Regex): List<String>
 
+  /// Splits this string around the first [limit] matches of [pattern].
```
holzensp, line 1370: Technically, around the first `limit-1` matches, no?

How much of a win (in what context/application) is this compared to `str.split(pat).take(limit)`?

HT154: This came up in the context of inputs like `key=value=with=equals` where eg. you want to split `key` from the remainder after the first `=`. This is easy to accomplish like this:
```pkl
input = "key=value=with=equals"
delimiter = "="
key = input.split(delimiter).first
value = input.split(delimiter).drop(1).join(delimiter)
```

This gets significantly trickier when the split separator is a `Regex`, i.e. you must support both `key=value` and `key:value`
```pkl
input = "key=value=with=equals"
delimiter = Regex("[:=]")
key = input.split(delimiter).first
value = input.split(delimiter).drop(1).join(<???>) // <- no possible always-valid input here
```

You have to approach this differently, eg.
```pkl
input = "key=value=with=equals"
delimiter = Regex("[:=]")
key = input.split(delimiter).first
value = input.drop(input.indexOf(delimiter) + 1)
```

This requires reason about character indexes and match length. In this case the match is always a single character, but more complex cases (eg. the delimiter can be multiple or a variable number of characters) require additional handling. And—subjectively—is more difficult to read fluently.

With `splitLimit` this becomes much simpler to write and reason about:
```
input = "key=value=with=equals"
delimiter = Regex("[:=]")
key = input.split(delimiter).first
value = input.splitLimit(delimiter, 2).last
```

HT154: Docstring updated!

stdlib/base.pkl
```diff
@@ -1367,6 +1367,12 @@ external class String extends Any {
   /// Splits this string around matches of [pattern].
   external function split(pattern: String|Regex): List<String>
 
+  /// Splits this string matches of [pattern], up to [limit] substrings.
```
StefMa, line 1370: What if you use 0 or -1 as a limit? 🤔
Would it return nothing or a full list?

I guess it makes sense to document that behavior.

StefMa (commented):
Just a small question about special cases 🙃

About naming:
I somewhat don't like limit, however I don't have a better proposal 🙈
I like Go splitN, but this naming convention doesn't fit to pkl at all.
I also thought about splitCount, but actually we don't count anything 🤷‍♂️

stdlib/base.pkl
bioball: Unlike Java, we can define `limit` to _not_ include 0 or negative numbers! Perhaps this should be `limit: Int(this > 0)`

HT154: I think it might just be best (and simplest) to mimic (and document) the Java behavior here, which is 

> * If the limit `n` is greater than zero then the pattern will be applied at most `n - 1` times, the array's length will be no greater than `n`, and the array's last entry will contain all input beyond the last matched delimiter. > * If `n` is non-positive then the pattern will be applied as many times as possible and the array can have any length.
> * If `n` is zero then the pattern will be applied as many times as possible, the array can have any length, and trailing empty strings will be discarded.

StefMa: > Unlike Java, we can define `limit` to _not_ include 0 or negative numbers! Perhaps this should be `limit: Int(this > 0)`

The question is, if pkl needs that behvior ? 🤔 
Are there use cases for -1 and 0 (as described by @HT154)? 🤔 

Additionally, instead of making this function super complicated to understand (with these rules) we might want to introduce other functions for the "zero and negative number" behavior? 
Not sure if this is a good idea though...

bioball: `splitLimit(char, -1)` is the same as `split(char)`

`splitLimit(char, 0)` is, weirdly, `split(char).dropLastWhile((it) -> it.empty)`, which is weirdly specific (and strange).

In both of these cases, I don't really see a strong use-case for them. 

Side-note: many of our stdlib APIs expect positive values, for example, `List.drop()` and `List.take()`, but this check is implemented in the external method rather than using `UInt` as a type. We should probably fix them.

StefMa: I agree with you.
And since pkl is a configuration language (and not a full programming language), I do think it makes sense to add `limit > 0` ✅

This will remove a lot of complexity for things that are rarely (if even) used.

stdlib/base.pkl
holzensp: Ah, but in that case, I'd expect
```
  local str7 = "aabbccaabbccaabbccaabbcc"

   ["splitLimit()"] {
     str7.splitLimit(Regex("b+c"), 1)
```
to correspond to
```
["splitLimit()"] {
     List("aa", "caabbccaabbccaabbcc")
```
instead of the current
```
["splitLimit()"] {
     List("aa")
```

... but I must be misunderstanding something, because in your example, I would then expect
```
value = input.splitLimit(delimiter, 1).last
```
(if the `limit` counts the matches _of the separator_).

holzensp: Ignore that last bit; I hadn't absorbed

> If the limit n is greater than zero then the pattern will be applied at most n - 1 times, the array's length will be no greater than n, and the array's last entry will contain all input beyond the last matched delimiter.

stdlib/base.pkl
HT154: Updated. Is it also worth checking this in the external code?

stackoverflow: It's not necessary to check in the external code, if we ever get a limit <= 0 there it's a bug we should fix.

pkl-core/src/test/files/LanguageSnippetTests/output/api/string.pcf
```diff
@@ -109,6 +109,14 @@ examples {
     List("aa", "caa", "c")
     List("aabb", "aabb")
   }
+  ["splitLimit()"] {
+    List("aa")
+    List("aa", "caa")
+    List("aa", "caa", "caa", "caa")
+    List("aabbccaabbccaabbccaabbcc")
+    List("aabb", "aabbccaabbccaabbcc")
+    List("aabb", "aabb", "aabb", "aabbcc")
```
holzensp, line 118: Separate comment (for visibility, because former is on `Outdated` code), but also to put in a `suggestion`;

I played with Java's `String::split`, GoLang's `strings.SplitN`, and Python's `str.split`. For each of them, the following invariant holds:

```
let (result = string.splitLimit(separator, limit))
  string.endsWith(separator) | string.endsWidth(result.last)
```

Another invariant diverges between, on the one hand, Java & GoLang and, on the other, Python. 
```
string.splitLimit(separator, limit).length.isBetween(0, if (isPython) limit + 1 else limit)
```
(i.e. Java & GoLang count maximum post-split chunks, Python counts maximum separator occurrences)

Here, I was confused by the lack of the first invariant I mentioned, but upon closer inspection, that's only for the `Regex` case; your `String` case behaves like I would expect.
```suggestion
    List("aabbccaabbccaabbccaabbcc")
    List("aa", "caabbccaabbccaabbcc")
    List("aa", "caa", "caa", "caabbcc")
    List("aabbccaabbccaabbccaabbcc")
    List("aabb", "aabbccaabbccaabbcc")
    List("aabb", "aabb", "aabb", "aabbcc")
```

(Can you tell I would really _love_ property-based testing for Pkl?)

pkl-core/src/main/java/org/pkl/core/stdlib/base/StringNodes.java
```diff
@@ -739,6 +739,21 @@ protected VmList eval(String self, VmRegex separator) {
...
+    @Specialization
+    protected VmList eval(String self, String separator, long limit) {
+      var parts = self.split(Pattern.quote(separator), (int) limit);
+      return VmList.create(parts);
+    }
+
+    @TruffleBoundary
+    @Specialization
+    protected VmList eval(String self, VmRegex separator, long limit) {
+      return VmList.create(separator.getPattern().split(self), (int) limit);
```
holzensp, line 753: It took me far too long to find this, but considering my comment on your snippet test expectation, I _believe_ this should be...
```suggestion
      return VmList.create(separator.getPattern().split(self, (int) limit));
```
I don't mind either way, but the difference would have been easier to spot if this was written in the same form as `eval(String,String,long)`;
```
var parts = separator.getPattern().split(self, (int) limit);
return VmList.create(parts);
```

holzensp (changes requested):
Unless I've misunderstood... I believe I caught an actual bug.

pkl-core/src/main/java/org/pkl/core/stdlib/base/StringNodes.java
HT154: Oof good catch. This is fixed now.

stdlib/base.pkl
```diff
@@ -1367,6 +1367,12 @@ external class String extends Any {
...
+  /// 
+  /// Returns a [List] with at most [limit] elements.
```
bioball, line 1372:
```suggestion
  /// Splits this string matches of [pattern], up to [limit] substrings.
  /// 
  /// Returns a [List] with at most [limit] elements.
  /// If the limit has been reached, the last entry will contain the un-split remainder of this string.
  ///
  /// Facts:
  /// ```
  /// "a.b.c".splitLimit(".", 2) == List("a", "b.c")
  /// "a.b.c".splitLimit(".", 1) == List("a.b.c")
  /// "a.b.c".splitLimit(".", 50) == List("a", "b", "c")
  /// "a.b:c".splitLimit(Regex("[.:]"), 3) == List("a", "b", "c")
  /// ```
```

bioball (changes requested):
I think let's beef up the docs a little bit, but I'm good with this addition to the string API. The name `splitLimit` seems fine to me too.

bioball (approved):
LGTM, over to @holzensp and @stackoverflow to review

stackoverflow (approved):
Looks good!

HT154:
Ah this needs a snippet rewrite for the CI check, will do that later today

bioball:
Run `OVERWRITE_SNIPPETS=1 ./gradlew pkl-core:check` to overwrite the snippet test results

holzensp (approved):
Ship it!

PR #672, opened 2024-10-09
Add properties and methods to `Mapping`/`Listing` API to resemble `Map`/`List` more

holzensp:


stdlib/base.pkl
```diff
@@ -1890,9 +1964,18 @@ class Mapping<out Key, out Value> extends Object {
 
   /// The keys contained in this mapping.
   external keys: Set<Key>
+  
+  /// The values contained in this mapping.
+  external values: List<Value>
+  
+  /// The entries contained in this mapping.
+  external entries: List<Pair<Key, Value>>
```
HT154, line 1972: Is there a particular reason why there are `List`s while `keys` is a `Set`? Mappings are ordered (and documented as such), so this sort of makes sense to me. Having this reasoning documented might be helpful.

StefMa:
While we are here, what is the reason that pkl has Mappings and Map and Listing and List? 🤔

bioball:
> While we are here, what is the reason that pkl has Mappings and Map and Listing and List? 

Several reasons (not comprehensive):

* Mapping and Listing are values of type `Object`; their members are lazy and late-bound, which has implications around how the object is represented in the underlying implementation. Map and List are a lot more performant for "algorithm-y" type logic; e.g. concat, map, filter, etc.
* Mapping and Listing have a special property called "default". It's not really clear how something like `.map()` and `.flatMap()` should work; what do you do with the special "default" property?

It'd be good to simplify the language and not have this dichotomy, but we'd need to at least solve the above challenges first.

stdlib/base.pkl
translatenix: IMHO these should be `Listing`s, and `keys` should be changed to `Listing` via a deprecation cycle.

pkl-core/src/main/java/org/pkl/core/runtime/VmMapping.java
```diff
@@ -127,6 +127,20 @@ public Map<Object, Object> toMap() {
 
     return properties;
   }
+  
+  public VmDynamic toDynamic() {
+    var keys = getAllKeys();
+    var builder = new VmObjectBuilder(keys.getLength());
+    for (var key: keys) {
+      var member = VmUtils.readMember(this, key);
```
translatenix, line 135: Instead of forcing and copying each value, the Dynamic could delegate to the Mapping via `addEntry(Object key, SharedMemberNode valueNode)`.
A similar overload could be added for `addProperty` (I didn't need it at the time).

PR #678, opened 2024-10-11
Add support for const object members

bioball:
Follows [SPICE-0011](https://github.com/apple/pkl-evolution/pull/12)

### DIFF
--- a/pkl-core/src/test/files/LanguageSnippetTests/input/basic/constModifier4.pkl
+++ b/pkl-core/src/test/files/LanguageSnippetTests/input/basic/constModifier4.pkl
@@ -1,0 +1,11 @@
+const baz = 15
+
+foo {
+  const local bar = baz
+    
+  const local function biz() = baz
+
+  qux = bar
+
+  corge = biz()
+}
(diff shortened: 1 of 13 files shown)


pkl-core/src/test/files/LanguageSnippetTests/input/basic/constModifier4.pkl
```diff
@@ -0,0 +1,11 @@
+const baz = 15
+
+foo {
+  const local bar = baz
```
holzensp, line 4:
```suggestion
  const local bar = biz()
```

^^ otherwise; you don't cover the `const function` consumption side.

holzensp (approved):
Please roll in the nit in the test, but otherwise LGTM

pkl-core/src/test/files/LanguageSnippetTests/input/basic/constModifier4.pkl
bioball: That's covered on line 10 (see `corge = biz()`)

bioball: Added another line to test calling a local method from a local property

PR #680, opened 2024-10-12
Update Let Expressions Examples

ManuW:
Because I did not understand what a `Let Expression` should be and the example code was more confusing as explanatory, I changed the format of the examples a little bit and changed the result.

- It should be clear that the example is just one code-line
- I tried out the example (`pkl eval...`) and get a little bit different result


holzensp (approved):
These are legit improvements. Thank you.

PR #683, opened 2024-10-14
Add methods from `List`/`Map` to `Listing`/`Mapping`

holzensp:
Methods that "project out" of the type can be added to `Listing` and `Mapping` to avoid having to use `toList()`/`toMap()`. Methods that result in a `Listing`, resp. `Mapping`, are not included for performance and (lazy) semantics reasons, but methods like `contains` etc. can be defined straightforwardly.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmMapping.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmMapping.java
@@ -15,14 +15,20 @@
  */
 package org.pkl.core.runtime;
 
+import com.oracle.truffle.api.CompilerDirectives;
 import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
+import com.oracle.truffle.api.frame.FrameDescriptor;
 import com.oracle.truffle.api.frame.MaterializedFrame;
 import java.util.Map;
 import java.util.Objects;
 import javax.annotation.concurrent.GuardedBy;
+import org.graalvm.collections.EconomicMap;
 import org.graalvm.collections.UnmodifiableEconomicMap;
+import org.pkl.core.ast.VmModifier;
+import org.pkl.core.ast.member.DelegateToExtraStorageObjOrParentNode;
 import org.pkl.core.ast.member.ListingOrMappingTypeCastNode;
 import org.pkl.core.ast.member.ObjectMember;
+import org.pkl.core.ast.member.UntypedObjectMemberNode;
 import org.pkl.core.util.CollectionUtils;
 import org.pkl.core.util.EconomicMaps;
 import org.pkl.core.util.LateInit;
@@ -126,6 +132,47 @@ public final class VmMapping extends VmListingOrMapping<VmMapping> {
         });
 
     return properties;
+  }
+
+  public VmDynamic toDynamic() {
+    EconomicMap<Object, ObjectMember> members = EconomicMaps.create(this.members.size());
+    iterateMemberValues(
+        (key, member, value) -> {
+          if (key instanceof String name) {
+            var identifier = Identifier.get(name);
+            if (isDefaultProperty(identifier)) {
+              CompilerDirectives.transferToInterpreter();
+              throw new VmExceptionBuilder().evalError("cannotConvertDefaultProperty").build();
+            }
+            var property =
+                new ObjectMember(
+                    VmUtils.unavailableSourceSection(),
+                    VmUtils.unavailableSourceSection(),
+                    VmModifier.NONE,
+                    identifier,
+                    name);
+            property.initMemberNode(
+                new UntypedObjectMemberNode(
+                    null,
+                    new FrameDescriptor(),
+                    property,
+                    new DelegateToExtraStorageObjOrParentNode(true)));
+            members.put(identifier, property);
+          } else {
+            assert isDefaultProperty(key);
+            members.put(key, member);
+          }
+
+          return true;
+        });
+    var result =
+        new VmDynamic(
+            VmUtils.createEmptyMaterializedFrame(),
+            BaseModule.getDynamicClass().getPrototype(),
+            members,
+            0);
+    result.setExtraStorage(this);
+    return result;
   }
 
   @Override
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmMapping.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmMapping.java
@@ -15,20 +15,14 @@
  */
 package org.pkl.core.runtime;
 
-import com.oracle.truffle.api.CompilerDirectives;
 import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
-import com.oracle.truffle.api.frame.FrameDescriptor;
 import com.oracle.truffle.api.frame.MaterializedFrame;
 import java.util.Map;
 import java.util.Objects;
 import javax.annotation.concurrent.GuardedBy;
-import org.graalvm.collections.EconomicMap;
 import org.graalvm.collections.UnmodifiableEconomicMap;
-import org.pkl.core.ast.VmModifier;
-import org.pkl.core.ast.member.DelegateToExtraStorageObjOrParentNode;
 import org.pkl.core.ast.member.ListingOrMappingTypeCastNode;
 import org.pkl.core.ast.member.ObjectMember;
-import org.pkl.core.ast.member.UntypedObjectMemberNode;
 import org.pkl.core.util.CollectionUtils;
 import org.pkl.core.util.EconomicMaps;
 import org.pkl.core.util.LateInit;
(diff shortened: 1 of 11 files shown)


stdlib/base.pkl
```diff
@@ -1901,9 +1984,22 @@ class Mapping<out Key, out Value> extends Object {
...
+  /// Tells if [predicate] holds for at least one entry of this mapping.
+  /// 
+  /// Returns [false] for an empty mapping.
+  external function any(predicate: (Key, Value) -> Boolean): Boolean
 
   /// Converts this mapping to a [Map].
   external function toMap(): Map<Key, Value>
+  
+  /// Converts this mapping to a [Dynamic] object.
+  external function toDynamic(): Dynamic
```
translatenix, line 2002: How about documenting that late binding is not preserved?
"Returns a Dynamic that delegates to this mapping" would be a clear explanation of a good implementation.

pkl-core/src/main/java/org/pkl/core/runtime/VmList.java
```diff
@@ -220,6 +220,11 @@ public VmCollection.Builder<VmList> builder() {
     return new Builder();
   }
 
+  @TruffleBoundary
+  public static VmCollection.Builder<VmList> newBuilder() {
```
translatenix, line 224: Existing code uses `VmList.EMPTY.builder()` instead.

pkl-core/src/main/java/org/pkl/core/runtime/VmMapping.java
```diff
@@ -128,6 +128,20 @@ public Map<Object, Object> toMap() {
     return properties;
   }
 
+  public VmDynamic toDynamic() {
+    var keys = getAllKeys();
+    var builder = new VmObjectBuilder(keys.getLength());
+    for (var key : keys) {
```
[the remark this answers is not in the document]
stackoverflow: Having a `Dynamic` delegate seems like a good idea. Would make this function almost a noop.

pkl-core/src/main/java/org/pkl/core/stdlib/base/MappingNodes.java
```diff
@@ -73,6 +75,28 @@ protected VmSet eval(VmMapping self) {
...
   }
 
+  public abstract static class values extends ExternalPropertyNode {
+    @Specialization
+    protected VmList eval(VmMapping self) {
+      var builder = VmList.newBuilder();
+      for (var key : self.getAllKeys()) {
+        builder.add(VmUtils.readMember(self, key));
+      }
+      return builder.build();
```
stackoverflow, line 85: I'm thinking if there's no way to keep this (and `entries`) lazy. Making the `VmList` just a delegate to the map values. Problem I see is the backing `EconomicMap` has just an iterator to the values, so you can't access them by index...

stdlib/base.pkl
```diff
@@ -1843,6 +1863,49 @@ class Listing<out Element> extends Object {
...
+  /// Throws if this listing is empty.
+  ///
+  /// Facts:
+  /// ```
+  /// new Listing { 1 ; 2 ; 3 }.first == 1
+  /// (import "pkl:test").catch(() -> new Listing {}.first)
+  /// ```
+  external first: Element
+
+  /// Same as [first] but returns [null] if this listing is empty.
```
stackoverflow, line 1877:
```suggestion
  /// The first element in this listing or [null] if this listing is empty.
  ///
  /// See: [first]
```
nit: not a big fan of `does the same as X` type of documentation. I'd rather just explain it.
Same for `lastOrNull`, `singleOrNull`.

stackoverflow (dismissed):

stackoverflow:
I'm missing `any` and `every` for `Listing`.

pkl-core/src/main/java/org/pkl/core/stdlib/base/MappingNodes.java
translatenix: A natural way to keep `values` and `entries` lazy would be to return a Listing that delegates to the Mapping.
These more complicated properties/methods could be tackled in a separate PR.
Perhaps `toMap()` should remain the only way to directly transition to eager collections.
(`values.toList()` and `entries.toList()` could also be used.)

bioball (changes requested):
I have mostly nits, but also: I think we should punt these methods to another PR:

* `toDynamic`
* `Mapping.entries`
* `Mapping.values`

Also, there's still many methods that's missing! Not blocking, but it would be good to have. Here's what I see that I think we can add:

* `indexOf`
* `indexOfOrNull`
* `lastIndexOf`
* `lastIndexOfOrNull`
* `find`
* `findOrNull`
* `findLast`
* `findLastOrNull`
* `findIndex`
* `findIndexOrNull`
* `findLastIndex`
* `findLastIndexOrNull`
* `count`
* `foldBack`
* `foldIndexed`
* `reduce`
* `min`
* `minOrNull`
* `minBy`
* `minByOrNull`
* `minWith`
* `minWithOrNull`
* `max`
* `maxOrNull`
* `maxBy`
* `maxByOrNull`
* `maxWith`
* `maxWithOrNull`
* `indexOf`

stdlib/base.pkl
holzensp: I see that, but I also see a point about consistency. This is the common style in the documentation throughout the stdlib (I'm ditto not a fan of "**Tells if** ..." instead of just "Whether ..." for boolean properties, but that's simply the style as is). Punting this suggestion to later work.

stdlib/base.pkl
```diff
@@ -1843,6 +1863,49 @@ class Listing<out Element> extends Object {
...
   external distinct: Listing<Element>
 
+  /// The first element in this listing.
+  ///
+  /// Throws if this listing is empty.
+  ///
+  /// Facts:
+  /// ```
+  /// new Listing { 1 ; 2 ; 3 }.first == 1
+  /// (import "pkl:test").catch(() -> new Listing {}.first)
```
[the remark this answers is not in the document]
holzensp: Copy-paste artefact... No idea where this came from, but it happens throughout `base.pkl`. Adjusted everywhere.

pkl-core/src/test/files/LanguageSnippetTests/input/api/listing.pkl
```diff
@@ -58,6 +70,72 @@ facts {
...
+  ["first"] {
+    module.catch(() -> empty.first) == "Expected a non-empty Listing."
+    base.first == base[0]
+    derived.first == base[0]
+  }
+  
+  ["firstOrNull"] {
+    empty.firstOrNull == null
+    base.firstOrNull == base[0]
+    derived.firstOrNull == base[0]
```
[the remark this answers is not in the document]
holzensp: I did this on purpose to make clearer nothing had changed at all (although, fair enough, this doesn't test object-identity-equality)

pkl-core/src/main/java/org/pkl/core/stdlib/base/ListingNodes.java
```diff
@@ -116,6 +186,52 @@ protected VmListing eval(VmListing self, VmFunction selector) {
...
 
+  public abstract static class every extends ExternalMethod1Node {
+    @Child private ApplyVmFunction1Node applyNode = ApplyVmFunction1Node.create();
+
+    @Specialization
+    protected boolean eval(VmListing self, VmFunction predicate) {
+      var result = new MutableBoolean(true);
+      self.iterateMemberValues(
+          (key, member, unforcedValue) -> {
+            var value = unforcedValue != null ? unforcedValue : VmUtils.readMember(self, key);
```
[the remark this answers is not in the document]
holzensp: Fair enough on not introducing a new variable... not so much `forcedValue`... It's _unforced_ and you can tell whether it happened to have been evaluated (without force) by seeing it isn't `null`.

stdlib/base.pkl
holzensp: > "Returns a Dynamic that delegates to this mapping" would be a clear explanation of a good implementation.

This would be the only case where we document _implementation_ as opposed to _outcome_ in the standard library.

pkl-core/src/main/java/org/pkl/core/stdlib/base/MappingNodes.java
```diff
@@ -86,6 +89,21 @@ protected boolean eval(VmMapping self, Object key) {
     }
   }
 
+  public abstract static class containsValue extends ExternalMethod1Node {
+    @Specialization
+    protected boolean eval(VmMapping self, Object value) {
+      MutableBoolean foundValue = new MutableBoolean(false);
```
bioball, line 95:
```suggestion
      var foundValue = new MutableBoolean(false);
```

pkl-core/src/main/java/org/pkl/core/stdlib/base/MappingNodes.java
```diff
@@ -86,6 +89,21 @@ protected boolean eval(VmMapping self, Object key) {
...
+      self.iterateMemberValues(
+          (key, member, unforcedMemberValue) -> {
+            var memberValue =
+                unforcedMemberValue != null ? unforcedMemberValue : VmUtils.readMember(self, key);
+            foundValue.set(value.equals(memberValue));
```
bioball, line 100:
```suggestion
          (key, member, memberValue) -> {
            if (memberValue == null) {
              memberValue =  VmUtils.readMember(self, key);
            }
            foundValue.set(value.equals(memberValue));
```

pkl-core/src/main/java/org/pkl/core/stdlib/base/MappingNodes.java
```diff
@@ -110,6 +128,38 @@ protected Object eval(VmMapping self, Object initial, VmFunction function) {
...
+  public abstract static class every extends ExternalMethod1Node {
+    @Child private ApplyVmFunction2Node applyLambdaNode = ApplyVmFunction2NodeGen.create();
+
+    @Specialization
+    protected boolean eval(VmMapping self, VmFunction function) {
+      var result = new MutableBoolean(true);
+      self.iterateMemberValues(
+          (key, member, unforcedValue) -> {
+            var value = unforcedValue != null ? unforcedValue : VmUtils.readMember(self, key);
+            result.set(applyLambdaNode.executeBoolean(function, key, value));
```
bioball, line 140:
```suggestion
          (key, member, value) -> {
            if (value == null) {
              value = VmUtils.readMember(self, key);
            }
            result.set(applyLambdaNode.executeBoolean(function, key, value));
```

pkl-core/src/main/java/org/pkl/core/stdlib/base/MappingNodes.java
```diff
@@ -110,6 +128,38 @@ protected Object eval(VmMapping self, Object initial, VmFunction function) {
...
+    @Child private ApplyVmFunction2Node applyLambdaNode = ApplyVmFunction2NodeGen.create();
+
+    @Specialization
+    protected boolean eval(VmMapping self, VmFunction function) {
+      var result = new MutableBoolean(false);
+      self.iterateMemberValues(
+          (key, member, unforcedValue) -> {
+            var value = unforcedValue != null ? unforcedValue : VmUtils.readMember(self, key);
+            result.set(applyLambdaNode.executeBoolean(function, key, value));
+            return !result.get();
```
bioball, line 157:
```suggestion
          (key, member, value) -> {
            if (value == null) {
              value = VmUtils.readMember(self, key);
            }
            result.set(applyLambdaNode.executeBoolean(function, key, value));
            return !result.get();
```

bioball (approved):
Some more suggestions, but otherwise LGTM

PR #688, opened 2024-10-15
Use case-insensitive comparison for module URI schemes

translatenix:
Previously, some factories used case-sensitive comparison, for no apparent reason.


bioball (approved):
LGTM, good catch

PR #689, opened 2024-10-15
Update dependencies

translatenix:
- Update dependencies with `gw updateDependencyLocks`. Most updated dependencies are test dependencies.
- Adapt code to breaking changes in library commonmark.
- Fix test to close PackageServer exactly once (problem went unnoticed with earlier JUnit versions).


gradle/libs.versions.toml
```diff
@@ -38,15 +39,16 @@ ktfmt = "0.44"
 # replaces nuValidator's log4j dependency
 # something related to log4j-1.2-api is apparently broken in 2.17.2
 log4j = "2.17.1"
-msgpack = "0.9.0"
-nexusPublishPlugin = "1.3.0"
+msgpack = "0.+"
```
holzensp, line 42: Not _entirely_ comfortable with a `0.+`, given the instability of `0` major. @bioball / @stackoverflow; thoughts?

holzensp (commented):
Overall, LGTM; just not entirely sure of `0.+` dependency spec; leaving to @bioball and/or @stackoverflow to sign off.

gradle/libs.versions.toml
bioball: I believe @holzensp is saying that we should declare this as `0.9.+`, rather than `0.+`, because per semver rules, new `0.x` versions can have breaking changes.

translatenix: The way I read the semver spec, all bets are off for 0.x.y versions, and 0.9.8 -> 0.9.9 may break everything.
0.9.+ risks missing that 0.10 exists. If you prefer it nevertheless, I don’t mind making this change.

bioball: Ah, you're right. And, looking at their [release history](https://github.com/msgpack/msgpack-java/releases), they have breaking changes in patch versions.

Maybe it's best to fully specify the version here; `msgpack = "0.9.8"`.

translatenix: Personally I'm more worried about falling way behind the latest version than about the build failing to catch an incompatible change. But to move this forward, I'll change back to `msgpack = "0.9.8"`.

By the way, I wasn't happy with the state of msgpack's Java library.
(They also didn't reply to the issues and PRs I opened.)
Eventually I wrote a replacement that I'll soon turn into a public GitHub project. From the README:

## MiniPack at a Glance

* Complete implementation of the MessagePack binary serialization format.
* Clean and modern API with [JSpecify](https://github.com/jspecify/jspecify) nullness annotations.
* Designed to be correct, efficient, and extensible.
* Embraces Java NIO: Heap and direct byte buffers, channels, channel transfers, gathering writes, buffer pooling.
* Fast, correct, and customizable string encoding/decoding with java.nio.charset.CharsetEncoder/CharsetDecoder.
* Optimized handling of strings used as identifiers.
* Small JAR size (~50 KB).
* No use of reflection or internal/unsafe JDK classes.
* Ships as Java (JPMS) module.
* No dependencies other than [JSpecify](https://central.sonatype.com/artifact/org.jspecify/jspecify) (~3 KB).
* Compatible with [GraalVM Native Image](https://www.graalvm.org/latest/reference-manual/native-image/).
* Thoroughly benchmarked with [JMH](https://github.com/openjdk/jmh).
* Thoroughly tested with [jqwik](https://jqwik.net/).

bioball: Sounds good! Happy to entertain switching from msgpack-java to minipack.

translatenix: @bioball ready to squash and merge from my side.

pkl-config-java/gradle.lockfile
```diff
@@ -1,19 +1,18 @@
...
-com.github.ajalt.clikt:clikt-jvm:3.5.1=pklCodegenJava
-com.github.ajalt.clikt:clikt:3.5.1=pklCodegenJava
+com.github.ajalt.clikt:clikt-jvm:3.5.4=pklCodegenJava
+com.github.ajalt.clikt:clikt:3.5.4=pklCodegenJava
 com.squareup:javapoet:1.13.0=pklCodegenJava
-com.tunnelvisionlabs:antlr4-runtime:4.9.0=default,pklCodegenJava,runtimeClasspath,testRuntimeClasspath
-io.leangen.geantyref:geantyref:1.3.14=default
-io.leangen.geantyref:geantyref:1.3.15=compileClasspath,implementationDependenciesMetadata,runtimeClasspath,testCompileClasspath,testImplementationDependenciesMetadata,testRuntimeClasspath
+com.tunnelvisionlabs:antlr4-runtime:4.9.0=pklCodegenJava,runtimeClasspath,testRuntimeClasspath
+io.leangen.geantyref:geantyref:1.3.16=compileClasspath,implementationDependenciesMetadata,runtimeClasspath,testCompileClasspath,testImplementationDependenciesMetadata,testRuntimeClasspath
```
bioball, line 8: Looks like geantyref doesn't have a changelog or release notes right now; filed https://github.com/leangen/geantyref/issues/31

Looking at the git diffs, though, should be fine to bump to 1.3.16.

stdlib/gradle.lockfile
```diff
@@ -1,24 +1,4 @@
...
-org.jetbrains.kotlinx:kotlinx-html-jvm:0.8.1=pkldoc
-org.jetbrains.kotlinx:kotlinx-serialization-bom:1.5.1=pkldoc
-org.jetbrains.kotlinx:kotlinx-serialization-core-jvm:1.5.1=pkldoc
-org.jetbrains.kotlinx:kotlinx-serialization-core:1.5.1=pkldoc
-org.jetbrains.kotlinx:kotlinx-serialization-json-jvm:1.5.1=pkldoc
-org.jetbrains.kotlinx:kotlinx-serialization-json:1.5.1=pkldoc
-org.jetbrains:annotations:13.0=pkldoc
-org.organicdesign:Paguro:3.10.3=pkldoc
-org.snakeyaml:snakeyaml-engine:2.5=pkldoc
-empty=archives,default,signatures
```
bioball, line 24: Pretty weird that these ended up in the lockfile here in the first place.

bioball (approved):
LGTM, thanks for the contributions, including addressing breaking changes!

PR #693, opened 2024-10-16
Set lower limit for glob pattern resolution in test mode

### LINKED ISSUE #691: Snippet test invalidGlobImport6.pkl is too slow
This test takes 15 seconds to run on my machine, hampering developer productivity.

translatenix:
Motivation:
Speed up the test that verifies enforcement of the limit for glob pattern resolution (invalidGlobImport6.pkl).

Changes:
- Set a lower limit if test mode is enabled.
- Change static field to static method to prevent compile-time evaluation by native-image.

Result:
Test takes 0.5 instead of 15 seconds to run (on my machine).

Fixes #691.


bioball (approved):
LGTM

PR #695, opened 2024-10-16
Add analyze imports libs (SPICE-0001)

bioball:
This adds a new feature to build a dependency graph of Pkl programs, following the SPICE outlined in https://github.com/apple/pkl-evolution/pull/2.

It adds:
* CLI command `pkl analyze imports`
* Java API `org.pkl.core.Analyzer`
* Pkl stdlib module `pkl:analyze`
* pkl-gradle extension `analyze`

In addition, it also deprecates `transitiveModules` from pkl-gradle. It instead derives transitive modules by building an import graph of those tasks' source modules.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/Analyzer.java
+++ b/pkl-core/src/main/java/org/pkl/core/Analyzer.java
@@ -1,0 +1,121 @@
+/**
+ * Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.core;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.file.Path;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import org.graalvm.polyglot.Context;
+import org.pkl.core.http.HttpClient;
+import org.pkl.core.http.HttpClientInitException;
+import org.pkl.core.module.ModuleKeyFactory;
+import org.pkl.core.module.ProjectDependenciesManager;
+import org.pkl.core.packages.PackageLoadError;
+import org.pkl.core.packages.PackageResolver;
+import org.pkl.core.project.DeclaredDependencies;
+import org.pkl.core.runtime.ModuleResolver;
+import org.pkl.core.runtime.ResourceManager;
+import org.pkl.core.runtime.VmContext;
+import org.pkl.core.runtime.VmException;
+import org.pkl.core.runtime.VmImportAnalyzer;
+import org.pkl.core.runtime.VmUtils;
+import org.pkl.core.util.Nullable;
+
+/** Utility library for static analysis of Pkl programs. */
+public class Analyzer {
+  private final StackFrameTransformer transformer;
+  private final SecurityManager securityManager;
+  private final @Nullable Path moduleCacheDir;
+  private final @Nullable DeclaredDependencies projectDependencies;
+  private final ModuleResolver moduleResolver;
+  private final HttpClient httpClient;
+
+  public Analyzer(
+      StackFrameTransformer transformer,
+      SecurityManager securityManager,
+      Collection<ModuleKeyFactory> moduleKeyFactories,
+      @Nullable Path moduleCacheDir,
+      @Nullable DeclaredDependencies projectDependencies,
+      HttpClient httpClient) {
+    this.transformer = transformer;
+    this.securityManager = securityManager;
+    this.moduleCacheDir = moduleCacheDir;
+    this.projectDependencies = projectDependencies;
+    this.moduleResolver = new ModuleResolver(moduleKeyFactories);
+    this.httpClient = httpClient;
+  }
+
+  /**
+   * Builds a graph of imports from the provided source modules.
+   *
+   * <p>For details, see {@link ImportGraph}.
+   */
+  public ImportGraph importGraph(URI... sources) {
+    var context = createContext();
+    try {
+      context.enter();
+      var vmContext = VmContext.get(null);
+      var results = VmImportAnalyzer.analyze(sources, vmContext);
+      return new ImportGraph(results.first, results.second);
+    } catch (SecurityManagerException
+        | IOException
+        | URISyntaxException
+        | PackageLoadError
+        | HttpClientInitException e) {
+      throw new PklException(e.getMessage(), e);
+    } catch (PklException err) {
+      throw err;
+    } catch (VmException err) {
+      throw err.toPklException(transformer);
+    } catch (Exception e) {
+      throw new PklBugException(e);
+    } finally {
+      context.leave();
+      context.close();
+    }
+  }
+
+  private Context createContext() {
+    var packageResolver =
+        PackageResolver.getInstance(
+            securityManager, HttpClient.builder().buildLazily(), moduleCacheDir);
+    return VmUtils.createContext(
+        () -> {
+          VmContext vmContext = VmContext.get(null);
+          vmContext.initialize(
+              new VmContext.Holder(
+                  transformer,
+                  securityManager,
+                  httpClient,
+                  moduleResolver,
+                  new ResourceManager(securityManager, List.of()),
+                  Loggers.noop(),
+                  Map.of(),
+                  Map.of(),
+                  moduleCacheDir,
+                  null,
+                  packageResolver,
+                  projectDependencies == null
+                      ? null
+                      : new ProjectDependenciesManager(
+                          projectDependencies, moduleResolver, securityManager)));
+        });
+  }
+}
--- a/pkl-core/src/main/java/org/pkl/core/Analyzer.java
+++ b/pkl-core/src/main/java/org/pkl/core/Analyzer.java
@@ -106,7 +106,7 @@ public class Analyzer {
                   httpClient,
                   moduleResolver,
                   new ResourceManager(securityManager, List.of()),
-                  Loggers.noop(),
+                  Loggers.stdErr(),
                   Map.of(),
                   Map.of(),
                   moduleCacheDir,
--- a/pkl-core/src/main/java/org/pkl/core/Analyzer.java
+++ b/pkl-core/src/main/java/org/pkl/core/Analyzer.java
@@ -72,8 +72,7 @@ public class Analyzer {
     try {
       context.enter();
       var vmContext = VmContext.get(null);
-      var results = VmImportAnalyzer.analyze(sources, vmContext);
-      return new ImportGraph(results.first, results.second);
+      return VmImportAnalyzer.analyze(sources, vmContext);
     } catch (SecurityManagerException
         | IOException
         | URISyntaxException
--- a/pkl-core/src/main/java/org/pkl/core/ImportGraph.java
+++ b/pkl-core/src/main/java/org/pkl/core/ImportGraph.java
@@ -1,0 +1,33 @@
+/**
+ * Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.core;
+
+import java.net.URI;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Java representation of {@code pkl.analyze#ImportGraph}.
+ *
+ * @param imports The graph of imports declared within the program.
+ *     <p>Each key is a module inside the program, and each value is the module URIs decalred as
+ *     imports inside that module. The set of all modules initialized within the program is the set
+ *     of keys in this map.
+ * @param resolvedImports A mapping of a module's in-language URI, and the URI that it resolves to.
+ *     <p>For example, a local package dependency is represented with scheme {@code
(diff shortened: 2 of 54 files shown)


pkl-core/src/main/java/org/pkl/core/runtime/VmImportAnalyzer.java
```diff
@@ -0,0 +1,116 @@
...
+import org.pkl.core.SecurityManagerException;
+import org.pkl.core.ast.builder.ImportsAndReadsParser;
+import org.pkl.core.ast.builder.ImportsAndReadsParser.Entry;
+import org.pkl.core.util.GlobResolver;
+import org.pkl.core.util.GlobResolver.InvalidGlobPatternException;
+import org.pkl.core.util.GlobResolver.ResolvedGlobElement;
+import org.pkl.core.util.IoUtils;
+import org.pkl.core.util.Pair;
+
+public class VmImportAnalyzer {
```
bioball, line 37: There is a `org.pkl.core.Analyzer` API, and this one. This one is an internal, private API, whereas `org.pkl.core.Analyzer` is a public API.

stdlib/analyze.pkl
```diff
@@ -0,0 +1,44 @@
...
+@ModuleInfo { minPklVersion = "0.27.0" }
+module pkl.analyze
+
+// used by doc comments
+import "pkl:reflect"
+
+/// Given a set of Pkl module URIs, returns a graph of imports declared by these modules.
+///
+/// The resulting graph includes transitive imports.
+external function importGraph(moduleUris: Set<String>): ImportGraph
```
translatenix, line 31: Why not `Set<Uri>`?

translatenix: If this is intended, it's worth a comment.

stdlib/analyze.pkl
```diff
@@ -0,0 +1,44 @@
...
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+/// A library for statically analyzing Pkl modules.
+///
+/// These tools differentiate from [pkl:reflect][reflect] in that they parse Pkl modules, but do not
+/// execute any code within these modules.
+@Since { version = "0.27.0" }
+@ModuleInfo { minPklVersion = "0.27.0" }
+module pkl.analyze
```
translatenix, line 23: This seems to be the first stdlib module name that's clearly a verb instead of a noun. Bringing this up because of consistency.

stdlib/analyze.pkl
```diff
@@ -0,0 +1,44 @@
...
+  /// The imports declared within a Pkl program.
+  ///
+  /// Each entry is a module URI, mapped to the set of imports declared in that module.
+  ///
+  /// The set of all modules in the graph can be obtained via its keys.
+  imports: Map<Uri, Set<Uri>>
+
+  /// Mappings of modules from their in-language URI, to the URI that is used to fetch the module's
+  /// contents.
+  resolvedImports: Map<Uri, Uri>(keys == imports.keys)
```
translatenix, line 43: From the doc comments, I couldn't tell how `resolveImports` relates to `imports` ("each entry" vs. "mappings", "module URI" vs. "in-language URI"). It only clicked when I saw the constraint.

stdlib/analyze.pkl
```diff
@@ -0,0 +1,44 @@
...
+
+// used by doc comments
+import "pkl:reflect"
+
+/// Given a set of Pkl module URIs, returns a graph of imports declared by these modules.
+///
+/// The resulting graph includes transitive imports.
+external function importGraph(moduleUris: Set<String>): ImportGraph
+
+class ImportGraph {
```
translatenix, line 33: should this have a doc comment?

pkl-gradle/src/main/java/org/pkl/gradle/spec/ModulesSpec.java
```diff
@@ -23,6 +23,8 @@
 public interface ModulesSpec extends BasePklSpec {
   ListProperty<Object> getSourceModules();
 
+  /** As of Pkl 0.27, this setting is ignored. */
```
translatenix, line 26: Would help to explain how users should deal with this change. Presumably this is just inferred automatically now?

pkl-core/src/main/java/org/pkl/core/stdlib/analyze/AnalyzeNodes.java
```diff
@@ -0,0 +1,101 @@
...
+    @TruffleBoundary
+    protected Object eval(@SuppressWarnings("unused") VmTyped self, VmSet moduleUris) {
+      var uris = new URI[moduleUris.getLength()];
+      var idx = 0;
+      for (var moduleUri : moduleUris) {
+        URI uri;
+        try {
+          uri = new URI((String) moduleUri);
+        } catch (URISyntaxException e) {
+          CompilerDirectives.transferToInterpreter();
```
translatenix, line 78: I can't spot any other code that transfers to interpreter behind a Truffle boundary. Does this have any effect?

stdlib/analyze.pkl
bioball: We also have "reflect"

pkl-gradle/src/main/java/org/pkl/gradle/spec/ModulesSpec.java
bioball: Actually, I think I'm going to bring this back. Analyzing imports is an extra step that adds time to the best case, that maybe users want to opt-out of.

I think the behavior should be: if not set, compute the import graph.

translatenix: Giving users the option to trade correctness for performance sounds a bit scary. Is it worth it?

stdlib/analyze.pkl
translatenix: In my mind, "reflect" is short for "reflection". What worries me is that `analyze.importGraph()` inverts the convention of using nouns for classes/modules and verbs for functions.

pkl-gradle/src/main/java/org/pkl/gradle/spec/ModulesSpec.java
bioball: If users want to specify transitive modules on their own, they can provide a very coarse set, something like this (as you pointed out actually):

```
transitiveModules = fileTree("src") { include("**/*.pkl" }
```

It sounds like a good idea to compute this by default, but provide knobs for users that know what they're doing and want to override it. Does that still sound scary?

translatenix: I just wonder if it's worth keeping a manual knob that adds complexity and can cause correctness issues. I can't really tell, just wanted to bring it up.

bioball: Our ANTLR parser is quite slow (it takes ~170ms to parse `pkl:base` with jpkl on my machine). We're looking into swapping this out with a hand-rolled parser. Until then, running this as an extra step can be painful for some users, especially if there's no way to opt-out.

pkl-core/src/main/java/org/pkl/core/runtime/VmWrappedEvalException.java
```diff
@@ -0,0 +1,59 @@
...
+        cause,
+        isExternalMessage,
+        messageArguments,
+        programValues,
+        location,
+        sourceSection,
+        memberName,
+        hint,
+        insertedStackFrames);
+    this.wrappedException = wrappedException;
```
holzensp, line 53: What's the difference between a "wrapped exception" and an "exception with a cause"? In other words; do we really need a separate class for this?

pkl-core/src/main/java/org/pkl/core/stdlib/analyze/AnalyzeNodes.java
holzensp: I believe you're right; no transfers necessary (or useful) behind a boundary.

pkl-core/src/test/files/LanguageSnippetTests/input/api/analyze1.pkl
```diff
@@ -0,0 +1,33 @@
...
+
+examples {
+  ["basic"] {
+    analyze.importGraph(Set(reflect.Module(a).uri))
+  }
+  ["cycles"] {
+    analyze.importGraph(Set(reflect.Module(cyclicalA).uri))
+  }
+  ["globs"] {
+    analyze.importGraph(Set(reflect.Module(globImport).uri))
```
holzensp, line 18: Considering that such examples are also often read as HOWTOs, I'd prefer just putting the paths in. This is a little "stilted," if you ask me.

pkl-gradle/src/test/kotlin/org/pkl/gradle/AnalyzeImportsTest.kt
```diff
@@ -0,0 +1,108 @@
...
+        analyze {
+          imports {
+            analyzeMyImports {
+              sourceModules = ["input.pkl"]
+            }
+          }
+        }
+      }
+    """
+        .trimIndent()
```
holzensp, line 42:
```suggestion
      """
        plugins {
          id "org.pkl-lang"
        }

        pkl {
          analyze {
            imports {
              analyzeMyImports {
                sourceModules = ["input.pkl"]
              }
            }
          }
        }
      """
        .trimIndent()
```

pkl-gradle/src/test/kotlin/org/pkl/gradle/AnalyzeImportsTest.kt
```diff
@@ -0,0 +1,108 @@
...
+          imports {
+            analyzeMyImports {
+              sourceModules = ["input.pkl"]
+              outputFile = file("myFile.pcf")
+            }
+          }
+        }
+      }
+    """
+        .trimIndent()
```
holzensp, line 69:
```suggestion
      """
        plugins {
          id "org.pkl-lang"
        }

        pkl {
          analyze {
            imports {
              analyzeMyImports {
                sourceModules = ["input.pkl"]
                outputFile = file("myFile.pcf")
              }
            }
          }
        }
      """
        .trimIndent()
```

pkl-gradle/src/test/kotlin/org/pkl/gradle/AnalyzeImportsTest.kt
```diff
@@ -0,0 +1,108 @@
...
+          imports {
+            analyzeMyImports {
+              sourceModules = ["input.pkl"]
+              outputFormat = "json"
+            }
+          }
+        }
+      }
+    """
+        .trimIndent()
```
holzensp, line 96:
```suggestion
      """
        plugins {
          id "org.pkl-lang"
        }

        pkl {
          analyze {
            imports {
              analyzeMyImports {
                sourceModules = ["input.pkl"]
                outputFormat = "json"
              }
            }
          }
        }
      """
        .trimIndent()
```

pkl-gradle/src/test/kotlin/org/pkl/gradle/AnalyzeImportsTest.kt
```diff
@@ -0,0 +1,108 @@
...
+    )
+    val result = runTask("analyzeMyImports")
+    assertThat(result.output)
+      .contains(
+        """
+      {
+        "imports": {
+    """
+          .trimIndent()
```
holzensp, line 105:
```suggestion
        """
          {
            "imports": {
        """
          .trimIndent()
```

stdlib/analyze.pkl
bioball: I actually read "reflect" as a verb, and `reflect.Class` as an imperative (e.g. "please reflect this class") 🤷

stdlib/analyze.pkl
bioball: No; this does need a fully qualified URI. Will change this to `Set<Uri>`.

pkl-core/src/main/java/org/pkl/core/stdlib/analyze/AnalyzeNodes.java
bioball: Good catch!

pkl-core/src/test/files/LanguageSnippetTests/input/api/analyze1.pkl
bioball: Not totally following this. What do you mean by paths?

stdlib/analyze.pkl
translatenix: I read `function importGraph()` as "import the graph", but `analyze.importGraph()` as "analyze the import graph". I guess that's why this doesn't feel quite right to me.

pkl-core/src/main/java/org/pkl/core/Analyzer.java
```diff
@@ -0,0 +1,121 @@
...
+        () -> {
+          VmContext vmContext = VmContext.get(null);
+          vmContext.initialize(
+              new VmContext.Holder(
+                  transformer,
+                  securityManager,
+                  httpClient,
+                  moduleResolver,
+                  new ResourceManager(securityManager, List.of()),
+                  Loggers.noop(),
```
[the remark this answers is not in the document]
bioball: Fair!

stdlib/analyze.pkl
bioball: I beefed up the docs for this property quite a bit.

stackoverflow (approved):
LGTM. Some nits.

pkl-core/src/test/files/LanguageSnippetTests/input/api/analyze1.pkl
holzensp: Addressed in other comment; I thought `analyze.importGraph("./other.pkl")` was intended legal, but it isn't.

pkl-core/src/main/java/org/pkl/core/stdlib/analyze/AnalyzeNodes.java
```diff
@@ -0,0 +1,98 @@
...
+          throw exceptionBuilder()
+              .evalError("invalidModuleUri", moduleUri)
+              .withHint(e.getMessage())
+              .build();
+        }
+        if (!uri.isAbsolute()) {
+          throw exceptionBuilder().evalError("cannotAnalyzeRelativeModuleUri", moduleUri).build();
+        }
+        uris[idx] = uri;
+        idx++;
```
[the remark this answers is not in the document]
bioball: Definitely a preference thing, but I find your suggestion harder to read.

pkl-core/src/main/java/org/pkl/core/ImportGraph.java
```diff
@@ -0,0 +1,92 @@
...
+import org.pkl.core.util.json.Json.JsArray;
+import org.pkl.core.util.json.Json.JsonParseException;
+import org.pkl.core.util.json.Json.MappingException;
+
+/**
+ * Java representation of {@code pkl.analyze#ImportGraph}.
+ *
+ * @param imports The graph of imports declared within the program.
+ *     <p>Each key is a module inside the program, and each value is the module URIs decalred as
+ *     imports inside that module. The set of all modules initialized within the program is the set
```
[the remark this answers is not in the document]
bioball: Good catch, reworded

PR #696, opened 2024-10-16
Fix handling of `file:` module URIs with non-ASCII characters

### LINKED ISSUE #653: Error occurs when processing files with Japanese names
I was trying to use the `pkl` command to process files with Japanese names. However, an error occurs when attempting to evaluate the file.

#### Information for Bug Report
- **pkl --version**: 
```
```

- **Complete Error Message**:
```
```

#### Steps to Reproduce
1. Check the version of `pkl`:
    ```shell
    $ pkl --version
    ```

2. Create a file `en.pkl` and check the content:
    ```shell
    $ cat en.pkl
    name = ""
    ```

3. Evaluate the file with `pkl`:
    ```shell
    $ pkl eval en.pkl
    name = ""
    ```

4. Copy the file and rename it to a Japanese name:
    ```shell
    $ cp en.pkl 日本語.pkl
    ```

5. Attempt to evaluate the Japanese-named file:
    ```shell
    $ pkl eval 日本語.pkl
    ```

HT154:
Resolves #653


bioball (changes requested):
Thanks for the fix!

We should also fix this in `org.pkl.core.module.FileResolver#listElements(java.net.URI)`, and `org.pkl.core.module.FileResolver#hasElement(java.net.URI)`.

That would handle cases like:

```
read("file:///path/to/日本語.pkl")
```

HT154:
I refactored this into `IoUtils`, adopted it in `FileResolved` and also `StackFrameTransformer`, and added some more test coverage.

I suspect most other usage of `Path.of` in the codebase is at risk of similar issues. Really quite an inconvenient way for this API to work!

pkl-core/src/test/files/LanguageSnippetTests/output/basic/readGlob.pcf
```diff
@@ -109,11 +109,15 @@ examples {
...
     new {
+      ["prop:inputFile"] = "/$snippetsDir/input/basic/readGlob.pkl"
+      ["prop:inputFileDirectory"] = "/$snippetsDir/input/basic/readGlob.pkl"
       ["prop:name1"] = "value1"
       ["prop:name2"] = "value2"
     }
     new {
       ["prop:/foo/bar"] = "foobar"
+      ["prop:inputFile"] = "/$snippetsDir/input/basic/readGlob.pkl"
+      ["prop:inputFileDirectory"] = "/$snippetsDir/input/basic/readGlob.pkl"
```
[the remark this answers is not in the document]
HT154: Yup, just forgot to push the updated snippets. Fixed now!

pkl-cli/src/test/kotlin/org/pkl/cli/CliEvaluatorTest.kt
```diff
@@ -1457,6 +1457,49 @@ result = someLib.x
...
+      readDir = read(".").text
+      readDirFile = read("file:$tempDir/🤬").text
+      readOne = read("日本語.pkl").text.split("\n").first
+      readOneFile = read("file:$tempDir/🤬/日本語.pkl").text.split("\n").first
+      readGlob = read*("./日*.pkl").keys
+      readGlobFile = read*("file:$tempDir/**/*.pkl").keys.map((it) -> it.replaceAll("$tempDir", ""))
+      importOne = import("日本語.pkl").readOne
+      importOneFile = import("file:$tempDir/🤬/日本語.pkl").日本語
+      importGlob = import*("./日*.pkl").keys
+      importGlobFile = import*("file:$tempDir/**/*.pkl").keys.map((it) -> it.replaceAll("$tempDir", ""))
```
bioball, line 1474: This will fail on Windows thanks to different filepath separators.

Better to do:

```
val tempDirUri = tempDir.toUri()
```

And use that in the code snippet instead, e.g.

```
read("$tempDirUri/日本語.pkl")
```

HT154: Ah good catch. I think I've got this sorted, but I have no way to test on windows to be sure.

bioball (approved):
LGTM pending CI checks.

You need to run `./gradlew spotlessApply` to fix some formatting woes.

HT154:
Hmm I think this may be a legitimate difference in how things are handled on different OSs:
```
Changed content at line 6:
expecting:
  ["at 日本語_error#日本語 (file:///$snippetsDir/input/modules/%E6%97%A5%E6%9C%AC%E8%AA%9E_error.pkl)"]
but was:
  ["at 日本語_error#日本語 (file:///$snippetsDir/input/modules/日本語_error.pkl)"]
```

Is there a way to provide a different expectation or exclude this particular test on Windows? Or would it be best to just move this to `CliEvaluatorTest` where the URI diff can be substituted out of existence?

bioball:
Let me look into that one for you; I have access to a Windows machine so it's much easier for me to figure this one out.

pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
```diff
@@ -88,6 +88,20 @@ public static URI toUri(String str) throws URISyntaxException {
     return new URI(null, null, str, null);
   }
 
+  /** Converts a URI to a Path, normalizing any non-ASCII characters. */
+  public static Path pathOfURI(URI uri) {
```
translatenix, line 92:
```suggestion
  public static Path pathOfUri(URI uri) {
```

bioball:
Looks like the issue is that `Path#toUri()` behaves differently on Windows. I added just that one test to the set of excluded tests on Windows.

PR #697, opened 2024-10-16
Fix double unary minus

### LINKED ISSUE #650: Wrong result of unary `-` operation
Pkl will return the wrong result for the unary `-` operation like below.

```pkl
// foo.pkl
a = --1
```

The result of this operation should be `1` but got `-1`.

image

### LINKED ISSUE #653: Error occurs when processing files with Japanese names

HT154:
If `AstBuilder` is processing `UnaryMinus -> [Parenthesized]* -> UnaryMinus -> IntLiteral`, the "inner" `UnaryMinus` helpfully defers inversion to the `IntLiteral` and returns it directly. The "_outer_" `UnaryMinus`, none the wiser, then does the exact same thing. There needs to be a way to force only the inner-most `UnaryMinus` to delegate to the literal node, which seems tricky in context.

This is my naive attempt at that, but this feels messy. More than happy to take suggestions on a better way to do this!

Resolves #650


translatenix:
@HT154 Here is a simpler fix:

```
  @Override
  public ExpressionNode visitUnaryMinusExpr(UnaryMinusExprContext ctx) {
    var childCtx = ctx.expr();
    var childExpr = visitExpr(childCtx);
    if (childCtx instanceof IntLiteralContext || childCtx instanceof FloatLiteralContext) {
      // negation already handled (see visitIntLiteral/visitFloatLiteral)
      return childExpr;
    }
    return UnaryMinusNodeGen.create(createSourceSection(ctx), childExpr);
  }
```

HT154:
@translatenix Slick, thanks!

bioball (approved):
LGTM!

PR #700, opened 2024-10-16
Error when JSON/YAML parsing would set property `default` in a Dynamic

### LINKED ISSUE #561: pkl:json.Parser should throw an error when `useMapping = false` and key "default" is encountered
The behavior as-is is very surprising! Check this out:
```pkl
import "pkl:json"

hidden jsonString = """
  {
    "hello": "world",
    "default": "greeting"
  }
  """

asDynamic = new json.Parser {}.parse(jsonString)
asMapping = new json.Parser { useMapping = true }.parse(jsonString)
```

Result:
```pkl
asDynamic {
  hello = "world"
}
asMapping {
  ["hello"] = "world"
  ["default"] = "greeting"
}
```

I think throwing an error (or at least tracing) here instead of returning unexpected results is desirable. Something like "encountered object key `default` when parsing JSON, resulting `Dynamic` will be invalid" might make sense.

N.B.: Attempting to access the default yields surprising results:
```pkl
asDynamicDefault = asDynamic.default
```
Results in
```
❯ pkl eval test.pkl
–– Pkl Error ––
Expected value of type `Function1`, but got type `String`.
Value: "greeting"

1741 | hidden default: (unknown) -> Any = (_) -> new Dynamic {}
                       ^^^^^^^^^^^^^^^^
at pkl.base#Dynamic.default (https://github.com/apple/pkl/blob/0.26.0/stdlib/base.pkl#L1741)

1 | 
    ^
at generated (source:unavailable)

11 | asDynamicDefault = asDynamic.default
                        ^^^^^^^^^^^^^^^^^
at test#asDynamicDefault (file:///Users/jbasch/src/tsukemono/test.pkl, line 11)

106 | text = renderer.renderDocument(value)
at pkl.base#Module.output.text (https://github.com/apple/pkl/blob/0.26.0/stdlib/base.pkl#L106)
```

HT154:
Resolves #561

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/stdlib/json/ParserNodes.java
+++ b/pkl-core/src/main/java/org/pkl/core/stdlib/json/ParserNodes.java
@@ -172,6 +172,11 @@ public final class ParserNodes {
 
     @Override
     public void startObjectValue(@Nullable EconomicMap<Object, ObjectMember> members, String name) {
+      if (!useMapping && "default".equals(name)) {
+        // https://github.com/apple/pkl/issues/561
+        throw new ParseException(
+            "Cannot parse object key `default` into a `Dynamic` value", getLocation());
+      }
       currPath.push(Identifier.get(name));
     }
 
--- a/pkl-core/src/main/java/org/pkl/core/stdlib/json/ParserNodes.java
+++ b/pkl-core/src/main/java/org/pkl/core/stdlib/json/ParserNodes.java
@@ -27,6 +27,7 @@ import org.pkl.core.runtime.*;
 import org.pkl.core.stdlib.ExternalMethod1Node;
 import org.pkl.core.stdlib.PklConverter;
 import org.pkl.core.util.EconomicMaps;
+import org.pkl.core.util.ErrorMessages;
 import org.pkl.core.util.Nullable;
 import org.pkl.core.util.json.JsonHandler;
 import org.pkl.core.util.json.JsonParser;
@@ -172,12 +173,13 @@ public final class ParserNodes {
 
     @Override
     public void startObjectValue(@Nullable EconomicMap<Object, ObjectMember> members, String name) {
-      if (!useMapping && "default".equals(name)) {
+      var identifier = Identifier.get(name);
+      if (!useMapping && identifier == Identifier.DEFAULT) {
         // https://github.com/apple/pkl/issues/561
         throw new ParseException(
-            "Cannot parse object key `default` into a `Dynamic` value", getLocation());
+            ErrorMessages.create("jsonParseErrorDynamicPropertyDefault"), getLocation());
       }
-      currPath.push(Identifier.get(name));
+      currPath.push(identifier);
     }
 
     @Override
--- a/pkl-core/src/main/java/org/pkl/core/stdlib/yaml/ParserNodes.java
+++ b/pkl-core/src/main/java/org/pkl/core/stdlib/yaml/ParserNodes.java
@@ -483,6 +483,12 @@ public final class ParserNodes {
           var memberName =
               convertedKey instanceof String string && !useMapping ? Identifier.get(string) : null;
 
+          // https://github.com/apple/pkl/issues/561
+          if (memberName != null && "default".equals(convertedKey)) {
+            throw new YamlEngineException(
+                "Cannot parse object key `default` into a Dynamic value. Node: " + node);
+          }
+
           var member =
               new ObjectMember(
                   sourceSection,
--- a/pkl-core/src/main/java/org/pkl/core/stdlib/yaml/ParserNodes.java
+++ b/pkl-core/src/main/java/org/pkl/core/stdlib/yaml/ParserNodes.java
@@ -31,6 +31,8 @@ import org.pkl.core.runtime.*;
 import org.pkl.core.stdlib.ExternalMethod1Node;
 import org.pkl.core.stdlib.PklConverter;
 import org.pkl.core.util.EconomicMaps;
+import org.pkl.core.util.ErrorMessages;
+import org.pkl.core.util.yaml.ParseException;
 import org.pkl.core.util.yaml.snake.YamlUtils;
 import org.snakeyaml.engine.v2.api.ConstructNode;
 import org.snakeyaml.engine.v2.api.Load;
@@ -116,6 +118,8 @@ public final class ParserNodes {
               .evalError("yamlParseErrorTooManyAliases", getMaxCollectionAliases(self))
               .build();
         }
+        throw exceptionBuilder().evalError("yamlParseError").withHint(e.getMessage()).build();
+      } catch (ParseException e) {
         throw exceptionBuilder().evalError("yamlParseError").withHint(e.getMessage()).build();
       }
 
@@ -484,9 +488,10 @@ public final class ParserNodes {
               convertedKey instanceof String string && !useMapping ? Identifier.get(string) : null;
 
           // https://github.com/apple/pkl/issues/561
-          if (memberName != null && "default".equals(convertedKey)) {
-            throw new YamlEngineException(
-                "Cannot parse object key `default` into a Dynamic value. Node: " + node);
+          if (memberName == Identifier.DEFAULT) {
+            throw new ParseException(
+                ErrorMessages.create("yamlParseErrorDynamicPropertyDefault"),
+                keyNode.getStartMark().isEmpty() ? null : keyNode.getStartMark().get());
           }
 
           var member =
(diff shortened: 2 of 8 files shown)


bioball (changes requested):
Thanks for the PR! Just some comments about error messages.

CC @holzensp @stackoverflow; merging this is a breaking change.

pkl-core/src/main/java/org/pkl/core/stdlib/yaml/ParserNodes.java
```diff
@@ -469,6 +469,12 @@ private void addMembers(MappingNode node, VmObject object) {
           var memberName =
               convertedKey instanceof String string && !useMapping ? Identifier.get(string) : null;
 
+          // https://github.com/apple/pkl/issues/561
+          if (memberName != null && "default".equals(convertedKey)) {
+            throw new YamlEngineException(
+                "Cannot parse object key `default` into a Dynamic value. Node: " + node);
```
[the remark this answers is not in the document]
HT154: Added a `ParseException` type that takes and renders SnakeYAML's `Mark`, eg.
```
–– Pkl Error ––
Error parsing YAML document.

3 | x = new yaml.Parser{}.parse("""
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
at test#x (file:///Users/jbasch/src/pkl/test.pkl, line 3)

org.pkl.core.util.yaml.ParseException: Cannot parse object with key `default` into ` Dynamic`.
Try parsing into `Mapping` instead with `useMapping = true` in `pkl.yaml#Parser`. in input_string, line 1, column 1:
    default: null
    ^

106 | text = renderer.renderDocument(value)
at pkl.base#Module.output.text (https://github.com/apple/pkl/blob/109101e/stdlib/base.pkl#L106)
```

It's a little wonky, but serviceable

bioball (approved):
LGTM; pending thumbs from @stackoverflow and @holzensp

pkl-core/src/main/java/org/pkl/core/stdlib/json/ParserNodes.java
```diff
@@ -172,6 +173,11 @@ public void endObject(@Nullable EconomicMap<Object, ObjectMember> members) {
 
     @Override
     public void startObjectValue(@Nullable EconomicMap<Object, ObjectMember> members, String name) {
+      if (!useMapping && "default".equals(name)) {
+        // https://github.com/apple/pkl/issues/561
+        throw new ParseException(
+            ErrorMessages.create("jsonParseErrorDynamicPropertyDefault"), getLocation());
+      }
       currPath.push(Identifier.get(name));
```
translatenix, line 181:
```suggestion
      var identifier = Identifier.get(name);
      if (!useMapping && identifier == Identifier.DEFAULT) {
        // https://github.com/apple/pkl/issues/561
        throw new ParseException(
            "Cannot parse object key `default` into a `Dynamic` value", getLocation());
      }
      currPath.push(identifier);
```

pkl-core/src/main/java/org/pkl/core/stdlib/json/ParserNodes.java
```diff
@@ -172,6 +173,11 @@ public void endObject(@Nullable EconomicMap<Object, ObjectMember> members) {
 
     @Override
     public void startObjectValue(@Nullable EconomicMap<Object, ObjectMember> members, String name) {
+      if (!useMapping && "default".equals(name)) {
+        // https://github.com/apple/pkl/issues/561
+        throw new ParseException(
```
translatenix, line 178: Instead of failing, an entry with string key `"default"` could be added to the dynamic object. Not sure which behavior is more appropriate.

pkl-core/src/main/java/org/pkl/core/stdlib/yaml/ParserNodes.java
```diff
@@ -469,6 +473,13 @@ private void addMembers(MappingNode node, VmObject object) {
           var memberName =
               convertedKey instanceof String string && !useMapping ? Identifier.get(string) : null;
 
+          // https://github.com/apple/pkl/issues/561
+          if (memberName != null && "default".equals(convertedKey)) {
```
translatenix, line 477:
```suggestion
          if (memberName == Identifier.DEFAULT) {
```

pkl-core/src/main/java/org/pkl/core/stdlib/yaml/ParserNodes.java
```diff
@@ -469,6 +473,13 @@ private void addMembers(MappingNode node, VmObject object) {
...
+            throw new ParseException(
```
translatenix, line 478: All other error handling code in this file throws `YamlEngineException`. Why throw a custom exception here and duplicate the code that catches `YamlEngineException` in the same file? (I know this was suggested in the review, but I don't see the benefit of adding all this code.)

pkl-core/src/main/java/org/pkl/core/stdlib/json/ParserNodes.java
HT154: I'm not sure that's any less surprising to users than throwing an exception here, but it at least broadens the set of valid inputs rather than shrinks it. Would love some input from the core team on this.

bioball: We considered introducing this behavior in the past (before we added the `useMapping` option to the parser).

I don't feel strongly about this either way; curious what @holzensp or @stackoverflow think.

pkl-core/src/main/java/org/pkl/core/stdlib/yaml/ParserNodes.java
bioball: `YamlEngineException` doesn't feel like the right class to throw our _own_ errors with tailored messages. I think the direction we should move here is to switch all of our own throws from `YamlEngineException` to `org.pkl.core.util.yaml.ParseException`.

translatenix: Throwing ParseException everywhere is an improvement. But since these exceptions are implementation details and thrown and caught in the same file, introducing your own exception feels unnecessary (adds code for no benefit).

Same applies to json.ParserNodes: It throws org.pkl.core.util.json.ParseException, which also isn't your own exception.

pkl-core/src/main/java/org/pkl/core/stdlib/json/ParserNodes.java
stackoverflow: Throwing an exception is the more principled approach.
Having `default` be an entry feels very hacky. It also has a much worse discoverability: people who don't know this behavior may try calling `obj.default` and get errors or subtle bugs that are hard to debug.
Also the exception even tells you how to fix the problem.

stackoverflow (approved):
LGTM!

pkl-core/src/main/java/org/pkl/core/stdlib/yaml/ParserNodes.java
holzensp: I agree that the extra class seems needlessly verbose and that we should probably move away from `YamlEngineException`s. I don't quite understand `org.pkl.core.util.json.ParseException` not being "your own exception." What do you mean by that, @translatenix?

Is there any reason not to move `org.pkl.core.util.json.ParseException` up one level (into `util`) and use it for either?

holzensp (approved):
Nits, but no strong objections. Approving to unblock; exception discussion is worth seeing through, but we can also follow up.

pkl-core/src/main/java/org/pkl/core/stdlib/yaml/ParserNodes.java
bioball: The idea is that errors coming from `ParseException` can be presented to users as-is; because they are already tailored to be presented as error messages.

`YamlEngineException` messages come from SnakeYAML, and should be decorated with a generic "malformed yaml" message.

I think a good north star here is to not surface messages from `YamlEngineException` at all, and instead present all parse errors as our own tailored error messages. But, for now, it's helpful to have two different exception classes so we know what comes from us, and what comes from our parser library.

translatenix: > I don't quite understand org.pkl.core.util.json.ParseException not being "your own exception." What do you mean by that, @translatenix?

`org.pkl.core.util.json.ParseException` comes from the external JSON parser library that was copied into Pkl's codebase. As such it's the equivalent of `YamlEngineException` .

> YamlEngineException messages come from SnakeYAML, and should be decorated with a generic "malformed yaml" message.

If that's what you want, having your own exception makes sense, assuming you can't throw a `VmException` to begin with. You may also want to make this change for the JSON parser.

PR #701, opened 2024-10-16
Replace some usages of java.lang.String.replaceAll with String.replace

translatenix:
Both methods replace all occurrences.
The difference is that String.replaceAll interprets its first argument as a regex.


bioball (approved):
LGTM

PR #704, opened 2024-10-17
Fix error in spread documentation

djarnis73:


bioball (approved):
Good catch, thank you!

PR #705, opened 2024-10-17
codegen-kotlin: Fix generation of copy() methods

### LINKED ISSUE #569: Koltin generator fails with "'copy' overrides nothing." for subclasses of abstract class
Try generating Kotlin classes for this template:
```
abstract class Day {
  day: Int
  month: Int
  year: Int
}

class Weekend extends Day {}
class Workday extends Day {}

days: Listing<Day>
```

The classes `Weekend` and `Workday` contain copy methods:
```
    override fun copy(
      day: Long,
      month: Long,
      year: Long
    ): Workday = Workday(day, month, year)
```
but the class `Day` which they extend:
```
class Workday(
    day: Long,
    month: Long,
    year: Long
  ) : Day(day, month, year) {
```
does not define `copy` method, which leads to a compilation error:
```
e: file:///home/whoever/whatever/app/build/generated/pkl/template/kotlin/Calendar.kt:52:5 'copy' overrides nothing.
```

It looks like the logic to [skip generation](https://github.com/apple/pkl/blob/5cc2ea2d003a1e2c6e856363684e2aa085ed38a9/pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt#L282) of `copy` methods for abstract classes is flawed.

When making the hierarchy deeper:
```
open class TimeInterval {}

abstract  class Day extends TimeInterval {
...
```
it even fails to build with "Cannot create an instance of an abstract class." (apart from `'copy' overrides nothing.`). This case should also be considered when fixing the implementation of `copy`.

Marking `Day` as `open` instead of `abstract` can be used to work around this issue.

translatenix:
- don't generate copy methods for abstract classes
- precisely determine if copy method needs override modifier (tricky)

Fixes #569

### DIFF
--- a/pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
+++ b/pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
@@ -274,19 +274,12 @@ class KotlinCodeGenerator(
       return methodBuilder.addCode(codeBuilder.build()).build()
     }
 
-    fun nearestNonAbstractAncestorDeclaresCopyMethodWithSameArity(): Boolean {
-      if (properties.isNotEmpty()) {
-        // assertion: pClass declares properties that are generated
-        return false
-      }
-      for (currClass in generateSequence(pClass.superclass) { it.superclass }) {
-        if (!currClass.isAbstract) return true
-        if (currClass.properties.values.any { !it.isHidden }) {
-          // assertion: currClass declares properties that are generated
-          return false
-        }
-      }
-      return false
+    fun inheritsCopyMethodWithSameArity(): Boolean {
+      val nearestNonAbstractAncestor =
+        generateSequence(pClass.superclass) { it.superclass }.firstOrNull { !it.isAbstract }
+          ?: return false
+      return nearestNonAbstractAncestor.allProperties.values.count { !it.isHidden } ==
+        allProperties.size
     }
 
     // besides generating copy method for current class,
@@ -303,8 +296,7 @@ class KotlinCodeGenerator(
 
         // avoid generating multiple methods with same no. of parameters
         if (currParameters.size < prevParameterCount) {
-          val isOverride =
-            currClass !== pClass || nearestNonAbstractAncestorDeclaresCopyMethodWithSameArity()
+          val isOverride = currClass !== pClass || inheritsCopyMethodWithSameArity()
           typeBuilder.addFunction(generateCopyMethod(currParameters, isOverride))
           prevParameterCount = currParameters.size
         }
--- a/pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
+++ b/pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
@@ -274,9 +274,27 @@ class KotlinCodeGenerator(
       return methodBuilder.addCode(codeBuilder.build()).build()
     }
 
+    fun nearestNonAbstractAncestorDeclaresCopyMethodWithSameArity(): Boolean {
+      if (properties.isNotEmpty()) {
+        // assertion: pClass declares properties that are generated
+        return false
+      }
+      for (currClass in generateSequence(pClass.superclass) { it.superclass }) {
+        if (!currClass.isAbstract) return true
+        if (currClass.properties.values.any { !it.isHidden }) {
+          // assertion: currClass declares properties that are generated
+          return false
+        }
+      }
+      return false
+    }
+
     // besides generating copy method for current class,
     // override copy methods inherited from parent classes
     fun generateCopyMethods(typeBuilder: TypeSpec.Builder) {
+      // copy methods don't make sense for abstract classes
+      if (pClass.isAbstract) return
+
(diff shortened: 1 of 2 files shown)


pkl-codegen-kotlin/src/test/kotlin/org/pkl/codegen/kotlin/KotlinCodeGeneratorTest.kt
```diff
@@ -661,6 +661,184 @@ class KotlinCodeGeneratorTest {
     assertCompilesSuccessfully(kotlinCode)
   }
 
+  // https://github.com/apple/pkl/issues/569
+  @Test
+  fun abstractClasses() {
```
bioball, line 666:
```suggestion
  fun `abstract classes`() {
```

pkl-codegen-kotlin/src/test/kotlin/org/pkl/codegen/kotlin/KotlinCodeGeneratorTest.kt
```diff
@@ -661,6 +661,184 @@ class KotlinCodeGeneratorTest {
...
+        .trimMargin(),
+      kotlinCode
+    )
+
+    assertCompilesSuccessfully(kotlinCode)
+  }
+
+  // https://github.com/apple/pkl/issues/569
+  @Test
+  fun abstractClassThatExtendsOpenClass() {
```
bioball, line 727:
```suggestion
  fun `abstract class that extends open class`() {
```

pkl-codegen-kotlin/src/test/kotlin/org/pkl/codegen/kotlin/KotlinCodeGeneratorTest.kt
```diff
@@ -661,6 +661,184 @@ class KotlinCodeGeneratorTest {
...
+        .trimMargin(),
+      kotlinCode
+    )
+
+    assertCompilesSuccessfully(kotlinCode)
+  }
+
+  // https://github.com/apple/pkl/issues/569
+  @Test
+  fun abstractClassThatExtendsOpenClassWithoutAddingProperties() {
```
bioball, line 791:
```suggestion
  fun `abstract class that extends open class without adding properties`() {
```

pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
```diff
@@ -274,9 +274,27 @@ class KotlinCodeGenerator(
...
+    fun nearestNonAbstractAncestorDeclaresCopyMethodWithSameArity(): Boolean {
+      if (properties.isNotEmpty()) {
+        // assertion: pClass declares properties that are generated
+        return false
+      }
+      for (currClass in generateSequence(pClass.superclass) { it.superclass }) {
+        if (!currClass.isAbstract) return true
+        if (currClass.properties.values.any { !it.isHidden }) {
+          // assertion: currClass declares properties that are generated
+          return false
```
bioball, line 286: Does this ever get checked? I put a breakpoint here and ran the entire test suite, and never hit this line. Plus, with your new change, we now don't generate copy methods for abstract classes, right?

pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
```diff
@@ -285,7 +303,8 @@ class KotlinCodeGenerator(
 
         // avoid generating multiple methods with same no. of parameters
         if (currParameters.size < prevParameterCount) {
-          val isOverride = currClass !== pClass || superclass != null && properties.isEmpty()
+          val isOverride =
+            currClass !== pClass || nearestNonAbstractAncestorDeclaresCopyMethodWithSameArity()
```
bioball, line 307: [nit] Either make `nearestNonAbstractAncestorDeclaresCopyMethodWithSameArity` a lazy value, or cache this so we don't call it on every iteration

bioball (changes requested):
Just a few nits and comments, but otherwise looks good!

pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
translatenix: Good catch. This implementation is neither obvious nor correct, so I rewrote it.

pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
translatenix: It's only called once due to `currClass !== pClass ||`.

translatenix:
@bioball ready from my side

pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
bioball: Ah, yeah, that was a little confusing, but you're right

bioball (approved):
LGTM, thank you for the fix!

PR #708, opened 2024-10-18
Store `PklProject` annotations in the project metadata

stackoverflow:
Store `PklProject` annotations in the project metadata and use these annotations to inform doc generation.


bioball (changes requested):
Good start!

pkl-core/src/main/java/org/pkl/core/packages/DependencyMetadata.java
```diff
@@ -423,8 +512,84 @@ private void write() throws IOException {
...
+            throw new PklException(
+                "Error serializing annotation for PklProject:\n"
+                    + "  cannot render map with non-string key: "
+                    + key);
+          }
+          writeGenericObject(kv.getValue());
+        }
+        jsonWriter.endObject();
+      } else if (value instanceof Pattern pattern) {
+        jsonWriter.value(pattern.pattern());
```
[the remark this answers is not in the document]
stackoverflow: Added support for the missing cases and disambiguated `Pattern`, `Set`, `Map`, etc.
I still don't support `PClass` and `TypeAlias` because these types are too complex (sometimes recursive) and I think it's fine to not support them for metadata.
Also added a test to roudtrip all new types.

pkl-core/src/main/java/org/pkl/core/packages/DependencyMetadata.java
```diff
@@ -423,8 +584,129 @@ private void write() throws IOException {
...
+        for (var kv : map.entrySet()) {
+          var key = kv.getKey();
+          if (key instanceof String s) {
+            jsonWriter.name(s);
+          } else {
+            throw new PklException(
+                "Error serializing annotation for PklProject:\n"
+                    + "  cannot render map with non-string key: "
+                    + key);
+          }
```
bioball, line 673: How about we serialize maps like so?

```
[
  {
    "key": ...
    "value": ...
  }
]
```

bioball (changes requested):
Just one comment, but otherwise I think this is ready!

pkl-core/src/main/java/org/pkl/core/packages/DependencyMetadata.java
stackoverflow: Good shout. So we can support non-string keys.

PR #710, opened 2024-10-18
codegen-java: Change condition for generating equals/hashCode/toString/with/Serializable

### LINKED ISSUE #706: codegen-java: Don't generate equals/hashCode/toString/with/Serializable for abstract classes
Is there a good reason for generating these methods? As far as I can tell, they are never called from generated subclasses.

translatenix:
Motivation:
* Currently, the condition for generating `equals/hashCode/toString` methods is "generated class declares properties". As a result, classes that don't declare properties inherit these methods from their generated superclasses. However, generated `equals` and `toString` methods aren't designed to be inherited and will either fail or produce wrong results when called for a subclass.
* Currently, the condition for generating `with` methods is "class is not abstract". However, it isn't useful to generate `with` methods for non-instantiable non-abstract classes.
* Currently, the condition for making classes serializable is "class is not a module class". However, it isn't useful to make non-instantiable non-module classes serializable, and it is useful to make instantiable module classes serializable.

Changes:
Change the condition for generating `equals/hashCode/toString/with/Serializable` to "class is instantiable". This is a breaking change.
(A generated class is instantiable, i.e., declares a public constructor, if it is neither abstract nor stateless. This behavior remains unchanged.)

Result:
* Fixes all motivating issues.
* Fixes #706.

### DIFF
--- a/pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/JavaCodeGenerator.kt
+++ b/pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/JavaCodeGenerator.kt
@@ -237,15 +237,14 @@ class JavaCodeGenerator(
       )
     }
 
-    fun generateConstructor(): MethodSpec {
+    fun generateConstructor(isInstantiable: Boolean): MethodSpec {
       val builder =
         MethodSpec.constructorBuilder()
           // choose most restrictive access modifier possible
           .addModifiers(
             when {
-              pClass.isAbstract -> Modifier.PROTECTED
-              allProperties.isNotEmpty() -> Modifier.PUBLIC // if `false`, has no state
-              pClass.isOpen -> Modifier.PROTECTED
+              isInstantiable -> Modifier.PUBLIC
+              pClass.isAbstract || pClass.isOpen -> Modifier.PROTECTED
               else -> Modifier.PRIVATE
             }
           )
@@ -542,7 +541,10 @@ class JavaCodeGenerator(
       val builder =
         TypeSpec.classBuilder(javaPoetClassName.simpleName()).addModifiers(Modifier.PUBLIC)
 
-      if (codegenOptions.implementSerializable && !isModuleClass) {
+      // stateless classes aren't instantiable by choice, i.e., no public ctor is generated
+      val isInstantiable = !pClass.isAbstract && allProperties.isNotEmpty()
+
+      if (codegenOptions.implementSerializable && isInstantiable) {
         builder.addSuperinterface(java.io.Serializable::class.java)
         builder.addField(generateSerialVersionUIDField())
       }
@@ -574,7 +576,7 @@ class JavaCodeGenerator(
         generateSpringBootAnnotations(builder)
       }
 
-      builder.addMethod(generateConstructor())
+      builder.addMethod(generateConstructor(isInstantiable))
 
       superclass?.let { builder.superclass(it.toJavaPoetName()) }
 
@@ -590,12 +592,12 @@ class JavaCodeGenerator(
             builder.addMethod(generateGetter(name, property, isOverridden))
           }
         }
-        if (!pClass.isAbstract) {
+        if (isInstantiable) {
           builder.addMethod(generateWithMethod(name, property))
         }
       }
 
-      if (properties.isNotEmpty()) {
+      if (isInstantiable) {
         builder
           .addMethod(generateEqualsMethod())
           .addMethod(generateHashCodeMethod())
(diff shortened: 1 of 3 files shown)


pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/JavaCodeGenerator.kt
```diff
@@ -542,7 +541,10 @@ class JavaCodeGenerator(
       val builder =
         TypeSpec.classBuilder(javaPoetClassName.simpleName()).addModifiers(Modifier.PUBLIC)
 
-      if (codegenOptions.implementSerializable && !isModuleClass) {
+      // stateless classes aren't instantiable by choice, i.e., no public ctor is generated
```
holzensp, line 544: Comment refers to only one of the applications of `isInstantiable`, located elsewhere. That makes it likely to go out-of-sync. Better to omit.
```suggestion
```

pkl-codegen-java/src/test/kotlin/org/pkl/codegen/java/JavaCodeGeneratorTest.kt
```diff
@@ -263,8 +263,8 @@ class JavaCodeGeneratorTest {
         |      return new ClassWithDeprecatedProperty(deprecatedProperty);
         |    }
       """
-        .trimMargin()
-    assertThat(javaCode).contains(expectedPropertyDef).contains(expectedWithMethodDef)
+    assertContains(expectedPropertyDef, javaCode)
+    assertContains(expectedWithMethodDef, javaCode)
```
holzensp, line 267: I actually don't like our `assertFoo` methods much; it reads entirely different than the (quite nice) `assertj` style. Not blocking (happy to follow up myself), but might I suggest replacing `assertContains` with
```
  private fun <T : AbstractStringAssert<T>> AbstractStringAssert<T>.containsTrimmed(part: String): AbstractStringAssert<T> {
    val trimmed = part.trim().trimMargin()
    if (!part.contains(trimmed)) {
      isEqualTo(trimmed)
    }
    return this
  }
```

and then
```suggestion
    assertThat(javaCode)
      .containsTrimmed(expectedPropertyDef)
      .containsTrimmed(expectedWithMethodDef)
```

[throughout]

pkl-codegen-java/src/test/kotlin/org/pkl/codegen/java/JavaCodeGeneratorTest.kt
```diff
@@ -336,77 +331,72 @@ class JavaCodeGeneratorTest {
...
-        else
-          """
-          |/**
-          | * @deprecated module class deprecation message
-          | */
+      @Deprecated{ message = "module class deprecation message" }
+      module DeprecatedModule
+      
+      propertyInDeprecatedModuleClass : Int = 42
+    """,
```
holzensp, line 347: Oh Spotless, wherefore art though so spotty :(
```suggestion
          @Deprecated{ message = "module class deprecation message" }
          module DeprecatedModule

          propertyInDeprecatedModuleClass : Int = 42
        """,
```

[throughout]

pkl-codegen-java/src/test/kotlin/org/pkl/codegen/java/JavaCodeGeneratorTest.kt
```diff
@@ -336,77 +331,72 @@ class JavaCodeGeneratorTest {
...
-    for (generateJavadoc in listOf(false, true)) {
-      val javaCode =
-        generateJavaCode(
-          """
-          @Deprecated{ message = "module class deprecation message" }
-          module DeprecatedModule
-          
-          propertyInDeprecatedModuleClass : Int = 42
+  @ParameterizedTest
+  @ValueSource(booleans = [false, true])
```
holzensp, line 338: Much nicer

holzensp (approved):
LGTM and thanks for the additional polish of `JavaCodeGeneratorTest`. Couple of suggestions about that, but non-blocking.

Looking for at least one other 👍 (cc @stackoverflow, @bioball)

pkl-codegen-java/src/test/kotlin/org/pkl/codegen/java/JavaCodeGeneratorTest.kt
translatenix: working on this (will take some time)

pkl-codegen-java/src/test/kotlin/org/pkl/codegen/java/JavaCodeGeneratorTest.kt
translatenix: In my latest commit, I formatted the code exactly as it falls out of the IntelliJ editor.
ktfmt successfully destroyed this formatting, and ktfmt is the boss.

translatenix:
Ready to squash and merge from my side.

bioball (approved):
LGTM pending CI checks

translatenix:
rebased and squashed, ready from my side

PR #716, opened 2024-10-20
Produce more helpful output when module output is overwritten

### LINKED ISSUE #709: Fatal error on mistyping `output`
I got a bit confused and tried to re-type what turned out to be the special `output` property. This seems to crash the CLI.

Given **whoops.pkl**
```pkl
output: String = "hi"
```

Upon evaluation, it asks that I file this bug report.
```shell
% pkl eval whoops.pkl 
% echo $?
1
```

```shell
% pkl --version
```

HT154:
Evaluating this module:
```pkl
output: String = "abc"
```

Now produces
```
–– Pkl Error ––
Expected `output` of module `file:///$snippetsDir/input/modules/invalidOutput1.pkl` to be of type `ModuleOutput`, but got type `String`.

x | output: String = "abc"
                     ^^^^^
at output (file:///$snippetsDir/input/modules/invalidOutput1.pkl)
```

Resolves #709


pkl-core/src/main/java/org/pkl/core/EvaluatorImpl.java
```diff
@@ -157,6 +157,19 @@ public Object evaluateOutputValue(ModuleSource moduleSource) {
...
 
+  private VmTyped readModuleOutput(VmTyped module) {
+    var value = VmUtils.readMember(module, Identifier.OUTPUT);
+    if (value instanceof VmTyped typedOutput) {
+      if (typedOutput.getVmClass().getPClassInfo() != PClassInfo.ModuleOutput) {
+        throw moduleOutputValueTypeMismatch(
+            module, PClassInfo.ModuleOutput, value, module, Identifier.OUTPUT, "output");
+      }
+      return typedOutput;
+    }
```
translatenix, line 168:
```suggestion
    if (value instanceof VmTyped typedOutput && typedOutput.getVmClass().getPClassInfo() == PClassInfo.ModuleOutput) {
      return typedOutput;
    }
```

pkl-core/src/test/files/LanguageSnippetTests/input/modules/invalidOutput1.pkl
```diff
@@ -0,0 +1 @@
+output: String = "abc"
```
translatenix, line 1: ideally this problem would be caught right here (override with different type), but I guess that's impossible for now

pkl-core/src/main/java/org/pkl/core/EvaluatorImpl.java
```diff
@@ -157,6 +157,15 @@ public Object evaluateOutputValue(ModuleSource moduleSource) {
         });
   }
 
+  private VmTyped readModuleOutput(VmTyped module) {
+    var value = VmUtils.readMember(module, Identifier.OUTPUT);
+    if (value instanceof VmTyped typedOutput && typedOutput.getVmClass().getPClassInfo() == PClassInfo.ModuleOutput) {
+      return typedOutput;
+    }
```
translatenix, line 164: Sorry, my suggestion needs reformatting:

```suggestion
    if (value instanceof VmTyped typedOutput
        && typedOutput.getVmClass().getPClassInfo() == PClassInfo.ModuleOutput) {
      return typedOutput;
    }
```

pkl-core/src/test/files/LanguageSnippetTests/input/modules/invalidOutput2.pkl
```diff
@@ -0,0 +1,3 @@
+class Test {}
+
+output: Test = new {}
```
bioball, line 3: Move this and `invliadOutput1` to `input/errors`.

I know we have some other error tests in `input/modules`, but those should probably be moved too. The intention is to put all testing around error messages in the `errors/` dir.

pkl-core/src/test/files/LanguageSnippetTests/input/modules/invalidOutput1.pkl
bioball: We'd like to make Pkl stricter here and enforce that subclasses are also subtypes. Maybe something to do before we cut a 1.0 release.

pkl-core/src/main/java/org/pkl/core/EvaluatorImpl.java
```diff
@@ -366,24 +377,30 @@ private void handleTimeout(@Nullable TimeoutTask timeoutTask) {
   }
 
   private VmException moduleOutputValueTypeMismatch(
-      VmTyped module, PClassInfo<?> expectedClassInfo, Object value, VmTyped output) {
+      VmTyped module,
+      PClassInfo<?> expectedClassInfo,
+      Object value,
+      VmTyped container,
+      Identifier propertyIdentifier,
+      String propertyName) {
```
bioball, line 385: There's a fair bit of logic here that's bespoke to reading `output.value`, so it's not too appropriate to be re-purposed. For example, it show different error messages based on whether `output.value` was explicitly assigned or not.

I think we should just copy some of the logic here that's useful, and inline that into `readModuleOutput`.

bioball (changes requested):
Thanks for the PR! Generally, this is the right approach. See comments for some minor issues here.

pkl-core/src/main/java/org/pkl/core/EvaluatorImpl.java
bioball: The amend hint is for things like: your module does not amend anything, and maybe it should have `amends "pkl:Project"`.

In this case, `output` should always be of type `ModuleOutput`, so that type of suggestion doesn't really make much sense.

Showing the source section of the assignment makes sense; I think you can either copy/paste it, or maybe refactor this into a separate method?

holzensp (approved):
LGTM. Needs a rebase & `spotlessApply`, because of the license header change.

bioball (approved):
LGTM, thanks for the fix!

PR #719, opened 2024-10-21
Add version of resolved Pkl distribution to `ExecutorException`

holzensp:
Only when catching an exception from evaluation can we decorate the `ExecutorException` with Pkl version information (all other occurrences are pre-resolution failures or failure-to-resolve). That's why this PR adds an extra constructor.

### DIFF
--- a/pkl-executor/src/main/java/org/pkl/executor/ExecutorException.java
+++ b/pkl-executor/src/main/java/org/pkl/executor/ExecutorException.java
@@ -19,11 +19,22 @@ package org.pkl.executor;
  * Indicates an {@link Executor} error. {@link #getMessage()} returns a user-facing error message.
  */
 public final class ExecutorException extends RuntimeException {
+  String pklVersion = "unresolved";
+
   public ExecutorException(String message) {
     super(message);
   }
 
   public ExecutorException(String message, Throwable cause) {
     super(message, cause);
+  }
+
+  public ExecutorException(String message, String version, Throwable cause) {
+    super(message, cause);
+    pklVersion = version;
+  }
+
+  public String getPklVersion() {
+    return pklVersion;
   }
 }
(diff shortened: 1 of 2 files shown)


pkl-executor/src/main/java/org/pkl/executor/ExecutorException.java
```diff
@@ -19,11 +19,22 @@
...
   }
+
+  public ExecutorException(String message, String version, Throwable cause) {
+    super(message, cause);
+    pklVersion = version;
+  }
+
+  public String getPklVersion() {
+    return pklVersion;
+  }
```
bioball, line 39: Some nits:

* We can probably just return `null` if there is no pkl version to relate.
* Add a doc comment to `getPklVersion()`
* Would be nice to add the Pkl version to `getMessage()`

```suggestion
  String pklVersion = null;

  public ExecutorException(String message) {
    super(message);
  }

  public ExecutorException(String message, Throwable cause) {
    super(message, cause);
  }

  public ExecutorException(String message, String version, Throwable cause) {
    super(message, cause);
    pklVersion = version;
  }

  /**
   * The selected Pkl version used to evaluate the module.
   *
   * Returns {@code null} if this exception does not originate from an underlying Pkl evaluator.
   */
  public String getPklVersion() {
    return pklVersion;
  }

  @Override
  public String getMessage() {
    var message = super.getMessage();
    if (pklVersion == null) {
      return message;
    }
    return message + "\nPkl version: " + pklVersion;
  }
```

bioball (approved):
No blockers, but have some suggestions!

pkl-executor/src/main/java/org/pkl/executor/ExecutorException.java
translatenix: * Should be `private final String pklVersion`.
* New constructor should append parameter: `public ExecutorException(String message, Throwable cause, String version)`

PR #721, opened 2024-10-21
codegen-kotlin: Support Java serialization of module classes

translatenix:
Motivation:
- Java serialization makes as much sense for module classes as it does for regular classes.
- Offer same Java serialization support as codegen-java.

Changes:
- Move generation of `appendProperty` method and serialization code
  into new method `generateCompanionRelatedCode`.
- Improve/fix generation of serialization code.

Result:
Java serialization is also supported for module classes.


bioball (approved):
LGTM, good fix. Thank you!

PR #722, opened 2024-10-21
Fix build performance issues of spotless tasks

translatenix:
Motivation:
Spotless tasks take minutes instead of seconds to complete. This greatly slows down `gw spotlessApply`, `gw check`, and `gw build`.

Changes:
- Use more efficient glob expressions for inclusion filters (`target()`).
- Avoid the need for exclusion filters (`targetExclude()`).

Result:
Spotless tasks complete in seconds instead of minutes. As a concrete example, repeated invocation of `gw spotlessApply` completes in 2 instead of 122 seconds on my laptop.


bioball:
Ahhh Windows. The spotless tasks are _quite_ snappy on macOS.

bioball (approved):
LGTM

PR #725, opened 2024-10-23
Fix typecheck error on Listing/Mapping

bioball:
The addresses the following case:

```
local l1: Listing<String> = new { 1 }
l2: Listing<Int> = (l1) { 2 }
```

The member in `l1` should be checked against its owner; to check against `Listing<String>`.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
@@ -298,6 +298,9 @@ public final class VmUtils {
           }
         }
       } else if (receiver instanceof VmListingOrMapping<?> vmListingOrMapping) {
+        if (owner != receiver && owner instanceof VmListingOrMapping<?> vmListingOrMappingOwner) {
+          ret = vmListingOrMappingOwner.typecastObjectMember(member, ret, callNode);
+        }
         ret = vmListingOrMapping.typecastObjectMember(member, ret, callNode);
       }
       receiver.setCachedValue(memberKey, ret, member);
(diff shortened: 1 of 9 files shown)


pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
```diff
@@ -298,6 +298,9 @@ public static Object doReadMember(
           }
         }
       } else if (receiver instanceof VmListingOrMapping<?> vmListingOrMapping) {
+        if (owner != receiver && owner instanceof VmListingOrMapping<?> vmListingOrMappingOwner) {
+          ret = vmListingOrMappingOwner.typecastObjectMember(member, ret, callNode);
```
stackoverflow, line 302: Why the `ret =` if in the next line you reassign to `ret`?

stackoverflow (approved):
Looks good.

pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
```diff
@@ -298,6 +298,9 @@ public static Object doReadMember(
...
+        }
         ret = vmListingOrMapping.typecastObjectMember(member, ret, callNode);
       }
```
translatenix, line 305: If the following suggestion doesn't make sense, a test is missing. :-)

```suggestion
      } else if (owner instanceof VmListingOrMapping<?> vmListingOrMappingOwner) {
        ret = vmListingOrMappingOwner.typecastObjectMember(member, ret, callNode);
      }
```

pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
holzensp: `ret` is also an argument to the function call in the line below; it doubles the cast. That said; @translatenix's suggestion should be sufficient.

holzensp (approved):
Approving to unblock; but take the suggestion or add the test ;)

pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
translatenix: Tests pass even after removing the first `ret =`.

PR #734, opened 2024-10-24
codegen-java: Make stateless classes instantiable

translatenix:
Last codegen PR for now. :-)

Motivation:
Currently, the condition for making a generated Java class instantiable is "class is neither abstract nor stateless".
(An instantiable class receives a public constructor and equals/hashCode/toString/with methods.)
This is inconsistent with Pkl and codegen-kotlin, both of which support instantiation of stateless classes.

Changes:
Widen condition for making a class instantiable to "class is neither abstract nor a stateless final module class".
This is consistent with codegen-kotlin, which generates object declarations (only) for stateless final module classes.

Result:
Stateless classes are instantiable in Pkl, codegen-java, and codegen-kotlin.

### DIFF
--- a/pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/JavaCodeGenerator.kt
+++ b/pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/JavaCodeGenerator.kt
@@ -541,8 +541,9 @@ class JavaCodeGenerator(
       val builder =
         TypeSpec.classBuilder(javaPoetClassName.simpleName()).addModifiers(Modifier.PUBLIC)
 
-      // stateless classes aren't instantiable by choice, i.e., no public ctor is generated
-      val isInstantiable = !pClass.isAbstract && allProperties.isNotEmpty()
+      // stateless final module classes are non-instantiable by choice
+      val isInstantiable =
+        !(pClass.isAbstract || (isModuleClass && !pClass.isOpen && allProperties.isEmpty()))
 
       if (codegenOptions.implementSerializable && isInstantiable) {
         builder.addSuperinterface(java.io.Serializable::class.java)
--- a/pkl-codegen-java/src/test/kotlin/org/pkl/codegen/java/JavaCodeGeneratorTest.kt
+++ b/pkl-codegen-java/src/test/kotlin/org/pkl/codegen/java/JavaCodeGeneratorTest.kt
@@ -776,6 +776,83 @@ class JavaCodeGeneratorTest {
   }
 
   @Test
+  fun `stateless classes`() {
+    val javaCode =
+      generateJavaCode(
+        """
+      module my.mod
+
+      class Foo
+      abstract class Bar
+      class Baz extends Bar
+    """
+          .trimIndent()
+      )
+
+    assertThat(javaCode)
+      .contains(
+        """
+        |  public static final class Foo {
+        |    public Foo() {
+        |    }
+      """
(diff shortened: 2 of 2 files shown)


pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/JavaCodeGenerator.kt
```diff
@@ -541,8 +541,8 @@ class JavaCodeGenerator(
       val builder =
         TypeSpec.classBuilder(javaPoetClassName.simpleName()).addModifiers(Modifier.PUBLIC)
 
-      // stateless classes aren't instantiable by choice, i.e., no public ctor is generated
-      val isInstantiable = !pClass.isAbstract && allProperties.isNotEmpty()
+      // stateless module classes are non-instantiable by choice
+      val isInstantiable = !pClass.isAbstract && !(isModuleClass && allProperties.isEmpty())
```
bioball, line 545: We should let empty modules to be instantiable too, unless I'm missing something obvious here?

```suggestion
      val isInstantiable = !pClass.isAbstract;
```

pkl-codegen-java/src/test/kotlin/org/pkl/codegen/java/JavaCodeGeneratorTest.kt
```diff
@@ -775,6 +775,55 @@ class JavaCodeGeneratorTest {
       .isEqualToResourceFile("Inheritance.jva")
   }
 
+  @Test
+  fun `stateless classes`() {
```
bioball, line 779: [nit] "empty" feels like a better word for this

```suggestion
  fun `empty classes`() {
```

pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/JavaCodeGenerator.kt
translatenix: I just narrowed the exception from "stateless module classes" to "stateless final module classes".
This is consistent with codegen-kotlin, which generates object declarations in such cases.
I guess this is trying to generate nicer code for the "I'm a pure schema" case.

bioball: I don't know if modules are any different here.

This module should be able to be mapped to Java, for example:

```pkl
module com.example.Foo
// I'm totally empty!
```

translatenix: With this PR, both code generators should behave exactly the same.
A future PR could remove the special handling of stateless final module classes.
This will generate some top-level code (equals/hashCode/toString/with) that sits next to (possibly above) the class declarations.

pkl-codegen-java/src/test/kotlin/org/pkl/codegen/java/JavaCodeGeneratorTest.kt
bioball: Ah, gotcha

pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/JavaCodeGenerator.kt
bioball: Okay, fair enough!

bioball (approved):
LGTM

PR #735, opened 2024-10-24
Fix test failures on main

HT154:
Just rebased some PRs on main and they're all failing in the pkl-gradle tests. This PR fixes the expectations, which changed with what I think is #695


holzensp (approved):

stackoverflow (approved):
Thanks for the fix!

PR #498, opened 2024-05-16
Improve testing with stats and errors per test section

jjmaestro:
Add some basic test stats to the test runner output for both SimpleReport and JUnitReport:

- number of tests that pass/fail
- % of success
- number of assertions within each test that pass / fail

PR #738, opened 2024-10-24
Polish test result running and reporting

bioball:
Changes:
* Move class `TestResults` to package `org.pkl.core`, because it is a public class (it's the result of `Evaluator#evaluateTest`)
* Change examples to treat individual examples as assertions in the same test. Previously, they were considered different tests with an incrementing number. This better aligns with how facts are treated.
* Change `TestResults` to be a record, and introduce builders.
* Remove "module" test result section (it is not really a section).
* Add javadoc to `TestResults`
* Formatting fix: prefix ✍️ emoji just like we do the ❌ and ✅ emojis 
* Consider writing examples as failures, not successes. `pkl test` will exit with code 2 if the only failing tests are due to writing examples. Primary motivation: not checking in `pkl-expected.pcf` files should not be a passing test in CI. The expected output file is part of the specification of the test, and users are meant to review them and check them into source control.


pkl-gradle/src/test/kotlin/org/pkl/gradle/TestsTest.kt
```diff
@@ -112,34 +112,33 @@ class TestsTest : AbstractTest() {
...
+               Expected: (file:///file, line x)
+               new {
+                 name = "Parrot"
+                 age = 35
+               }
+               Actual: (file:///file, line x)
+               new {
+                 name = "Welma"
+                 age = 35
+               }
```
jjmaestro, line 139: IMHO this is a "regression", in  #498, I added the "breaking out examples" to make testing examples (e.g. large example files) easier. An example can have "different facts", each object / part of the example becomes a fact and when it fails, you can quickly see that it's the Nth part of the example file.

This PR goes back to making "examples opaque", a failure in an examples file doesn't let you know how many other parts pass and doesn't give you info of which part of the example breaks 😐

bioball: Gotcha, I think it makes sense that you might want some details about _which_ example failed.

How about an output that looks like this? This would still be considered an assertion inside the test, but this makes it clearer _which_ assertion failed.

```
examples
  ✅ user 0
  ❌ user 1
    #1:
      (file:///file, line x)
      Expected: (file:///file, line x)
      new {
        name = "Parrot"
        age = 35
      }
      Actual: (file:///file, line x)
      new {
        name = "Welma"
        age = 35
      }
    #2:
      (file:///file, line x)
      Expected: (file:///file, line x)
      5
      Actual: (file:///file, line x)
      10
```

jjmaestro: Awesome, IMHO that's actually much better than my output, as it's closer to the `facts` output: it's green if all the assertions/examples pass and red when at least one fails, and then it shows the errors of the assertions / "example cases" that failed 👌 

Also, re. whitespace and alignment, one minor nit is I'd make sure everything is tabulated to the `user`. The output you pasted was "one-space-less" than what I'd use 😄.

Here's what I'd prefer :
```
examples
  ✅ user 0
  ❌ user 1
     #1:
       (file:///file, line x)
       Expected: (file:///file, line x)
       new {
         name = "Parrot"
         age = 35
       }
       Actual: (file:///file, line x)
       new {
         name = "Welma"
         age = 35
       }
     #2:
       (file:///file, line x)
       Expected: (file:///file, line x)
       5
       Actual: (file:///file, line x)
       10
```

Finally, and this is something that I didn't do in my PR but maybe you can consider doing, is to remove one of the `file:///file, line x` from the example's error output, because I think it's duplicated in `Actual`. That way you save one line per error as well 😅 :
```
examples
  ✅ user 0
  ❌ user 1
     #1:
       Expected: (file:///file, line x)
       new {
         name = "Parrot"
         age = 35
       }
       Actual: (file:///file, line x)
       new {
         name = "Welma"
         age = 35
       }
     #2:
       Expected: (file:///file, line x)
       5
       Actual: (file:///file, line x)
       10
```
 
 And If you decide not to remove it, you could also put it next to the `#<ID>`:
 ```
 examples
  ✅ user 0
  ❌ user 1
     #1: (file:///file, line x)
       Expected: (file:///file, line x)
       new {
         name = "Parrot"
         age = 35
       }
       Actual: (file:///file, line x)
       new {
         name = "Welma"
         age = 35
       }
     #2: (file:///file, line x)
       Expected: (file:///file, line x)
       5
       Actual: (file:///file, line x)
       10
```

Comment (bioball) [reply pkl-gradle/src/test/kotlin/org/pkl/gradle/TestsTest.kt:139]: It's not duplicated, actually! There's three source sections to show:

1. Within the test
2. Within `pkl-expected.pcf`
3. Within `pkl-actual.pcf`

But good idea re: putting it next to the counter.

Comment (jjmaestro) [reply pkl-gradle/src/test/kotlin/org/pkl/gradle/TestsTest.kt:139]: Aaaah true, true, you are probably right, I was just thinking it would duplicate it in the case where examples are defined in-file, in which case I hought `Actual` would match with the `#<nth>` example.

File: pkl-cli/src/test/kotlin/org/pkl/cli/CliTestRunnerTest.kt [hunk 2]
```diff
@@ -429,18 +429,55 @@ class CliTestRunnerTest {
...
-      module test
-        examples
-          ❌ nums
-            
-             Output mismatch: Expected "nums" to contain 1 examples, but found 2
-      ❌ 0.0% tests pass [1/1 failed], 0.0% asserts pass [1/1 failed]
+        module test
+          examples
+            ❌ nums
+
```
jjmaestro, line 435: nit: maybe the extra whitespace / line could be removed too? :-? Or maybe it's an artifact of the tests, when removing the location string? I'm trying to find the exact place where the additional `\n` is added but I see a `location` that's not showing in here :-?

pkl-core/src/test/kotlin/org/pkl/core/EvaluateTestsTest.kt
```diff
@@ -344,21 +344,21 @@ class EvaluateTestsTest {
...
+          Expected: (/tempDir/example.pkl-expected.pcf)
+          new {
+            name = "Alice"
+            age = 45
+          }
+          Actual: (/tempDir/example.pkl-actual.pcf)
+          new {
+            name = "Bob"
+            age = 33
+          }
```
jjmaestro, line 360: I think these need an extra space so that the `Expected` and `Actual` align with the `(` column

pkl-core/src/main/java/org/pkl/core/runtime/TestRunner.java
```diff
@@ -387,74 +388,85 @@ private static String getDisplayUri(ModuleInfo moduleInfo) {
...
+    String message = sourceSection.getCharacters().toString() + " " + renderLocation(location);
+    return new Failure("Fact Failure", message);
   }
 
   private static Failure exampleLengthMismatchFailure(
       String location, String property, int expectedLength, int actualLength) {
     String msg =
-        "Output mismatch: Expected \""
+        renderLocation(location)
+            + "\n"
```
jjmaestro, line 399: I think this is the extra new-line I mentioned before that I would remove

pkl-core/src/main/java/org/pkl/core/runtime/TestRunner.java
```diff
@@ -387,74 +388,85 @@ private static String getDisplayUri(ModuleInfo moduleInfo) {
...
+            + ":\n  "
+            + "Expected: "
+            + renderLocation(expectedLocation)
+            + "\n  "
+            + expectedValue.replaceAll("\n", "\n  ")
+            + "\n  "
             + "Actual: "
-            + "("
-            + actualLocation
-            + ")"
```
jjmaestro: Here I think you need an extra space in all of the `"\n  "`, as mentioned in the example output, so that it aligns with the "`(` column" (i.e. 3 spaces following the new line instead of the current two spaces).

jjmaestro (commented):
Just some nits re. column alignment and whitespace / new-line in the test output

pkl-core/src/main/java/org/pkl/core/runtime/TestRunner.java
bioball: Yup; this is stripped by `stripFileAndLines`

pkl-core/src/main/java/org/pkl/core/runtime/TestRunner.java
jjmaestro: The +1 space IMHO is not really "inconsistent" but a side-effect because the "fields" are not fixed-width. So, when you have "multiple fields" in one line (`#<nth>` and the `(file:...)`) and the fields are not fixed-width, then it's inevitable.

That's also why I think the "inconsistent" spacing is OK, because we use the inconsistencies to actually make it nice / aesthetically pleasing by aligning to the natural vertical whitespace.

Re. the width of numbers, you are right, I didn't consider it!  But, even so, since most cases will probably have well under 100 cases you can easily cover 90%+ of the actual usage by accounting for double digits.

E.g.

```
          examples
            ✅ user 0
            ❌ user 1
               #1  (file:///file, line x):
                   Expected: (file:///file, line x)
                     new {
                       name = "Parrot"
                       age = 35
                     }
                   Actual: (file:///file, line x)
                     new {
                       name = "Welma"
                       age = 35
                     }
               #91 (file:///file, line x):
                   Expected: (file:///file, line x)
                     new {
                       name = "Parrot"
                       age = 35
                     }
                   Actual: (file:///file, line x)
                     new {
                       name = "Welma"
                       age = 35
                     }
        ❌ 50.0% tests pass [2/4 failed], 66.7% asserts pass [2/6 failed]

```

That would almost surely work for most cases and I think it looks muuuch better than the `E/A` from `Expected` / `Actual` "mis-aligned` and taking over the vertical whitespace ... that really triggers me 😅

But I agree, a big part of this is a matter of preference, so  🤷.

holzensp (approved):

pkl-core/src/test/kotlin/org/pkl/core/EvaluateTestsTest.kt
jjmaestro: I've discussed it more [here](https://github.com/apple/pkl/pull/738#discussion_r1819607141), TL;DR yes, it's a moving target but IMHO it could be solved to a great extent just by forcing alignment with double-digit examples.

The suggestion is aesthetic but I think it's quite reasonable and helpful UX-wise. Maybe it's just me but I do find misalignments (in this case, the breaking of the vertical whitespace) distracting when looking at code / logs 😅

pkl-cli/src/main/kotlin/org/pkl/cli/CliTestRunner.kt
```diff
@@ -100,7 +102,8 @@ constructor(
         }
       }
       if (failed) {
-        throw CliTestException(ErrorMessages.create("testsFailed"))
+        val exitCode = if (isExampleWrittenFailure) 2 else 1
```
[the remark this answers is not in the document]
bioball: Yeah; maybe in the future we can have a table of exit codes and their meanings. I think this is okay for now though.

I'll also change this to `10`; a bit of padding here might be useful.

pkl-core/src/main/java/org/pkl/core/runtime/TestRunner.java
bioball: I think I'm going to keep my version as-is--not totally sold that indenting this much is a good thing. Sorry! But yes, there's no right way to do this, and I don't think my version is necessarily better than yours. However, I am concerned about the amount of indentation we already have, and I think we can improve this output even more when we have text coloring.

jjmaestro: Cool, no worries, thanks for considering it!

PR #739, opened 2024-10-24
Try fixing eval of non-ascii URI on Windows

bioball:


bioball:
I think the issue here is that `Path.of("日本语.pkl")` is turning into path `???.pkl` on Windows.

This is throwing:

https://github.com/apple/pkl/blob/8b0118fec5d62da0f0949b59c9feb759a0c9afbd/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt#L52

Not sure why we're getting a bunch of question marks, though.

translatenix:
@bioball The main branch builds fine for me.
Otherwise you'd have heard from me already. :-)
On my laptop, the `???` seems to be a terminal encoding issue:

```
jshell> Path.of("日本语.pkl").toString()
$13 ==> "???.pkl"

jshell> $13.equals("日本语.pkl")
$14 ==> true
```

translatenix:
@bioball Please ignore my earlier (now deleted) comment. I'm getting the same error as CI.
I don't know more yet, but I'll investigate a bit.

translatenix:
This looks much like: https://github.com/oracle/graal/issues/8593

The solution described here makes `gw testNative` pass on my laptop:
https://github.com/oracle/graal/issues/8593#issuecomment-2009163353

Not sure what this means for Pkl's CI build.

pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
```diff
@@ -90,11 +90,14 @@ public static URI toUri(String str) throws URISyntaxException {
 
   /** Converts a URI to a Path, normalizing any non-ASCII characters. */
   public static Path pathOf(URI uri) {
-    // Path.of(URI) throws on non-ASCII characters so the module URI here must be normalized to
-    // ASCII
+    // Path.of(URI) on unix throws on non-ASCII characters so the module URI here must be normalized
```
holzensp, line 93:
```suggestion
    // Path.of(URI) on POSIX throws on non-ASCII characters so the module URI here must be normalized
```

pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
```diff
@@ -90,11 +90,14 @@ public static URI toUri(String str) throws URISyntaxException {
...
+    // to ASCII.
     // Unfortunately there's no way to go from URI -> ASCII URI directly
     // so this must transform URI -> ASCII String -> ASCII URI
     try {
+      if (IoUtils.isWindows()) {
+        return Path.of(uri);
+      }
       return Path.of(new URI(uri.toASCIIString()));
```
holzensp, line 101:
```suggestion
      return Path.of(IoUtils.isWindows() ? uri : new URI(uri.toASCIIString()));
```

.circleci/config.pkl
```diff
@@ -21,7 +21,9 @@ import "jobs/GradleCheckJob.pkl"
 import "jobs/DeployJob.pkl"
 import "jobs/SimpleGradleJob.pkl"
 
-local prbJobs: Listing<String> = gradleCheckJobs.keys.toListing()
+local prbJobs: Listing<String> = (gradleCheckJobs.keys.toListing()) {
+  "pkl-cli-windows-amd64-snapshot"
```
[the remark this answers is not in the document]
bioball: This is just temporary code so we can test the windows native executable as a PRB

translatenix:
> Looks like we're blocked by the GraalVM folks on this.

In the meantime, you *could* try to find a different solution. For example, you could try `chcp 65001`. But from what I read, this should already be enable on GitHub runners. Probably there's also a registry equivalent of "Enable experimental UTF8 support", which solved the problem for me.

bioball:
I'll give that a shot, but if that doesn't work, I'm not going to spend too much time debugging this one.

translatenix:
I think that every Java app running on Windows faces this issue, not just Graal native images.
Until I enabled experimental UTF8 support, I'd get scrambled output from spotless, JMH, etc.

bioball:
The same test is passing in our JVM-based language snippet tests on Windows, so, it seems to be working at least when not in native-image.

translatenix:
> The same test is passing in our JVM-based language snippet tests on Windows,

That's because JVM based snippet tests don't pass filenames as command line arguments, which is where the problem lies.

bioball:
Hm, interesting.

I'm merging this for now. I think we can re-visit this again when that issue has been resolved, but if you have any other ideas what what to try in CI, I'm happy to fire up another PR just for testing purposes.

PR #740, opened 2024-10-24
Fix lazy type checking of UInt types

translatenix:
For example, `Listing<UInt16>` is equivalent to `Listing<UInt16>`, but not to `Listing<UInt32>`.


bioball (approved):
Ah! Good catch!

PR #743, opened 2024-10-27
Add isBase64 string property

jjuliano:
This PR adds isBase64 string property.


StefMa:
If we add this, do we need this for the other hash functions as well? Sha, md5 etc? 🤔

jjuliano:
> If we add this, do we need this for the other hash functions as well? Sha, md5 etc? 🤔

Maybe it makes sense only for base64 due to the base64Decoded string property, or make base64Decoded detect if a string is a base64 through `str.base64?.base64Decoded`.

bioball:
What's the use-case for this? Is it for defining a constraint?

Today, you can do this in-language:

```pkl
function isBase64(str: String) = str.match(Regex("[-A-Za-z0-9+/]*={0,3}"))
```

Or, a more precise version, and also a little bit more expensive:

```pkl
import "pkl:test"

function isBase64(str: String) = test.catch(() -> str.base64Decoded) == null
```

PR #746, opened 2024-10-28
Add color to error formatting

holzensp:


pkl-core/src/main/java/org/pkl/core/runtime/OutputFormatterColor.java
```diff
@@ -0,0 +1,183 @@
...
+    return escape(0);
+  }
+
+  private OutputFormatterColor escape(int i) {
+    return a('\033').a('[').a(i).a('m');
+  }
+
+  private OutputFormatterColor escape(String s) {
+    return a('\033').a('[').a(s).a('m');
+  }
```
bioball, line 149: Delete unused

pkl-core/src/main/java/org/pkl/core/evaluatorSettings/PklEvaluatorSettings.java
```diff
@@ -38,6 +38,7 @@ public record PklEvaluatorSettings(
     @Nullable Map<String, String> env,
     @Nullable List<Pattern> allowedModules,
     @Nullable List<Pattern> allowedResources,
+    @Nullable Boolean color,
```
bioball, line 41: Blocking: feels like this should be `@Nullable Color color` (where we move enum `Color` from `pkl-commons-cli`)

I think we should turn that enum into a boolean color right before passing this option to `EvaluatorBuilder`.

pkl-core/src/main/java/org/pkl/core/project/Project.java
```diff
@@ -506,6 +509,7 @@ public EvaluatorSettings(
         @Nullable Map<String, String> env,
         @Nullable List<Pattern> allowedModules,
         @Nullable List<Pattern> allowedResources,
+        @Nullable Boolean color,
```
bioball, line 512: Let's remove this line; this class is deprecated and should be considered as not receiving updates.

pkl-core/src/main/java/org/pkl/core/runtime/OutputFormatterColor.java
```diff
@@ -0,0 +1,183 @@
...
+
+  private OutputFormatterColor a(char ch) {
+    builder.append(ch);
+    return this;
+  }
+
+  private OutputFormatterColor a(Object obj) {
+    builder.append(obj);
+    return this;
+  }
```
bioball, line 133: [nit] delete these methods and replace their usage with `append`

pkl-core/src/main/java/org/pkl/core/runtime/OutputFormatterColor.java
```diff
@@ -0,0 +1,183 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.core.runtime;
+
+import java.io.PrintWriter;
+import org.pkl.core.OutputFormatter;
+import org.pkl.core.util.StringBuilderWriter;
+
+public class OutputFormatterColor extends OutputFormatter<OutputFormatterColor> {
```
bioball, line 22: Suggestion: `AnsiFormatter`

pkl-core/src/main/java/org/pkl/core/OutputFormatter.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public abstract SELF margin(String marginMatter);
+
+  public abstract SELF hint(String hint);
+
+  public abstract SELF newline();
+
+  public abstract SELF repetitions(int counter);
+
+  public abstract SELF text(String text);
```
bioball, line 37: Delete; replace usages with `append`

pkl-core/src/main/java/org/pkl/core/OutputFormatter.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public abstract SELF createBlank();
+
+  public abstract SELF margin(String marginMatter);
+
+  public abstract SELF hint(String hint);
+
+  public abstract SELF newline();
+
+  public abstract SELF repetitions(int counter);
```
bioball, line 35:
```suggestion
  public abstract SELF stackOverflowLoopCount(int counter);
```

pkl-core/src/main/java/org/pkl/core/OutputFormatter.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.PrintWriter;
+import org.pkl.core.runtime.OutputFormatterColor;
+import org.pkl.core.runtime.OutputFormatterPlain;
+
+public abstract class OutputFormatter<SELF extends OutputFormatter<SELF>> {
+  public static OutputFormatter<?> create(boolean usingColor) {
+    return usingColor ? new OutputFormatterColor() : new OutputFormatterPlain();
+  }
+
+  public abstract SELF createBlank();
```
bioball, line 27:
```suggestion
  public abstract SELF newInstance();
```

stdlib/EvaluatorSettings.pkl
```diff
@@ -67,6 +67,9 @@ allowedModules: Listing<String(isRegex)>?
 /// ```
 allowedResources: Listing<String(isRegex)>?
 
+/// Whether errors should be formatted with ANSI color.
+color: ("never"|*"auto"|"always")?
```
bioball, line 71: Nullable _plus_ default here doesn't do anything, does it?

```suggestion
/// When to format messages with ANSI color codes.
///
/// Possible values:
///
/// - `"never"`: Never format 
/// - `"auto"`: Format if the process' stdin, stdout, or stderr are connected to a console.
/// - `"always"`: Always format
@Since { version = "0.27.0" }
color: ("never"|"auto"|"always")?
```

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -128,6 +129,16 @@ class BaseOptions : OptionGroup() {
...
 
+  val color: Color? by
+    option(
+        names = arrayOf("--color"),
+        metavar = "<when>",
+        help =
+          "Whether to format error messages in ANSI color. Possible values of <when> are \"never\", \"auto\", and \"always\"."
+      )
+      .enum<Color> { it.value }
+      .single()
```
bioball, line 140:
```suggestion
  val color: Color by
    option(
        names = arrayOf("--color"),
        metavar = "<when>",
        help =
          "Whether to format messages in ANSI color. Possible values of <when> are \"never\", \"auto\", and \"always\"."
      )
      .enum<Color> { it.value }
      .single()
      .default(Color.AUTO)
```

docs/modules/pkl-gradle/partials/gradle-common-properties.adoc
```diff
@@ -67,6 +67,13 @@ The cache directory for storing packages.
 If `null`, defaults to `~/.pkl/cache`.
 ====
 
+.color: Property<Boolean>
+[%collapsible]
+====
+Default: `false` +
+Format errors using ANSI color.
```
bioball, line 74:
```suggestion
Format messages using ANSI color.
```

bioball (changes requested):
Generally looks good! I have mostly nits but some blocking comments too.

Something else that's missing: docs for pkl-cli

pkl-core/src/main/java/org/pkl/core/OutputFormatter.java
```diff
@@ -0,0 +1,56 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.core;
+
+import java.io.PrintWriter;
+import org.pkl.core.runtime.OutputFormatterColor;
+import org.pkl.core.runtime.OutputFormatterPlain;
+
+public abstract class OutputFormatter<SELF extends OutputFormatter<SELF>> {
```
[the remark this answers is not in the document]
holzensp: Mostly done as suggested; part of the intent is to improve them in the very near future (TODO comments added).

If, for example, we move margin-matter-management into the formatter (`pushMargin(String matter)` / `popMargin()`), then `newline()` can write the required margin matter. Similarly; currently all ANSI coloring defaults to bookending all semantic units (e.g. `errorHeader`) with escape codes (setting color before and resetting after). By doing state tracking in the formatter, we can significantly reduce the amount of escape codes produced.

stdlib/EvaluatorSettings.pkl
holzensp: I felt it was a reasonable way to make it self-documenting that `"auto"` was the default setting, but sure; it's not functional.

pkl-core/src/main/java/org/pkl/core/Analyzer.java
```diff
@@ -49,12 +51,14 @@ public class Analyzer {
 
   public Analyzer(
       StackFrameTransformer transformer,
+      TextFormatter<?> textFormatter,
```
bioball, line 54:
```suggestion
      boolean color,
```

pkl-core/src/main/java/org/pkl/core/project/Project.java
```diff
@@ -192,6 +195,7 @@ private static List<List<URI>> findImportCycle(ModuleSource moduleSource) {
     var analyzer =
         new Analyzer(
             StackFrameTransformers.defaultTransformer,
+            TextFormatter.create(false),
```
bioball, line 198:
```suggestion
            builder.getColor(),
```

pkl-core/src/main/java/org/pkl/core/project/ProjectPackager.java
```diff
@@ -108,6 +110,7 @@ public ProjectPackager(
       Path workingDir,
       String outputPathPattern,
       StackFrameTransformer stackFrameTransformer,
+      TextFormatter<?> textFormatter,
```
bioball, line 113:
```suggestion
      boolean color,
```

pkl-core/src/main/java/org/pkl/core/runtime/VmException.java
```diff
@@ -117,8 +117,16 @@ public String toString() {
...
 
   @TruffleBoundary
-  public PklException toPklException(StackFrameTransformer transformer) {
-    var renderer = new VmExceptionRenderer(new StackTraceRenderer(transformer));
+  public PklException toPklException(
+      StackFrameTransformer transformer, TextFormatter<?> formatter) {
+    var renderer = new VmExceptionRenderer(new StackTraceRenderer(transformer), formatter);
+    var rendered = renderer.render(this);
+    return new PklException(rendered);
+  }
```
bioball, line 125: Let's remove this method; every usage should instead pass a `boolean color` in.

pkl-core/src/main/java/org/pkl/core/runtime/AnsiFormatter.java
```diff
@@ -0,0 +1,144 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.core.runtime;
+
+/*
+   TODO:
+     * Manage "active state" to reduce "escape code spew" in output
+*/
```
bioball, line 21: Given the current API, not sure if this is possible?

pkl-core/src/main/java/org/pkl/core/runtime/TextFormatter.java
```diff
@@ -0,0 +1,103 @@
...
+
+  protected SELF a(char ch) {
+    builder.append(ch);
+    return self();
+  }
+
+  protected SELF a(Object obj) {
+    builder.append(obj);
+    return self();
+  }
```
bioball, line 102: I think we should just call this `append`, and get rid of `.object`, `.text`, etc.

pkl-core/src/main/java/org/pkl/core/runtime/TextFormatter.java
```diff
@@ -0,0 +1,103 @@
...
+
+  public abstract SELF lineNumber(String line);
+
+  public abstract SELF text(String text);
+
+  public abstract SELF object(Object obj);
+
+  public abstract SELF errorHeader(String header);
+
+  public abstract SELF error(String message);
```
bioball, line 70: General note: I think a better API is to just have methods that describe _what_ the format is. For example:

```java
out
  .bold()
  .color(Color.RED)
  .append("–– Pkl Error ––")
  .append("\n")
```

This feels much easier to use when we want to extend colors to other places.

If the goal is to be able to theme this later on, that can happen in when these messages are built.

```java
out
  .bold()
  .color(errorHeaderColor)
  .append("–– Pkl Error ––")
  .reset()
  .append("\n")
```

Here, `errorHeaderColor` can come from some config file or something.

Also: I don't know if we really need multiple subclasses; it seems good enough to just have a text formatter that accepts `boolean color` as a constructor arg.

Given this is all private, we don't need to figure this out right now. But, I'm tempted to say that we should refactor this later.

bioball (changes requested):
Another pass!

Some nits, and some blocking comments around passing around `TextFormatter` instead of `boolean`.

pkl-core/src/main/java/org/pkl/core/runtime/AnsiFormatter.java
holzensp: It was (`fg` and `fgBright` could just check current state and emit an escape or not), but I think the new approach covers this also.

pkl-core/src/main/java/org/pkl/core/runtime/TextFormatter.java
bioball: If colors are disabled, `.color(Color.RED)` would just be a no-op!

docs/modules/pkl-cli/partials/cli-common-options.adoc
```diff
@@ -19,6 +19,19 @@ Patterns are matched against the beginning of resource URIs.
...
+[[color]]
+.--color
+[%collapsible]
+====
+Default: `auto` +
+When to format messages with ANSI color codes.
+Possible values:
+
+- `"never"`: Never format
+- `"auto"`: Format if the process' `stdin`, `stdout`, or `stderr` are connected to a console.
```
madrob, line 31: `process'` is a britishism, maybe collapse to
> Format if `stdin`...

(same for EvaluatorSettings)

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -141,6 +143,17 @@ class BaseOptions : OptionGroup() {
       )
       .associateProps()
 
+  val color: Color by
+    option(
+        names = arrayOf("--color"),
+        metavar = "<when>",
+        help =
+          "Whether to format messages in ANSI color. Possible values of <when> are \"never\", \"auto\", and \"always\"."
```
madrob, line 151: style: single quotes wouldn't need escaping

pkl-core/src/main/java/org/pkl/core/runtime/VmException.java
```diff
@@ -117,8 +117,8 @@ public String toString() {
   }
 
   @TruffleBoundary
-  public PklException toPklException(StackFrameTransformer transformer) {
-    var renderer = new VmExceptionRenderer(new StackTraceRenderer(transformer));
+  public PklException toPklException(StackFrameTransformer transformer, boolean color) {
```
madrob, line 120: Instead of each caller having to keep track of color selection, it feels like something that the Renderer should hold onto. That's likely a bigger change though.

bioball: This is an internal method; our public APIs throw `PklException`s with messages that are either colored or not, depending on how they've configured the evaluator

pkl-core/src/main/java/org/pkl/core/evaluatorSettings/Color.java
```diff
@@ -0,0 +1,30 @@
...
+  AUTO,
+  ALWAYS;
+
+  public boolean hasColor() {
+    return switch (this) {
+      case AUTO -> System.console() != null;
+      case NEVER -> false;
+      case ALWAYS -> true;
+    };
+  }
```
bioball, line 29: [nit] would be good to move this method elsewhere (data objects shouldn't be concerned with business logic)

pkl-core/src/main/java/org/pkl/core/project/Project.java
```diff
@@ -517,6 +519,7 @@ public EvaluatorSettings(
               env,
               allowedModules,
               allowedResources,
+              Color.AUTO,
```
bioball, line 522:
```suggestion
              null,
```

pkl-core/src/main/java/org/pkl/core/runtime/VmExceptionRenderer.java
```diff
@@ -16,66 +16,77 @@
...
   /**
    * Constructs an error renderer with the given stack trace renderer. If stack trace renderer is
    * {@code null}, stack traces will not be included in error output.
    */
-  public VmExceptionRenderer(@Nullable StackTraceRenderer stackTraceRenderer) {
+  public VmExceptionRenderer(
+      @Nullable StackTraceRenderer stackTraceRenderer, TextFormatter textFormatter) {
     this.stackTraceRenderer = stackTraceRenderer;
+    this.textFormatter = textFormatter;
+  }
```
bioball, line 39: Delete; replace all usages with just passing in a boolean

pkl-core/src/main/java/org/pkl/core/repl/ReplServer.java
```diff
@@ -83,7 +83,9 @@ public ReplServer(
     this.workingDir = workingDir;
     this.securityManager = securityManager;
     this.moduleResolver = new ModuleResolver(moduleKeyFactories);
-    this.errorRenderer = new VmExceptionRenderer(new StackTraceRenderer(frameTransformer));
+    this.errorRenderer =
+        new VmExceptionRenderer(
+            new StackTraceRenderer(frameTransformer), TextFormatter.create(false));
```
bioball, line 88: `--color` is an arg that's also available to `pkl repl`. So, let's add `color` as a field and constructor arg to `ReplServer`, and the `color` option through in `CliRepl`.

Also (per my other comment): change this to just accept a boolean.

```suggestion
    this.errorRenderer =
        new VmExceptionRenderer(
            new StackTraceRenderer(frameTransformer), color);
```

bioball (approved):
Left a couple more comments, but approving to unblock!

docs/modules/pkl-cli/partials/cli-common-options.adoc
holzensp: TIL! (for my future reference; the britishism is that there no `s` after the `'`, right?)

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
holzensp: Fair shout. Force of habit from referring to Pkl enum values (where the `"`s are part of the actual value).

docs/modules/pkl-cli/partials/cli-common-options.adoc
madrob: Yea `process's` (en-us) vs. `process'` (en-uk)

bioball: TIL too; I also use a single apostrophe for this.

PR #748, opened 2024-10-29
Convert `org.pkl.core.resource` POJOs to `record` classes

KushalP:


pkl-core/src/main/java/org/pkl/core/resource/ResourceReaders.java
```diff
@@ -690,16 +680,9 @@ public void close() {
...
-    private final ExternalResourceResolver.Spec readerSpec;
-    private final ExternalResourceResolver resolver;
-
-    public ExternalResolver(
-        ExternalResourceResolver.Spec readerSpec, ExternalResourceResolver resolver) {
-      this.readerSpec = readerSpec;
-      this.resolver = resolver;
-    }
-
+  private record ExternalResolver(
```
bioball, line 683: These don't make too much sense as a record; they're not data objects.

pkl-core/src/main/java/org/pkl/core/resource/Resource.java
```diff
@@ -20,22 +20,13 @@
...
-
-  /** Constructs a resource. */
-  public Resource(URI uri, byte[] bytes) {
-    this.uri = uri;
-    this.bytes = bytes;
-  }
-
-  /** Returns the URI of this resource. */
+public record Resource(URI uri, byte[] bytes) {
+  @Deprecated
```
bioball, line 24:
```suggestion
  /**
   * @deprecated As of 0.28.0, replaced by {@link #uri()}
   */
  @Deprecated(forRemoval = true)
```

pkl-core/src/main/java/org/pkl/core/resource/Resource.java
```diff
@@ -20,22 +20,13 @@
...
   public URI getUri() {
     return uri;
   }
 
+  @Deprecated
```
bioball, line 29:
```suggestion
  /**
   * @deprecated As of 0.28.0, replaced by {@link #bytes()}
   */
  @Deprecated(forRemoval = true)
```

bioball (changes requested):
Two quick comments, otherwise this looks good

pkl-core/src/main/java/org/pkl/core/resource/ResourceReaders.java
```diff
@@ -341,12 +341,7 @@ public Optional<Object> read(URI uri) throws IOException {
     }
   }
 
-  private static final class ClassPathResource implements ResourceReader {
-    private final ClassLoader classLoader;
-
-    public ClassPathResource(ClassLoader classLoader) {
-      this.classLoader = classLoader;
-    }
+  private record ClassPathResource(ClassLoader classLoader) implements ResourceReader {
```
bioball, line 344: Let's undo this one and `ModulePathResource`; these aren't POJOs

bioball (changes requested):

bioball (approved):
LGTM!

PR #750, opened 2024-10-29
Convert `org.pkl.core.util` POJOs to `record` classes

KushalP:


pkl-core/src/main/java/org/pkl/core/util/GlobResolver.java
```diff
@@ -45,26 +45,24 @@ public InvalidGlobPatternException(String message) {
...
 
+    @Deprecated
     public URI getUri() {
       return uri;
     }
 
+    @Deprecated
     public boolean isDirectory() {
       return isDirectory;
     }
```
bioball, line 68: The `util` package is not a public API, so no need to add deprecated methods.

bioball, line 68: The util package is not a public API, so no need to add deprecated methods.

```suggestion
  public record ResolvedGlobElement(String path, URI uri, boolean isDirectory) {
  }
```

bioball (changes requested):
Thanks! Just one comment.

pkl-core/src/main/java/org/pkl/core/util/GlobResolver.java
bioball: Oh, we definitely want that, my bad!

pkl-core/src/main/java/org/pkl/core/util/yaml/snake/YamlResolver.java
```diff
@@ -48,15 +48,7 @@ public Tag resolve(String value, Boolean implicit) {
...
-
-    ResolverTuple(Tag tag, Pattern regexp) {
-      this.tag = tag;
-      this.regexp = regexp;
-    }
-
+  private record ResolverTuple(Tag tag, Pattern regexp) {
     @Override
     public String toString() {
       return "Tuple tag=" + tag + " regexp=" + regexp;
```
KushalP, line 54: Some inconsistency here with the `toString()` format. Might be worth deleting this and using the one that comes with `record` ?

bioball: The `org.pkl.core.util.yaml.snake` package contains source code that is vendored from the SnakeYAML library.

bioball (approved):
LGTM

PR #751, opened 2024-10-29
 Correct EmbeddedExecutor to throw ExecutorException instead of PklException when project evaluation fails

HT154:
There are two fixes here:
* pkl-core's `ExecutorSpiImpl` now properly handles project evaluation errors
* pkl-executor's `EmbeddedExecutor` now catches all `RuntimeException`s and transforms them to `ExecutorException` so the `Executor` API contract is upheld.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/service/ExecutorSpiImpl.java
+++ b/pkl-core/src/main/java/org/pkl/core/service/ExecutorSpiImpl.java
@@ -122,7 +122,7 @@ public final class ExecutorSpiImpl implements ExecutorSpi {
         var project =
             Project.loadFromPath(
                 options.getProjectDir().resolve(PKL_PROJECT_FILENAME),
-                SecurityManagers.defaultManager,
+                securityManager,
                 null,
                 transformer,
                 System.getenv());
--- a/pkl-core/src/main/java/org/pkl/core/service/ExecutorSpiImpl.java
+++ b/pkl-core/src/main/java/org/pkl/core/service/ExecutorSpiImpl.java
@@ -125,7 +125,7 @@ public final class ExecutorSpiImpl implements ExecutorSpi {
                 securityManager,
                 null,
                 transformer,
-                System.getenv());
+                options.getEnvironmentVariables());
         builder.setProjectDependencies(project.getDependencies());
       }
 
--- a/pkl-core/src/main/java/org/pkl/core/service/ExecutorSpiImpl.java
+++ b/pkl-core/src/main/java/org/pkl/core/service/ExecutorSpiImpl.java
@@ -116,13 +116,22 @@ public final class ExecutorSpiImpl implements ExecutorSpi {
             .setTimeout(options.getTimeout())
             .setOutputFormat(options.getOutputFormat())
             .setModuleCacheDir(options.getModuleCacheDir());
-    if (options.getProjectDir() != null) {
-      var project = Project.loadFromPath(options.getProjectDir().resolve(PKL_PROJECT_FILENAME));
-      builder.setProjectDependencies(project.getDependencies());
-    }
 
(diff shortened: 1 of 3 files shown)


pkl-core/src/main/java/org/pkl/core/service/ExecutorSpiImpl.java
```diff
@@ -116,13 +116,22 @@ public String evaluatePath(Path modulePath, ExecutorSpiOptions options) {
...
-    }
 
-    try (var evaluator = builder.build()) {
-      return evaluator.evaluateOutputText(ModuleSource.path(modulePath));
+    try {
+      if (options.getProjectDir() != null) {
+        var project =
+            Project.loadFromPath(
+                options.getProjectDir().resolve(PKL_PROJECT_FILENAME),
+                SecurityManagers.defaultManager,
```
bioball, line 125: While we're here; this should really use the same security manager as the normal evaluator

```suggestion
                securityManager,
```

pkl-executor/src/main/java/org/pkl/executor/EmbeddedExecutor.java
```diff
@@ -227,6 +227,13 @@ String evaluatePath(Path modulePath, ExecutorOptions options) {
         return executorSpi.evaluatePath(modulePath, options.toSpiOptions());
       } catch (ExecutorSpiException e) {
         throw new ExecutorException(e.getMessage(), e.getCause(), executorSpi.getPklVersion());
+      } catch (RuntimeException e) {
+        // This branch would ideally never be hit, but older Pkl releases (<0.27) may trigger this
+        // by bubbling PklException up to this point in cases where project evaluation fails.
+        // This catch-all ensures the API contract of Executor is upheld.
+        // Unfortunately the PklException thrown by the ExecutorSpi is from a different class loader
+        // so the host program cannot directly catch PklException specifically
+        throw new ExecutorException(e.getMessage(), e.getCause());
```
bioball, line 236: Good fix! But probably better to only catch `PklException`.

```suggestion
        // This branch would ideally never be hit, but older Pkl releases (<0.27) erroneously throw
        // PklException in some cases.
        // Can't catch PklException directly because pkl-executor cannot depend on pkl-core.
        if (e.getClass().getName().equals("org.pkl.core.PklException")) {
          throw new ExecutorException(e.getMessage(), e.getCause());
        }
        throw e;
```

pkl-executor/src/test/kotlin/org/pkl/executor/EmbeddedExecutorTest.kt
```diff
@@ -415,6 +420,60 @@ class EmbeddedExecutorTest {
...
+    assertThat(e.message)
+      .contains("Cannot find module")
+      // ensure module file paths are relativized
+      .contains("/nonexistent/PklProject")
+
+    // legacy distribution does not handle these errors with the correct stack frame transformer
+    // only assert on this for newer distributions
+    if (!executor.toString().contains("Distribution1")) {
+      assertThat(e.message).doesNotContain(tempDir.toString())
+    }
```
bioball, line 474: Not so sure what this assertion is about, but it's failing on my machine

HT154: You might see this if you haven't built  since switching to this branch since EmbeddedExecutorTest loads the built `pkl-config-java/build/libs/pkl-config-java-all-SNAPSHOT.jar`. The tests error if this file is not present, but not if it's out of date.

bioball: Ah, gotcha!

bioball (approved):
LGTM, thanks for the fix!

PR #752, opened 2024-10-30
Eagerly check listing/mapping in iterables

bioball:
When we introduced lazy listing/mapping typechecking, we widened an existing bug around resolving for-generator variables in https://github.com/apple/pkl/issues/741.

To prevent this bug from being widened, this introduces logic to eagerly evaluate any `Listing<V>` or `Mapping<K, V>` found in the iterable.

We're not sure what the best approach is yet to fix the bug. In the meantime, this prevents the existing bug from being any more prevalent than it currently is.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
@@ -908,7 +908,7 @@ public final class AstBuilder extends AbstractAstBuilder<Object> {
         createSourceSection(ctx),
         expr,
         ctx.QSPREAD() != null,
-        symbolTable.getCurrentScope().isVisitingIterable());
+        visitingIterable);
   }
 
   private void insertWriteForGeneratorVarsToFrameSlotsNode(@Nullable MemberNode memberNode) {
--- a/pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
@@ -904,8 +904,7 @@ public final class AstBuilder extends AbstractAstBuilder<Object> {
     scope.setVisitingIterable(true);
     var expr = visitExpr(ctx.expr());
     scope.setVisitingIterable(visitingIterable);
-    return GeneratorSpreadNodeGen.create(
-        createSourceSection(ctx), expr, ctx.QSPREAD() != null, visitingIterable);
+    return GeneratorSpreadNodeGen.create(createSourceSection(ctx), expr, ctx.QSPREAD() != null);
   }
 
   private void insertWriteForGeneratorVarsToFrameSlotsNode(@Nullable MemberNode memberNode) {
@@ -1207,7 +1206,7 @@ public final class AstBuilder extends AbstractAstBuilder<Object> {
 
           var modifier =
               scope.isVisitingIterable()
-                  ? VmModifier.ELEMENT | VmModifier.ITERABLE_MEMBER
+                  ? VmModifier.ELEMENT | VmModifier.IS_IN_ITERABLE
                   : VmModifier.ELEMENT;
           var member =
               new ObjectMember(
@@ -1266,7 +1265,7 @@ public final class AstBuilder extends AbstractAstBuilder<Object> {
     return scope -> {
       var modifier =
           scope.isVisitingIterable()
-              ? VmModifier.ENTRY | VmModifier.ITERABLE_MEMBER
+              ? VmModifier.ENTRY | VmModifier.IS_IN_ITERABLE
               : VmModifier.ENTRY;
       var member =
           new ObjectMember(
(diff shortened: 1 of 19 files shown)


holzensp (approved):
Sad, but true.

holzensp (changes requested):
Sorry, wasn't paying attention when I reviewed that. As discussed off-line; this surfacing another bug, I think close it out and have a go at tracking a marker in AstBuilder.

bioball:
Okay, this is now using the AstBuilder approach. This works a lot better; take another look!

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -899,8 +904,16 @@ public GeneratorMemberNode visitObjectEntry(ObjectEntryContext ctx) {
...
+    var visitingIterable = scope.isVisitingIterable();
+    scope.setVisitingIterable(true);
+    var expr = visitExpr(ctx.expr());
+    scope.setVisitingIterable(visitingIterable);
     return GeneratorSpreadNodeGen.create(
-        createSourceSection(ctx), visitExpr(ctx.expr()), ctx.QSPREAD() != null);
+        createSourceSection(ctx),
+        expr,
+        ctx.QSPREAD() != null,
+        symbolTable.getCurrentScope().isVisitingIterable());
```
holzensp, line 916: If this scope is not aligned with the above, something went wrong, no?
```suggestion
        visitingIterable);
```

pkl-core/src/main/java/org/pkl/core/ast/member/ObjectMember.java
```diff
@@ -27,6 +27,9 @@
 import org.pkl.core.util.Nullable;
 
 public final class ObjectMember extends Member {
+
+  private final boolean isInIterable;
```
holzensp, line 31: Seems costly... not worth rolling it into the member's `modifiers`? Probably worth profiling (both `jpkl` and native) on large corpora.

pkl-core/src/main/java/org/pkl/core/ast/member/PropertyTypeNode.java
```diff
@@ -61,9 +61,30 @@ public String getName() {
...
 
+  private boolean isInIterable(VirtualFrame frame) {
+    if (frame.getArguments().length < 4) {
+      return false;
+    }
+    if (frame.getArguments()[3] instanceof Boolean isInIterable) {
+      return isInIterable;
+    }
+    return false;
+  }
```
holzensp, line 72:
```suggestion
  private boolean isInIterable(VirtualFrame frame) {
    var args = frame.getArguments();
    return args.length >= 4 && args[3] instanceof Boolean b && b;
  }
```

^^ less branching ;)

pkl-core/src/main/java/org/pkl/core/ast/member/PropertyTypeNode.java
```diff
@@ -61,9 +61,30 @@ public String getName() {
...
+
   @Override
   public Object execute(VirtualFrame frame) {
     try {
+      if (isInIterable(frame)) {
+        // There is currently a bug around resolving variables within the iterable of a for
+        // generator or spread syntax (https://github.com/apple/pkl/issues/741)
+        //
+        // Normally, mappings/listings are type-checked lazily. However, this results in the said
+        // bug getting widened, for any object members declared in the iterable.
```
holzensp, line 82:
```suggestion
        // Normally, mappings/listings are type-checked lazily. However, this results in said
        // bug getting widened, for any object members declared in the iterable.
```
^^ Only still used in British English, esp. legalese;

> **The said.** As used in legal writing, the word said is a Middle-English sibling aforesaid, having the sense "above-stated." Originally legal writers would write _the said defendant_ － and still do in BrE － just as they would write _the aforesaid defendant_ or _the above-stated defendant_. In AmE, however, _the_ was dropped before _said_, which has come to act almost as an article.

 -- A Dictionary of Modern Legal Usage, 2nd edition, Bryan A. Garner

pkl-core/src/main/java/org/pkl/core/ast/member/TypedPropertyNode.java
```diff
@@ -26,6 +26,7 @@
 /** A property definition that has a type annotation. */
 public final class TypedPropertyNode extends RegularMemberNode {
   @Child private DirectCallNode typeCheckCallNode;
+  private final ObjectMember member;
```
holzensp, line 29: If booleans can be tightly packed, the cost concern in `ObjectMember` (above) goes away. In that case, I'd also just grab the single boolean here, because a pointer won't be.

pkl-core/src/main/java/org/pkl/core/ast/member/ObjectMember.java
```diff
@@ -35,9 +38,11 @@ public ObjectMember(
       SourceSection headerSection,
       int modifiers,
       @Nullable Identifier name,
-      String qualifiedName) {
+      String qualifiedName,
+      boolean isInIterable) {
```
holzensp, line 42: There's quite a lot of extra code throughout, all with a default `false` case. I don't care much, considering it's likely rather temporary, but might it be worth an extra constructor that defaults this to `false`?

holzensp (approved):
No blockers, but I'd profile a bit before adding a member to _every object member_.

pkl-core/src/main/java/org/pkl/core/ast/member/ObjectMember.java
bioball: Good idea! Will make this a modifier instead.

PR #753, opened 2024-10-30
Replace code tick with italic

StefMa:
Right now it looks like that (red underline):
Screenshot 2024-10-30 at 10 39 57 AM

I removed the code ticks and replaced them with an underscore to better match the style (see the blue circle 🙃).


bioball (approved):
LGTM, thanks for the fix!

PR #762, opened 2024-10-31
Refine external reader API

HT154:
Thanks to @translatenix for identifying these in #759. This is the subset of items that were either low-hanging fruit or definitely need be addressed before this becomes public API.

* Encapsulate message transport by removing `ExternalReaderProcess.getTransport` and adding `getModuleResolver` and `getResourceResolver` methods
* Reuse `Random` instances within `ExternalReaderProcessImpl` and module/resource resolvers
* Externalize all `ExternalReaderProcessException` messages
* Add some missing doc comments to `ModuleKeyFactories` and `ResourceReaders` methods for external readers
* Move `org.pkl.core.util.Readers` to `org.pkl.core.Readers`


bioball (approved):
LGTM, good cleanups!

translatenix:
@HT154 Did you consider addressing this one?
* org.pkl.core.module.ResourceReaders API is (too) tightly coupled to reader protocol via ModuleReaderSpec

Perhaps I'm missing something, but a high-level user-facing API that exports a low-level internal API is a grave breach of modularity.

HT154:
@translatenix Yup, needed some more time to chew on these

> Did you consider addressing this one?
> 
> * org.pkl.core.module.ResourceReaders API is (too) tightly coupled to reader protocol via ModuleReaderSpec
> 
> Perhaps I'm missing something, but a high-level user-facing API that exports a low-level internal API is a grave breach of modularity.

The only way I can think of to do this is to move `ModuleKeys.ExternalResolver`, `ModuleKeyFactories.ExternalProcess`, `ResourceReaders.ExternalResolver`, and `ResourceReaders.ExternalProcess` to `org.pkl.core.messaging`. Some early feedback on the original PR requested moving these into their current locations. I'm definitely open to doing this, but I'm interested in hearing from @bioball and the crew before I jump into it.

> One more idea: `Readers.closeQuietly()` is totally generic, so it might as well be `Closeables.closeQuietly()`. Perhaps it's needed somewhere else in the future.

I don't feel too strongly about this, but happy to take it on if there's desire from the core team.

> Last but not least, I thought about renaming `ExternalReaderProcess` to `ExternalReader`, as the API doesn't seem to mandate a process.

I am planning on opening a separate PR for this tomorrow.

bioball:
One way to hide the `messaging` package is to define `ModuleReaderSpec` as an interface in `externalreader`, and use that as the return type of `org.pkl.core.externalreader.ExternalReaderProcess#getModuleReaderSpec`.

Then, have record `org.pkl.core.messaging.Messages.ModuleReaderSpec` implement that interface. And `ExternalReaderProcessImpl` would just return an instance of that record.

This refactor sounds reasonable, so, go for it @HT154.

> I don't feel too strongly about this, but happy to take it on if there's desire from the core team.

Yeah `Closeables` sounds like a better name.

PR #763, opened 2024-11-01
Vendor msgpack-java

bioball:
This adds msgpack-java's sources to pkl-core, which allows us to exclude this as a dependency.

This excludes APIs that we do not need (direct byte buffers, android support).

This follows a pattern that we have with a few other libraries (JSON, YAML).

Note: package `org.pkl.core.util.msgpack` is considered private, but yet is used in `pkl-server`. In the future, we can fix this by moving all serializing/deserializing to pkl-core.]

Note 2: this increases pkl-core's jar size from 3.3M to 3.5M.


translatenix:
Vendoring and optional dependency each have their own drawbacks. As msgpack is large and rarely needed by Pkl Java library users, my instinct is to make it optional for now (`compileOnly`) and document in which cases users need to declare the dependency themselves. Even if `BinaryEvaluator` is moved, most users of pkl-config-java etc. still won't need msgpack.

My vision is that one day, there will be a pkl-evaluator module (subset of pkl-core) that communicates with the outside world (even pkl-config-java etc.) purely through (a more efficient version of) the binary protocol. At that point, it might be a good idea to vendor MiniPack, which is one-tenth the size of msgpack. :-)

bioball:
One downside with making msgpack a `compileOnly` dependency is that it introduces `ClassNotFoundException`s at runtime. As a library user, you don't get any compiler errors if msgpack-java is not included. That sounds kind of dangerous to me.

msgpack-java is large, but, 0.2M seems palatable? But also, 1/10 the size with MiniPack sounds good if/when that becomes available.

> My vision is that one day, there will be a pkl-evaluator module (subset of pkl-core) that communicates with the outside world (even pkl-config-java etc.) purely through (a more efficient version of) the binary protocol. At that point, it might be a good idea to vendor MiniPack, which is one-tenth the size of msgpack. :-)

This is something we've talked about in the past, too. This would be nice for folks who can't use Java 17+, or can be used for the pkl-executor use-case.

Also, something else we've talked about is having an in-language binary API. So, you can do things like: use Pkl to write the binary output of a module, then deserialize that in Java/Swift/Go. Or go the other way around, too.

translatenix:
Turns out that the msgpack-core JAR is only 123 KB. I have no idea how I came to believe it's 500 KB. (MiniPack is ~50 KB.)
I'll leave it to you whether to vendor, make optional, or shade msgpack-core.

translatenix:
> This is something we've talked about in the past, too. This would be nice for folks who can't use Java 17+, or can be used for the pkl-executor use-case.

In fact I'd like to see pkl-evaluator developed against the latest Java LTR and latest Graal/Truffle. Among other things, it's the only way to keep up with supported Graal/Truffle versions. If pkl-evaluator only communicates over the binary protocol, everything built on top can be run with the pkl-evaluator Java library, native executable, or native library. This offers a lot of flexibility. For example, shipping a small Pkldoc executable becomes trivial. Also, the latest Java LTR is only required at run time if pkl-evaluator is run as a Java library, which should become the exception due to very poor performance and isolation. A JVM that loads thousands of Truffle classes for one-time config evaluation isn't a happy JVM.

> Also, something else we've talked about is having an in-language binary API. So, you can do things like: use Pkl to write the binary output of a module, then deserialize that in Java/Swift/Go. Or go the other way around, too.

This makes a lot of sense. It means that for some use cases, no evaluator is required at run time, which can be a big plus.

translatenix:
Turns out that there is already a `pkl-config-java-all` fat JAR with shaded dependencies. So probably even a regular dependency on msgpack-core is fine. Sorry that I didn't notice this sooner.

holzensp:
At 500KB, I would agree that the dependency is prohibitively cumbersome. At 123 KB, less so.

I think we should generally look to modularize Pkl much further, to give users a better granularity of dependencies. I was mostly arguing this for the binary (having a model similar to `git` or `protoc` where, for example, some subcommands are delegated to external binaries that need to be installed separately). I think doing the same for the JVM uses makes sense, too. Capabilities, like External Readers, could then be turned on/off by including/excluding the corresponding dependencies.

In my opinion, the direct stdout/stderr (as discussed on #759) is a bigger problem for `pkl-config-java` as a dependency. It's fine that it's there for developer ergonomics and should never happen in practical use cases, but code analysis tools will still pick it up and cause people to shy away from flagged dependencies.

pkl-cli/gradle.lockfile
```diff
@@ -74,10 +74,14 @@ org.jetbrains.kotlin:kotlin-scripting-common:1.7.10=kotlinCompilerPluginClasspat
...
+org.jetbrains.kotlin:kotlin-stdlib-common:1.8.22=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
+org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.7.10=apiDependenciesMetadata,implementationDependenciesMetadata,testImplementationDependenciesMetadata
+org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.8.22=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
+org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.7.10=apiDependenciesMetadata,implementationDependenciesMetadata,testImplementationDependenciesMetadata
+org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.8.22=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
+org.jetbrains.kotlin:kotlin-stdlib:1.7.10=apiDependenciesMetadata,implementationDependenciesMetadata,kotlinCompilerClasspath,kotlinCompilerPluginClasspathMain,kotlinCompilerPluginClasspathTest,kotlinKlibCommonizerClasspath,testImplementationDependenciesMetadata
+org.jetbrains.kotlin:kotlin-stdlib:1.8.22=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
```
holzensp, line 84: This should be avoidable; let's make sure we only have one Kotlin stdlib version in our dependencies.

holzensp (changes requested):
Is a dependency on only `msgpack-core` doable instead? Vendoring this seems selective; for other dependencies we either require our users to include them, or to use the shadowing FAT jars. I see that External Readers maybe relevant to `pkl-config-java` use-cases, unlike `pkl-server` (previously the only MsgPack dependent part), but I would strongly prefer an opt-in model. Could we not catch `ClassNotFoundException`s to see whether they're about MsgPack and adjust our errors accordingly?

holzensp (commented):
Should have looked ahead at #764: yes, it's possible. I prefer that approach.

translatenix:
@holzensp @bioball In case you missed my conclusion buried in this thread:

> Turns out that there is already a pkl-config-java-all fat JAR with shaded dependencies. So even a regular dependency on msgpack-core should be fine. Sorry for not noticing this sooner.

bioball:
Yeah, chatted offline with @holzensp a bit about this, and we're on the same page. Closing this!

PR #766, opened 2024-11-01
More external reader refinements

HT154:
* Rename `org.pkl.core.Readers` -> `org.pkl.core.Closeables`
* Remove coupling between `pkl.core.module`/`resource` and `pkl.core.messaging`


pkl-core/src/main/java/org/pkl/core/messaging/MessageTransportModuleResolver.java
HT154, line 1: Not in love with the naming here, but it's necessary to disambiguate and accurately describes the functionality. These classes are a straight move, though git doesn't show that.

translatenix:
I'd love to rein in the cyclic package dependencies between messaging, externalreader, resource, and module.
Is there time for me to investigate?

bioball:
> I'd love to rein in the cyclic package dependencies between messaging, externalreader, resource, and module.
Is there time for me to investigate before the 0.27 release?

We are planning on doing a release on Monday. When you think you can get a PR in for this?

translatenix:
> We are planning on doing a release on Monday. When you think you can get a PR in for this?

I can send a PR (or give up) by tonight. Probably makes sense to first merge this PR.

pkl-core/src/main/java/org/pkl/core/messaging/MessageTransportModuleResolver.java
```diff
@@ -0,0 +1,129 @@
...
+import org.pkl.core.SecurityManager;
+import org.pkl.core.SecurityManagerException;
+import org.pkl.core.messaging.Messages.ListModulesRequest;
+import org.pkl.core.messaging.Messages.ListModulesResponse;
+import org.pkl.core.messaging.Messages.ReadModuleRequest;
+import org.pkl.core.messaging.Messages.ReadModuleResponse;
+import org.pkl.core.module.ExternalModuleResolver;
+import org.pkl.core.module.PathElement;
+
+public class MessageTransportModuleResolver implements ExternalModuleResolver {
```
bioball, line 36: I don't think this needed to be moved? Package `externalreader` was not exporting `MessageTransport`.

HT154: I think I get it? I'm still tripping over how Java does things.

The idea here is to ensure that nothing in `org.pkl.core.messaging` ends up in the `org.pkl.core.module`/`resource` public API, right? Does that include the constructor for the resolver classes? Or am I missing something here?

bioball: Yeah; none of signatures of the public methods (including constructors) were using types from `org.pkl.core.messaging `, unless I missed something.

It's totally fine for private methods use to `org.pkl.core.messaging`.

bioball (approved):
LGTM!

pkl-core/src/main/java/org/pkl/core/messaging/MessageTransportModuleResolver.java
HT154: The constructors for the resolver classes take a `MessageTransport` so I these do need to move?

bioball: Ah! My bad!

I just merged this PR. Can you re-submit the other change as a new PR?

bioball:
> I can send a PR (or give up) by tonight. Probably makes sense to first merge this PR.

Sounds good!

PR #778, opened 2024-11-04
Eagerly check function arguments when called from inside iterable

bioball:
This mitigates an issue where lazy mappings and listings widen an existing bug.

This is a follow-up to https://github.com/apple/pkl/pull/752.

Ideally, when we fix the scoping bug, we also revert this commit.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/ast/expression/member/InvokeMethodVirtualNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/expression/member/InvokeMethodVirtualNode.java
@@ -36,6 +36,7 @@ import org.pkl.core.runtime.VmClass;
 import org.pkl.core.runtime.VmFunction;
 
 /** A virtual method call. */
+@SuppressWarnings("DuplicatedCode")
 @ImportStatic(Identifier.class)
 @NodeChild(value = "receiverNode", type = ExpressionNode.class)
 @NodeChild(value = "receiverClassNode", type = GetClassNode.class, executeWith = "receiverNode")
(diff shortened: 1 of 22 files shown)


pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -94,6 +94,14 @@ public Object executeEagerly(VirtualFrame frame, Object value) {
     return execute(frame, value);
   }
 
+  /**
+   * Checks if {@code value} conforms to this type.
+   *
+   * <p>If {@code value} is conforming, sets {@code slot} to {@code value}. Otherwise, throws a
+   * {@link VmTypeMismatchException}.
+   */
+  public abstract Object executeEagerlyAndSet(VirtualFrame frame, Object value);
```
bioball, line 103: This method is used by a function's parameter type node.

This node:
```
function foo(bar: Listing<Int>) =
                  ^^^^^^^^^^^^
```

We need to add the method because we need to capture the logic of: 
1. Run an eager typecheck
2. Set that value on the frame.

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InvokeMethodVirtualNode.java
```diff
@@ -36,6 +36,7 @@
 import org.pkl.core.runtime.VmFunction;
 
 /** A virtual method call. */
+@SuppressWarnings("DuplicatedCode")
```
holzensp, line 39: What does this overlap with?

holzensp (approved):
Sad but necessary, I guess :(

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InvokeMethodVirtualNode.java
bioball: The various specializations here. Adding just one more line to each method triggered the duplicate code detection warning.

pkl-core/src/main/java/org/pkl/core/ast/member/FunctionNode.java
```diff
@@ -44,7 +44,7 @@ public final class FunctionNode extends RegularMemberNode {
   // For VmObject receivers, the owner is the same as or an ancestor of the receiver.
   // For other receivers, the owner is the prototype of the receiver's class.
   // The chain of enclosing owners forms a function/property's lexical scope.
-  private static final int IMPLICIT_PARAM_COUNT = 2;
+  private static final int IMPLICIT_PARAM_COUNT = 3;
```
stackoverflow, line 47: We should update the docs which still state there are "two implicit arguments".

stackoverflow (approved):
Looks good. We can improve it once we have parse time variables.

PR #779, opened 2024-11-05
Polish ANSI formatting and underlying code

translatenix:
- change line numbers from blue to faint to improve legibility on dark backgrounds
- use EnumSet throughout
- move all ANSI classes into package core.util (no need to split them over util and runtime)
- rename AnsiCodingStringBuilder to AnsiStringBuilder

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/util/AnsiTheme.java
+++ b/pkl-core/src/main/java/org/pkl/core/util/AnsiTheme.java
@@ -17,7 +17,7 @@ package org.pkl.core.util;
 
 import java.util.EnumSet;
 import java.util.Set;
-import org.pkl.core.runtime.AnsiCodingStringBuilder.AnsiCode;
+import org.pkl.core.util.AnsiStringBuilder.AnsiCode;
 
 public final class AnsiTheme {
   private AnsiTheme() {}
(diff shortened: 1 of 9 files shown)


pkl-core/src/main/java/org/pkl/core/util/AnsiTheme.java
```diff
@@ -28,7 +28,7 @@ private AnsiTheme() {}
 
   public static final AnsiCode STACK_FRAME = AnsiCode.FAINT;
   public static final AnsiCode STACK_TRACE_MARGIN = AnsiCode.YELLOW;
-  public static final AnsiCode STACK_TRACE_LINE_NUMBER = AnsiCode.BLUE;
+  public static final AnsiCode STACK_TRACE_LINE_NUMBER = AnsiCode.BRIGHT_BLUE;
```
bioball, line 31: How about `AnsiCode.FAINT`? The bright blue draws attention to the line number, but it's arguably one of the less important parts of the error message.

Screenshot 2024-11-04 at 4 31 06 PM

translatenix: Looks even better (updated).

By the way, is there a reason why I can't set colors in `settings.pkl`?

bioball: I haven't put too much thought into it so far. I don't see any reason why it shouldn't be there, though.

bioball (approved):
LGTM; looking for one more thumb from either @holzensp or @stackoverflow

stackoverflow (approved):
LGTM!

PR #784, opened 2024-11-05
Fix broken link to documentation site in release notes

GUI:
The "number literals" link on the [0.27 release notes](https://pkl-lang.org/main/current/release-notes/0.27.html) is a broken link pointing to a local `file://` URL. I think this should fix it to point to the live docs site (similar to how other doc links are setup).


bioball (approved):
Thanks!

PR #786, opened 2024-11-06
Fix a possible deadlock during external reader process close

HT154:


pkl-core/src/main/java/org/pkl/core/externalreader/ExternalReaderProcessImpl.java
```diff
@@ -157,6 +157,8 @@ public void close() {
         return;
       }
 
+      var forceDestroyTimer = new Timer();
```
translatenix, line 160: A timer shouldn't be necessary here. I'd expect something like the following to work.
The `!= null` checks are to guard against a process/transport that wasn't initialized.
`CLOSE_TIMEOUT` was changed to `java.time.Duration`.
```
  @Override
  public void close() {
    synchronized (lock) {
      if (closed) return;
      closed = true;
      try {
        if (transport != null && process != null && process.isAlive()) {
          transport.send(new CloseExternalProcess());
          process.waitFor(CLOSE_TIMEOUT.toMillis(), TimeUnit.MILLISECONDS);
        }
      } catch (Exception ignored) {
      } finally {
        if (process != null) {
          process.destroyForcibly();
        }
        if (transport != null) {
          transport.close();
        }
      }
    }
  }
```

HT154: Nice! Really appreciate your feedback on these. It's especially valuable in places like this where my unfamiliarity with the java stdlib really show.

holzensp (approved):
LGTM. Thanks @HT154 & @translatenix!

PR #792, opened 2024-11-08
codegen-java: Support not annotating constructor parameters

translatenix:
Motivation:
Spring Boot configuration classes neither require nor benefit from annotating constructor parameters with their name. The same is true for pkl-config-java configuration classes compiled with `-parameter`.

Changes:
- add a `noParamsAnnotation` parameter to CLI, Gradle plugin, CliJavaCodeGeneratorOptions, and JavaCodegenOptions
- default to annotating constructor parameters unless `generateSpringBootConfig` is set
- fail if both `paramsAnnotation` and `noParamsAnnotation` are set
- add tests
- update docs

Result:
Generated code does not contain unnecessary annotations.


docs/modules/pkl-gradle/pages/index.adoc
```diff
@@ -373,14 +373,32 @@ Example: `generateGetters = true` +
...
 ====
-Default: `false` +
-Example: `preferJavaxInjectAnnotation = true` +
-Whether to annotate constructor parameters with `@javax.inject.Named` instead of `@org.pkl.config.java.mapper.Named`.
-If `true`, the generated code will have a compile dependency on `javax.inject:javax.inject:1`.
+Default: `"org.pkl.config.java.mapper.Named"` +
+Example: `paramsAnnotation = "org.project.MyAnnotation"` +
+Fully qualified name of the annotation type to use for annotating constructor parameters with their name. +
+The specified type must have a `value` parameter of type `String` or the generated code may not compile. +
+This option cannot be combined with `noParamsAnnotation`.
```
bioball, line 383: Thanks for filling these in!

docs/modules/java-binding/pages/codegen.adoc
```diff
@@ -152,16 +152,26 @@ Flag that indicates to generate Javadoc based on doc comments for Pkl modules, c
...
+The specified type must have a `value` parameter of type `String` or the generated code may not compile. +
+This option cannot be combined with `--no-params-annotation`.
+====
+
+.--no-params-annotation
+[%collapsible]
+====
+Default: (flag not set unless `--generate-spring-boot` is set) +
+Flag that indicates not to annotate constructor parameters with their name. +
+This option cannot be combined with `--params-annotation`.
```
bioball, line 165: Would be good to add a blurb explaining how to use this.

```suggestion
Default: (flag not set unless `--generate-spring-boot` is set) +
Flag that indicates not to annotate constructor parameters with their name. +
This option cannot be combined with `--params-annotation`.

If `true`, requires that Java is compiled with `-parameters` flag, or that these classes are loaded by Spring Boot.
```

bioball (approved):
LGTM; makes sense!

Looking for at least one more thumb from @stackoverflow or @holzensp.

translatenix:
Today I discovered a reason *not* to generate `@NonNull` annotations, which begs the question if `--no-` flags/APIs are the right solution.

My first attempt was to change `paramsAnnotation = null` to mean "no annotation" and set the default `paramsAnnotation = "org.pkl.config.java.mapper.Named"`. However, I discovered a few downsides of this approach:

* the default logic has to be duplicated in every layer (CLI/Gradle/CliGenerator/Generator)
* `--params-annotation=null` isn't great (perhaps `--params-annotation=` would work?)
* explicitly setting `paramsAnnotation = null` changes meaning (breaking change)

But perhaps this is nevertheless a better approach? In my opinion, we should support disabling both `--params-annotation` and `--non-null-annotation`, and I'm hoping to add `--nullable-annotation`.

bioball:
What is the reason to not generate `@NonNull`?

translatenix:
I’d like to support `@Nullable`, and I think many users will want one or the other.

bioball:
Gotcha. The reason we went with `@NonNull` instead of `@Nullable` is because we only need to use one annotation, rather than two (`@Nullable` implies a `@NonNullByDefault` type of annotation too).

I'd be happy to accept a PR for this, though.

stackoverflow (approved):
Looks good!

translatenix:
After doing more research, I've come to the conclusion that disabling annotations with `--params-annotation=none` and `paramsAnnotation = null` is a better approach. I believe the same approach should be used for `--non-null-annotation` and a future `--nullable-annotation`. Please have another look. See commit message for details.

bioball (approved):
LGTM

PR #793, opened 2024-11-08
codegen-kotlin: Generate toString() methods consistent with data classes

### LINKED ISSUE #717: codegen-kotlin: Use same toString() representation for data classes and regular classes
codegen-kotlin generates data classes whenever possible and regular classes otherwise.

For data classes, the default `toString()` implementation of data classes is used.
Example:
```
Person(name=Bob, address=Address(city=London))
`````

For regular classes, a `toString()` method is generated whose output somewhat resembles Pkl syntax.
Example:
```
Person {
  name = Bob
  address = Address {
    city = London
  }
}
```

It would be more consistent to use the same `toString()` representation for data classes and regular classes.
This could be achieved in one of the following ways:
1. Generate` toString()` methods for both data classes and regular classes.
2. Generate `toString()` methods for regular classes whose output matches that of data classes.

(1) enables a `toString()` representation that somewhat resembles Pkl syntax, whereas (2) generates much less code.

### PR DESCRIPTION
**Pull request by translatenix**
Note to reviewers: The second commit is a rename refactoring.

codegen-kotlin: Generate toString() methods consistent with data classes

Motivation:
codegen-kotlin generates a mix of data classes and regular classes.
For regular classes, toString() methods are also generated.
However, the output of generated toString() methods
differs from the output of default toString() methods of data classes.

Changes:
Generate toString() methods that produce the same output
as default toString() methods of data classes.

Result:
- Less and simpler code is generated.
- Object graphs containing a mix of data classes and regular classes are formatted consistently.
- Fixes #717.

### DIFF
--- a/pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
+++ b/pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
@@ -58,19 +58,6 @@ class KotlinCodeGenerator(
   private val options: KotlinCodegenOptions,
 ) {
   companion object {
-    // Prevent class name from being replaced with shaded name
-    // when pkl-codegen-kotlin is shaded and embedded in pkl-tools
-    // (requires circumventing kotlinc constant folding).
-    private val KOTLIN_TEXT_PACKAGE_NAME = buildString {
-      append("kot")
-      append("lin.")
-      append("text")
-    }
-
-    // `StringBuilder::class.asClassName()` generates "java.lang.StringBuilder",
-    // apparently because `StringBuilder` is an `expect class`.
-    private val STRING_BUILDER = ClassName(KOTLIN_TEXT_PACKAGE_NAME, "StringBuilder")
-
     private val STRING = String::class.asClassName()
     private val ANY_NULL = ANY.copy(nullable = true)
     private val NOTHING = Nothing::class.asClassName()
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt [hunk 1]
```diff
@@ -385,34 +345,25 @@ class KotlinCodeGenerator(
...
-          // to avoid making user code depend on kotlin-reflect
-          "builder.append(%T::class.java.simpleName).append(\" {\")",
-          kotlinPoetClassName
+          "return %P",
+          buildString {
+            append(pClass.simpleName)
+            append("(")
+            var isFirst = true
+            for (propertyName in allProperties.keys) {
+              if (isFirst) isFirst = false else append(", ")
```
bioball, line 358: [nit]: I think this is much easier to read:

```suggestion
              if (isFirst) {
                isFirst = false 
              } else {
                append(", ")
              }
```

pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
```diff
@@ -385,34 +345,25 @@ class KotlinCodeGenerator(
...
+              append(propertyName)
+              append("=$")
+              append(propertyName)
```
bioball, line 361: This needs to handle possibly quoted identifiers. Played around with it and this seems to do the trick:

```suggestion
              val propertyNameQuoted = CodeBlock.of("%N", propertyName).toString()
              if (propertyName == propertyNameQuoted) {
                append(propertyName)
              } else {
                append("{")
                append(propertyNameQuoted)
                append("}")
              }
```

bioball (changes requested):
Thanks! Just one issue; we should handle quoted identifiers (see comment). Also, it would be good to have a test for it.

pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
translatenix: Good catch. I ended up using a `CodeBlock` for the entire interpolated string (as recommended in docs).

pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
translatenix: I replaced `isFirst` (here and in one other location) with a nicer construct.

translatenix:
@bioball ready from my side

pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/KotlinCodeGenerator.kt
```diff
@@ -344,24 +338,25 @@ class KotlinCodeGenerator(
...
-            append("(")
-            var isFirst = true
-            for (propertyName in allProperties.keys) {
-              if (isFirst) isFirst = false else append(", ")
-              append(propertyName)
-              append("=$")
-              append(propertyName)
+          CodeBlock.builder()
+            .apply {
+              add("%L", pClass.toKotlinPoetName().simpleName)
```
translatenix, line 350: added `toKotlinPoetName()` here, which I think is necessary

bioball: Ah, yup; good catch

Issue #780, opened 2024-11-05
Listing with computed index has wrong length


translatenix:
The following program prints `length = 2`:

```
local l1 = new Listing {
  "foo"
}

local l2 = (l1) {
  [0 + 0] = "bar"
}

length = l2.length
```

If `[0 + 0]` is replaced with `[0]`, the program prints the expected `length = 1`.

bioball:
Good catch!

PR #797, opened 2024-11-10
Fix length of listings with computed index

translatenix:
Motivation:
The following expression evaluates to 2 instead of 1:
`new Listing { "value" } { [0 + 0] = "override" }.length`

Changes:
- fix length computation in EntriesLiteralNode
- improve `api/listing` tests
- make snippet test failures diffable in IntelliJ

Result:
- fixes https://github.com/apple/pkl/issues/780
- improved dev experience in IntelliJ


holzensp (approved):
Had just come across this, seeing that the _expected_ `length` was not always the one given to the `VmListing` and `VmDynamic` constructors. Thanks for the test coverage!

HT154:
Does this need to be backported to `release/0.27`?

PR #800, opened 2024-11-11
Update Gradle to 8.11

translatenix:
Changes:
- Update wrapper by running the following command: ./gradlew wrapper --gradle-version 8.11 --gradle-distribution-sha256-sum 57dafb5c2622c6cc08b993c85b7c06956a2f53536432a30ead46166dbca0f1e9
- Verify wrapper JAR integrity according to: https://docs.gradle.org/current/userguide/gradle_wrapper.html# manually_verifying_the_gradle_wrapper_jar
- Replace usages of deprecated method `Project.exec` with `ExecOperations.exec`
- Convert extension function `Task.configureInstallGraalVm` to task class `InstallGraalVm`
  - a task class is the cleanest way to get hold of `ExecOperations`
- Move extension property `BuildInfo.GraalVm.downloadFile` into class `GraalVm`
- Don't apply plugin `pklGraalVm` in project `bench` (unnecessary, now causes error)

### DIFF
--- a/buildSrc/src/main/kotlin/InstallGraalVm.kt
+++ b/buildSrc/src/main/kotlin/InstallGraalVm.kt
@@ -1,0 +1,85 @@
+/*
+ * Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import java.nio.file.Files
+import java.nio.file.Paths
+import java.nio.file.StandardCopyOption
+import java.util.*
+import javax.inject.Inject
+import kotlin.io.path.createDirectories
+import org.gradle.api.DefaultTask
+import org.gradle.api.internal.file.FileOperations
+import org.gradle.api.provider.Property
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/kotlin/InstallGraalVm.kt
```diff
@@ -0,0 +1,85 @@
...
+abstract class InstallGraalVm
+@Inject
+constructor(
+  private val fileOperations: FileOperations,
+  private val execOperations: ExecOperations
+) : DefaultTask() {
+  @get:Input abstract val graalVm: Property<BuildInfo.GraalVm>
+
+  init {
+    @Suppress("LeakingThis") onlyIf("GraalVM not installed") { !graalVm.get().installDir.exists() }
```
StefMa, line 38: Doesn't this destroy the laziness of the input? 🤔 Having get() in the init block?
In this vase we can also unwrap the GraalVm from the Property...

translatenix: > Doesn't this destroy the laziness of the input?

No, because the `onlyIf` closure will only be called at execution time.

I did look for a more appropriate API to use from a task class but couldn't find one. 
I tried throwing `StopExecutionException` from `run()`, which should be roughly the same as `onlyIf`.
This way I could include the install location in the message.
However, the downside was that the task no longer showed as "skipped".

If you know a better way, let me know!

buildSrc/src/main/kotlin/InstallGraalVm.kt
```diff
@@ -0,0 +1,85 @@
...
+        throw e
+      }
+    } catch (e: Exception) {
+      try {
+        fileOperations.delete(distroDir)
+      } catch (ignored: Exception) {}
+      throw e
+    }
+  }
+}
```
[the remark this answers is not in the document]
translatenix: As stated in the commit message, `project.exec` was [deprecated](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#exec(groovy.lang.Closure)) in Gradle 8.11 for removal in Gradle 9.

PR #802, opened 2024-11-12
codegen-java/kotlin: Fix generation of hashCode methods

translatenix:
codegen-java:
- use `pattern.pattern()` instead of `pattern` in hashCode method
  (consistent with equals method)

codegen-kotlin:
- use `regex.pattern` instead of `regex` in hashCode method
  (consistent with equals method)
- if a data class has a Regex property, generate not only
  an equals method but also a hashCode method


bioball (approved):
LGTM, thank you!

PR #810, opened 2024-11-14
codegen-kotlin: Deprecate method `toKotlinCodegenOptions` without replacement

translatenix:
Motivation:
`CliKotlinCodeGeneratorOptions.toKotlinCodegenOptions` is an internal method that isn't useful to clients.


bioball (approved):
LGTM, thanks for the follow up!

PR #819, opened 2024-11-14
Add `math.atan2`

gordonbondon:


pkl-core/src/test/files/LanguageSnippetTests/input/api/mathModule.pkl
```diff
@@ -130,6 +130,10 @@ examples {
     math.atan(math.tan(-2.34))
   }
 
+  ["atan2"] {
+    math.atan2(4, -3)
```
bioball, line 134: Let's add some more examples that exercise all permutations.

```suggestion
    math.atan2(4, -3)
    math.atan2(4.5, -3.5)
    math.atan2(4.5, -3)
    math.atan2(4.5, 3.5)
```


Also: maybe a test for correspondence with PI, in the `facts` section:

```pkl
facts {
  ["atan2"] {
    math.atan2(1, 0) == math.pi / 2
    math.atan2(0, -1) == math.pi
  }
}
```

stdlib/math.pkl
```diff
@@ -145,6 +145,11 @@ external function acos(x: Number): Float
 /// [1]: https://en.wikipedia.org/wiki/Inverse_trigonometric_functions
 external function atan(x: Number): Float
 
+/// Returns the [4-quadrant inverse tangent][1] of [x] and [y].
+///
+/// [1]: https://en.wikipedia.org/wiki/Atan2
```
bioball, line 150: Since we're linking to Wikipedia, let's use verbiage that aligns with it.

Also, need a `@Since` annotation.

```suggestion
/// Returns the [2-argument arctangent][1] of [x] and [y].
///
/// [1]: https://en.wikipedia.org/wiki/Atan2
@Since { version = "0.28.0" }
```

It might also be helpful to have one more paragraph to describe this function's purpose. Maybe something like: "The result is `θ` (theta) when converting cartesian coordinates [x] and [y] to polar coordinates `r` and `θ`."

bioball (changes requested):

holzensp (approved):

holzensp:
Needs a confirmation approval from @bioball (having requested changes); they look addressed to me.

bioball (approved):
LGTM!

Issue #791, opened 2024-11-08
pkl-gradle: cannot use pkldoc generator on package URIs


bioball:
This is a regression:

In the following:

```kt
pkl {
  pkldocGenerators {
    register("pkldoc") {
      sourceModules = listOf(uri("package://pkg.pkl-lang.org/pkl-pantry/pkl.toml@1.0.0"))
    }
  }
}
```

The implicit `<taskName>GetImports` task will attempt to analyze the imports of a package, which fails with:

```
* What went wrong:
Execution failed for task ':pkldocGatherImports'.
> –– Pkl Error ––
  Module URI `package://pkg.pkl-lang.org/pkl-pantry/pkl.toml@1.0.0` has invalid syntax.
```

The workaround for now is to set `transitiveModules` explicitly to something, which disables the `GetImports` task dependency.

```kt
pkl {
  pkldocGenerators {
    register("pkldoc") {
      sourceModules = listOf(uri("package://pkg.pkl-lang.org/pkl-pantry/pkl.toml@1.0.0"))
      transitiveModules = files("some_file.txt")
    }
  }
}
```

To fix this: we only need to analyze the imports of `file:` URIs. There is little chance that non-file URIs can import file URIs, because Pkl's default security manager disallows this.

PR #821, opened 2024-11-17
Exclude non file-based modules from synthesized *GatherImports task

bioball:
This fixes an issue where certain modules tasks fail due to the plugin attempting to analyze their imports, but the arguments may not actually be Pkl modules.

For example, the pkldoc task accepts entire packages in its "sourceMoules" property.

This changes the gather imports logic to only analyze file-based modules. This is also a performance improvement; non file-based modules are unlikely to import files due to insufficient trust levels.

Closes https://github.com/apple/pkl/issues/791

### DIFF
--- a/pkl-gradle/src/main/java/org/pkl/gradle/task/BasePklTask.java
+++ b/pkl-gradle/src/main/java/org/pkl/gradle/task/BasePklTask.java
@@ -17,9 +17,6 @@ package org.pkl.gradle.task;
 
 import java.io.File;
 import java.net.URI;
-import java.net.URISyntaxException;
-import java.net.URL;
-import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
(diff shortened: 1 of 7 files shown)


pkl-gradle/src/main/java/org/pkl/gradle/task/BasePklTask.java
```diff
@@ -199,100 +194,6 @@ protected List<Path> parseModulePath() {
...
-   */
-  private URI parsedModuleNotationToUri(Object notation) {
-    if (notation instanceof File file) {
-      return IoUtils.createUri(file.getPath());
-    } else if (notation instanceof URI uri) {
-      return uri;
-    }
-    throw new IllegalArgumentException("Invalid parsed module notation: " + notation);
-  }
-
```
bioball, line 295: These methods got moved to `PluginUtils.java`, we are now using them in `PklPlugin.java` too.

stackoverflow (approved):
LGTM

Issue #785, opened 2024-11-05
Semantics of Listing typechecks have changed in 0.27


odenix:
The following program fails in 0.26.3 but passes in 0.27.0.
(My understanding is that the goal was to preserve semantics of Listing typechecks.)

```
local a = new Listing { new Listing { 0 } }
// 0.26.3 fails here, 0.27.0 doesn't
local b = a as Listing<Listing<String>> 
local c = (b) { new Listing { 1 } }
local d = c as Listing<Listing<Int>>

result = d
```

bioball:
Ah, I think this is a regression in our typechecking logic. It shouldn't be possible to amend `b` and get members of `Listing<Int>`.

PR #822, opened 2024-11-17
Fix: check computed members against listing/mapping owner

bioball:
This addresses a regression where a value can possibly not be type-casted correctly.

Closes https://github.com/apple/pkl/issues/785


translatenix:
Only executing `owner`'s and `receiver`'s type cast isn't sufficient. New test:

```
local a = new Listing { new Foo {} }
local b = (a) { new Bar {} }
local c = b as Listing<Bar>
local d = (c) { new Foo {} }
local e = d as Listing<Foo>

result = e

open class Foo
class Bar extends Foo
```

I think the best fix is to switch to my implementation. :-)

bioball:
Ah, yup, you're right! I see what you mean now by it being insufficient to just check the owner and receiver.

I'll take a closer look at your PR.

PR #824, opened 2024-11-18
Added support for an alternative current dir mode in pkldoc

netvl:
Some systems, for example, GitHub Pages, have trouble with handling symlinks, which breaks the `current` directory links created by Pkldoc. In this PR, we add an alternative mode which creates a full copy of the latest published version contents in the `current` directory instead.

### DIFF
--- a/docs/modules/pkl-doc/pages/index.adoc
+++ b/docs/modules/pkl-doc/pages/index.adoc
@@ -232,8 +232,19 @@ Relative URIs are resolved against the working directory.
 [%collapsible]
 ====
 Default: (none) +
-Example: `pkldoc`
+Example: `pkldoc` +
 The directory where generated documentation is placed.
+====
+
+.--no-symlinks
+[%collapsible]
+====
+Create copies of files and directories instead of symbolic links.
+Affects, in particular, how the "current" directories containing documentation content
+for the last generated version should be created. By default, a symbolic link is created
+pointing to the last generated version. If symlinks are disabled, a full copy
+of the last generated version is created. Symbolic links are not processed correctly
+by some systems (e.g. GitHub Pages), so the disabling them is occasionally useful.
 ====
 
 Common CLI options:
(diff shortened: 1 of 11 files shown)


docs/modules/pkl-doc/pages/index.adoc
```diff
@@ -232,10 +232,22 @@ Relative URIs are resolved against the working directory.
 [%collapsible]
 ====
 Default: (none) +
-Example: `pkldoc`
+Example: `pkldoc` +
 The directory where generated documentation is placed.
 ====
 
+.--current-directory-mode
```
translatenix, line 239: Unless you foresee a third mode (I can’t think of one), a flag would be easier to use and understand.

netvl: Yes, I think that it might be possible to eventually have a mode to disable `current` generation at all.

Also, I'm not very sure about a binary flag to be easier to understand to be honest. It is not disabling or enabling something here, it is changing the logic of operation between two states. Enums are supposed to be the tool for this ([one of discussions](https://stackoverflow.com/questions/4337942/using-enum-vs-boolean)).

bioball: I agree with @translatenix here; I think this should be called `--no-symlinks`, and default to false. It's not necessarily about how the "current" dir is treated; the intention is to disable symlinks.

bioball (changes requested):
Thanks for the PR!

Per the [comments](https://github.com/apple/pkl/pull/824#discussion_r1847457065), can we change this to a boolean flag? I think a name like `--no-symlinks` would be sensible here.

docs/modules/pkl-doc/pages/index.adoc
odenix: That's a good point that I had failed to notice.

netvl:
@odenix @bioball fixed according to review comments, sorry it took so long!

bioball (approved):
LGTM, thanks!

PR #830, opened 2024-11-27
fix regression where typealiases are not executed in their original context

### LINKED ISSUE #823: Regression: incorrect scope in T of typealias of Listing<T>/Mapping<K, T>
Given:

```pkl
// mod1.pkl
const local lastName = "Smith"

typealias Smiths = Listing<String(endsWith(lastName))>
```

This breaks:

```pkl
import "mod1.pkl"

res: mod1.Smiths = new { "Lily Smith" }
```

Error:

```
–– Pkl Error ––
Cannot find property `lastName`.

3 | typealias Smiths = Listing<String(endsWith(lastName))>
                                               ^^^^^^^^
```

stackoverflow:
Closes #823

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/ast/member/ListingOrMappingTypeCastNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/member/ListingOrMappingTypeCastNode.java
@@ -36,25 +36,21 @@ public class ListingOrMappingTypeCastNode extends PklRootNode {
   private final @Nullable Object receiver;
 
   public ListingOrMappingTypeCastNode(
-      VmLanguage language, FrameDescriptor descriptor, TypeNode typeNode, String qualifiedName) {
-    super(language, descriptor);
-    this.typeNode = typeNode;
-    this.qualifiedName = qualifiedName;
-    owner = null;
-    receiver = null;
-  }
-
-  public ListingOrMappingTypeCastNode(
       VmLanguage language,
       FrameDescriptor descriptor,
       TypeNode typeNode,
       String qualifiedName,
-      VirtualFrame frame) {
+      @Nullable VirtualFrame originalFrame) {
     super(language, descriptor);
     this.typeNode = typeNode;
     this.qualifiedName = qualifiedName;
-    owner = VmUtils.getOwner(frame);
-    receiver = VmUtils.getReceiver(frame);
+    if (originalFrame != null) {
+      owner = VmUtils.getOwner(originalFrame);
+      receiver = VmUtils.getReceiver(originalFrame);
+    } else {
+      owner = null;
+      receiver = null;
+    }
   }
 
   public TypeNode getTypeNode() {
--- a/pkl-core/src/main/java/org/pkl/core/ast/member/ListingOrMappingTypeCastNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/member/ListingOrMappingTypeCastNode.java
@@ -23,6 +23,8 @@ import org.pkl.core.ast.PklRootNode;
 import org.pkl.core.ast.type.TypeNode;
 import org.pkl.core.ast.type.VmTypeMismatchException;
 import org.pkl.core.runtime.VmLanguage;
+import org.pkl.core.runtime.VmObjectLike;
+import org.pkl.core.runtime.VmUtils;
 import org.pkl.core.util.Nullable;
 
 /** Performs a typecast on a Mapping entry value, or a Listing element. */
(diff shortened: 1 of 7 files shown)


pkl-core/src/main/java/org/pkl/core/ast/member/ListingOrMappingTypeCastNode.java
```diff
@@ -54,6 +73,20 @@ public SourceSection getSourceSection() {
...
+      VmUtils.setOwner(frame, owner);
+      VmUtils.setReceiver(frame, receiver);
+      return executeInternal(frame);
+    } finally {
+      VmUtils.setOwner(frame, prevOwner);
+      VmUtils.setReceiver(frame, prevReceiver);
+    }
+  }
+
+  private Object executeInternal(VirtualFrame frame) {
```
holzensp, line 89: mild preference to just inline, myself
```suggestion
      return typeNode.execute(frame, frame.getArguments()[2]);
    } catch (VmTypeMismatchException e) {
      CompilerDirectives.transferToInterpreter();
      throw e.toVmException();
    } finally {
      VmUtils.setOwner(frame, prevOwner);
      VmUtils.setReceiver(frame, prevReceiver);
    }
  }
```

(GH suggestion can't span out of the diff, alas)

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -1342,6 +1373,14 @@ public Object executeEagerly(VirtualFrame frame, Object value) {
       throw typeMismatch(value, BaseModule.getMapClass());
     }
 
+    @Override
+    public Object executeInDefinitionContext(VirtualFrame frame, Object value) {
+      if (value instanceof VmMap vmMap) {
+        return evalInDefinitionContext(frame, vmMap);
+      }
+      throw typeMismatch(value, BaseModule.getMapClass());
```
holzensp, line 1381: For consistency and to save on single use (albeit `private`) helpers; why not follow the same pattern as in `VmList`?

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -1427,7 +1483,15 @@ public Object execute(VirtualFrame frame, Object value) {
...
-      return doTypeCast(frame, vmListing);
+      return doTypeCast(frame, vmListing, false);
+    }
+
+    @Override
+    public Object executeInDefinitionContext(VirtualFrame frame, Object value) {
+      if (!(value instanceof VmListing vmListing)) {
+        throw typeMismatch(value, BaseModule.getListingClass());
+      }
+      return doTypeCast(frame, vmListing, true);
```
holzensp, line 1494: ^^ same re:`VmList`

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -1560,15 +1634,26 @@ public TypeNode getValueTypeNode() {
...
+                  valueTypeNode,
+                  getRootNode().getName(),
+                  frame);
+        } else {
+          listingOrMappingTypeCastNode =
+              new ListingOrMappingTypeCastNode(
+                  VmLanguage.get(this),
+                  getRootNode().getFrameDescriptor(),
+                  valueTypeNode,
+                  getRootNode().getName());
```
holzensp, line 1655: It's this kind of repetition that would almost make me suggest to just make passing a `null` part of the API of `ListingOrMappingTypeCastNode`

holzensp (approved):
Solves a problem, but less than ideally pretty (I think we agree). We may want to do a clean-up pass on how we do frame management.

pkl-core/src/main/java/org/pkl/core/ast/member/ListingOrMappingTypeCastNode.java
stackoverflow: Inlining would make it slower, right? I'd always have to get and reset the owner/receiver in the frame, while right now it's short-circuiting for the 99% case where it doesn't come from a typealias.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
stackoverflow: I did this for consistency with `execute` and `executeEager`. I don't think it makes sense to only inline it here. Either we remove `eval` and `evalEager` or we leave all three.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
stackoverflow: Improved.

pkl-core/src/main/java/org/pkl/core/ast/member/ListingOrMappingTypeCastNode.java
holzensp: You are entirely right... I had missed that top line somehow

odenix:
I implemented a simpler fix that passes your tests for my "lazy-check" branch: https://github.com/apple/pkl/pull/789/commits/bc10eba3b27336caba257333e928da1a3c6d8c81

PS: As far as I know, having multiple execute methods with the same signature isn't a common Truffle idiom and can cause problems with `@Specialization`. Instead of having `TypeNode.execute` and `TypeNode.executeEagerly`, it might be better to have `ListingTypeNode` and `EagerListingTypeNode` (same for mappings).

odenix:
It would be great if we could prioritize https://github.com/apple/pkl/pull/789 to avoid parallel development in this part of the codebase.

stackoverflow:
@odenix I'll review you PR and then, hopefully, we can close this one.

PR #831, opened 2024-11-27
Delay instantiation of `cachedValues` to save memory

holzensp:
I'm putting this up as a discussion piece, more than anything. Wins are minimal, at best.

The thinking was that every `VmObject` gets its `cachedValues` initialised by default, but only the ones that are forced get those values filled. However, `cachedValues.entries` is `null` by default. There appears to be a bigger possible win to address the fact that a lot of `EconomicMap`s have `entries` be an oversized array (`Object[8]` by default).

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmDynamic.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmDynamic.java
@@ -75,8 +75,7 @@ public final class VmDynamic extends VmObject {
   @Override
   @TruffleBoundary
   public PObject export() {
-    var properties =
-        CollectionUtils.<String, Object>newLinkedHashMap(EconomicMaps.size(cachedValues));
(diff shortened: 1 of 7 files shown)


translatenix:
If you're looking for small wins, there are several places in the codebase where `EconomicMap.create()` could be replaced with `EconomicMap.emptyMap()` because entries will never be added.

holzensp:
I agree. I _am_ still looking into replacing `members` with something a little more compact, because of frequent over-allocation there. I'm skeptical it'll be significant, but it's worth the experiment, at least.

odenix:
Note that `members` is often shared between objects created by the same new/amend expression, so you're often optimizing AST memory usage here. Objects whose members are created on the fly should probably use a different representation altogether.

To reduce AST memory usage, I'd take a look at `SourceSection`. I believe GraalJS, GraalPY, etc. use fewer `SourceSection`s and also instantiate them on demand to reduce memory usage. I don't know how much this matters at Pkl scale.

PR #836, opened 2024-12-02
Convert `org.pkl.executor.Version.Identifier` POJO to `record` class

KushalP:


pkl-executor/src/main/java/org/pkl/executor/EmbeddedExecutor.java
```diff
@@ -322,21 +322,14 @@ static URL[] toUrls(Path pklFatJar) {
     }
   }
 
-  private static final class ConcatenatedEnumeration<E> implements Enumeration<E> {
-    final Enumeration<E> e1;
-    final Enumeration<E> e2;
-
+  private record ConcatenatedEnumeration<E>(Enumeration<E> e1, Enumeration<E> e2)
```
odenix, line 325: imho not a great candidate for record (not really a pojo that exposes its state)

bioball (approved):
LGTM, thanks!

odenix (approved):

PR #837, opened 2024-12-02
Fix exception handling of PklRootNode's

odenix:
Motivation:
- Perform same exception handling for every implementation of PklRootNode.execute().
- Avoid code duplication.

Changes:
- Change PklRootNode.execute() to be a final method that performs exception handling and calls abstract method executeImpl(), which is implemented by subclasses.
- Remove executeBody() methods, which served a similar purpose but were more limited.
- Remove duplicate exception handling code.

Result:
- More reliable exception handling. This should fix known problems such as misclassifying stack overflows as internal errors and displaying errors without a stack trace.
- Less code duplication.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/ast/MemberNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/MemberNode.java
@@ -16,7 +16,6 @@
 package org.pkl.core.ast;
 
 import com.oracle.truffle.api.frame.FrameDescriptor;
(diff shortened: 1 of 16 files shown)


holzensp (approved):
*Yes* please! Thanks.

One more 👍 from either @bioball or @stackoverflow and we're good.

stackoverflow (approved):
This is a welcome simplification. Thanks!

bioball (approved):
Good improvement, thanks!

PR #851, opened 2024-12-18
Download JDK for windows build

bioball:
Don't use the system Java on Windows builds, instead download them from Adoptium.

Also:

* Fail job if curl returns a 4xx status code
* Add java version to `GradleJob`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -742,7 +742,10 @@ workflows:
         type: approval
     - pr-approval/authenticate:
         context: pkl-pr-approval
-    - pkl-cli-windows-amd64-snapshot:
+    - gradle-check-jdk17:
+        requires:
+        - hold
+    - gradle-check-jdk21:
         requires:
         - hold
     - gradle-check-jdk17-windows:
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -39,8 +39,8 @@ jobs:
             && rm -rf /var/cache/dnf
 
           # install jdk
-          curl -L \
-           https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.9%2B9/OpenJDK17U-jdk_x64_linux_hotspot_17.0.9_9.tar.gz -o /tmp/jdk.tar.gz
+          curl -Lf \
+           https://github.com/adoptium/temurin17-binaries/releases/download/jdk-jdk-17.0.9%2B9.1/OpenJDK17U-jdk_x64_linux_hotspot_17.0.9_9.tar.gz -o /tmp/jdk.tar.gz
```
stackoverflow, line 43: These new URLs are returning 404 for me.

bioball: Good catch, fixed

PR #868, opened 2024-12-24
Make Test Report locale independent 

### LINKED ISSUE #861: Running `test` fails
Hey guys,
greetings from Germany. We typically write numbers with `,` instead of `.` 🤓 
✅  1,20 Euro 
❌  1.20 Euro.

Why is this interesting? Cause right now I get errors when running
```
./gradlew test
```

```
expected: 
  "module test
    facts
      ✘ fail
         4 == 9 (/tempDir/test.pkl, line xx)
  
  0.0% tests pass [1/1 failed], 50.0% asserts pass [1/2 failed]
  "
 but was: 
  "module test
    facts
      ✘ fail
         4 == 9 (/tempDir/test.pkl, line xx)
  
  0,0% tests pass [1/1 failed], 50,0% asserts pass [1/2 failed]
  "
```

Right now these tests are failing:
```
CliTestRunner fail test(Path)
CliTestRunner succeed test(Path)
CliTestRunner with thrown error in examples -- existing expected output(Path)
CliTestRunner with thrown error in examples -- no expected output(Path)
CliTestRunner with thrown error in facts(Path)
example length mismatch(Path)
```

StefMa:
fixes #861

According to AI (and javadoc 😅) the problem is that String.format takes Locale.getDefault if not specified. Indeed I have set de-EN on my machine so it takes comma decimals instead of dot decimals. 

~I haven't tested this yet, just created it on mobile...~

This is my macobok setting (not sure if relevant or the correct screen):
Screenshot 2024-12-24 at 7 05 15 AM

If I run the newly added test without `Locale.ROOT` in `SimpleReport`, I get the following error:
Screenshot 2024-12-24 at 7 06 36 AM

Running with that `Locale.ROOT`:
`BUILD SUCCESSFUL in 37s` 🎉 

---

AI summary 

This pull request includes changes to improve locale independence in the `CliTestRunner` and to ensure consistent locale usage in the `SimpleReport` class. The most important changes include adding a locale independence test for `CliTestRunner` and specifying the root locale for formatting in `SimpleReport`.

Locale independence improvements:

* [`pkl-cli/src/test/kotlin/org/pkl/cli/CliTestRunnerTest.kt`](diffhunk://#diff-37144b29baadf10353a60ee1a542ada8b1988d561989383347319eb2fbce7c06R490-R530): Added a new test `CliTestRunner locale independence test` to verify that `CliTestRunner` works correctly regardless of the default locale.
* [`pkl-core/src/main/java/org/pkl/core/stdlib/test/report/SimpleReport.java`](diffhunk://#diff-66bf95f097f7f82c74ddfb08bd576eca3353c82e16a1cd1fa905562152f92c1eR20): Imported `java.util.Locale` to use for specifying locale in formatting.
* [`pkl-core/src/main/java/org/pkl/core/stdlib/test/report/SimpleReport.java`](diffhunk://#diff-66bf95f097f7f82c74ddfb08bd576eca3353c82e16a1cd1fa905562152f92c1eL147-R148): Modified the `makeStatsLine` method to use `Locale.ROOT` in `String.format` to ensure consistent formatting regardless of the default locale.

### DIFF
--- a/pkl-cli/src/test/kotlin/org/pkl/cli/CliTestRunnerTest.kt
+++ b/pkl-cli/src/test/kotlin/org/pkl/cli/CliTestRunnerTest.kt
@@ -35,6 +35,7 @@ import org.pkl.commons.readString
 import org.pkl.commons.toUri
 import org.pkl.commons.writeString
 import org.pkl.core.Release
+import java.util.Locale
 
 class CliTestRunnerTest {
   @Test
@@ -512,8 +513,8 @@ class CliTestRunnerTest {
     runner.run()
 
     assertThat(out.toString().stripFileAndLines(tempDir))
-        .isEqualTo(
-            """
+      .isEqualTo(
+        """
             module test
               facts
                 ✔ localeTest
@@ -521,12 +522,12 @@ class CliTestRunnerTest {
             100.0% tests pass [1 passed], 100.0% asserts pass [1 passed]
 
             """
-                .trimIndent()
-        )
+          .trimIndent()
+      )
     assertThat(err.toString()).isEqualTo("")
 
     Locale.setDefault(originalLocale)
-}
+  }
 
   private fun String.stripFileAndLines(tmpDir: Path): String {
     // handle platform differences in handling of file URIs
--- a/pkl-cli/src/test/kotlin/org/pkl/cli/CliTestRunnerTest.kt
+++ b/pkl-cli/src/test/kotlin/org/pkl/cli/CliTestRunnerTest.kt
@@ -487,6 +487,47 @@ class CliTestRunnerTest {
       )
   }
 
+  @Test
+  fun `CliTestRunner locale independence test`(@TempDir tempDir: Path) {
+    val originalLocale = Locale.getDefault()
+    Locale.setDefault(Locale.GERMANY)
+
+    val code =
+        """
+        amends "pkl:test"
+
+        facts {
+            ["localeTest"] {
+                1 == 1
+            }
+        }
+        """
+            .trimIndent()
+    val input = tempDir.resolve("test.pkl").writeString(code).toString()
+    val out = StringWriter()
+    val err = StringWriter()
+    val opts = CliBaseOptions(sourceModules = listOf(input.toUri()), settings = URI("pkl:settings"))
+    val testOpts = CliTestOptions()
+    val runner = CliTestRunner(opts, testOpts, consoleWriter = out, errWriter = err)
+    runner.run()
+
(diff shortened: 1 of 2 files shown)


StefMa:
@bioball I also updated the description with (maybe) hopefuly information 🙃

pkl-cli/src/test/kotlin/org/pkl/cli/CliTestRunnerTest.kt
```diff
@@ -487,6 +488,47 @@ class CliTestRunnerTest {
...
+                ✔ localeTest
+
+            100.0% tests pass [1 passed], 100.0% asserts pass [1 passed]
+
+            """
+          .trimIndent()
+      )
+    assertThat(err.toString()).isEqualTo("")
+
+    Locale.setDefault(originalLocale)
```
bioball, line 529: These assertions will throw errors, which means that `Locale.setDefault(...)` won't get called if any of these tests fail.

Let's put lines 496 through 527 in a `try/finally`, where `Locale.setDefault` is in the finally so we can ensure that this test cleans up the state that it's mucking with

bioball (changes requested):
Thanks! Left one comment.

Also, run `./gradlew spotlessApply` to fix formatting errors.

pkl-cli/src/test/kotlin/org/pkl/cli/CliTestRunnerTest.kt
odenix: Also note that mutating global JVM state is only safe as long as test execution is single-threaded (which is the default in Gradle).

StefMa: > These assertions will throw errors, which means that Locale.setDefault(...) won't get called if any of these tests fail.

In case the test fails, won't the jvm "shutdown" anyways (and report the error)?
So doesn't this even matter to restore the default in that case? 

> Also note that mutating global JVM state is only safe as long as test execution is single-threaded (which is the default in Gradle).

Do we have the default here? 😅 
When I running tests, it looks like Gradle runs the test parallel.
At least for different Gradle modules.
So I guess changing the Locale might also change the behviour on other modules/tests that runs at that time, correct? 🤔

odenix: By default, each `Test` task starts one JVM and runs tests in a single thread. I'm not aware of Pkl overriding this default. Multiple `Test` tasks may nevertheless run in parallel.

bioball (approved):
LGTM, thank you!

PR #871, opened 2025-01-02
Update license year

bioball:
This adds logic to update the license header year _only_ when source code files change.

For new files, they will receive the current year as the license year.

For existing files, they will receive a range from the initial year to the current year (e.g. 2024-2025 for the set of files currently in the repo).


buildSrc/src/main/kotlin/pklAllProjects.gradle.kts
```diff
@@ -132,7 +132,11 @@ private fun KotlinGradleExtension.configureFormatter() {
   licenseHeaderFile(licenseHeaderFile, "([a-zA-Z]|@file|//)")
 }
 
+val originalRemoteName = System.getenv("PKL_ORIGINAL_REMOTE_NAME") ?: "origin"
+
 spotless {
+  ratchetFrom = "$originalRemoteName/main"
```
bioball, line 138: I'd love to just say "the remote whose URL is git@github.com:apple/pkl.git", but it doesn't look like Spotless supports this.

bioball: Yeah, true!

stackoverflow (approved):

PR #879, opened 2025-01-10
`YamlRenderer`: allow all primitive scalar types as map keys

### LINKED ISSUE #878: Not possible to render mapping with Int keys in YAML
```
ll = new Mapping<Int, String>{
  [1] = ""
}
```

when trying to render it, I get an error:

```
–– Pkl Error ––
Cannot render object with non-string key as XML.
Object: new Mapping { [1] = "" }
Key   : 1
```

HT154:
[YAML (1.2) explicitly allows _any_ value as a mapping key](https://yaml.org/spec/1.2.2/#3211-nodes), just like Pkl! This PR allows all of Pkl's primitive scalars (those that can already be natively rendered in YAML: Int, Float, Boolean, Null) to be rendered as mapping keys by `YamlRenderer`.

This PR does not attempt to allow non-scalar YAML mapping keys (see Example 2.11 at https://yaml.org/spec/1.2.2/#22-structures). This doesn't appear to be a commonly used YAML feature—I've never seen it in the wild—but may be interesting for Pkl to support. If this is desired, it could be taken on as follow-up work, but it might require significant modifications to `YamlRendererNodes.YamlRenderer`.

Resolves #878

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/stdlib/base/YamlRendererNodes.java
+++ b/pkl-core/src/main/java/org/pkl/core/stdlib/base/YamlRendererNodes.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


pkl-core/src/main/java/org/pkl/core/stdlib/base/YamlRendererNodes.java
```diff
@@ -278,8 +278,24 @@ protected void visitEntryKey(Object key, boolean isFirst) {
...
-        emitter.emit(string, currIndent, true);
+      if (key instanceof String stringKey) {
+        emitter.emit(stringKey, currIndent, true);
+        builder.append(':');
+        return;
+      } else if (key instanceof Long longKey) {
+        emitter.emit(longKey);
+        builder.append(':');
+        return;
+      } else if (key instanceof Double doubleKey) {
```
StefMa, line 289: I am not sure with that, but what about Int(eger)? And why not just use Number instead?

StefMa (commented):
Just one comment 🤓
On mobile you can't just comment on PRs, it is always a review 😡

pkl-core/src/main/java/org/pkl/core/stdlib/base/YamlRendererNodes.java
HT154: Pkl's `Int` types are all represented by Java `long` internally (likewise, Pkl `Float` is always Java `double`). In this case, `emitter` has two method overloads `emit(long)` and `emit(double)` so having a single case for `Number` doesn't work here.

Issue #225, opened 2024-02-21
Cosinder adding dependabot for automatic dependency updates 


StefMa:
It's just a simple config file 🙃.
However I'm not sure if you want to deal with automatic dependency updates at this point.

holzensp:
We're waiting until we're fully on GraalVM23 (for `0.26.0-dev` we're now dropping JDK11 support) and Gradle 8.x, but then, yes, definitely want dependabot

translatenix:
I prefer renovate-bot (used by JUnit etc.), but either bot is a big step forward. Sending manual PRs with dependency updates is time-consuming and should be automated away as much as possible.

StefMa:
Dependabot can handle lock files now 🎉
https://github.blog/changelog/2025-06-24-dependabot-support-for-gradle-lockfiles-is-now-generally-available/

PR #885, opened 2025-01-15
Use GitHub Actions for checks

### LINKED ISSUE #152: Automatically run CI builds for external contributions
When an external contributor submits a PR, CI builds are on hold until manually authorized by a team member. This complicates external contributions as we won't know the build outcome for a long time. Is there a way to alleviate this pain?

StefMa:
Hello everyone ✌️ 

this PR might fix #152 and therefore also contributes to #225.

Two reasons why this would be a step forward:
* Forks can simply check in their own repo if something isn't broken. Right now they have to check stuff locally and we all know this might not happen all the time 😅 
* PRs, depending on the [setting](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories), can run automatically run checks without an approval.
* Since this is the "start" of using GitHub actions, we can also start implemeting #225 as dependabot would also run on actions and we can create workflows that "do something after they run" or after "dependabot created a PR"... See https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions
* ???

I just converted the `checks` from the circleci config for now because 
1. I wanted to hear your opinion about this first (generally)
2. What do you think about using my pkl-gha project 🙃 
3. Other comments 🙃 
Le me know if this is "fine" or if I also convert all the other jobs.
On the other side we can also do this as follow-ups in other PRs. 
Let me know what you prefer.

Because using a matrix is super easy to setup on actions, I also decided to run jdk 21 on windows 🤷 
This isn't in the circleci config but doesn't hurt i guess?! 

If you want to see how this looks like in action, checkout this run on my fork:
* [Check Pkl Actions Converted](https://github.com/StefMa/pkl/actions/runs/12783766972)
* [Checks](https://github.com/StefMa/pkl/actions/runs/12783766976)

### DIFF
--- a/.github/pkl-workflows/CheckActionsConverted.pkl
+++ b/.github/pkl-workflows/CheckActionsConverted.pkl
@@ -1,0 +1,37 @@
+amends "package://pkg.pkl-lang.org/github.com/stefma/pkl-gha/com.github.action@0.0.4#/GitHubAction.pkl"
+
+name = "Check Pkl Actions Converted"
+
+on {
+  push {}
+}
+
+jobs {
+  ["check-actions-converted"] {
+    name = "Check Actions converted"
+    `runs-on` = "ubuntu-latest"
+    steps {
+      new {
+        name = "Checkout"
+        uses = "actions/checkout@v4"
+      }
+      new {
+        name = "Install pkl"
+        uses = "pkl-community/setup-pkl@v0"
+        with {
+          ["pkl-version"] = "0.27.0"
+        }
+      }
+      new {
+        name = "Convert pkl actions to yaml"
+        run = """
+          pkl eval .github/pkl-workflows/*.pkl -o .github/workflows/%{moduleName}.generated.yml
+          """
+      }
+      new {
+        name = "Verify if pkl actions are converted"
+        run = "git diff --exit-code"
+      }
+    }
+  }
+}
--- a/.github/pkl-workflows/Checks.pkl
+++ b/.github/pkl-workflows/Checks.pkl
@@ -3,11 +3,7 @@ amends "package://pkg.pkl-lang.org/github.com/stefma/pkl-gha/com.github.action@0
 name = "Checks"
 
 on {
-  push {
-    branches {
-      "main"
-    }
-  }
+  push {}
   pull_request {}
 }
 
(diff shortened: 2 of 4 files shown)


bioball:
Hi @StefMa, thanks for the PR!

We're considering switching to GitHub Actions, but we're not quite ready yet to accept a commit for this. However, you can always maintain this commit on your own fork for now as a way to build and test Pkl yourself.

However, this is quite useful for us (I've never used GHA before), so, let's keep this PR around so we can go from here when we're ready!

madrob:
@bioball Actions are currently disabled on the repo, but would it be useful to merge anyway so that other folks who are forking can run the tests? I wouldn't want this to turn into a maintenance burden for the team to have untested code in the repo, but it could also be a nice quality of life improvement for contributors.

sgammon:
In case it helps forks in the meantime, [here](https://github.com/elide-dev/pkl/tree/chore/build-upgrades/.github/workflows) are some workflows that I've used to test PRs before submitting them to Pkl upstream.

The [build summary](https://github.com/elide-dev/pkl/actions/runs/13022097491?pr=6) shows how jobs are arranged, with Gradle and Native Image build reports, and so on. The workflows use [StepSecurity's actions](https://github.com/step-security/harden-runner) to harden execution (disables `sudo` access and offers firewalling to protect from supply chain attacks). Actions use principle-of-least-privilege and pin to specific commits for each step.

I'd be glad to offer these upstream if the project is interested, or on this PR, but I suspect the Pkl project authors will want to render these configs with Pkl, as is done for GHA

StefMa:
> let's keep this PR around so we can go from here when we're ready!

Sure, take your time. This is going to nowhere 🙃 

> Also, I'm open to using pkl-gha :)

Nice to hear! If you have any feedback to it, let me know.

---

I tend  to agree with @madrob.
You can simply disable Actions in the [repo settings](https://github.com/apple/pkl/settings/actions).
Screenshot 2025-01-29 at 9 01 19 AM
Could be quite easy enabled and improved by the community over time, while you still exploring GH Actions.. 🤔 
On the other side, it of course require a bit of maintainance...

---

@sgammon nice, thanks for sharing that!
I initially started with just PRs to give the maintainers here an first glimps/overview about this and gather feedback.
With that approach we can move slowly away from CircleCI to GH Actions and don't put anything at once here and addiotnally some other "nice|good-to-have" actions. Lets start with the pure basics and we can iterate from here.

PR #892, opened 2025-01-20
Document missing reserved keywords

sorcix:
It's not possible to use external without backticks.

### DIFF
--- a/docs/modules/language-reference/pages/index.adoc
+++ b/docs/modules/language-reference/pages/index.adoc
@@ -5269,6 +5269,8 @@ They cannot be used as a regular identifier, and currently do not have any meani
 
 To use these names in an identifier, <<quoted-identifiers, surround them with backticks>>.
 
+For a complete list of reserved keywords, consult link:{uri-github-PklLexer}[PklLexer.g4].
+
 [[blank-identifiers]]
 === Blank Identifiers
 
--- a/docs/modules/language-reference/pages/index.adoc
+++ b/docs/modules/language-reference/pages/index.adoc
@@ -5269,7 +5269,7 @@ They cannot be used as a regular identifier, and currently do not have any meani
 
 To use these names in an identifier, <<quoted-identifiers, surround them with backticks>>.
 
-For a complete list of reserved keywords, consult link:{uri-github-PklLexer}[PklLexer.g4].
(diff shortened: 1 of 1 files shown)


kristofvandam:
same applies for:
* `local`
* `open`
* `class`
* ...

sorcix:
Added the ones @kristofvandam mentioned.

sorcix:
Isn't it useful to have a complete list of reserved keywords, though? I wasted quite some time today trying to figure out a vague error:

```
No viable alternative at input `short = "text here."\n\t}\n\texternal:`
```

After finding the reserved keyword list in the documentation I scrapped that as a possible cause. It's weird that there's a list if it's incomplete.

The text above the list doesn't mention anything about other words with special meaning:

> The following keywords are reserved in the language.
> They cannot be used as a regular identifier, and currently do not have any meaning.

That looks like something that should be complete, no?

HT154:
The gotcha here is that `external` does actually have meaning! It's used in the standard library to mark declarations that are implemented in the language runtime (Java) instead of directly in the stdlib (Pkl).

It might make more sense to directly link Pkl's lexer definition here as that will always be up to date: https://github.com/apple/pkl/blob/main/pkl-core/src/main/antlr/PklLexer.g4

sorcix:
The `external` keyword is not listed in the language reference document at all. That's why I thought it didn't do anything.

I've changed it to a link now. I think it's a bit strange to have a language reference document when it's not the reference, though. :smile:

docs/modules/language-reference/pages/index.adoc
```diff
@@ -5269,6 +5269,8 @@ They cannot be used as a regular identifier, and currently do not have any meani
 
 To use these names in an identifier, <<quoted-identifiers, surround them with backticks>>.
 
+For a complete list of reserved keywords, consult link:{uri-github-PklLexer}[PklLexer.g4].
```
bioball, line 5272:
```suggestion
For a complete list of keywords, consult link:{uri-github-PklLexer}[PklLexer.g4].
```

bioball:
> I've changed it to a link now. I think it's a bit strange to have a language reference document when it's not the reference, though. 😄

It's partially because we're not yet at 1.x, and as such, don't have a language specification yet (which would naturally contain a list of keywords).

bioball (approved):
Thanks!

PR #898, opened 2025-01-22
Only run Gradle compatibility tests against minimum and maximum release

bioball:
Currently, we run against every release version, which is taking 20 minutes to run in CI, and this will grow as Gradle creates more releases.

This limits the amount of Gradle versions being tested against; theoretically, only testing against the minimum version and the maximum version should ensure that we are compatible with every version in the range.

### DIFF
--- a/buildSrc/src/main/kotlin/pklGradlePluginTest.gradle.kts
+++ b/buildSrc/src/main/kotlin/pklGradlePluginTest.gradle.kts
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


stackoverflow (approved):

StefMa:
What about introducing parallelism instead of hoping (?) that each Gradle version in between works?

This would also be scalable. No need to worry that this goes out of hand in the future...

bioball:
@StefMa from what I can tell, Gradle is actually kind of bad at handling multiple tasks within a single project in parallel.

It can run multiple test classes in parallel (via `org.gradle.api.tasks.testing.Test#setMaxParallelForks`), and multiple tasks of different subprojects (via the `--parallel` flag).

But if you want multiple tasks of the same project to be run in parallel, you need to use the [worker api](https://docs.gradle.org/current/userguide/worker_api.html).

This PR implements a simple workaround, but I'd be happy to accept a follow-up PR that parallelizes these :)

PR #900, opened 2025-01-23
Update Kotlin to 2.0

odenix:
- update Kotlin from 1.7.10 to 2.0.21
  - Kotlin 1.6 dependencies in Gradle lock files are expected because kotlinc, which is also used by some tests, internally uses some 1.6 dependencies for backwards compatibility reasons.
- update kotlinx-html and kotlinx-serialization
- adapt Kotlin code where necessary
- use Kotlin stdlib Path APIs where possible
- fix IntelliJ Kotlin inspection warnings
- reformat code with `./gradlew spotlessApply`
  - ktfmt adds lots of trailing commas

@KushalP If you'd like to take over from here, please go ahead. Status:
* Authored in early November, rebased today
* `./gradlew build buildNative` passes for me
* I'm not aware of any mandatory work left (could try to fix more inspection warnings)
* IntelliJ shows "unresolved reference" error for `import org.pkl.commons.readString` in class `AbstractTest`
  * looks bogus, couldn't find a way to fix (update: found a workaround, see second commit)


bioball (approved):
LGTM; this all looks good!

I wonder why ktfmt's formatting rules changed; trailing commas in Kotlin has been supported since 1.4.

odenix:
According to ktfmt's changelog, trailing commas are now "managed" by default. What I don't understand is why this is also influenced by which Kotlin version is used.

bioball:
It would make sense if they are disabled pre Kotlin 1.4 because they weren't supported before then. But yeah, the diff here is pretty strange.

Issue #907, opened 2025-01-26
Downstream `native-image` embedders are broken


sgammon:
https://github.com/apple/pkl/blob/258eda86301c2daee548395aed7fdd028a33e39b/pkl-cli/pkl-cli.gradle.kts#L87-L88

This exclusion drops key classes from the `pkl-cli` JAR which are needed to build Pkl native binaries. While the comment is correct, these classes are inert for JVM use at runtime, embedders will Have A Bad Time if these classes are excluded from a `native-image` classpath downstream, when using the `pkl-cli-*` JAR.

These classes can be dropped from fat JARs, because shading with Truffle is only supported on JVM anyway.

PR #914, opened 2025-01-28
fix: downstream `native-image` builds

sgammon:
## Summary

SVM (SubstrateVM) compile configuration classes must be included within the `cli` jar to prevent downstream `native-image` builds from failing. Fat JARs cannot be used with `native-image`, so these classes can still safely be excluded here.

Fixes and closes apple/pkl#907

## Changelog

- fix: don't exclude svm classes from `cli` jar
- fix: include svm exclusions in fatjar config


bioball (approved):
LGTM, thanks!

sgammon:
@bioball It's mad about formatting. Fixed

Issue #508, opened 2024-05-29
Optimization: `const` members should be cached for all children in prototype chain


bioball:
The result of a `const` member cannot be late-bound, so its values should be cached for all children in the prototype chain.

As an observable effect, this should only result in one trace. Today, this has two.

```groovy
class MyClass {
  const prop = trace("hello")
  res: Int
}

a = new MyClass { res = 1 }
b = (a) { res = 2 }
```

PR #915, opened 2025-01-29
Optimization: execute const object bodies and typechecks only once

bioball:
Closes https://github.com/apple/pkl/issues/508

If the object member is const, it only needs to be executed once, and all children in the prototype chain can use its cached value.

There is a possible other optimization here, not included in this PR: we can avoid adding these values to the child object's `cachedMembers`.


stackoverflow (approved):
Good one!

PR #910, opened 2025-01-27
Upgrade gradle to 8.12.1 (from 8.12) and fix some build warnings

KushalP:
### Upgrade gradle to 8.12.1 (from 8.12)

Command used:

```
./gradlew wrapper \
          --gradle-version=8.12.1 \
          --gradle-distribution-sha256-sum=8d97a97984f6cbd2b85fe4c60a743440a347544bf18818048e611f5288d46c94
```

Release notes: https://docs.gradle.org/8.12.1/release-notes.html

### Return tasks.register as tasks.create is deprecated

Fixes the following warning:

```
buildSrc/src/main/kotlin/pklGradlePluginTest.gradle.kts:94:16 'create(String, Class<T!>, Action<in T!>): T' is deprecated. Deprecated in Java
```

### Use property accessor .files instead of function .files()

Fixes warning:

```
pkl-executor/pkl-executor.gradle.kts:71:45: 'files(vararg Dependency!): (Mutable)Set<File!>!' is deprecated. Deprecated in Java
```

### Return `tasks.registering` as `tasks.creating` is deprecated

Fixes the following warning:

```
pkl-tools/pkl-tools.gradle.kts:55:27: 'creating(KClass<U>, U.() -> Unit): PolymorphicDomainObjectContainerCreatingDelegateProvider<Task!, U>' is deprecated. Use registering instead. See https://docs.gradle.org/current/userguide/task_configuration_avoidance.html for more information.
```

PR #916, opened 2025-01-29
Simplify inputs for `javadocJar`

KushalP:
Follow up to #910.

### DIFF
--- a/pkl-tools/pkl-tools.gradle.kts
+++ b/pkl-tools/pkl-tools.gradle.kts
@@ -22,7 +22,7 @@ plugins {
   signing
 }
 
-val dummy: SourceSet by sourceSets.creating
+val placeholder: SourceSet by sourceSets.creating
 
 val firstPartySourcesJars by configurations.existing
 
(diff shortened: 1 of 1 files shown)


StefMa:
We could further "simplify" this even more by adding
```
 archiveClassifier.set("javadoc")
```
to it and below (line 64), we just add the task. Nothing more. No `flatMap` and no `config`.
I guess this would also work, right?

madrob:
Can we call it a placeholder instead of a dummy? The former translates better for non native speakers.

bioball (approved):
LGTM

pkl-tools/pkl-tools.gradle.kts
```diff
@@ -51,15 +51,16 @@ dependencies {
...
+val javadocPlaceholder by tasks.registering(Javadoc::class) { source = placeholder.allJava }
 
 java { withJavadocJar() }
 
 val javadocJar by
   tasks.existing(Jar::class) {
-    from(javadocDummy.get().outputs.files)
+    from(javadocPlaceholder)
     archiveBaseName.set("pkl-tools-all")
+    archiveClassifier.set("javadoc")
```
netvl, line 63: Note that you don't need to set the classifier explicitly as `withJavadocJar()` will create the javadocJar task with the correct classifier. No need to fix it, just wanted to explain it for clarity :)

PR #917, opened 2025-01-29
Replace ANTLR with hand-rolled parser

### LINKED ISSUE #888: Parser accepts wrong string escapes
```pkl
foo = "foo \#(1)"
```

The above code renders foo as `foo 1` even though this is not syntactically valid Pkl.
In fact any number of `#` can be used and Pkl will still parse the interpolation.

The correct code would be:
```pkl
foo = #"foo \#(1)"#
```

### LINKED ISSUE #906: Replace ANTLR with new parser
The current parser is _quite_ slow. On my machine (macOS on M1 Max), it takes almost half a second to parse `pkl:base` in Java.

In some large projects, about 40% of the wall clock time is taken by ANTLR parsing.

It's time to replace it with a hand-rolled parser.

### LINKED ISSUE #927: ANTLR incompatibilities
Pkl's use of ANTLR is sometimes problematic, especially on the fork it uses, which is no longer maintained. When compiling under classpaths that embed a similar version of ANTLR (but not identical, especially from mainline) causes failures which are hard to diagnose; Pkl's ANTLR serialized data uses format `v3`, but latest versions use `v4`, causing an error that surfaces at runtime and not compile-time.

This is especially complicated because the mainline ANTLR version uses the same package as the forked ANTLR, so class loaders can easily get confused. Newer versions of GraalVM also [shade ANTLR](https://central.sonatype.com/artifact/org.graalvm.shadowed/antlr4). It uses `v4`.

All of this will be fixed anyway by #917, so I am mostly filing this for documentation purposes.

### LINKED ISSUE #931: Doc comments with interleaving comments result in an error
Adding a line or block comment between a doc comment results in an error, even though it should be ignored.

Example:
```pkl
/// doc comment
// line comment
/// doc continuation
foo = 1
```

Fails with:
```bash
–– Pkl Error ––
No viable alternative at input `/// doc comment\n// line comment\n/// doc continuation\n`.

3 | /// doc continuation
    ^^^^^^^^^^^^^^^^^^^^^
at foo (.../foo.pkl:3)

```

### LINKED ISSUE #932: Spread elements inside an object body don't need separators
Pkl requires that object members are separated by white space, semicolons or newlines, but this check is not working for spread syntax.

```pkl
foo {
  1 2 3...IntSeq(4, 10)...IntSeq(11, 20)
}
```

This should fail with:
```bash
–– Pkl Error ––
Object members must be separated by whitespace, newline, or semicolon.
```

stackoverflow:
Fixes #906
Fixes #888
Fixes #927
Fixes #931
Fixes #932

Incompatible changes:
- Object entries in the same line now require a semicolon (properties and elements are unaffected):
```pkl
// not valid anymore
foo { ["1"] = 1 ["2"] = 2 }

// requires ;
foo { ["1"] = 1; ["2"] = 2 }
```
Reason: not requiring semicolons or newlines requires too much backtracking in the parser, degrading performance. Also both our IntelliJ plugin and the LSP don't allow this syntax and show an error.

- Some invalid string escapes are not accepted anymore (see #888)
- Unseparated object members require a space, semicolon or newline (see #932)

Our current ANTLR parser is quite slow, and, depending on the codebase, can be the slowest part of running Pkl.

Some results from parsing all snippet tests in pkl-core showing ~100x performance improvement:
ANTLR elapsed: 7122ms (7.1s)
New parser elapsed: 78.73ms

This is still a draft, many tests are still failing and repl parsing is still using the old parser.


pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilderNew.java
```diff
@@ -0,0 +1,3043 @@
...
+    var radix = 10;
+    if (text.startsWith("0x") || text.startsWith("0b") || text.startsWith("0o")) {
+      var type = text.charAt(1);
+      if (type == 'x') {
+        radix = 16;
+      } else if (type == 'b') {
+        radix = 2;
+      } else {
+        radix = 8;
+      }
```
KushalP, line 590:
```suggestion
      radix = switch (type) {
        case 'x' -> 16;
        case 'b' -> 2;
        default -> 8;
      };
```

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilderNew.java
```diff
@@ -0,0 +1,3043 @@
...
+                  .build();
+            }
+            bodyNode =
+                doVisitObjectBody(
+                    body,
+                    new ReadSuperPropertyNode(
+                        unavailableSourceSection(),
+                        scope.getName(),
+                        // Never need a const check for amends declarations. In `foo { ... }`:
+                        // 1. if `foo` is const (i.e. `const foo { ... }`, `super.foo` is required
```
KushalP, line 2364:
```suggestion
                        // 1. if `foo` is const, i.e. `const foo { ... }`, `super.foo` is required
```

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilderNew.java
```diff
@@ -0,0 +1,3043 @@
...
+          .withSourceSection(createSourceSection(expr.newSpan()))
+          .build();
+    }
+
+    return doVisitObjectBody(expr.getBody(), inferredParentNode);
+  }
+
+  @Override
+  public ExpressionNode visitAmendsExpr(Amends expr) {
+    // parentExpr is always New, Amends or Parenthesized. The parser makes sure of it
```
KushalP, line 909: I like this comment, but it would be useful to link through to the method where this behaviour is expected in the parser.

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilderNew.java
```diff
@@ -0,0 +1,3043 @@
...
+    var params = ctx.getParams();
+    var size = params.size();
+    var result = new ArrayList<TypeParameter>(size);
+    for (var i = 0; i < size; i++) {
+      var paramCtx = params.get(i);
+      Variance variance;
+      var nodeVariance = paramCtx.getVariance();
+      if (nodeVariance == null) {
+        variance = TypeParameter.Variance.INVARIANT;
+      } else {
```
KushalP, line 2070: This could've been a `case` in the `switch` below if we either enabled preview of [JEP406](https://openjdk.org/jeps/406) (JDK17) or were on JDK18+ 🥹

pkl-core/src/test/kotlin/org/pkl/core/newparser/ParserComparisonTest.kt
```diff
@@ -0,0 +1,129 @@
...
+        path.extension == "pkl" &&
+          !exceptions.any { pathStr.endsWith(it) } &&
+          !regexExceptions.any { it.matches(pathStr) }
+      }
+      .toList()
+  }
+
+  companion object {
+    // tests that are not syntactically valid Pkl
+    private val exceptions =
```
KushalP, line 108: What are the plans for these files? Shouldn't we aim to fix the syntax in use so that they can no longer be excluded?

pkl-core/src/test/files/LanguageSnippetTests/input/basic/amendsChains.pkl
```diff
@@ -29,5 +29,5 @@ baz = new Listing<Listing<Listing<Number>>> {
       inner.fold(_acc, (__acc, it) -> __acc.add(it))))
 
 qux {
-  (foo.bar) { "world" "!" } { [0] = "Goodbye " [1] = "cruel " }.toList().join("")
+  (foo.bar) { "world" "!" } { [0] = "Goodbye "; [1] = "cruel " }.toList().join("")
```
KushalP, line 32: There's a change in behaviour here that should be called out in the PR description. Like how we call out wrong string escapes.

pkl-core/src/main/java/org/pkl/core/newparser/Parser.java
```diff
@@ -0,0 +1,1328 @@
...
+      builder.append(ch);
+    }
+    return builder.toString();
+  }
+
+  private record FullToken(
+      Token token, Span span, int textOffset, int textSize, boolean newLineBetween) {
+    String text(Lexer lexer) {
+      return lexer.textFor(textOffset, textSize);
+    }
```
KushalP, line 1326: Appears to be unused?

KushalP, line 1326: Appears to be unused?

pkl-core/src/test/files/LanguageSnippetTests/input/basic/amendsChains.pkl
stackoverflow: Added to the description.

pkl-core/src/main/java/org/pkl/core/newparser/Parser.java
stackoverflow: It's used now.

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilderNew.java
stackoverflow: There are a bunch of places in this PR where I have to use a chain of `if`s because 17 doesn't support patterns in switches.

sgammon:
you can fixes-and-closes #927 too :) filed from the sidelines where I am cheering this PR on

pkl-core/src/main/java/org/pkl/core/parser/Lexer.java
```diff
@@ -95,4 +735,88 @@ public static boolean isRegularIdentifier(String identifier) {
...
+  }
+
+  @SuppressWarnings("SuspiciousArrayMethodCall")
+  private static Token getKeywordOrIdent(String text) {
+    var index = Arrays.binarySearch(KEYWORDS, text);
+    if (index < 0) return Token.IDENT;
+    return KEYWORDS[index].token;
+  }
+
+  private static final KeywordEntry[] KEYWORDS = new KeywordEntry[43];
```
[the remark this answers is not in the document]
stackoverflow: RIght. No idea how I forgot that...

pkl-core/src/main/java/org/pkl/core/parser/cst/Expr.java
```diff
@@ -0,0 +1,2530 @@
...
+      return Objects.equals(number, that.number) && Objects.equals(span, that.span);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(number, span);
+    }
+  }
+
+  final class StringConstant implements Expr {
```
bioball, line 440: String constants aren't expressions; this should just be a normal node.

pkl-core/src/main/java/org/pkl/core/parser/cst/Expr.java
```diff
@@ -0,0 +1,2530 @@
...
+      return Objects.equals(strParts, that.strParts) && Objects.equals(span, that.span);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(strParts, span);
+    }
+  }
+
+  final class InterpolatedString implements Expr {
```
bioball, line 504: Can we call this `SingleLineStringLiteral`?

pkl-core/src/main/java/org/pkl/core/parser/cst/Expr.java
```diff
@@ -0,0 +1,2530 @@
...
+          && Objects.equals(span, that.span);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(parts, startDelimiterSpan, endDelimiterSpan, span);
+    }
+  }
+
+  final class InterpolatedMultiString implements Expr {
```
bioball, line 595: Can we call this `MultiLineStringLiteral`?

pkl-core/src/main/java/org/pkl/core/parser/cst/Expr.java
```diff
@@ -0,0 +1,2530 @@
...
+      return Objects.equals(expr, trace.expr) && Objects.equals(span, trace.span);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(expr, span);
+    }
+  }
+
+  final class ImportExpr implements Expr {
```
bioball, line 814: It's a bit strange that `read()`, `read?()`, and `read*()` are parsed as different classes, but `import()` and `import*()` are parsed as the same class. Can we be consistent here and parse as either one, or multiple classes for both cases?

pkl-core/src/main/java/org/pkl/core/parser/cst/Expr.java
```diff
@@ -0,0 +1,2530 @@
...
+      return Objects.equals(parts, that.parts)
+          && Objects.equals(startDelimiterSpan, that.startDelimiterSpan)
+          && Objects.equals(endDelimiterSpan, that.endDelimiterSpan)
+          && Objects.equals(span, that.span);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(parts, startDelimiterSpan, endDelimiterSpan, span);
+    }
```
bioball, line 592: Are `equals` and `hashCode` on these nodes used for anything?

pkl-core/src/main/java/org/pkl/core/parser/Lexer.java
```diff
@@ -91,8 +729,89 @@ public static boolean isRegularIdentifier(String identifier) {
...
+  }
+
+  @SuppressWarnings("SuspiciousArrayMethodCall")
+  private static Token getKeywordOrIdent(String text) {
+    var index = Arrays.binarySearch(KEYWORDS, text);
+    if (index < 0) return Token.IDENT;
+    return KEYWORDS[index].token;
+  }
+
+  private static final KeywordEntry[] KEYWORDS = {
```
bioball, line 749: Let's add a unit test to make sure this is sorted.

This test should do the trick:

```
@Test
  fun `lexer keywords are sorted`() {
    assertThat(Lexer.KEYWORDS).isSortedAccordingTo { a, b -> a.compareTo(b.name) }
  }
```

Will need to change visibility. Maybe make this and `KeywordEntry` package-private?

bioball (changes requested):
Did a first pass! Still have much more to review.

pkl-core/src/main/java/org/pkl/core/parser/Comment.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.core.parser;
+
+public record Comment(String raw, Token type, Span span) {}
```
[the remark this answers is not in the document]
stackoverflow: Right now we don't. Once we implement the formatter, for example, we will. But I removed it anyway, together with the shebang, because comments are better implemented as a prefix/suffix to actual nodes than as a separated list. We can worry about this later. Also not storing the comment text will reduce the parser memory consumption.

pkl-core/src/main/java/org/pkl/core/parser/cst/ClassPropertyEntry.java
```diff
@@ -0,0 +1,488 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import org.pkl.core.parser.ParserVisitor;
+import org.pkl.core.parser.Span;
+import org.pkl.core.util.Nullable;
+
+@SuppressWarnings("DuplicatedCode")
+public sealed interface ClassPropertyEntry extends Node {
```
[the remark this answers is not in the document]
stackoverflow: The reason to subclass was to prevent invalid states. Having both the expression and the body list not null is not a valid state. But I think the simplicity of a having a single class here outweighs the downsides.
Changed.

pkl-core/src/main/java/org/pkl/core/parser/cst/package-info.java
```diff
@@ -0,0 +1,4 @@
+@NonnullByDefault
+package org.pkl.core.parser.cst;
```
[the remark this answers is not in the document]
stackoverflow: It's kinda of a middle ground. It doesn't really abstract much of anything. We can rename it to `ast`.

pkl-core/src/main/java/org/pkl/core/parser/cst/Expr.java
stackoverflow: Made them all into one and added an enum.

pkl-core/src/main/java/org/pkl/core/parser/Lexer.java
stackoverflow: Good idea. Added.

pkl-core/src/main/java/org/pkl/core/parser/cst/Expr.java
stackoverflow: For now, no. But it makes sense for these classes to have at least equals/hashCode. They were records before, and, if not for `parent`, which is mutable, they would still be, which would create equals/hashCode/toString anyway.

pkl-core/src/main/java/org/pkl/core/parser/cst/package-info.java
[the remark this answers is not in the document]
bioball: Alternatively, avoid either name and call this `syntax` instead?

IMO: this being a middle-ground between a CST and an AST, is still an AST heh.

pkl-core/src/main/java/org/pkl/core/parser/cst/ObjectBody.java
```diff
@@ -0,0 +1,103 @@
...
+    this.parent = parent;
+  }
+
+  @Override
+  public List<Node> children() {
+    var children = new ArrayList<Node>(pars.size() + members.size());
+    children.addAll(pars);
+    children.addAll(members);
+    return children;
+  }
```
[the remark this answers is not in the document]
stackoverflow: I kept the `children()` method there but refactored the visitor not call it anymore. So no allocations are being made in our visitors.

pkl-core/src/main/java/org/pkl/core/parser/cst/Expr.java
bioball: Okay, that's fair

pkl-core/src/main/java/org/pkl/core/parser/Parser.java
```diff
@@ -15,218 +15,1551 @@
...
+    }
+    var bodies = parseBodyList();
+    var end = bodies.get(bodies.size() - 1).span();
+    return new ObjectMemberNode.ObjectEntry(key, null, bodies, start.endWith(end));
+  }
+
+  private ObjectMemberNode.ObjectSpread parseObjectSpread() {
+    if (lookahead != Token.SPREAD && lookahead != Token.QSPREAD) {
+      throw parserError("unexpectedToken2", "...", "...?");
+    }
```
bioball, line 706: This is guaranteed to be true

```suggestion
```

pkl-core/src/main/java/org/pkl/core/parser/Parser.java
```diff
@@ -15,218 +15,1551 @@
...
+    var peek = next();
+    boolean isNullable = peek.token == Token.QSPREAD;
+    var expr = parseExpr("}");
+    return new ObjectMemberNode.ObjectSpread(expr, isNullable, peek.span.endWith(expr.span()));
+  }
+
+  private ObjectMemberNode.WhenGenerator parseWhenGenerator() {
+    var start = expect(Token.WHEN, "unexpectedToken", "when").span;
```
bioball, line 714: This is guaranteed to be true

```suggestion
```

pkl-core/src/main/java/org/pkl/core/parser/Parser.java
```diff
@@ -15,218 +15,1551 @@
...
+    if (lookahead == Token.ELSE) {
+      next();
+      elseBody = parseObjectBody();
+      end = elseBody.span();
+    }
+    return new ObjectMemberNode.WhenGenerator(pred, body, elseBody, start.endWith(end));
+  }
+
+  private ObjectMemberNode.ForGenerator parseForGenerator() {
+    var start = expect(Token.FOR, "unexpectedToken", "for").span;
```
bioball, line 730: This is guaranteed to be true

```suggestion
```

pkl-core/src/main/java/org/pkl/core/parser/Parser.java
```diff
@@ -15,218 +15,1551 @@
...
+        header.annotations,
+        header.modifiers,
+        identifier,
+        typePars,
+        type,
+        startSpan.endWith(type.span()));
+  }
+
+  private Class parseClass(MemberHeader header) {
+    var start = expect(Token.CLASS, "unexpectedToken", "class").span;
```
bioball, line 360: This is guaranteed to be true

```suggestion
```

pkl-core/src/main/java/org/pkl/core/parser/Parser.java
```diff
@@ -15,218 +15,1551 @@
...
+        if (lookahead.isKeyword()) {
+          throw parserError("keywordNotAllowedHere", lookahead.text());
+        }
+        throw parserError("invalidTopLevelToken");
+      }
+    }
+  }
+
+  private TypeAlias parseTypeAlias(MemberHeader header) {
+    var start = expect(Token.TYPE_ALIAS, "unexpectedToken", "typealias").span;
```
bioball, line 340: This is guaranteed to be true

```suggestion
```

pkl-core/src/main/java/org/pkl/core/parser/Parser.java
```diff
@@ -15,218 +15,1551 @@
...
+      end = next().span;
+    } else {
+      args = parseListOfParameter(Token.COMMA);
+      end = expect(Token.RPAREN, "unexpectedToken2", ",", ")").span;
+    }
+    return new ParameterList(args, start.endWith(end));
+  }
+
+  private List<ObjectBody> parseBodyList() {
+    var bodies = new ArrayList<ObjectBody>();
```
bioball, line 1309:
```suggestion
    expect(Token.LBRACE, "unexpectedToken2", "{", "=");
    var bodies = new ArrayList<ObjectBody>();
```

pkl-core/src/test/files/LanguageSnippetTests/output/errors/delimiters/unbalancedEntryBrackets4.err
```diff
@@ -1,6 +1,6 @@
 –– Pkl Error ––
-Unmatched delimiter `]`.
+Unexpected token. Expected `{`.
```
bioball, line 2: This error is incorrect; should expect either `{` or `=`

pkl-core/src/test/files/LanguageSnippetTests/output/errors/parser10.err
```diff
@@ -1,5 +1,5 @@
 –– Pkl Error ––
-Mismatched input: `[`. Expected one of: `module`, `nothing`, `unknown`, `(`, `{`, `*`, SLQuote, Identifier
+Unexpected token. Expected `{`.
```
bioball, line 2: This should be something like: "Expected a type, or `{`."

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1126,3 +1130,45 @@ External {0} reader does not support scheme `{1}`.
...
+Unexpected token.
+
+unexpectedTokenForType=\
+Unexpected token. Expected a type.
+
+typeAnnotationInAmends=\
+Expected expression but got object body. Amended properties cannot have type annotations.
+
+invalidProperty=\
+Invalid property definition. Expected type annotation, assignment or amending.
```
bioball, line 1168:
```suggestion
Invalid property definition. Expected a type annotation, `=`, or `{`.
```

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1126,3 +1130,45 @@ External {0} reader does not support scheme `{1}`.
...
+
+unexpectedCharacter=\
+Unexpected character `{0}`. Did you mean {1}?
```
bioball, line 1171: Should do the backticking in the error message, and remove the backticks from the message argument in `Lexer.java`

```suggestion
Unexpected character `{0}`. Did you mean `{1}`?
```

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1126,3 +1130,45 @@ External {0} reader does not support scheme `{1}`.
...
+Keyword `{0}` is not allowed here. (If you must use this name as identifier, enclose it in backticks.)
+
+unexpectedEndOfFile=\
+Unexpected end of file.
+
+wrongHeaders=\
+{0} cannot have doc comments, annotations or modifiers.
+
+invalidTopLevelToken=\
+Invalid token at position. Expected class, typealias, method or property.
```
bioball, line 1144:
```suggestion
Invalid token at position. Expected a class, typealias, method, or property.
```

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1126,3 +1130,45 @@ External {0} reader does not support scheme `{1}`.
...
 External reader process has already terminated.
+
+keywordNotAllowedHere=\
+Keyword `{0}` is not allowed here. (If you must use this name as identifier, enclose it in backticks.)
+
+unexpectedEndOfFile=\
+Unexpected end of file.
+
+wrongHeaders=\
+{0} cannot have doc comments, annotations or modifiers.
```
bioball, line 1141: Missing snippet test

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1126,3 +1130,45 @@ External {0} reader does not support scheme `{1}`.
...
+
+invalidTopLevelToken=\
+Invalid token at position. Expected class, typealias, method or property.
+
+interpolationInConstant=\
+String constant cannot have interpolated values.
```
bioball, line 1147: Missing snippet test

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1126,3 +1130,45 @@ External {0} reader does not support scheme `{1}`.
...
+
+unexpectedToken=\
+Unexpected token. Expected `{0}`.
+
+unexpectedToken2=\
+Unexpected token. Expected `{0}` or `{1}`.
+
+unexpectedTokenMany=\
+Unexpected token. Expected one of {0}.
```
bioball, line 1156: Remove, since we don't need this anymore (see comments above)

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1126,3 +1130,45 @@ External {0} reader does not support scheme `{1}`.
...
+
+unexpectedTokenForExpression=\
+Unexpected token.
+
+unexpectedTokenForType=\
+Unexpected token. Expected a type.
+
+typeAnnotationInAmends=\
+Expected expression but got object body. Amended properties cannot have type annotations.
```
bioball, line 1165: Missing snippet test.

Also, perhaps:

```suggestion
Properties with type annotations cannot have object bodies.

To define both a type annotation and an object body, try using assignment instead.
For example: `obj: Bird = new { ... }`.
```

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -797,7 +800,8 @@ Modifier `const` cannot be applied to property `{0}`.\n\
 Property `{0}` cannot be declared const, because it overrides a non-const property on parent class `{1}`.
 
 unseparatedObjectMembers=\
-Object members must be separated by whitespace, newline, or semicolon.
+Object properties and elements must be separated by whitespace, newline, or semicolon.\n\
```
bioball, line 803: There's also for generators, spreads, when, etc.

I think `Object members` is better. We only present this error message in the case of object members that we can successfully parse on the same line.

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1126,3 +1130,45 @@ External {0} reader does not support scheme `{1}`.
 
 externalReaderAlreadyTerminated=\
 External reader process has already terminated.
+
+keywordNotAllowedHere=\
+Keyword `{0}` is not allowed here. (If you must use this name as identifier, enclose it in backticks.)
```
bioball, line 1135: I think the parentheses message should just be a hint

```suggestion
Keyword `{0}` is not allowed here.

If you must use this name as identifier, enclose it in backticks.
```

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1126,3 +1130,45 @@ External {0} reader does not support scheme `{1}`.
...
+Expected expression but got object body. Amended properties cannot have type annotations.
+
+invalidProperty=\
+Invalid property definition. Expected type annotation, assignment or amending.
+
+unexpectedCharacter=\
+Unexpected character `{0}`. Did you mean {1}?
+
+invalidCharacter=\
+Invalid identifier `{0}`.
```
bioball, line 1174: Missing snippet test

pkl-core/src/test/files/LanguageSnippetTests/input/parser/lineCommentBetween.pkl
```diff
@@ -0,0 +1,7 @@
+/// doc comment
+// line comments can appear between doc comments
+/*
+ block comments also
+*/
+/// doc continuation
+foo = 1
```
bioball, line 7: We're parsing this incorrectly; change this test to:

```pkl
import "pkl:reflect"

/// doc comment
// line comments can appear between doc comments
/*
 block comments also
*/
/// doc continuation
foo = 1

theComment = reflect.Class(module.getClass()).properties["foo"].docComment
```

Output:

```pkl
foo = 1
theComment = """
  doc comment
  // line comments can appear between doc comments
  /*
   block comments also
  */
  doc continuation
  """
```

pkl-core/src/test/files/LanguageSnippetTests/output/parser/spacesBetweenDocComments.err
```diff
@@ -0,0 +1,6 @@
+–– Pkl Error ––
+Invalid token at position. Expected class, typealias, method or property.
```
bioball, line 2: Nit: it would be good to customize error messages for doc comments.

Something like:

```
Dangling documentation comment.

x | /// doc comment continuation
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Documentation comments must be attached to modules, classes, typealiases, methods, or properties.
```

pkl-core/src/test/kotlin/org/pkl/core/parser/ParserComparisonTestInterface.kt
```diff
@@ -0,0 +1,79 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.SoftAssertions
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.parallel.Execution
+import org.junit.jupiter.api.parallel.ExecutionMode
+import org.pkl.core.parser.antlr.PklLexer
+import org.pkl.core.parser.antlr.PklParser
+
+@Execution(ExecutionMode.CONCURRENT)
+interface ParserComparisonTestInterface {
```
bioball, line 32: Why create an interface? These methods can just go directly into `ParserComparisonTest`.

pkl-core/src/test/kotlin/org/pkl/core/parser/ParserComparisonTestInterface.kt
```diff
@@ -0,0 +1,79 @@
...
+  @Test
+  @Execution(ExecutionMode.CONCURRENT)
+  fun compareSnippetTests() {
```
bioball, line 35: Nit: this would be a good candidate for `@ParameterizedTest`.

But, this is temporary code anyways, so, it's not a big deal.

bioball (changes requested):
Did another pass!

pkl-core/src/main/java/org/pkl/core/parser/cst/Expr.java
```diff
@@ -0,0 +1,2350 @@
...
+      return parts;
+    }
+
+    public Span getStartDelimiterSpan() {
+      return startDelimiterSpan;
+    }
+
+    public Span getEndDelimiterSpan() {
+      return endDelimiterSpan;
+    }
```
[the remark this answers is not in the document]
stackoverflow: The warning is gone for me after the refactor.

pkl-core/src/main/java/org/pkl/core/parser/cst/Expr.java
```diff
@@ -0,0 +1,2350 @@
...
+      return visitor.visitMultiLineStringLiteral(this);
+    }
+
+    public List<StringPart> getParts() {
+      return parts;
+    }
+
+    public Span getStartDelimiterSpan() {
+      return startDelimiterSpan;
+    }
```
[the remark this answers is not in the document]
stackoverflow: Same as above.

pkl-core/src/main/java/org/pkl/core/parser/cst/ObjectBody.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import org.pkl.core.parser.ParserVisitor;
+import org.pkl.core.parser.Span;
+import org.pkl.core.util.Nullable;
+
+public final class ObjectBody implements Node {
+  private final List<Parameter> parameters;
```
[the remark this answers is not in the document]
stackoverflow: The `*List` classes keep around the span of their delimiters which we use for error reporting.
This cannot be a `ParameterList` because this is not surrounded by parentheses.

pkl-core/src/main/java/org/pkl/core/parser/cst/package-info.java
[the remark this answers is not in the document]
stackoverflow: Yeah, I don't disagree. Will rename to ast.

pkl-core/src/main/java/org/pkl/core/parser/cst/ObjectBody.java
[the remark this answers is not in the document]
stackoverflow: I just refactored the ast to store everything in the children's list. Calling `children()` or using the visitor doesn't allocate anything.

pkl-core/src/main/java/org/pkl/core/parser/cst/ObjectMemberNode.java
```diff
@@ -0,0 +1,882 @@
...
+package org.pkl.core.parser.cst;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import org.pkl.core.parser.ParserVisitor;
+import org.pkl.core.parser.Span;
+import org.pkl.core.util.Nullable;
+
+public sealed interface ObjectMemberNode extends Node {
```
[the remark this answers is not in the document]
stackoverflow: Yeah, it clashes with a truffle node, but it's not the only thing that clashes. Will rename.
I thought about adding the `Pkl` prefix to every ast node.

pkl-core/src/main/java/org/pkl/core/parser/cst/TypeAnnotation.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+package org.pkl.core.parser.cst;
+
+import java.util.List;
+import java.util.Objects;
+import org.pkl.core.parser.ParserVisitor;
+import org.pkl.core.parser.Span;
+import org.pkl.core.util.Nullable;
+
+public class TypeAnnotation implements Node {
```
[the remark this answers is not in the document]
stackoverflow: `TypeAnnotation` keeps around the span of the whole annotation which we use for error reporting.
I could remove it, but error messages would change and only point to the type.
Also we would lose information such as the `:` was on a different line than the type, which could be useful for linting later.

pkl-core/src/main/java/org/pkl/core/parser/Parser.java
```diff
@@ -15,218 +15,1551 @@
...
+    expect(Token.STRING_START, "unexpectedToken", "\"");
+    var parts = new ArrayList<StringConstantPart>();
+    while (lookahead != Token.STRING_END) {
+      switch (lookahead) {
+        case STRING_PART -> {
+          var tk = next();
+          var text = tk.text(lexer);
+          parts.add(new StringConstantPart.ConstantPart(text, tk.span));
+        }
+        case STRING_ESCAPE_NEWLINE -> parts.add(new StringEscape(EscapeType.NEWLINE, next().span));
```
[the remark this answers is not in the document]
stackoverflow: I think you are mixing `STRING_ESCAPE_NEWLINE` with `STRING_NEWLINE`. The former is possible in a string constant, the later is not. Though I'm handling `STRING_NEWLINE` also here, but it's not needed.

pkl-core/src/main/java/org/pkl/core/parser/Parser.java
stackoverflow: `expect` consumes the token so this wouldn't work.

pkl-core/src/test/files/LanguageSnippetTests/input/parser/lineCommentBetween.pkl
stackoverflow: The parsing is correct and only stores the spans of the actual doc comments. The problem is our truffle nodes expect doc comments to be a single source section. So either we forbid normal comments between doc comments or we refactor our truffle nodes to represent doc comments using multiple sections.

pkl-core/src/main/java/org/pkl/core/parser/Parser.java
```diff
@@ -15,218 +15,1551 @@
...
+          var ret = parseType(expectation);
+          typ = new Type.FunctionType(types, ret, tk.span.endWith(end));
+        } else {
+          typ = new ParenthesizedType(types.get(0), tk.span.endWith(end));
+        }
+      }
+      case STAR -> {
+        var tk = next();
+        var type = parseType(true, expectation);
+        typ = new DefaultUnionType(type, tk.span.endWith(type.span()));
```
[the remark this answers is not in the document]
stackoverflow: Got rid of it and now union types have a flatten list of types. The logic to check union are now in the parser instead of the ast builder.

bioball (commented):
BTW: we should also remove "Grammar Definition" from the language spec, because after this change, ANTLR is no longer the source of truth. But, we also don't really have another spec to point people at yet.

pkl-core/src/main/java/org/pkl/core/parser/BaseParserVisitor.java
```diff
@@ -0,0 +1,497 @@
...
+  private @Nullable T visitChildren(Node node) {
+    T result = defaultValue();
+    var children = node.children();
+    if (children == null) return result;
+    for (var child : children) {
+      if (child != null) {
+        result = aggregateResult(result, child.accept(this));
+      }
+    }
+    return result;
```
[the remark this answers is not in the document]
stackoverflow: Makes sense. Got rid of these methods.

pkl-core/src/test/files/LanguageSnippetTests/input/parser/lineCommentBetween.pkl
stackoverflow: Changed the `docComment` in the truffle nodes to an array of `SourceSection`s.

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -71,7 +261,8 @@
 import org.pkl.core.util.Nullable;
 import org.pkl.core.util.Pair;
 
-public final class AstBuilder extends AbstractAstBuilder<Object> {
+@SuppressWarnings("DataFlowIssue")
```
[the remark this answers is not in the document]
stackoverflow: Fixed.

DEVELOPMENT.adoc
```diff
@@ -84,7 +84,7 @@ based on version information from https://search.maven.org, https://plugins.grad
 * Truffle code generation is performed by Truffle's annotation processor, which runs as part of task `:pkl-core:compileJava`
 ** Output dir is `generated/truffle/`
 * ANTLR code generation is performed by task `:pkl-core:generateGrammarSource`
```
bioball, line 86:
```suggestion
* ANTLR code generation is performed by task `:pkl-core:generateTestGrammarSource`
```

bioball (approved):
Awesome work!! This is going to be a massive improvement for users.

pkl-core/src/main/java/org/pkl/core/parser/Parser.java
```diff
@@ -15,218 +15,1551 @@
...
+                case STRING_NEWLINE -> temp.add(new StringNewline(next().span));
+                case STRING_ESCAPE_NEWLINE ->
+                    temp.add(new StringEscape(EscapeType.NEWLINE, next().span));
+                case STRING_ESCAPE_TAB -> temp.add(new StringEscape(EscapeType.TAB, next().span));
+                case STRING_ESCAPE_QUOTE ->
+                    temp.add(new StringEscape(EscapeType.QUOTE, next().span));
+                case STRING_ESCAPE_BACKSLASH ->
+                    temp.add(new StringEscape(EscapeType.BACKSLASH, next().span));
+                case STRING_ESCAPE_RETURN ->
+                    temp.add(new StringEscape(EscapeType.RETURN, next().span));
```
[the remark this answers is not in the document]
stackoverflow: The parser doesn't remove `_` from numbers anymore or backticks from identifiers.

PR #919, opened 2025-01-31
feat: kotlin abi checks, initial dump

sgammon:
## Summary

**[Kotlin ABI Check](https://github.com/Kotlin/binary-compatibility-validator)** Checks Kotlin public library symbols for drift. Run `apiCheck` to make sure a change does not break downstream Java/Kotlin library consumers; run `apiDump` to re-seal public API lockfiles.

Initial API lockfiles were generated as well. `apiCheck` runs on `./gradlew check` by default, preventing downstream API/ABI breakage when Pkl is used as a Maven library.


buildSrc/src/main/kotlin/pklKotlinLibrary.gradle.kts
```diff
@@ -38,3 +41,7 @@ dependencies {
 tasks.compileKotlin {
   enabled = true // disabled by pklJavaLibrary
 }
+
+tasks.withType<KotlinJvmCompile>().configureEach {
+  compilerOptions { jvmTarget = JvmTarget.JVM_17 }
+}
```
sgammon, line 47: this will rebase away when #876 is in; plus, this block is better customized by #920

sgammon (commented):
Notes:

1) Nullability of API-exposed types may be impacted by #920, so ideally this should go in first to detect any change that might cause

2) Pure Java modules do not have an API lock because this is a Kotlin-specific plugin. I'll look into a Java equivalent or some way to pin those, because `pkl-core`'s API is of course important too

build.gradle.kts
```diff
@@ -79,3 +80,10 @@ val formattedMessage =
   message.replace("\n====", "\n" + "=".repeat(message.lines().maxByOrNull { it.length }!!.length))
 
 logger.info(formattedMessage)
+
+apiValidation {
+  ignoredProjects += listOf("bench", "docs")
+  nonPublicMarkers += "org.pkl.commons.annotations.PklExperimental"
+
+  @OptIn(kotlinx.validation.ExperimentalBCVApi::class) klib { enabled = true }
+}
```
sgammon, line 89: It's best to configure this plugin in the root `build.gradle.kts`, because then projects can be ignored by name. I've enabled [experimental KLib validation](https://github.com/Kotlin/binary-compatibility-validator?tab=readme-ov-file#experimental-klib-abi-validation-support) since Pkl is well past the minimum of `1.9.20`.

pkl-commons/src/main/kotlin/org/pkl/commons/annotations/PklExperimental.kt
```diff
@@ -0,0 +1,25 @@
...
+package org.pkl.commons.annotations
+
+/**
+ * Marks experimental APIs in Kotlin or Java. Such APIs may be publicly exposed by the Pkl project,
+ * but are not provided with a stability guarantee, and require opt-in use from Kotlin.
+ */
+@RequiresOptIn(level = RequiresOptIn.Level.WARNING)
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION, AnnotationTarget.PROPERTY)
+@Retention(AnnotationRetention.BINARY)
+annotation class PklExperimental
```
sgammon, line 25: this annotation is added as an escape hatch for public API pins. it can be used like this:
```kotlin
@PklExperimental
public class SomeClass {
  // ...
}
```
Then, in user code:
```kotlin
@OptIn(PklExperimental::class)
fun someUserFunction {
  // usage of an experimental type
  val x = SomeClass()
}
```
Omission of the `@OptIn` generates a compile-time warning, and any type annotated with `@PklExperimental` is omitted from API pins.

PR #920, opened 2025-01-31
JSR305 integration for `kotlinc`

sgammon:
## Summary

Applies `-Xjsr305=strict` support for Pkl's own annotations, `org.pkl.core.util.Nullable` and `org.pkl.core.util.Nonnull`. Java parameter names are retained in Kotlin bytecode, enabling smoother Java/Kotlin interop and debugging. Fixes deprecated/outdated use of Kotlin Gradle Plugin.

### DIFF
--- a/buildSrc/src/main/kotlin/pklKotlinLibrary.gradle.kts
+++ b/buildSrc/src/main/kotlin/pklKotlinLibrary.gradle.kts
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


bioball (changes requested):
We already have some of these flags set, see https://github.com/apple/pkl/blob/408242a44cdc7bab60363c8b2fc9c17959b92555/buildSrc/src/main/kotlin/pklAllProjects.gradle.kts#L51-L56

As far as I can tell, we don't need `-Xjsr305=@org.pkl.core.util.Nullable:strict` nor `-Xjsr305=@org.pkl.core.util.Nonnull:strict`, and we already get compiler errors on the Kotlin side. 

For example:

```
diff --git a/pkl-cli/src/main/kotlin/org/pkl/cli/CliEvaluator.kt b/pkl-cli/src/main/kotlin/org/pkl/cli/CliEvaluator.kt
index 3ae23de..c72c201 100644
--- a/pkl-cli/src/main/kotlin/org/pkl/cli/CliEvaluator.kt
+++ b/pkl-cli/src/main/kotlin/org/pkl/cli/CliEvaluator.kt
@@ -93,6 +93,7 @@ constructor(
    */
   override fun doRun() {
     val builder = evaluatorBuilder()
+    println(builder.securityManager.checkReadResource(URI("foo:bar")))
     try {
       if (options.multipleFileOutputPath != null) {
         writeMultipleFileOutput(builder)
```

Can you expand on `javaParameters`? How does this help debugging?

sgammon:
@bioball Good point. Let me do some testing to verify my assumptions and I'll get back to you.

Issue #918, opened 2025-01-30
Failed type constraint error messages do not show forced members


bioball:
This is a regression.

This snippet:

```pkl
foo: Listing(firstOneIsSandy) = new {
  new Bird { name = "Bob" }
  new Bird { name = "Bob" }
}

firstOneIsSandy = (it: Listing<Bird>) -> it[0].name == "Sandy"

class Bird { name: String }
```

Produces the following error message:

```
–– Pkl Error ––
Type constraint `firstOneIsSandy` violated.
Value: new Listing { ?; ? }

1 | foo: Listing(firstOneIsSandy) = new {
                 ^^^^^^^^^^^^^^^
```

In Pkl 0.26, we get more details:

```
–– Pkl Error ––
Type constraint `firstOneIsSandy` violated.
Value: new Listing { new Bird { name = "Bob" }; new Bird { name = ? } }

1 | foo: Listing(firstOneIsSandy) = new {
                 ^^^^^^^^^^^^^^^
```

bioball:
This has to do with the listing/mapping typecheck changes.

Removing the type annotation on the lambda parameter give a clearer message:

```diff
 foo: Listing(firstOneIsSandy) = new {
   new Bird { name = "Bob" }
   new Bird { name = "Bob" }
 }
 
-firstOneIsSandy = (it: Listing<Bird>) -> it[0].name == "Sandy"
+firstOneIsSandy = (it) -> it[0].name == "Sandy"
 
 class Bird { name: String }
```

bioball:
Closed by #964

cceneag:
👍

PR #922, opened 2025-01-31
WIP: constraint details

bioball:
Fixes https://github.com/apple/pkl/issues/918

This causes Pkl to eagerly check type nodes that are executed through a constraint.

In the following snippet, the lambda parameter type (`Listing<Person>`) because an eager typecheck.

```pkl
foo: Listing(firstOneIsSandy)

local firstOneIsSandy = (it: Listing<Person>) -> it[0].name == "Sandy"
```

This adds a new implicit frame argument to every frame: the 0th frame argument is now `EnumSet<FrameMarkers>`.

A frame's argument layout changes to>

| index | description |
| --- | --- |
| 0 | Frame markers |
| 1 | Receiver |
| 2 | Owner |
| 3+ | Generic arguments |

Frame markers get propagated through child call nodes.


bioball:
Closing this in favor of #964

PR #926, opened 2025-02-03
fix: upgrade ci jobs to jdk 21

sgammon:
Fixes jobs like `bench` and CLI native builds which now need to build against JDK 21. Followup to #876.


bioball (approved):
LGTM, thanks!

PR #928, opened 2025-02-04
Run a simple eval script

bioball:
Execute `1 + 1` instead of eval'ing `circleci.pkl` (might cause spurious failures due to requirement on network connection).


sgammon (approved):
by the power not vested in me

stackoverflow (approved):

PR #929, opened 2025-02-04
fix: native gvm ci

sgammon:
**Fixes:**

- Checksum taken during #876 ended up out of date due to the release of GraalVM `21.1.2` (the PR was built against `21.1.1`), which happened while the PR was being authored. The checksum for `linux-aarch64` is updated.

- JDK 21 isn't installed within the environment under macOS. Windows and Linux already install it. Thus, a setup script is added to CI macOS jobs to install Temurin JDK 21.


.circleci/config.yml
```diff
@@ -9,6 +9,26 @@ jobs:
...
+            && cat /tmp/jdk.tar.gz | tar --strip-components=1 -xzC .
+
+          export JAVA_ROOT="/Library/Java/JavaVirtualMachines/jdk-21.jdk"
+          export JAVA_HOME="$JAVA_ROOT/Contents/Home"
+          export PATH="$JAVA_HOME/bin:$PATH"
+          mv $HOME/jdk $JAVA_ROOT
+          /usr/libexec/java_home -V
+          java -version
+          mkdir -p $HOME/.gradle
+          echo "org.gradle.java.home=$JAVA_HOME" >> $HOME/.gradle/gradle.properties
```
sgammon, line 29: Some notes:

- `JAVA_ROOT` vs `JAVA_HOME` accounts for the `../Contents/Home` path prepended to mac JDKs.
- `/jdk` is not writable on modern macOS, so the JDK is unpacked at `$HOME/jdk`
- I am not sure `export JAVA_HOME=...` will impact subsequent steps, so `org.gradle.java.home=...` is set instead

PR #935, opened 2025-02-05
Turn CLI commands into objects, statically register subcommands

bioball:
This is a minor improvement: all `Command` are now objects, and they register their own subcommands.


pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/Main.kt
```diff
@@ -100,13 +100,13 @@ class PklJavaCodegenCommand :
...
     """
-          .trimIndent()
+          .trimIndent(),
     )
 
   private val implementSerializable: Boolean by
     option(
         names = arrayOf("--implement-serializable"),
-        help = "Whether to generate classes that implement java.io.Serializable."
+        help = "Whether to generate classes that implement java.io.Serializable.",
```
bioball, line 109: Spotless Kotlin is a little funky; touching a single file will add trailing commas everywhere

stackoverflow (approved):

PR #941, opened 2025-02-06
Correctly set allowed modules/resoures when external reader scheme contain regex control characters

HT154:


bioball (approved):
LGTM, thanks for the fix!

PR #944, opened 2025-02-10
Improve documentation about Set/Map ordering

bioball:
The language reference and stdlib are both slightly incorrect; maps and sets are ordered when iterated on, but unordered in terms of equality.

### DIFF
--- a/docs/modules/language-reference/pages/index.adoc
+++ b/docs/modules/language-reference/pages/index.adoc
@@ -3539,9 +3539,7 @@ res6 = list.map((n) -> n * 3) // <6>
 [[sets]]
 === Sets
 
-A value of type link:{uri-stdlib-Set}[Set] is an ordered collection of unique _elements_.
-
-A set's elements are eagerly evaluated.
+A value of type link:{uri-stdlib-Set}[Set] is a collection of unique _elements_.
 
 Sets are constructed with the `Set()` methodfootnote:soft-keyword[]:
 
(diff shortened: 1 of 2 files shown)


docs/modules/language-reference/pages/index.adoc
```diff
@@ -3557,6 +3555,20 @@ res4 = Set(1, "x", 5.min, List(1, 2, 3)) // <4>
 <3> result: same set of length 3
 <4> result: heterogeneous set that contains a list as its last element
 
+Sets retain the order of elements when constructed, which impacts how they are iterated over.
+However, this order is not considered when determining equality of two sets.
```
HT154, line 3559:
```suggestion
However, this order is not considered when determining equality or hashing of two sets.
```

bioball: Hashing is only an implementation detail; it's not part of the language.

HT154: Fair, though it may be worth explicitly adding that differently ordered sets are also equivalent map/mapping keys.

stackoverflow (approved):

docs/modules/language-reference/pages/index.adoc
bioball: That feels oddly specific? I think the equality bit also implies that they are treated as equivalent mapping keys.

PR #946, opened 2025-02-11
Make commands classes instead of objects

bioball:
Making these classes caused native-image to statically initialize them at build time, which included CLI argument default values (like working dir).

This turns them back into classes.

### DIFF
--- a/pkl-cli/src/test/kotlin/org/pkl/cli/CliMainTest.kt
+++ b/pkl-cli/src/test/kotlin/org/pkl/cli/CliMainTest.kt
@@ -32,28 +32,28 @@ import org.pkl.commons.writeString
 import org.pkl.core.Release
 
 class CliMainTest {
+  private val rootCmd = RootCommand()
 
   @Test
   fun `duplicate CLI option produces meaningful error message`(@TempDir tempDir: Path) {
     val inputFile = tempDir.resolve("test.pkl").writeString("").toString()
 
     assertThatCode {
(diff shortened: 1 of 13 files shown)


pkl-cli/src/test/kotlin/org/pkl/cli/CliMainTest.kt
```diff
@@ -111,7 +111,7 @@ class CliMainTest {
   @Test
   fun `invalid URIs are not accepted`() {
     val ex =
-      assertThrows<BadParameterValue> { RootCommand.parse(arrayOf("eval", "file:my file.txt")) }
+      assertThrows<BadParameterValue> { RootCommand().parse(arrayOf("eval", "file:my file.txt")) }
```
stackoverflow, line 114:
```suggestion
      assertThrows<BadParameterValue> { rootCmd.parse(arrayOf("eval", "file:my file.txt")) }
```
nit

stackoverflow (approved):

KushalP:
Is it possible to create an integration test of the built CLI to validate this works as expected? Feels like a basic sanity check that would be useful to catch this in the future.

bioball:
> Is it possible to create an integration test of the built CLI to validate this works as expected? Feels like a basic sanity check that would be useful to catch this in the future.

Yes, perhaps, although, we did catch this through (internal) tests.

I'm going to merge this, and maybe can follow up with a test.

PR #947, opened 2025-02-12
Bump clikt to 5.0.3

gordonbondon:
Previously `bash` completion generated by `pkl` was causing errors like:
```
-bash: /opt/homebrew/etc/bash_completion.d/pkl: line 347: syntax error near unexpected token `<'
-bash: /opt/homebrew/etc/bash_completion.d/pkl: line 347: `    <modules>)'
```

This was due to clikt bug https://github.com/ajalt/clikt/issues/563 and was fixed in [clikt 5.0.3](https://github.com/ajalt/clikt/releases/tag/5.0.3)

This PR:
1. Bumps clikt to 5.0.3
2. Updates code following clikt migration guide https://ajalt.github.io/clikt/migration/ 
3. Removes unneeded `<..>` around metavars, as clikt now correctly generates them inside help output.

Bonus: This comes with added colours and "markdown" support inside help:
Before:
Screenshot 2025-02-12 at 17 28 04

After: 
Screenshot 2025-02-12 at 17 28 18


pkl-cli/pkl-cli.gradle.kts
```diff
@@ -73,6 +73,7 @@ dependencies {
     exclude(group = "org.jetbrains.kotlin", module = "kotlin-stdlib-jdk8")
     exclude(group = "org.jetbrains.kotlin", module = "kotlin-stdlib-common")
   }
+  implementation(libs.cliktMarkdown)
```
bioball, line 76: Move to `pkl-commons-cli`

pkl-cli/src/main/kotlin/org/pkl/cli/commands/RootCommand.kt
```diff
@@ -15,25 +15,28 @@
...
-  ) {
+class RootCommand : NoOpCliktCommand(name = "pkl") {
+  override val printHelpOnEmptyArgs = true
+
+  override fun helpEpilog(context: Context) = "For more information, visit $helpLink"
+
   init {
     versionOption(Release.current().versionInfo, names = setOf("-v", "--version"), message = { it })
 
+    installMordantMarkdown()
```
bioball, line 36: Should to share this in all of the other CLIs (pkldoc, pkl-codegen-java, pkl-codegen-kotlin).

Maybe introduce an abstract class called `AbstractRootCommand` with this init block, and have this inherit from it.

bioball: Also: currently, we use a bold style to highlight code spans in our error messages.

Played around with it, and we can do the same with Clikt with:

```
  companion object {
    private val theme = Theme {
      styles["markdown.code.span"] = TextStyle(bold = true)
    }
  }

  init {
    context {
      terminal = Terminal(theme = theme)
    }
  }
```

Should also apply this style to our other CLI commands too

bioball (changes requested):
Thanks, this looks really good! Love the colors introduced by this version bump.

Can you rebase off main? There are a lot of dependency bumps introduced by this PR that have since already been updated in main.

pkl-cli/pkl-cli.gradle.kts
gordonbondon: Done.

pkl-cli/src/main/kotlin/org/pkl/cli/commands/RootCommand.kt
gordonbondon: Created a separate `BaseRootCommand` inheriting from `NoOpCliktCommand`, but then noticed that other cli (generate and docs) are inheriting mostly from `ModulesCommand>BaseCommand>CliktCommand` so cant use `BaseRootCommand` for all of clis. Should I also add the same help setup inside `BaseCommand` ?

gordonbondon: I also did not see much change in the help after applying the theme, but I'm not sure if it's just my terminal theme. Could you check on your side to see if bold text still works?
Screenshot 2025-02-14 at 15 20 41

bioball: > Created a separate BaseRootCommand inheriting from NoOpCliktCommand, but then noticed that other cli (generate and docs) are inheriting mostly from ModulesCommand>BaseCommand>CliktCommand so cant use BaseRootCommand for all of clis. Should I also add the same help setup inside BaseCommand ?

Ah, yeah, I missed that. In that case, I don't think we should have a `BaseRootCommand`, because there's not much to share there.

Instead, we can create an extension method to apply the common logic shared between each of the root commands.

I think this works:

```kt
// extensions.kt
package org.pkl.commons.cli.commands

import com.github.ajalt.clikt.core.BaseCliktCommand
import com.github.ajalt.clikt.core.context
import com.github.ajalt.clikt.core.installMordantMarkdown
import com.github.ajalt.clikt.core.terminal
import com.github.ajalt.clikt.parameters.options.versionOption
import com.github.ajalt.mordant.rendering.TextStyle
import com.github.ajalt.mordant.rendering.Theme
import com.github.ajalt.mordant.terminal.Terminal
import org.pkl.core.Release

private val theme = Theme { styles["markdown.code.span"] = TextStyle(bold = true) }

fun  <T : BaseCliktCommand<T>> T.installCommonOptions() {
  installMordantMarkdown()

  versionOption(Release.current().versionInfo, names = setOf("-v", "--version"), message = { it })

  context {
    terminal = Terminal(theme = theme)
  }
}
```

Then you can just add this to each root command:

```
init {
  installCommonOptions()
```

> I also did not see much change in the help after applying the theme, but I'm not sure if it's just my terminal theme. Could you check on your side to see if bold text still works?

Take a look at how the output of `pkl project package --help` changes

gordonbondon: Updated with extensions approach!

Now seeing bold applied to help as expected:
Screenshot 2025-02-18 at 13 15 08

pkl-tools/gradle.lockfile
```diff
@@ -19,22 +38,23 @@ org.graalvm.truffle:truffle-api:24.1.2=runtimeClasspath,testRuntimeClasspath
 org.graalvm.truffle:truffle-compiler:24.1.2=runtimeClasspath,testRuntimeClasspath
 org.graalvm.truffle:truffle-runtime:24.1.2=runtimeClasspath,testRuntimeClasspath
 org.jetbrains.kotlin:kotlin-reflect:2.0.21=runtimeClasspath,testRuntimeClasspath
-org.jetbrains.kotlin:kotlin-stdlib-common:2.0.21=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
 org.jetbrains.kotlin:kotlin-stdlib-jdk7:2.0.21=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
 org.jetbrains.kotlin:kotlin-stdlib-jdk8:2.0.21=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
-org.jetbrains.kotlin:kotlin-stdlib:2.0.21=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
+org.jetbrains.kotlin:kotlin-stdlib:2.1.0=compileClasspath,runtimeClasspath,testCompileClasspath,testRuntimeClasspath
```
bioball, line 43: This looks off. We pin Kotlin to 2.0.21; I wouldn't expect to find 2.1.0 here.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseCommand.kt
```diff
@@ -16,9 +16,17 @@
 package org.pkl.commons.cli.commands
 
 import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.core.Context
 import com.github.ajalt.clikt.parameters.groups.provideDelegate
+import org.pkl.core.Release
+
+abstract class BaseCommand(name: String) : CliktCommand(name = name) {
+  open val helpLink: String = "${Release.current().documentation.homepage}pkl-cli/index.html#usage"
```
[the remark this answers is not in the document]
gordonbondon: `BaseCommand` is used in a bunch of places for `pkl` subcommands, so we'd have to override help link in all of those places.

For now I did override value of help link for gen clis and pkldoc. Let me know if this works.

bioball (approved):
LGTM, thanks for the contribution! This looks great!

PR #953, opened 2025-02-13
Add parser benchmarks

bioball:
The current benchmark is not too useful now that we have a linear parser.

This changes the benchmarks to parse the standard library, and also the language snippet tests.

Sample output:

```
Benchmark                           Mode  Cnt    Score    Error  Units
ParserBenchmark.parseSnippetTests  thrpt    5   52.435 ±  2.514  ops/s
ParserBenchmark.parseStdlib        thrpt    5  586.766 ± 95.770  ops/s
```


bench/src/jmh/java/org/pkl/core/parser/ParserBenchmark.java
```diff
@@ -15,46 +15,50 @@
...
+        new Parser().parseModule(moduleSource);
+      } catch (IOException e) {
+        throw new UncheckedIOException(e);
+      }
+    }
+  }
+
+  @Benchmark
+  public void parseSnippetTests() {
+    var parser = new Parser();
```
stackoverflow, line 50:
```suggestion
```

stackoverflow (approved):
Good idea!

PR #955, opened 2025-02-14
Add docs for various keywords

bioball:
CC @HT154

### DIFF
--- a/docs/modules/language-reference/pages/index.adoc
+++ b/docs/modules/language-reference/pages/index.adoc
@@ -2037,27 +2037,7 @@ greeting2 = greetPigeon(parrot)  // <4>
 <4> Call module method (on `this`).
 
 Like other object-oriented languages, methods defined on extended classes and modules may be overridden.
-The parent type's method may be called via the `super` keyword.
-
-[source%parsed,{pkl}]
-----
-open class Bird {
-  name: String
-  function canEat(food: String): Boolean = food == "seeds"
-}
-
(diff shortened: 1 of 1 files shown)


docs/modules/language-reference/pages/index.adoc
```diff
@@ -4851,6 +4835,161 @@ The predicate, enclosed in double brackets (`\[[...]]`), is matched against each
...
+polly = (lawyerBird) {
+  name = "Polly" // <1>
+}
+----
+<1> Polly has title `"Polly, Esq."`.
+
+[[outer-keyword]]
+=== `outer` keyword
+
+The `outer` keyword references the <<receiver,receiver>> of the outer lexical object.
```
HT154, line 4893: Is this right?
```suggestion
The `outer` keyword references the <<receiver,receiver>> of the immediately outer lexical object.
```

In the past I think I've expected the normal name resolution behavior but _starting at_ the next outer lexical object, but was surprised that it was _only_ the next outer lexical object.

HT154 (approved):
This is awesome!! Thanks for putting this together. Just one question.

StefMa:
This is super helpful! Thank you for writing this down.

stackoverflow (approved):
Very helpful!

docs/modules/language-reference/pages/index.adoc
bioball: Yeah, this is correct. Good clarification!

PR #956, opened 2025-02-14
Fix incorrect glob example

romacafe:
Found a mistake in the glob pattern documentation: https://pkl-lang.org/main/current/language-reference/index.html#examples

> `*.pc[lf]`  --  Anything suffixed by `.pkl` or `.pcf`

Changed the pattern to `*.pk[lg]`, matching `.pkl` and `.pkg`


bioball (approved):
Good find, thank you!

PR #962, opened 2025-02-18
move handling of strings to parser

stackoverflow:
The parser now makes sure that all strings are syntactically valid.


pkl-core/src/main/java/org/pkl/core/parser/ast/StringPart.java
```diff
@@ -31,15 +32,42 @@ public StringPart(Span span, @Nullable List<? extends @Nullable Node> children)
     return visitor.visitStringPart(this);
   }
 
-  public static final class StringConstantParts extends StringPart {
-    public StringConstantParts(List<StringConstantPart> parts, Span span) {
-      super(span, parts);
+  public static final class StringConstantPart extends StringPart {
```
bioball, line 35: Maybe call this `StringChars`? `StringConstantPart` sounds too similar to the `StringConstant` node.

bioball (approved):
LGTM!

Not blocking: I think we can remove `AbstractAstBuilder` now, because the remaining methods there can just be static helper methods (like we can put them in `VmUtils` perhaps). `AstBuilder` can now simply implement `ParserVisitor` instead of `BaseParserVisitor`, and `ImportsAndReadsParser can just implement `BaseParserVisitor`.

stackoverflow:
> Not blocking: I think we can remove AbstractAstBuilder now

Let's do this in a separate PR.

PR #964, opened 2025-02-19
Execute typechecks eagerly when within a constraint

bioball:
This changes the language to check all types eagerly when within a type constraint.

This addresses a regression where error messages might show objects that are missing details (see https://github.com/apple/pkl/issues/918).

Another implication of this change is: type constraints are stricter. Now, type annotations will check `Mapping` and `Listing` members if the annotation is executed from a constraint.

This currently does not throw, and will after this change:

```pkl
import "pkl:test"

const local isBirds = (it: Listing<Bird>) -> true

birds: Listing(isBirds) = new { 1; 2; 3 }

class Bird
```

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
@@ -72,10 +72,29 @@ public abstract class TypeNode extends PklNode {
   public abstract TypeNode initWriteSlotNode(int slot);
 
   /**
+   * Checks if {@code value} conforms to this type.
+   *
+   * <p>Possibly returns a new object with type-casted members, in the case of {@link
+   * MappingTypeNode} or {@link ListingTypeNode}.
+   *
+   * <p>If {@link VmLocalContext#shouldEagerTypecheck()} is true, this method will always do an
+   * eager check.
+   *
+   * <p>If not, throws a {@link VmTypeMismatchException}.
+   */
+  public final Object execute(VirtualFrame frame, Object value) {
+    var localContext = VmLanguage.get(this).localContext.get();
+    if (localContext.shouldEagerTypecheck()) {
+      return executeEagerly(frame, value);
+    }
+    return executeLazily(frame, value);
+  }
+
+  /**
    * Checks if {@code value} conforms to this type, and possibly casts it in the case of {@link
    * MappingTypeNode} or {@link ListingTypeNode}.
    */
-  public abstract Object execute(VirtualFrame frame, Object value);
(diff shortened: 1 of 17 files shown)


pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2514,21 +2537,24 @@ public TypeNode initWriteSlotNode(int slot) {
     }
 
     @ExplodeLoop
-    public Object execute(VirtualFrame frame, Object value) {
-      if (customThisSlot == -1) {
-        CompilerDirectives.transferToInterpreterAndInvalidate();
-        // deferred until execution time s.t. nodes of inlined type aliases get the right frame slot
-        customThisSlot =
-            frame.getFrameDescriptor().findOrAddAuxiliarySlot(CustomThisScope.FRAME_SLOT_ID);
-      }
```
bioball, line 2523: As part of this fix, I also realized that this logic is incorrect. We don't know that this node will always be called from the same frame descriptor, because this same node might be called from different root nodes.

Added a test (constraints14.pkl) that covers this edge case.

pkl-core/src/test/files/LanguageSnippetTests/input/errors/constraintDetails2.pkl
```diff
@@ -0,0 +1,11 @@
+// Error message should include `new Bird { name = "Bob" }`
+birds: Listing(
+    let (myself: Listing<Bird> = this)
```
stackoverflow, line 3: Is the `let` important here? Couldn't it be `this[0].name == "Sandy"`?

stackoverflow (approved):
Looks good. Left some small questions.

pkl-core/src/test/files/LanguageSnippetTests/input/errors/constraintDetails2.pkl
bioball: Yeah, this is testing that the type annotation inside the let expression is executed eagerly.

pkl-core/src/test/files/LanguageSnippetTests/input/errors/constraintDetails3.pkl
```diff
@@ -0,0 +1,10 @@
+// typechecks within child frames should also be eagerly checked
+foo: Listing(toList().every((it: Listing<Bird>) -> it[0].name == "Bob")) = new {
+  new Listing {
+    new Bird {
```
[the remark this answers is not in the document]
bioball: Added another one

PR #973, opened 2025-02-21
Remove dead links to ANTLR parser grammar

bioball:


docs/modules/language-reference/pages/index.adoc
```diff
@@ -5451,8 +5449,6 @@ They cannot be used as a regular identifier, and currently do not have any meani
 
 To use these names in an identifier, <<quoted-identifiers, surround them with backticks>>.
 
-For a complete list of keywords, consult link:{uri-github-PklLexer}[PklLexer.g4].
-
```
bioball, line 5455: Should replace this with something else, but, there's nothing good to point to right now.

stackoverflow: We could point to `pkl-core/src/main/java/org/pkl/core/parser/Lexer.java` at `Lexer.KEYWORDS`.

stackoverflow (approved):

docs/modules/language-reference/pages/index.adoc
bioball: Ah, thanks! Will update.

PR #1001, opened 2025-03-01
Allow native-image flags to be passed through from Gradle properties

### LINKED ISSUE #1000: Cannot build Pkl without permissions to write to cache dir
The build fails with file system exceptions, due to native-image attempting to create folders there

This is causing homebrew to fail the build: https://github.com/Homebrew/homebrew-core/actions/runs/13549918432/job/37876670696?pr=209038#step:3:246

Should allow builds to configure the cache dir (native-image has flag `-Dpolyglot.engine.userResourceCache=writableFolder` for this).

(Yay 1000!)

bioball:
Closes #1000

### DIFF
--- a/pkl-cli/pkl-cli.gradle.kts
+++ b/pkl-cli/pkl-cli.gradle.kts
@@ -237,6 +237,17 @@ fun Exec.configureExecutable(
   workingDir(outputFile.map { it.asFile.parentFile })
   executable = "${graalVm.baseDir}/bin/$nativeImageCommandName"
 
+  // For any system properties starting with `pkl.native`, strip off that prefix and pass the rest
+  // through as arguments to native-image.
+  //
+  // Allow setting args using flags like
+  // (-Dpkl.native-Dpolyglot.engine.userResourceCache=/my/cache/dir) when building through Gradle.
+  val extraArgsFromProperties =
+    System.getProperties().mapNotNull { (key, value) ->
+      if (!key.toString().startsWith("pkl.native")) null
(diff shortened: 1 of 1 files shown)


stackoverflow (approved):

pkl-cli/pkl-cli.gradle.kts
```diff
@@ -237,6 +237,17 @@ fun Exec.configureExecutable(
...
   executable = "${graalVm.baseDir}/bin/$nativeImageCommandName"
 
+  // For any system properties starting with `pkl.native`, strip off that prefix and pass the rest
+  // through as arguments to native-image.
+  //
+  // Allow setting args using flags like
+  // (-Dpkl.native-Dpolyglot.engine.userResourceCache=/my/cache/dir) when building through Gradle.
+  val extraArgsFromProperties =
+    System.getProperties().mapNotNull { (key, value) ->
+      if (!key.toString().startsWith("pkl.native")) null
```
StefMa, line 247: I don't like the negation here 🫣
Could we use filter with just map instead?
I guess it reads a bit better
`filter { key.startsWith(pkl.native) }`.
Or does it have to be filterNot? 🤔 😂

StefMa (commented):
Just a small nit picky one 🙃 sorry

pkl-cli/pkl-cli.gradle.kts
bioball: Yeah; fair, switched!

PR #1013, opened 2025-03-05
Add optimization for generator bodies that don't introduce new members

bioball:
If a generator object literal doesn't add any object members, we can simply use the parent in its place.


stackoverflow (approved):
Good find!

pkl-core/src/test/files/LanguageSnippetTests/input/generators/generatorNoMembers.pkl
```diff
@@ -0,0 +1,46 @@
+// should only result in once trace in the output; generator object literals that don't add any
```
KushalP, line 1:
```suggestion
// should only result in one trace in the output; generator object literals that don't add any
```

PR #1024, opened 2025-03-14
move parser out of pkl-core

stackoverflow:


pkl-parser/src/main/java/org/pkl/parser/util/ErrorMessages.java
```diff
@@ -0,0 +1,36 @@
...
+import java.text.MessageFormat;
+import java.util.Locale;
+import java.util.ResourceBundle;
+
+public final class ErrorMessages {
+  private ErrorMessages() {}
+
+  public static String create(String messageName, Object... args) {
+    var locale = Locale.getDefault();
+    String errorMessage =
```
bioball, line 27:
```suggestion
    var errorMessage =
```

pkl-parser/pkl-parser.gradle.kts
```diff
@@ -0,0 +1,91 @@
...
+  pklAllProjects
+  pklJavaLibrary
+  pklPublishLibrary
+  antlr
+  idea
+}
+
+val generatorSourceSet = sourceSets.register("generator")
+
+sourceSets { test { java { srcDir(file("testgenerated/antlr")) } } }
```
bioball, line 26: Remove; this is used to support org.pkl.core.generator.MemberRegistryGenerator in pkl-core.

Also, remove `annotationProcessor()` from the dependencies block, and also `tasks.named("generateGeneratorGrammarSource") { enabled = false }
` later down in this same file

bioball (approved):
LGTM pending tests passing, and also my comments addressed

pkl-parser/pkl-parser.gradle.kts
stackoverflow: I removed the `annotationProcessor()` and the `tasks.named("generateGeneratorGrammarSource") { enabled = false }`, but if I remove this line, the build fails because it can't find the antlr classes.

stackoverflow: Done!

PR #1052, opened 2025-04-25
Add command to generate shell completion

gordonbondon:


gordonbondon:
Yes, after this is released, I plan to submit brew fromula change to autoinstall completion.

pkl-cli/src/main/kotlin/org/pkl/cli/commands/RootCommand.kt
```diff
@@ -48,6 +49,11 @@ class RootCommand : NoOpCliktCommand(name = "pkl") {
       ProjectCommand(),
       DownloadPackageCommand(),
       AnalyzeCommand(),
+      CompletionCommand(
+        name = "completion",
+        help = "Generate shell completion scripts",
```
bioball, line 54: This will be a pretty rarely used command, so, I'd prefer if this had a more specific name (in case we have something else called complete in the future).

I think `shell-completion` or `generate-completion` (the default name from Clikt) work.

Curious what @stackoverflow and @holzensp think

bioball (changes requested):
Looks good! Except, I think we want to rename the command?

pkl-cli/src/main/kotlin/org/pkl/cli/commands/RootCommand.kt
gordonbondon: Updated to `shell-completion`, which seems to be the most unambiguous.

pkl-cli/src/main/kotlin/org/pkl/cli/commands/RootCommand.kt
```diff
@@ -48,6 +49,11 @@ class RootCommand : NoOpCliktCommand(name = "pkl") {
       ProjectCommand(),
       DownloadPackageCommand(),
       AnalyzeCommand(),
+      CompletionCommand(
+        name = "shell-completion",
+        help = "Generate shell completion scripts",
```
bioball, line 54:
```suggestion
        help = "Generate a completion script for the given shell",
```

bioball (approved):
LGTM! Curious what @stackoverflow thinks

pkl-cli/src/test/kotlin/org/pkl/cli/CliShellCompletionTest.kt
```diff
@@ -0,0 +1,35 @@
...
+import org.pkl.cli.commands.RootCommand
+
+class CliShellCompletionTest {
+  @Test
+  fun `shell completion command supports required shells`() {
+    val shellList = listOf("bash", "zsh", "fish")
+
+    for (shell in shellList) {
+      val command = RootCommand()
+      val result = command.test("shell-completion $shell")
```
stackoverflow, line 30:
```suggestion
      val result = RootCommand().test("shell-completion $shell")
```
nit

stackoverflow (approved):
Looks good!

gordonbondon:
@madrob this is now available during brew install on macos https://github.com/Homebrew/homebrew-core/pull/231522

PR #1053, opened 2025-04-25
Add `getOrDefault` method to `Listing` and `Mapping`

HT154:
This PR adds methods to `Listing` and `Mapping` that can be used to retrieve members. If the member for the index/key isn't present, it applies `default` to the index/key. In both cases, this is essentially sugar for `getOrNull(<index/key>) ?? default.apply(<index/key>)`.

I considered adding another method called something like `getOrElse` to these types (as well as `List` and `Map`). This two-argument method would work like `getOrNull` but in the event of no member being present would return the second argument.

I opted not to implement `getOrElse` since the null coalescing operator `??` already implements this in a more convenient way: `getOrNull(<index/key>) ?? <default value>`. I think the pattern of applying `default` should have its own method: for many users it's not always obvious that `default` is a lambda, especially when it's defined via the sugared anonymous function object body syntax where the argument may be elided if unused.

I'm definitely open to naming suggestions here!

I'm also unsure if `Dynamic` warrants a similar change. I'm currently of the mind that `Dynamic.default` should be considered harmful due to #700.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/stdlib/base/ListingNodes.java
+++ b/pkl-core/src/main/java/org/pkl/core/stdlib/base/ListingNodes.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@ import com.oracle.truffle.api.CompilerDirectives;
 import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
 import com.oracle.truffle.api.dsl.Specialization;
 import com.oracle.truffle.api.nodes.LoopNode;
+import org.pkl.core.PklBugException;
 import org.pkl.core.ast.PklNode;
 import org.pkl.core.ast.lambda.*;
 import org.pkl.core.runtime.*;
@@ -57,6 +58,21 @@ public final class ListingNodes {
     protected Object eval(VmListing self, long index) {
       if (index < 0 || index >= self.getLength()) {
         return VmNull.withoutDefault();
+      }
+      return VmUtils.readMember(self, index);
+    }
+  }
+
+  public abstract static class getOrDefault extends ExternalMethod1Node {
+    @Child private ApplyVmFunction1Node applyNode = ApplyVmFunction1Node.create();
+
+    @Specialization
+    protected Object eval(VmListing self, long index) {
+      if (index < 0 || index >= self.getLength()) {
+        if (VmUtils.readMember(self, Identifier.DEFAULT) instanceof VmFunction defaultFn) {
+          return applyNode.execute(defaultFn, index);
+        }
+        throw PklBugException.unreachableCode();
       }
       return VmUtils.readMember(self, index);
     }
--- a/pkl-core/src/main/java/org/pkl/core/stdlib/base/ListingNodes.java
+++ b/pkl-core/src/main/java/org/pkl/core/stdlib/base/ListingNodes.java
@@ -18,8 +18,8 @@ package org.pkl.core.stdlib.base;
 import com.oracle.truffle.api.CompilerDirectives;
 import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
 import com.oracle.truffle.api.dsl.Specialization;
+import com.oracle.truffle.api.nodes.IndirectCallNode;
 import com.oracle.truffle.api.nodes.LoopNode;
-import org.pkl.core.PklBugException;
 import org.pkl.core.ast.PklNode;
 import org.pkl.core.ast.lambda.*;
 import org.pkl.core.runtime.*;
@@ -64,15 +64,14 @@ public final class ListingNodes {
   }
 
   public abstract static class getOrDefault extends ExternalMethod1Node {
+    @Child private IndirectCallNode callNode = IndirectCallNode.create();
     @Child private ApplyVmFunction1Node applyNode = ApplyVmFunction1Node.create();
 
     @Specialization
     protected Object eval(VmListing self, long index) {
       if (index < 0 || index >= self.getLength()) {
-        if (VmUtils.readMember(self, Identifier.DEFAULT) instanceof VmFunction defaultFn) {
-          return applyNode.execute(defaultFn, index);
-        }
-        throw PklBugException.unreachableCode();
+        var defaultFunction = (VmFunction) VmUtils.readMember(self, Identifier.DEFAULT, callNode);
+        return applyNode.execute(defaultFunction, index);
       }
       return VmUtils.readMember(self, index);
     }
--- a/pkl-core/src/main/java/org/pkl/core/stdlib/base/MappingNodes.java
+++ b/pkl-core/src/main/java/org/pkl/core/stdlib/base/MappingNodes.java
@@ -19,7 +19,6 @@ import com.oracle.truffle.api.CompilerDirectives.TruffleBoundary;
 import com.oracle.truffle.api.dsl.Specialization;
 import com.oracle.truffle.api.nodes.IndirectCallNode;
 import java.util.HashSet;
-import org.pkl.core.PklBugException;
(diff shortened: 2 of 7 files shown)


StefMa:
Kotlin has getOrElse(index, default).
Could also be an option.

I don't like the other names suggested by @bioball 👀 🙃

HT154:
@StefMa in `Listing<Element>` and `Mapping<Key, Value>`, the `default` isn't a concrete value, it's a `Function1<Int, Element>` or `Function1<Key, Value>`, respectively. The lambda syntax is often sugared and the arg omitted when unused so it's easy to miss this! In cases where the index/key is not used applying the `default` and returning the result _is_ "returning the default value".

StefMa:
@HT154 thanks for the explanation!
I already deleted my comment. Was too early in the morning 🫣

I agree with @bioball that getOrDefault is something different in other languages. This might be confusing. See my original comment 🤦‍♂️.

Still "apply" sounds a bit technical. `gefOrUseDefault` 🤔 `getOrUse[Listings/Mappings]Default`.

One questions though, what if default is also not defined? What does it return? 🤔
If it's crashing "as the element is also not available".

HT154:
If `default` isn't explicitly defined on the Listing/Mapping, the `default` is the default value of element/value type. That's why this works:
```pkl
class Foo {
  bar: String = "bar"
  baz: String
}

value = new Mapping<String, Foo> {
  ["a"] {
    // this amends new Foo
    baz = "hello world"
  }
}
// new Mapping {
//   ["a"] = new Foo {
//     bar = "bar"
//     baz = "hello world"
//   }
// }
```

If no element/value type is present, then `new Dynamic {}` is used instead.

I've simplified a little here, but these rules are discussed in a little more detail here: https://pkl-lang.org/main/current/language-reference/index.html#type-defaults

HT154:
And in case it helps think about this, here's how things desugar:

```pkl
// rock candy
new Listing<Foo> {
  default {
    baz = "abc"
  }
}

// but default is a lambda
new Listing<Foo> {
  default { index ->
    baz = "abc"
  }
}

// not so sweet
new Listing<Foo> {
  default = (index) -> (super.default.apply(index)) {
    baz = "abc"
  }
}
```

And you might think of `Listing` and `Mapping` as defined something like this:
```pkl
class Listing<Element = Dynamic> { // default type argument, not really a thing, this is fake syntax
  default: Function1<Int, Element> = (_) -> new Element {}
}

class Mapping<Key = Any, Value = Dynamic> { // more fake syntax
  default: Function1<Key, Value> = (_) -> new Value {}
}
```

stackoverflow:
I think `getOrDefault` is a good name. In Java, the signature is `getOrDefault(index, default)`. So it's already clear by the signature that the Pkl version is doing something different, and it's picking the default somewhere else.
The other suggestions still have `default` in the name, which won't help with the confusion, so may as well use the better name.

b0900062030 (approved):

Abdullah4345 (approved):

pkl-core/src/main/java/org/pkl/core/stdlib/base/ListingNodes.java
```diff
@@ -62,6 +63,21 @@ protected Object eval(VmListing self, long index) {
...
+    protected Object eval(VmListing self, long index) {
+      if (index < 0 || index >= self.getLength()) {
+        if (VmUtils.readMember(self, Identifier.DEFAULT) instanceof VmFunction defaultFn) {
+          return applyNode.execute(defaultFn, index);
+        }
+        throw PklBugException.unreachableCode();
+      }
+      return VmUtils.readMember(self, index);
+    }
+  }
```
bioball: Two changes:

1. Create a cached IndirectCallNode, and pass it to `readMember` (share the same call node for each instance of `getOrDefault()`)
2. Just assert the return type as a `VmFunction` (no need for `PklBugException.unreachableCode()`)

```suggestion
  public abstract static class getOrDefault extends ExternalMethod1Node {
    @Child private IndirectCallNode callNode = IndirectCallNode.create();
    @Child private ApplyVmFunction1Node applyNode = ApplyVmFunction1Node.create();

    @Specialization
    protected Object eval(VmListing self, long index) {
      if (index < 0 || index >= self.getLength()) {
        var defaultFunction = (VmFunction) VmUtils.readMember(self, Identifier.DEFAULT, callNode);
        return applyNode.execute(defaultFunction, index);
      }
      return VmUtils.readMember(self, index, callNode);
    }
  }
```

pkl-core/src/main/java/org/pkl/core/stdlib/base/MappingNodes.java
```diff
@@ -114,6 +116,23 @@ protected Object eval(VmMapping self, Object key) {
...
+    @Specialization
+    protected Object eval(VmMapping self, Object key) {
+      var value = VmUtils.readMemberOrNull(self, key, callNode);
+      if (value != null) {
+        return value;
+      }
+      if (VmUtils.readMember(self, Identifier.DEFAULT) instanceof VmFunction defaultFn) {
+        return applyNode.execute(defaultFn, key);
+      }
+      throw PklBugException.unreachableCode();
```
bioball:
```suggestion
      var defaultFunction = (VmFunction) VmUtils.readMember(self, Identifier.DEFAULT, callNode);
      return applyNode.execute(defaultFunction, key);
```

stdlib/base.pkl
```diff
@@ -1835,6 +1835,13 @@ class Listing<out Element> extends Object {
   @Since { version = "0.27.0" }
   external function getOrNull(index: Int): Element?
 
+  /// Returns the element at [index].
+  ///
+  /// Returns [default] applied to [index] if [index] is outside the bounds of this listing.
+  /// This is equivalent to `getOrNull([index]) ?? default.apply([index])`.
```
bioball:
```suggestion
  /// Returns [default] applied to [index] if [index] is outside the bounds of this listing.
  ///
  /// This is equivalent to `getOrNull(index) ?? default.apply(index)`.
```

stdlib/base.pkl
```diff
@@ -1998,6 +2005,12 @@ class Mapping<out Key, out Value> extends Object {
   /// This is the nullable equivalent of the subscript operator (`object[key]`).
   external function getOrNull(key: Any): Value?
 
+  /// Returns the value associated with [key] or [default] applied to [key] if this mapping does not contain [key].
+  ///
+  /// This is equivalent to `getOrNull([key]) ?? default.apply([key])`.
```
bioball:
```suggestion
  /// Returns the value associated with [key], or [default] applied to [key] if this mapping does
  /// not contain [key].
  ///
  /// This is equivalent to `getOrNull(key) ?? default.apply(key)`.
```

stdlib/base.pkl
```diff
@@ -1835,6 +1835,14 @@ class Listing<out Element> extends Object {
   @Since { version = "0.27.0" }
   external function getOrNull(index: Int): Element?
 
+  /// Returns the element at [index].
+  ///
+  /// Returns [default] applied to [index] if [index] is outside the bounds of this listing.
+  /// 
+  /// This is equivalent to `getOrNull([index]) ?? default.apply([index])`.
```
bioball: The `[index]` notation creates a member link; it creates a code block and also a link. In a code snippet, we should just reference the parameter plainly.

```suggestion
  /// This is equivalent to `getOrNull(index) ?? default.apply(index)`.
```

stdlib/base.pkl
```diff
@@ -1998,6 +2006,13 @@ class Mapping<out Key, out Value> extends Object {
   /// This is the nullable equivalent of the subscript operator (`object[key]`).
   external function getOrNull(key: Any): Value?
 
+  /// Returns the value associated with [key] or [default] applied to [key] if this mapping does
+  /// not contain [key].
+  ///
+  /// This is equivalent to `getOrNull([key]) ?? default.apply([key])`.
```
bioball: (see comment above)

```suggestion
  /// This is equivalent to `getOrNull(key) ?? default.apply(key)`.
```

bioball (changes requested):
Just one last issue with the doc comments!

stackoverflow (approved):
LGTM

HT154:
Amy idea what's up with CI here? Looks like an infra failure downloading JDKs.

bioball (approved):
LGTM!

PR #1056, opened 2025-04-29
Aggregate junit report into one file

gordonbondon:
Some systems require junit report to be in a single file. For example `bazel` https://bazel.build/reference/test-encyclopedia needs single file to be available in `XML_OUTPUT_FILE` path.

To avoid implementing junit aggregation in pkl wrappers in different places this PR instead adds a `--junit-aggregate-reports` flag to return all junit reports as a single file.

Additional flag `--junit-suite-name` is added to allow overriding global test suite name from default `test`

### DIFF
--- a/docs/modules/pkl-cli/pages/index.adoc
+++ b/docs/modules/pkl-cli/pages/index.adoc
@@ -381,7 +381,7 @@ pkl eval -m . myFiles.pkl
 pkl eval -m "%{moduleName}" foo.pkl bar.pkl
 ----
 
-For additional details, see xref:language-reference:index.adoc#multiple-file-output[Multiple File Output] 
+For additional details, see xref:language-reference:index.adoc#multiple-file-output[Multiple File Output]
 in the language reference.
 ====
 
@@ -465,7 +465,29 @@ Default: (none) +
 Example: `./build/test-results` +
 Directory where to store JUnit reports.
 
+One file will be created for each test module.
+
 No JUnit reports will be generated if this option is not present.
+====
+
+[[junit-aggregate-reports]]
+.--junit-aggregate-reports
+[%collapsible]
+====
+Aggregate JUnit reports into a single file.
+
+By default it will be `test.xml` but you can override it using `--junit-suite-name` flag.
+====
+
+[[junit-suite-name]]
+.--junit-suite-name
+[%collapsible]
+====
+Default: (none) +
+Example: `my-tests` +
+The name of the root JUnit test suite.
+
+Used in combination with `--junit-aggregate-reports` flag.
 ====
 
 [[overwrite]]
@@ -553,6 +575,26 @@ Example: `./build/test-results` +
 Directory where to store JUnit reports.
 
 No JUnit reports will be generated if this option is not present.
+====
+
+[[junit-aggregate-reports]]
+.--junit-aggregate-reports
+[%collapsible]
+====
+Aggregate JUnit reports into a single file.
+
+By default it will be `test.xml` but you can override it using `--junit-suite-name` flag.
+====
+
+[[junit-suite-name]]
+.--junit-suite-name
+[%collapsible]
+====
+Default: (none) +
+Example: `my-tests` +
+The name of the root JUnit test suite.
+
+Used in combination with `--junit-aggregate-reports` flag.
 ====
 
 .--overwrite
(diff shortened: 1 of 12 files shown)


pkl-core/src/main/java/org/pkl/core/stdlib/test/report/TestReport.java
```diff
@@ -24,11 +24,11 @@
 import org.pkl.core.TestResults;
 import org.pkl.core.util.StringBuilderWriter;
 
-public interface TestReport {
+abstract class TestReport {
```
KushalP, line 27: Looks like you need to run `./gradlew spotlessApply` to fix formatting across all files. Running `./gradlew check` should show any issues for you locally.

gordonbondon: Fixed.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/CliTestOptions.kt
```diff
@@ -17,4 +17,9 @@ package org.pkl.commons.cli
 
 import java.nio.file.Path
 
-class CliTestOptions(val junitDir: Path? = null, val overwrite: Boolean = false)
+class CliTestOptions(
+  val junitDir: Path? = null,
+  val overwrite: Boolean = false,
+  val junitAggregateReports: Boolean = false,
+  val junitSuiteName: String? = null,
+)
```
bioball, line 25: How about setting the default name as `pkl-tests`?

Also, let's make `junitSuiteName` non-nullable, and declare the default here.

Also, maybe call it `junitAggregatedTestSuitesName"? This is the name of the `<testsuites>`,  not the `<testsuite>`.

```kt
class CliTestOptions(
  val junitDir: Path? = null,
  val overwrite: Boolean = false,
  val junitAggregateReports: Boolean = false,
  val junitAggregatedTestSuitesName: String = "pkl-tests",
) {
  companion object {
    val defaults = CliTestOptions()
  }
}
```

We should also make the flags in Gradle and the CLI non-optional (use `.convention` in Gradle, `.default` in Clikt), and refer to the defaults here.

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -465,9 +465,31 @@ Default: (none) +
 Example: `./build/test-results` +
 Directory where to store JUnit reports.
 
+One file will be created for each test module.
```
bioball, line 468:
```suggestion
By default, one file will be created for each test module. This behavior can be changed with `--junit-aggregate-reports`, which will instead create a single JUnit report file with all test results.
```

bioball (changes requested):
Thanks, good start!

pkl-core/src/main/java/org/pkl/core/stdlib/test/report/JUnitAggregateReport.java
```diff
@@ -0,0 +1,65 @@
...
+
+    var tests =
+        results.stream()
+            .map(r -> reporter.buildSuite(r))
+            .collect(Collectors.toCollection(ArrayList::new));
+
+    var suite = reporter.buildXmlElement("testsuites", attrs, tests.toArray(new VmDynamic[0]));
+
+    writer.append(JUnitReport.renderXML("    ", "1.0", suite));
+  }
```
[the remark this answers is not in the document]
gordonbondon: Done.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/CliTestOptions.kt
gordonbondon: Done.

pkl-core/src/main/java/org/pkl/core/stdlib/test/report/TestReport.java
```diff
@@ -24,11 +24,11 @@
...
 
-  void report(TestResults results, Writer writer) throws IOException;
+  public abstract void report(TestResults results, Writer writer) throws IOException;
```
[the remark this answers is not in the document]
gordonbondon: This is not needed anymore when we revert back to one class. Reverted.

stackoverflow (approved):
LGTM!

PR #1063, opened 2025-05-06
Add `mapNonNullIndexed` API to `List` and `Set`

HT154:


pkl-core/src/test/files/LanguageSnippetTests/input/api/list.pkl
```diff
@@ -332,6 +332,10 @@ examples {
     List(1, 2, 3).mapIndexed((i, n) -> i.isOdd && n.isEven)
   }
 
+  ["mapNonNullIndexed()"] {
+    List(1, 2, 3, 4).mapNonNullIndexed((i, n) -> if (n.isOdd) null else n * i)
```
bioball, line 336: Can you add a test case that exercises null members?

Same for `Set`.

stdlib/base.pkl
```diff
@@ -2441,6 +2441,16 @@ abstract external class Collection<out Element> extends Any {
   @AlsoKnownAs { names { "filterMap" } }
   abstract function mapNonNull<Result>(transform: (Element) -> Result): Collection<Result(this != null)>
 
+  /// Transforms this collection by applying [transform] to each element and removing resulting [null] elements.
+  ///
+  /// [transform] takes two arguments: the index of the element, and the element itself.
+  ///
+  /// Facts:
+  /// ```
+  /// List(1, 2, 3, 4).mapNonNullIndexed((i, n) -> if (n.isOdd) null else n * i) == List(2, 12)
```
bioball, line 2450: Let's add another example here that includes a null member

bioball (changes requested):
Thanks! Generally looks good, just two comments.

bioball (approved):
need to add the `@Since` annotation, but LGTM otherwise. Thanks!

bioball:
LGTM on my wide, looking for a thumbs from @stackoverflow

stackoverflow (approved):
Looks good!

PR #1065, opened 2025-05-08
docs: fixed referenced property in "fixed" section

sitaktif:


KushalP (approved):
I'll let one of @bioball/@stackoverflow do the final review/merge.

PR #1066, opened 2025-05-09
Simplify precedence and associativity parsing

stackoverflow:


pkl-parser/src/main/java/org/pkl/parser/Parser.java
```diff
@@ -800,54 +800,57 @@ private Expr parseExpr() {
...
+      case AND -> new OpInfo(4, true);
+      case EQ_EQ, NOT_EQ -> new OpInfo(5, true);
+      case IS, AS -> new OpInfo(6, true);
+      case LT, LTE, GT, GTE -> new OpInfo(7, true);
+      case PLUS, MINUS -> new OpInfo(8, true);
+      case MULT, DIV, INT_DIV, MOD -> new OpInfo(9, true);
+      case POW -> new OpInfo(10, false);
+      case DOT, QDOT -> new OpInfo(11, true);
+    };
+  }
```
bioball, line 819: We can avoid allocation here by creating these as static final's on record `OpInfo`, e.g.

```java
  private record OpInfo(int prec, boolean isLeftAssoc) {
    static final OpInfo NULL_COALESCE = new OpInfo(1, false);
    static final OpInfo PIPE = new OpInfo(2, true);
    // etc
```

Or, maybe even better, you can add these as fields to the `Operator` enum directly:

```java
public enum Operator {
  POW(10, false),
  // etc

  private final int prec;
  private final boolean isLeftAssoc;

  Operator(int prec, boolean isLeftAssoc) {
    this.opInfo = opInfo;
  }

  public int getPrec() {
    return prec;
  }

  public boolean getIsLeftAssoc() {
    return prec;
  }
}
```

bioball (approved):
I think we can optimize allocations here, but otherwise LGTM!

pkl-parser/src/main/java/org/pkl/parser/Parser.java
stackoverflow: Good idea!

PR #1067, opened 2025-05-11
Add external reader support in pkl-gradle

HT154:
Test on [windows] please

### DIFF
--- a/pkl-gradle/src/main/java/org/pkl/gradle/ExternalReader.java
+++ b/pkl-gradle/src/main/java/org/pkl/gradle/ExternalReader.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright © 2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.gradle;
+
+import java.io.Serializable;
+import java.util.List;
+import org.pkl.core.util.Nullable;
+
+public record ExternalReader(String executable, @Nullable List<String> arguments)
+    implements Serializable {
+  public ExternalReader(String executable) {
+    this(executable, null);
+  }
+}
--- a/pkl-gradle/src/main/java/org/pkl/gradle/ExternalReader.java
+++ b/pkl-gradle/src/main/java/org/pkl/gradle/ExternalReader.java
@@ -16,12 +16,11 @@
 package org.pkl.gradle;
 
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.List;
-import org.pkl.core.util.Nullable;
 
-public record ExternalReader(String executable, @Nullable List<String> arguments)
-    implements Serializable {
+public record ExternalReader(String executable, List<String> arguments) implements Serializable {
   public ExternalReader(String executable) {
-    this(executable, null);
+    this(executable, Collections.emptyList());
   }
 }
--- a/pkl-gradle/src/main/java/org/pkl/gradle/spec/BasePklSpec.java
+++ b/pkl-gradle/src/main/java/org/pkl/gradle/spec/BasePklSpec.java
@@ -22,6 +22,7 @@ import org.gradle.api.file.DirectoryProperty;
 import org.gradle.api.provider.ListProperty;
 import org.gradle.api.provider.MapProperty;
 import org.gradle.api.provider.Property;
+import org.pkl.gradle.ExternalReader;
 
 /** Configuration options shared between plugin features. Documented in user manual. */
 public interface BasePklSpec {
@@ -59,4 +60,8 @@ public interface BasePklSpec {
   ListProperty<String> getHttpNoProxy();
 
   MapProperty<URI, URI> getHttpRewrites();
+  
+  MapProperty<String, ExternalReader> getExternalModuleReaders();
+
+  MapProperty<String, ExternalReader> getExternalResourceReaders();
 }
--- a/pkl-gradle/src/main/java/org/pkl/gradle/spec/BasePklSpec.java
+++ b/pkl-gradle/src/main/java/org/pkl/gradle/spec/BasePklSpec.java
@@ -60,7 +60,7 @@ public interface BasePklSpec {
   ListProperty<String> getHttpNoProxy();
 
   MapProperty<URI, URI> getHttpRewrites();
-  
+
   MapProperty<String, ExternalReader> getExternalModuleReaders();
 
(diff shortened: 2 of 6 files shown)


pkl-gradle/src/main/java/org/pkl/gradle/task/BasePklTask.java
```diff
@@ -231,4 +241,17 @@ protected List<Pattern> patternsFromStrings(List<String> patterns) {
...
+  protected Map<String, PklEvaluatorSettings.ExternalReader> parseExternalReaders(
+      Provider<Map<String, String>> provider) {
+    return provider.getOrElse(Map.of()).entrySet().stream()
+        .collect(
+            Collectors.toMap(
+                Map.Entry::getKey,
+                (it) -> {
+                  var cmd = StringsKt.shlex(it.getValue());
+                  return new PklEvaluatorSettings.ExternalReader(
+                      cmd.getFirst(), cmd.subList(1, cmd.size()));
```
HT154, line 254: If the input value string is empty this throws an exception. Is there a better way to surface this validation of invalid input? The CLI does this:
```bash
> pkl eval pkl:base --external-resource-reader x=''
Usage: pkl eval [<options>] <modules>...

Error: invalid value for --external-resource-reader: List is empty.
```

pkl-gradle/src/main/java/org/pkl/gradle/spec/BasePklSpec.java
```diff
@@ -57,4 +57,8 @@ public interface BasePklSpec {
   Property<URI> getHttpProxy();
 
   ListProperty<String> getHttpNoProxy();
+
+  MapProperty<String, String> getExternalModuleReaders();
+
+  MapProperty<String, String> getExternalResourceReaders();
```
bioball, line 63: I think we should make this `MapProperty<String, ExternalReader>`, where `ExternalReader` is defined as as record class in package `org.pkl.gradle`:

```java
public record ExternalReader(String executable, @Nullable List<String> arguments) implements Serializable {
  public ExternalReader(String executable) {
    this(executable, null);
  }
}
```

And I _believe_ we should add `@Nested` as an annotation here.

@netvl: is this the right Gradle thing to do?

pkl-gradle/src/main/java/org/pkl/gradle/task/BasePklTask.java
bioball: You can validate a property by throwing here (e.g. by throwing a `GradleException`).

pkl-gradle/src/main/java/org/pkl/gradle/spec/BasePklSpec.java
HT154: Ahh I was wondering how to do this. Had a real hard time searching the gradle docs to figure out how to do that, so if you have a doc handy I'd love a link.

HT154: I updated this, but not really sure how to go about writing a test for this. Looking at the other pkl-gradle tests, it seems like they actually run evaluation, but in this case I'd need a working external reader executable to point at.

bioball:
@HT154 can we add some test coverage for this?

I think this would involve moving `TestExternalReaderProcess` to pkl-commons-test, and adding a corresponding test to `org.pkl.gradle.EvaluatorsTest`.

Hint: the Groovy syntax for maps is `[key: value]`.
External readers would look something like:

```groovy
externalModuleReaders = ["foo": new org.pkl.gradle.ExternalReader("bar")]
```

pkl-gradle/src/main/java/org/pkl/gradle/ExternalReader.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.gradle;
+
+import java.io.Serializable;
+import java.util.List;
+import org.pkl.core.util.Nullable;
+
+public record ExternalReader(String executable, @Nullable List<String> arguments)
```
netvl: Is there a material difference between `null` arguments and empty list of arguments? Because if not, I would recommend not marking this as nullable and instead use an empty list.

pkl-gradle/src/main/java/org/pkl/gradle/spec/BasePklSpec.java
netvl: @bioball yes, this is okay - `@Nested` is also not needed because it is needed for Gradle to generate proxies and inject values like `Property` instances or services, and in this case, it is just a plain data class.

pkl-gradle/src/main/java/org/pkl/gradle/task/BasePklTask.java
```diff
@@ -142,6 +145,14 @@ public Provider<String> getEvalRootDirPath() {
   @Optional
   public abstract ListProperty<String> getHttpNoProxy();
 
+  @Nested
+  @Optional
+  public abstract MapProperty<String, ExternalReader> getExternalModuleReaders();
+
+  @Nested
+  @Optional
+  public abstract MapProperty<String, ExternalReader> getExternalResourceReaders();
```
netvl: These should be marked as `@Input` properties, not `@Nested`. The [@Nested](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Nested.html) annotation in Gradle is used to make Gradle process values of a map property as beans. All `@Nested` properties are treated as `@Input`, so it will work as is, but semantically values of this map are just data carriers without any annotations which do not require further processing, so I think that `@Input` would be more appropriate here.

HT154: Thanks @netvl! Do you know of any documentation that covers these subtleties? I wasn't able to find anything when I searched.

netvl: This is all described in Gradle docs, but it may be hard to find, especially since Gradle docs constantly evolve. Most of it comes with experience with writing plugins and sifting through tons of github tickets in the Gradle repo :) 
On this particular thing, I believe, the description is here: https://docs.gradle.org/current/userguide/incremental_build.html#sec:task_input_output_annotations

pkl-gradle/src/main/java/org/pkl/gradle/task/BasePklTask.java
```diff
@@ -146,6 +148,14 @@ public Provider<String> getEvalRootDirPath() {
   @Optional
   public abstract MapProperty<URI, URI> getHttpRewrites();
 
+  @Input
+  @Optional
+  public abstract MapProperty<String, ExternalReader> getExternalModuleReaders();
+
+  @Input
+  @Optional
+  public abstract MapProperty<String, ExternalReader> getExternalResourceReaders();
```
netvl, line 155: Be aware that `@Optional`, according to [its documentation](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Optional.html), is intended to be used for file-based properties. It either will do nothing or will fail plugin validation if it is used with non-file properties (the exact behavior here might depend on Gradle version which is used to build the plugin, I think). I would recommend just dropping it - empty map is always a valid value (and is actually the default value) for map properties anyway.

netvl (approved):
Looks good!

HT154:
It looks like this hangs right now for reasons I'm not understanding. It's intermittent on *nix systems and seemingly consistent on Windows (but I don't have a system to test that locally)

PR #1081, opened 2025-05-29
Gradle adjustments for libpkl

bioball:
* Move native tests into `nativeTest` source set; they are now run with `./gradlew testNative` and skipped in `./gradlew test`.
* Fix native tests to be platform independent
* In `NativeImageBuild`: Add `abstract val sharedLibrary: Property<Boolean>` property to `NativeImageBuild`
* In `NativeImageBuild`: Make `mainClass` optional
* Introduce enum class `Target`, which enumerates over all the target machines that we support building
* Simplify build logic (move common logic into `configure()` extension methods)
* Remove `LibPkl.main`, which is no longer needed
* Run `libpkl` tests in CircleCI (let's run these for now on this branch, but let's remove it prior to merging).

### DIFF
--- a/buildSrc/src/main/kotlin/BuildInfo.kt
+++ b/buildSrc/src/main/kotlin/BuildInfo.kt
@@ -15,8 +15,8 @@
  */
 @file:Suppress("MemberVisibilityCanBePrivate")
 
-import Machine.Arch
-import Machine.OS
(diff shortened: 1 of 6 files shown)


KushalP (approved):
I like that you've added target-specific build pipelines here to call out the target-specific failure

PR #1087, opened 2025-06-02
Treat opaque file URIs as errors

bioball:
Opaque file URIs are URIs whose scheme-specific part does not start with `/`.
For example, `file:foo/bar.txt` is an opaque URI.

Currently, this has the unintentional behavior of: look for file `foo/bar.txt` from the process working directory.
These are effectively dynamics imports; from a single import, we can't statically analyze what it resolves as.

According to RFC-8089, File URIs must have paths that start with `/`. So, these are actually _not valid URIs_.
See the grammar defined in https://datatracker.ietf.org/doc/html/rfc8089#section-2

This changes Pkl's behavior so that these URIs are treated as errors.

### DIFF
--- a/pkl-core/src/test/files/LanguageSnippetTests/output/errors/invalidFileUri4.err
+++ b/pkl-core/src/test/files/LanguageSnippetTests/output/errors/invalidFileUri4.err
@@ -1,0 +1,13 @@
+–– Pkl Error ––
+Module URI `file:path/to/foo.pkl` has invalid syntax.
+
+x | res = import*("file:path/to/foo.pkl")
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+at invalidFileUri4#res (file:///$snippetsDir/input/errors/invalidFileUri4.pkl)
+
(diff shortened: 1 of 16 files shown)


pkl-core/src/test/files/LanguageSnippetTests/output/errors/invalidFileUri4.err
```diff
@@ -0,0 +1,13 @@
+–– Pkl Error ––
+Module URI `file:path/to/foo.pkl` has invalid syntax.
+
+x | res = import*("file:path/to/foo.pkl")
```
bioball, line 4: Ideally, this should have the same error message as `invalidFileUri2`; the error is that the glob pattern is invalid.

But, this is somewhat challenging to fix, because the `URISyntaxException` gets thrown before `ImportGlobNode` is initialized.

Regardless, this error is good enough (given this error message, users know what to do), and I think we can afford to be pragmatic here.

HT154 (approved):
Thanks for fixing this!!

stackoverflow (approved):
Looks good.

PR #1099, opened 2025-06-16
Fix native build

bioball:
* Add missing truffle boundaries
* Make Java snippet tests eval `output.bytes`, just like the CLI does
* Overwrite language snippet output files


stackoverflow (approved):
LGTM!

PR #1100, opened 2025-06-16
Adds traceMode evaluator setting to support trace() pretty printing

ssalevan:
Problem
-------
Calling `trace(...)` upon a complex Pkl data structure results in a compact, single-line evaluation of that structure, which is non-ideal when debugging complex rendering scenarios.

Solution
--------
Adds a new evaluator setting, `traceMode`, which supports the following values:

* `"hidden"`: No trace() calls will be emitted to stderr
* `"default"`: All structures passed to `trace(...)` will be emitted to stderr in compact, single-line form.
* `"pretty"`: All structures passed to `trace(...)` will be emitted to stderr in indented, multi-line form.

As this mechanism needs to be evaluated within the `TraceNode` visitor, this setting is ultimately plumbed into the `VmContext` to be accessible from an `executeGeneric()->doTrace()` call. This approach seemed to be the most idiomatic, but there may ultimately be more straightforward options for exposing this value at the `doTrace()` level.

Testing
-------
Added Pkl, Java and Kotlin tests across all applicable areas to ensure correct parsing of `traceMode` value alongside expected behavior. Ensured that all spotless formatting checks ran cleanly.


pkl-server/src/main/kotlin/org/pkl/server/Server.kt
```diff
@@ -225,6 +225,7 @@ class Server(private val transport: MessageTransport) : AutoCloseable {
       cacheDir,
       dependencies,
       message.outputFormat,
+      false,
```
HT154: Please do actually expose this flag in the message passing API! This will need PRs in pkl-go and pkl-swift to match.

stdlib/EvaluatorSettings.pkl
```diff
@@ -115,6 +115,10 @@ class Http {
   proxy: Proxy?
 }
 
+/// Whether to pretty print calls to trace(), producing indented output across multiple lines.
+@Since { version = "0.28.3" }
```
HT154: A change like this will land in a minor release, not a patch.
```suggestion
@Since { version = "0.29.0" }
```

HT154 (changes requested):

HT154:
There was recently some discussion of the ability to hide traces at the evaluator level. Maybe this would make sense as a `traceMode` setting with possible values `default` (current behavior), `pretty` (as implemented here), or `hidden` (trivial to add here or defer to a separate PR).

stackoverflow:
Agree with @HT154 here about having multiple options. You don't have to implement the `hidden` now.
Also, you have to add this to `org.pkl.commons.cli.commands.BaseOptions` too.

pkl-server/src/main/kotlin/org/pkl/server/Server.kt
ssalevan: Roger that, I've added logic to pkl-server to pass this information along and will create complementary PRs there.

stdlib/EvaluatorSettings.pkl
ssalevan: Updated to `0.29.0`.

ssalevan:
> There was recently some discussion of the ability to hide traces at the evaluator level. Maybe this would make sense as a `traceMode` setting with possible values `default` (current behavior), `pretty` (as implemented here), or `hidden` (trivial to add here or defer to a separate PR).

Got it, changed this to use the new `TraceMode` enum with the suggested values.

> Agree with @HT154 here about having multiple options. You don't have to implement the `hidden` now. Also, you have to add this to `org.pkl.commons.cli.commands.BaseOptions` too.

Added `TraceMode` parsing support to the CLI BaseOptions as specified.

pkl-core/src/main/java/org/pkl/core/ast/expression/unary/TraceNode.java
```diff
@@ -48,7 +57,12 @@ private void doTrace(Object value, VmContext context) {
...
     var sourceSection = valueNode.getSourceSection();
-    var renderedValue = renderer.render(value);
+    String renderedValue;
+    if (context.getTraceMode() == TraceMode.PRETTY) {
+      renderedValue = multiLineRenderer.render(value);
+    } else {
+      renderedValue = singleLineRenderer.render(value);
+    }
     var message =
         (sourceSection.isAvailable() ? sourceSection.getCharacters() : "<value")
```
stackoverflow, line 67: Not related to your PR, but I think `"<value"` here should have being `"<value>"`.
Also, something to think about: it may be a good idea, in the case of pretty rendering, to fiddle a bit with the format.
This is not very readable:
```bash
pkl: TRACE: new {
      ["Parrot"] {
        name = "Parrot"
        age = 1234
        bucketSettings = new {
          kicked = true
          ["fjordConfig"] = new {
            isPining = true
          }
        }
      }
    } = new Dynamic {
  ["Parrot"] {
    name = "Parrot"
    age = 1234
    bucketSettings {
      kicked = true
      ["fjordConfig"] {
        isPining = true
      }
    }
  }
} (file:///$snippetsDir/input/projects/prettyTraceMode/prettyTraceMode.pkl)
```

Perhaps adding some newlines and spaces:
```bash
pkl: TRACE:
  new {
      ["Parrot"] {
        name = "Parrot"
        age = 1234
        bucketSettings = new {
          kicked = true
          ["fjordConfig"] = new {
            isPining = true
          }
        }
      }
    }
=
  new Dynamic {
    ["Parrot"] {
      name = "Parrot"
      age = 1234
      bucketSettings {
        kicked = true
        ["fjordConfig"] {
          isPining = true
        }
      }
    }
  } (file:///$snippetsDir/input/projects/prettyTraceMode/prettyTraceMode.pkl)
```

HT154 (approved):
This looks good to me! I gave it a rebase and a little cleanup.

How are we feeling about this @bioball @stackoverflow?

pkl-core/src/main/java/org/pkl/core/runtime/VmContext.java
```diff
@@ -50,6 +51,7 @@ public static final class Holder {
     private final ModuleCache moduleCache;
     private final @Nullable PackageResolver packageResolver;
     private final @Nullable ProjectDependenciesManager projectDependenciesManager;
+    private final TraceMode traceMode;
```
stackoverflow, line 54:
```suggestion
    private final @Nullable TraceMode traceMode;
```

pkl-core/src/main/java/org/pkl/core/Analyzer.java
```diff
@@ -46,6 +47,7 @@ public class Analyzer {
   private final @Nullable DeclaredDependencies projectDependencies;
   private final ModuleResolver moduleResolver;
   private final HttpClient httpClient;
+  private final TraceMode traceMode;
```
stackoverflow, line 50:
```suggestion
  private final @Nullable TraceMode traceMode;
```

HT154:
I'm going to merge this as-is and open a new PR as follow-up addressing @stackoverflow's comments.

We discussed this and decided to back out the `hidden` trace mode in favor of addressing that as part of the logging system.

PR #1106, opened 2025-06-23
Expose collected superclass properties/methods in `pkl:reflect`

HT154:


pkl-core/src/test/files/LanguageSnippetTests/input/api/reflect4.pkl
```diff
@@ -0,0 +1,89 @@
...
+    name = "test"
+  }
+  annotated { // amending annotated property from parent, should not shadow parent annotations in allProperties
+    name = "test"
+  }
+
+  // methods always "shadow"
+  function overridden(): Int? = null // should shadow annotated parent in allMethods
+  function overriddenAnnotated(): String? = null // should shadow annotated parent in allMethods
+}
```
HT154: Please do validate that these are actually the correct behaviors here!

HT154: I wonder if it makes sense to instead build these methods to "merge" the modifiers and annotations from each parent, as that seems like the most relevant functionality to consumers of the reflection API. I'm thinking along the lines of "in the hierarchy of class A, is property B ever hidden?", and as written this PR doesn't really answer that question, especially since this case disagrees with evaluated output's handling of `hidden`:
```pkl
import "pkl:reflect"

class Nested {
	name: String
}

open class A {
	hidden x: Nested
}

class B extends A {
	x: Nested = new { name = "default" }
}

class C extends A {
	x { name = "default" }
}

testB: B = new {
	x { name = "test" }
}
testC: C = new {
	x { name = "test" }
}

bxMirror = reflect.Class(B).allProperties["x"].modifiers
cxMirror = reflect.Class(C).allProperties["x"].modifiers

```
Result:
```pkl
testB {}
testC {}
bxMirror = Set()
cxMirror = Set("hidden")
```

HT154: So, back to the drawing board I think

stackoverflow: I think we should have an `allModifiers` (and perhaps `allAnnotations`) which would work similarly to your `allProperties/Methods` and would go up the chain collecting modifiers.

pkl-core/src/main/java/org/pkl/core/ast/member/ClassProperty.java
```diff
@@ -67,8 +68,52 @@ public String getCallSignature() {
...
+        klazz = klazz.getSuperclass();
+      }
+
+      return VmModifier.getMirrors(mods, false);
+    }
+  }
+
+  public VmTyped getMirror(VmClass clazz) {
+    return MirrorFactories.propertyFactory.create(new Mirror(this, clazz));
   }
```
HT154, line 109: I'm not sure how to derive `allAnnotations`/`allModifiers` without somehow "currying" the `VmClass` into the mirror for the `ClassProperty`. If you've got any better ideas here, I'm all ears!

pkl-core/src/main/java/org/pkl/core/ast/member/ClassProperty.java
```diff
@@ -67,8 +68,52 @@ public String getCallSignature() {
...
+      while (klazz != null) {
+        var p = klazz.getDeclaredProperty(prop.getName());
+        if (p != null) {
+          annotations.addAll(p.getAnnotations());
+        }
+        klazz = klazz.getSuperclass();
+      }
+
+      return annotations;
+    }
```
stackoverflow:
```suggestion
    public List<VmTyped> getAllAnnotations() {
      var annotations = new ArrayList<VmTyped>();

      for (var klazz = clazz; klazz != null; klazz = klazz.getSuperclass()) {
        var p = klazz.getDeclaredProperty(prop.getName());
        if (p != null) {
          annotations.addAll(p.getAnnotations());
        }
      }

      return annotations;
    }
```
nit: we tend to use this pattern in a bunch of places.

stdlib/reflect.pkl
```diff
@@ -284,6 +292,16 @@ external class Property extends Declaration {
...
   /// [null] if this property does not have a default value.
   hidden defaultValue: Any
+
+  /// The modifiers of this property, merged with any from the containing class's superclasses.
+  @Since { version = "0.30.0" }
+  allModifiers: Set<Modifier>
+
+  /// The annotations of this module, appended with any from the containing class's superclasses.
+  /// 
+  /// Annotations are ordered starting with the current class and walking up the class hierarchy.
```
stackoverflow, line 302: You mentioned the order here, but `allProperties/Methods/Modifiers` also follow the same rule right?

HT154: I'm a little concerned with defining behavior for `allProperties`/`allMethods` because those are implemented in terms of the existing `VmClass.collectAll(Properties|Methods)` methods and iteration order isn't well-defined for them.

This also isn't applicable for `allModifiers` because the implementation builds the bitset with bitwise OR and calls `VmModifier.getMirrors`, which has a stable order for elements based on bits set.

PR #1107, opened 2025-06-24
Formatter

stackoverflow:
This is a draft for the canonical Pkl formatter based on this SPICE: https://github.com/apple/pkl-evolution/pull/15


pkl-formatter/src/test/files/FormatterSnippetTests/output/expr-binary.pkl
```diff
@@ -0,0 +1,19 @@
...
+  clazz.superclass == null ||
+    clazz.superclass.reflectee == Module ||
+    clazz.superclass.reflectee == Typed
+
+baz = new Listing {
+  1
+  2
+}
+  |> mixin1
+  |> mixin2
```
HT154: Hmm. This makes some of my code much more verbose in practice. It'd be nice to allow multiple mixin applications per line (up to the length limit), but use a (future) linter implementation to (ignorably) force one per line.

stackoverflow: The current rule is: the same operators in a chain are considered the same group for line breaking purposes.
So the `|>` only breaks line here because it doesn't fit in a single line (the first operand already doesn't fit).
We could remove this rule and make it so that the right hand operator is a group so it would be kept in a single line if it fits. But this may "break" other sequence of operators and make they look ugly. It's really hard to make everything look good unless you add tons of exceptions, but I rather not add that many exceptions because it makes it hard to understand what the formatter is doing, so some compromises have to be made.
You can try to changes some rules in `Builder.kt` (like `flattenBinaryOperatorExprs`) and see how expressions look. Maybe there's a better way to format chains of operators.

pkl-parser/src/main/java/org/pkl/parser/syntax/generic/GenNode.java
```diff
@@ -0,0 +1,94 @@
...
+ * limitations under the License.
+ */
+package org.pkl.parser.syntax.generic;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import org.pkl.parser.util.Nullable;
+
+public class GenNode {
```
bioball: [nit] `generic.GenNode` is a redundant name; just call this `Node`

pkl-parser/src/main/java/org/pkl/parser/syntax/generic/NodeType.java
```diff
@@ -0,0 +1,176 @@
...
+ * limitations under the License.
+ */
+package org.pkl.parser.syntax.generic;
+
+public enum NodeType {
+  TERMINAL,
+  // affixes
+  LINE_COMMENT(NodeKind.AFFIX),
+  BLOCK_COMMENT(NodeKind.AFFIX),
+  SHEBANG(NodeKind.AFFIX),
```
bioball: Not really affix? Shebang must be the first line of a file. It isn't really attached to any node.

bioball:
Run `./gradlew pkl-cli:javaExecutable`, and that will create an executable at `pkl-cli/build/executable/jpkl`.
You can then run `path/to/pkl-cli/build/executable/jpkl format apply **.pkl` to format code.

Note: I've found that if the glob pattern matches too many files, the jpkl version breaks, and I need to instead use the native executable (built using `./gradlew pkl-cli:buildNative`).

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -733,6 +733,29 @@ pkl shell-completion bash
...
+
+This command checks for format violations on the given files and print their names to stdout. +
+It returns a non-zero status code in case violations are found.
+
+[[command-format-apply]]
+=== `pkl format apply`
+
+*Synopsis*: `pkl format apply [<options>] <paths>`
+
+This command formats the given files overwriting them.
```
bioball, line 751:
```suggestion
This command formats the given files overwriting them.

If any path is a directory, recursively looks for files with a `.pkl` extension, or files named `PklProject`.
```

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -733,6 +733,29 @@ pkl shell-completion bash
 pkl shell-completion zsh
 ----
 
+[[command-format-check]]
+=== `pkl format check`
+
+*Synopsis*: `pkl format check <paths>`
+
+This command checks for format violations on the given files and print their names to stdout. +
+It returns a non-zero status code in case violations are found.
```
bioball, line 742:
```suggestion
It returns a non-zero status code in case violations are found.

If any path is a directory, recursively looks for files with a `.pkl` extension, or files named `PklProject`.
```

pkl-cli/src/main/kotlin/org/pkl/cli/commands/FormatterCommand.kt
```diff
@@ -0,0 +1,133 @@
...
+    argument(name = "path", help = "File or directory to check.")
+      .path(mustExist = true, canBeDir = true)
+
+  @OptIn(ExperimentalPathApi::class)
+  override fun run() {
+    var status = 0
+    val paths =
+      if (path.isDirectory()) {
+        path.walk().filter { it.extension == "pkl" || it.name == "PklProject" }
+      } else sequenceOf(path)
```
bioball: We can address this in a follow-up PR, but: we should probably have some sort of ignore file, so that non-src files are skipped (e.g. bazel folders).

pkl-formatter/src/main/kotlin/org/pkl/formatter/Formatter.kt
```diff
@@ -0,0 +1,44 @@
...
+  fun format(text: String): String {
+    val parser = GenericParser()
+    val builder = Builder(text)
+    val gen = Generator()
+    val ast = parser.parseModule(text)
+    val formatAst = builder.format(ast)
+    // force a line at the end of the file
+    gen.generate(Nodes(listOf(formatAst, ForceLine)))
+    return gen.toString()
+  }
```
bioball, line 52: These are public methods, so: can you add some doc comments to the methods and class here?

pkl-formatter/src/main/kotlin/org/pkl/formatter/Formatter.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.pkl.formatter.ast.ForceLine
+import org.pkl.formatter.ast.Nodes
+import org.pkl.parser.GenericParser
+
+class Formatter {
+  fun format(path: Path): String {
+    try {
+      return format(Files.readString(path))
+    } catch (e: IOException) {
+      throw RuntimeException("Could not format $path:", e)
```
bioball: Why rethrow this as a `RuntimeException`?

pkl-formatter/src/main/kotlin/org/pkl/formatter/Generator.kt
```diff
@@ -0,0 +1,149 @@
...
+import org.pkl.formatter.ast.Indent
+import org.pkl.formatter.ast.Line
+import org.pkl.formatter.ast.MultilineStringGroup
+import org.pkl.formatter.ast.Nodes
+import org.pkl.formatter.ast.Space
+import org.pkl.formatter.ast.SpaceOrLine
+import org.pkl.formatter.ast.Text
+import org.pkl.formatter.ast.Wrap
+
+class Generator {
```
bioball: This class is an implementation detail, right?

```suggestion
internal class Generator {
```

pkl-formatter/pkl-formatter.gradle.kts
```diff
@@ -0,0 +1,53 @@
...
+dependencies {
+  api(projects.pklParser)
+
+  testImplementation(projects.pklCommonsTest)
+}
+
+tasks.test {
+  configureTest()
+  useJUnitPlatform()
+}
```
bioball: Two comments:

1. `configureTest()` isn't abstracting much, might as well inline it
2. `useJunitPlatform()` is already provided via the `pklKotlinLibrary` plugin

```suggestion
tasks.test {
  inputs
    .dir("src/test/files/FormatterSnippetTests/input")
    .withPropertyName("formatterSnippetTestsInput")
    .withPathSensitivity(PathSensitivity.RELATIVE)
  inputs
    .dir("src/test/files/FormatterSnippetTests/output")
    .withPropertyName("formatterSnippetTestsOutput")
    .withPathSensitivity(PathSensitivity.RELATIVE)
}
```

pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -0,0 +1,1215 @@
...
+import org.pkl.formatter.ast.MultilineStringGroup
+import org.pkl.formatter.ast.Nodes
+import org.pkl.formatter.ast.Space
+import org.pkl.formatter.ast.SpaceOrLine
+import org.pkl.formatter.ast.Text
+import org.pkl.parser.syntax.Operator
+import org.pkl.parser.syntax.generic.GenNode
+import org.pkl.parser.syntax.generic.NodeType
+
+class Builder(sourceText: String) {
```
bioball: Same comment as before (make this internal because it's not a public API)

```suggestion
internal class Builder(sourceText: String) {
```

pkl-formatter/src/main/kotlin/org/pkl/formatter/NaturalOrderComparator.kt
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.formatter
+
+class NaturalOrderComparator(private val ignoreCase: Boolean = false) : Comparator<String> {
```
bioball:
```suggestion
internal class NaturalOrderComparator(private val ignoreCase: Boolean = false) : Comparator<String> {
```

pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -0,0 +1,1215 @@
...
+        if (n1.type == NodeType.TERMINAL) null else Line
+      }
+    return Group(newId(), first + listOf(Indent(nodes)))
+  }
+
+  private fun formatUnqualifiedAccessExpression(node: GenNode): FormatNode {
+    val children = node.children
+    if (children.size == 1) return format(children[0])
+    val firstNode = children[0]
+    return if (firstNode.text() == "Map") {
```
bioball, line 232: This has the same problem as `AstBuilder`:

1. It's possible that this is some other method and not the base module `Map`
2. This doesn't handle the case of:
    ```pkl
    import "pkl:base"

    myMap = base.Map(1, 2, 3, 4)
    ```

But, I can't think of a solution here outside of resolving variables at parse time.

pkl-formatter/src/test/files/FormatterSnippetTests/output/imports.pkl
```diff
@@ -0,0 +1,16 @@
...
+
+import "https://example.com/baz.pkl"
+import "package://example.com/myPackage@1.0.0#/Qux.pkl"
+import "pkl:math"
+import "pkl:reflect"
+
+import "@bar/Bar.pkl"
+import "@foo/Foo.pkl" as foo
+
+import* "**.pkl"
```
bioball: Hmm... I think we should actually have globbed imports be after regular imports; something like:

```
<absolute imports>

<dependency imports>

<relative imports>

<absolute glob imports>

<dependency glob imports>

<relative glob imports>
```

It seems a little weird that this is mixed in with the rest of the regular imports.

pkl-cli/src/main/kotlin/org/pkl/cli/commands/FormatterCommand.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    for (path in paths) {
+      val contents = Files.readString(path)
+      val (formatted, stat) = format(path, contents)
+      status = maxOf(stat, status)
```
bioball: If we change the error code to `-1`, `maxOf()` will do the incorrect thing. How about just:

```suggestion
      status = if (status == 0) stat else status
```

Same comment for `ApplyCommand`

pkl-formatter/src/test/files/FormatterSnippetTests/input/map-function.pkl
```diff
@@ -0,0 +1,2 @@
+
+foo = Map(1000, "some random string", 20000, "another random string", 30000, "yet another random string")
```
bioball, line 2: Although it's incorrect code, can we add an example of calling `Map()` with an odd number of arguments?

pkl-parser/src/main/java/org/pkl/parser/GenericParser.java
```diff
@@ -0,0 +1,1574 @@
...
+            ff(children);
+            // unary minus has higher precendence than most binary operators
+            children.add(parseExpr(expectation, 12));
+            yield new GenNode(NodeType.UNARY_MINUS_EXPR, children);
+          }
+          case NOT -> {
+            var children = new ArrayList<GenNode>();
+            children.add(makeTerminal(next()));
+            ff(children);
+            // unary minus has higher precendence than most binary operators
```
bioball:
```suggestion
            // logical not has higher precendence than most binary operators
```

pkl-parser/src/test/kotlin/org/pkl/parser/ParserComparisonTest.kt
```diff
@@ -0,0 +1,118 @@
...
+        val pathStr = path.toString().replace("\\", "/")
+        path.extension == "pkl" &&
+          !exceptions.any { pathStr.endsWith(it) } &&
+          !regexExceptions.any { it.matches(pathStr) }
+      }
+      .toList()
+  }
+
+  fun compare(code: String, path: String? = null, softly: SoftAssertions? = null) {
+    val (sexp, genSexp) = renderBoth(code)
```
bioball, line 62: Very cool!

pkl-cli/src/main/kotlin/org/pkl/cli/commands/FormatterCommand.kt
```diff
@@ -0,0 +1,133 @@
...
+      if (contents != formatted) {
+        println(path.toAbsolutePath().toString())
+        status = 1
+      }
+    }
+    exitProcess(status)
```
bioball: We should follow the pattern set in our other commands: for each CLI command, let's have an API-equivalent class that represents the same thing.

Here, we should have `CliFormatCheckCommand` and `CliFormatApplyCommand`. It should throw `CliException` if we want to exit with non-0 and present an error message to the user.

For reference, see `CliEvaluator`, `CliDocGenerator`, `CliTestRunner`, `CliImportAnalyzer`, and more.

The body of the `run()` methods should just be calling into these classes.

bioball (changes requested):
Amazing work! This is looks really, really great. I left some comments, but nothing major.

Some formatting issues that I think we should address (but not blocking; it'd be great to get this PR in sooner rather than later):

**trailing lambdas**

It would be better to format trailing lambdas to be on the same line. The formatter changes `stdlib/semver.pkl` in method `isPreReleaseLessThan` to this:

```pkl
preReleaseIdentifiers
  .zip(other.preReleaseIdentifiers)
  .fold(
    null,
    (result, next) ->
      if (result != null)
        result
      else if (next.first == next.second)
        null
      else if (next.first.getClass() == next.second.getClass())
        next.first < next.second
      else
        next.first is Int
  )
```

This looks much more readable:

```pkl
preReleaseIdentifiers
  .zip(other.preReleaseIdentifiers)
  .fold(null, (result, next) ->
    if (result != null)
      result
    else if (next.first == next.second)
      null
    else if (next.first.getClass() == next.second.getClass())
      next.first < next.second
    else
      next.first is Int
  )
```

**Multiline let**

It'd be good to keep the variable on the same line as the `let` keyword. This looks more readable:

```pkl
let (foo = 
  bar
   .baz
   .biz
)
  foo + 15
```

This looks less readable:

```pkl
let (
  foo = 
    bar
     .baz
     .biz
)
  foo + 15
```

**Trailing commas**

Multiline args/params should have trailing commas:

```
foo = bar(
  baz,
  biz,
  buz,
)
```

pkl-parser/src/main/java/org/pkl/parser/syntax/generic/NodeType.java
stackoverflow: This was from before the changes to shebang. I forgot to change it here.

pkl-cli/src/main/kotlin/org/pkl/cli/commands/FormatterCommand.kt
stackoverflow: Yeah, this requires some thought on how to do the filtering. Let's add it in a separate PR.

pkl-formatter/src/main/kotlin/org/pkl/formatter/Formatter.kt
stackoverflow: Added.

pkl-formatter/src/main/kotlin/org/pkl/formatter/Formatter.kt
stackoverflow: Indeed. There's not much to be gained vs the original IOException.

pkl-formatter/pkl-formatter.gradle.kts
stackoverflow: Good catch.

pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
stackoverflow: The formatter only has access to syntactical data, so we can't know for sure if this is `base.Map`.
The only options I can see is always separating 2 by 2 regardless (which we do now), or don't treat `Map` as a special construct.

pkl-formatter/src/test/files/FormatterSnippetTests/output/imports.pkl
stackoverflow: I'm fine with separating globs. Done.

pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
bioball: Yeah, I guess this is one of the costs of not having a special syntax for map entries. Food for thought for if we want to make this change in the future.

pkl-formatter/src/test/files/FormatterSnippetTests/input/map-function.pkl
stackoverflow: Added.

pkl-cli/src/main/kotlin/org/pkl/cli/commands/FormatterCommand.kt
stackoverflow: Changed it to use CLiCommands.

pkl-parser/src/main/java/org/pkl/parser/syntax/generic/Node.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(children, span, type);
+  }
+
+  @Override
+  public String toString() {
+    return "GenNode{type='" + type + "', span=" + span + ", children=" + children + '}';
```
HT154:
```suggestion
    return "Node{type='" + type + "', span=" + span + ", children=" + children + '}';
```

HT154 (approved):
Looks good! Let's get this merged so we can iterate as needed on smaller changes.

PR #1108, opened 2025-06-24
Formatter tests

srueg:


pkl-formatter/src/test/files/FormatterSnippetTests/output/for-amend.pkl
```diff
@@ -0,0 +1,14 @@
+module forgenerator.amend
+
+local app = new {
+  some = "a"
+  field = "b"
+}
+
+l = new Listing {
+  for (_item in (app) {
```
srueg, line 9: Honestly, I'm not sure how this _should_ look like. But what the current formatter produces is not ideal:

```pkl
        for
        (
        _item
        in
        (app) {
```

pkl-formatter/src/test/files/FormatterSnippetTests/output/dangling-doccomment.pkl
```diff
@@ -0,0 +1,8 @@
+module dangling
+
+/// Doc comment.
+///
+/// for this field
+// sepearted by a stray comment
```
srueg, line 6: Unsure for the fix here as well: most often the correct thing to do would be replacing `//` with `///`, but that brings us in the territory of fixing, not formatting.

pkl-formatter/src/test/files/FormatterSnippetTests/output/comment-interleaved.pkl
```diff
@@ -0,0 +1,6 @@
+module comment.interleaved
+
+test: Int|String = if (true)
+    1 // It's the same as "100%"
```
srueg, line 4: Another option would be to still bring it on one line, but use `/* comment */` instead. Again, goes in the territory of fixing.

PR #1110, opened 2025-06-25
Don't show 100% when number of failures is rounded up

gordonbondon:


pkl-core/src/main/java/org/pkl/core/stdlib/test/report/SimpleReport.java
```diff
@@ -142,14 +142,14 @@ private void makeStatsLine(
     var passed = total - failed;
     var passRate = total > 0 ? 100.0 * passed / total : 0.0;
 
+    if (failed > 0) {
+      // visual bug: 9,999/10,000 appears to round to 100% so nudge it back down
+      passRate = Math.min(passRate, 99.9);
```
stackoverflow: I think a better approach here is to use proper math. Something like:
```java
var passRate = 0.0;
if (total > 0) {
  var passed = new BigDecimal(total - failed);
  var hundred = new BigDecimal(100);
  var totalPrecise = new BigDecimal(total);
  passRate = hundred.multiply(passed).divide(totalPrecise, 1, RoundingMode.DOWN).doubleValue()
}
```
A scale of 1 should be enough for our use case.

madrob: At that point, I would consider toString() on it instead of going through the intermediate double.

stackoverflow: Yeah, you can go directly to String.

gordonbondon: Updated.

PR #1113, opened 2025-07-01
Add `iterate` and `iterateIndexed` methods for `Collections` in the stdlib

stackoverflow:


pkl-core/src/main/java/org/pkl/core/stdlib/base/ListNodes.java
```diff
@@ -785,6 +788,79 @@ protected Object eval(VmList self, Object initial, VmFunction function) {
...
+      return typeNode;
+    }
+
+    @Specialization
+    protected Object eval(VirtualFrame frame, VmList self, Object initial, VmFunction function) {
+      var iter = self.iterator();
+      var typeNode = getTypeNode();
+      var result = initial;
+      long index = 0;
+      var loop = 0;
```
HT154: What's the reason these are separate?

Is this style preferable?
```suggestion
      var index = 0L;
      var loop = 0;
```

HT154 (approved):
Looks good, just a couple questions

pkl-core/src/main/java/org/pkl/core/stdlib/base/ListNodes.java
stackoverflow: No reason really. I just copied this code from the `fold` implementation. But good point.

stdlib/base.pkl
```diff
@@ -2571,6 +2571,18 @@ abstract external class Collection<out Element> extends Any {
   /// The first parameter of [operator] is the zero-based index of the current element.
   abstract function foldIndexed<Result>(initial: Result, operator: (Int, Result, Element) -> Result): Result
 
+  /// Iterate over elements of this collection in order, accumulating a result.
+  /// The looping function returns a [Pair] where the first value is a [Boolean] indicating if the iteration
+  /// should keep going ([true]) or stop, and the second value is the accumulated result.
+  /// This function is similar to [fold], but it can short-circuit.
+  abstract function iterate<Result>(initial: Result, loopFunction: (Result, Element) -> Pair<Boolean, Result>): Result
```
bioball: Are there other libraries that also call this `iterate`?

I also wonder if we should use a clojure-style "return a `reduced` value" to short-circuit?

In addition to `fold`, we also have `foldBack` and `reduce`. It'd be strange if you can short-circuit here, but not in the other two methods.

stackoverflow: Yeah, that's a good point. I can change the type signature of the folding function to return `Element|Reduced<Element>` and just change the original functions, instead of creating new ones.

Edit: scratch that. I went with the Clojure approach of having a short-circuiting function that you can call from within any folding/reducing function.

PR #1114, opened 2025-07-07
Remove ANTLR from the repo

stackoverflow:
The new parser is already battle tested enough that we don't need to compare it to the older one.


bioball (approved):
LGTM

PR #1115, opened 2025-07-07
Add Kotlin support for "addGeneratedAnnotation" flag

bioball:
This adds logic so that the Kotlin code generator also supports the "addGeneratedAnnotation" flag.

Also:
* Add Antora documentation
* Adjust names (generated-annotation -> add-generated-annotation)
* Adjust doc comments

### DIFF
--- a/pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/Main.kt
+++ b/pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/Main.kt
@@ -57,7 +57,7 @@ class PklJavaCodegenCommand : ModulesCommand(name = "pkl-codegen-java", helpLink
 
   private val generatedAnnotation: Boolean by
     option(
-        names = arrayOf("--generated-annotation"),
+        names = arrayOf("--add-generated-annotation"),
         help = "Whether to add a @Generated annotation to the types to be generated.",
       )
       .flag()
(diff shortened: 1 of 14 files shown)


pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/Main.kt
```diff
@@ -57,7 +57,7 @@ class PklJavaCodegenCommand : ModulesCommand(name = "pkl-codegen-java", helpLink
 
   private val generatedAnnotation: Boolean by
```
stackoverflow:
```suggestion
  private val addGeneratedAnnotation: Boolean by
```

pkl-codegen-kotlin/src/main/kotlin/org/pkl/codegen/kotlin/Main.kt
```diff
@@ -79,6 +79,12 @@ class PklKotlinCodegenCommand : ModulesCommand(name = "pkl-codegen-kotlin", help
       )
       .flag()
 
+  private val addGeneratedAnnotation: Boolean by
+    option(
+        names = arrayOf("--add-generated-annotation"),
+        help = "Whether to generate classes that implement java.io.Serializable.",
```
stackoverflow: Why is the description here different from the Java one: `Whether to add a @Generated annotation to the types to be generated.`?

PR #1121, opened 2025-07-11
Switch publishing to sonatype central

bioball:
This configures the gradle-publish plugin to publish to sonatype central, because OSSRH is now EOL'd.

Note: we might want to switch to using jreleaser in the future. From [Sonatype's docs](https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/#getting-started-for-maven-api-like-plugins):

> The Portal OSSRH Staging API is a partial reimplementation of the OSSRH / Nexus Repository Manager 2 Staging APIs. The intent of this service is to enable publishers who are using existing plugins to have a smooth migration to the Central Publisher Portal. We hope that over time plugins will adopt the [Portal API](https://central.sonatype.org/publish/publish-portal-api/) rather than rely on this service.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -32,8 +32,8 @@ plugins {
 nexusPublishing {
   repositories {
     sonatype {
-      nexusUrl.set(uri("https://s01.oss.sonatype.org/service/local/"))
-      snapshotRepositoryUrl.set(uri("https://s01.oss.sonatype.org/content/repositories/snapshots/"))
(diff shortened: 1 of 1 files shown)


StefMa:
I haven't used the nexus publish gradle plugin to publish, but from my expirience is that this ossrh-staging-api repository is super unreliable.

Not sure if it was about me or something else, but moving from there to maven central leads to missing artifacts. They just didn't showed up 🤷‍♂️.

I went with the [nmcp](https://github.com/GradleUp/nmcp) plugin instead. For publishing to the snapshot repo, I just used the default maven-publish plugin provided by gradle.

PR #1133, opened 2025-07-22
Polish http rewrites

bioball:
* Add ability to set HTTP builder in ConfigEvaluatorBuilder
* Add ability to set rewrites in executor API
* Polish documentation

### DIFF
--- a/pkl-config-java/src/main/java/org/pkl/config/java/ConfigEvaluatorBuilder.java
+++ b/pkl-config-java/src/main/java/org/pkl/config/java/ConfigEvaluatorBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


pkl-config-java/src/main/java/org/pkl/config/java/ConfigEvaluatorBuilder.java
```diff
@@ -242,6 +243,15 @@ public ConfigEvaluatorBuilder setTimeout(Duration timeout) {
...
   }
 
+  /**
+   * Returns the currently set evaluation timeout.
+   *
+   * <p>This is a convenience method that delegates to the underlying evaluator builder.
+   */
+  public @Nullable Duration getTimeout() {
+    return evaluatorBuilder.getTimeout();
+  }
```
bioball, line 253: This isn't a new method; I just moved the getter next to the setter.

pkl-executor/src/test/kotlin/org/pkl/executor/EmbeddedExecutorTest.kt
```diff
@@ -107,7 +110,7 @@ class EmbeddedExecutorTest {
       }
     }
 
-    // a pkl-executor class loader that supports ExecutorSpiOptions up to v2
+    // a pkl-executor class loader that supports ExecutorSpiOptions up to v3
```
stackoverflow, line 113: nit: it's a bit annoying that `pklExecutorClassLoader2` (emphasis on 2) supports up to v3.
Same for `pklDistribution2`.

stackoverflow (approved):

pkl-executor/src/test/kotlin/org/pkl/executor/EmbeddedExecutorTest.kt
bioball: The SPI version will keep rising; I can't think of a better name here (`executorOldNew` is harder to read than `executor1_2` for example).

Issue #993, opened 2025-02-27
Provide a C library (libpkl) for Pkl


bioball:
The Swift and Go libraries should be able to bind to this library, rather than spawn a child process.

Docs on creating native shared libraries: https://www.graalvm.org/latest/reference-manual/native-image/guides/build-native-shared-library/

PR #1134, opened 2025-07-22
C library for Pkl

KushalP:
This PR introduces native C bindings for Pkl. See: https://github.com/apple/pkl/issues/993

## Dynamic Library

Using `org.graalvm.nativeimage` and the `native-image` binary we produce a dynamic `-shared` library (for each OS/Arch variant) that provides a way to initialise itself with a `graal_isolatethread_t`.

Methods are annotated with `@CEntryPoint` and exported.

This change results in an architecture and OS-specific directory being
created which now produces the headers for our shared library
functions:

```
❯ ls libpkl/build/native-libs/macos-aarch64/
graal_isolate_dynamic.h
graal_isolate.h
libpkl_internal_dynamic.h
libpkl_internal.dylib
libpkl_internal.h
libpkl.dylib                  <---- this is the interface we're exposing
libpkl.dylib.dSYM.            <---- debug symbols for the interface we're exposing
pkl.h                         <---- this is the interface we're exposing
```

## Gradle Tasks

`NativeImageBuild` has been extended to allow creating a shared library. Alongside this, a `:libpkl:nativeTest` task has been creating which uses JNA to drive the dynamic library from Java.

### JNA                                                                      

Testing of the produced `libpkl` dynamic library is done using [Java Native Access](https://github.com/java-native-access/jna) for ease. We provide an `interface` in Kotlin which JNA transposes against the `libpkl` dynamic library discoverable at the path that is discoverable with `jna.library.path`.

## `libpkl`

The produced `libpkl` dynamic library wraps the GraalVM C interface into something that is future-friendly for the needs of a Pkl integrator. It exports an interface which aligns with [SPICE-0015](https://github.com/apple/pkl-evolution/pull/16).

Note the last three items called out in the **Dynamic Library** section.

## Outstanding tasks

There's work remaining before this can be merged.

- [ ] Get review feedback from GraalVM maintainers on the implementation, specifically the `graal_attach_thread` behaviour
- [ ] Provide the `-library_name` flag to `native-image` to not require a fully-qualified path for library inclusion
- [ ] The implementation is inherently single-threaded, how can we allow this to be safe to use across multiple threads?
- [ ] Allow configuring a _release_ build
- [ ] Merge `libpkl_internal.dylib` and `libpkl.dylib` into a single artefact that is easier for users to consume as `libpkl.dylib`
- [ ] Remove architecture-specific builds that have been added for testing


KushalP:
Replaced with https://github.com/apple/pkl/pull/1135 as Circle CI only builds on PRs from forks.

PR #1137, opened 2025-07-22
SPICE-0019: Allow trailing commas in comma-separated syntax elements

### LINKED ISSUE #1132: Allow trailing commas in comma-separated syntax elements
Many languages—Swift included—allow trailing commas between comma-separated syntax elements.
Allowing this means that adding elements to the end or reordering elements doesn't incur additional diff to manage the comma absence on the new/former final element.

These syntax rules in Pkl should allow trailing commas:
* Function parameter lists (method definitions)
* Argument lists (method calls)
* Type argument lists (in declared type names)
* Type parameter lists (in class/module headers)
* Function type parameter lists (function type annotations)
* Type constraint lists
* Object body parameter lists (sugar'd lambdas)

Pkl's formatter should likely enforce trailing commas are present when the containing syntax rules are broken over multiple lines.
This will require support in Pkl's various grammars (pkl-intellij, tree-sitter-pkl).
This should probably also affect the `pkl.experimental.syntax` library in pkl-pantry.

HT154:
Implements https://github.com/apple/pkl-evolution/pull/21

This changes Pkl's grammar but is not breaking. It only admits inputs that were previously invalid.

Covers these syntax elements:
- Function parameter lists (method definitions) - tested in `parser/methodTrailingCommas.pkl`
- Argument lists (method calls) - tested in `parser/methodTrailingCommas.pkl`
- Type argument lists (in declared type names) - tested in `parser/trailingCommas.pkl`
- Type parameter lists (in class/module headers) - tested in `parser/trailingCommas.pkl`
- Type constraint lists - tested in `parser/constraintsTrailingComma.pkl`
- Function type parameter lists (function type annotations) - tested in `parser/lambdaTrailingCommas.pkl `
- Object body parameter lists (sugar'd lambdas) - tested in `parser/lambdaTrailingCommas.pkl`

Resolves #1132


pkl-parser/src/main/java/org/pkl/parser/Parser.java
```diff
@@ -1725,11 +1725,31 @@ private <T> List<T> parseListOf(Token separator, Supplier<T> parser) {
...
+      res.add(parser.get());
+    }
+    return res;
+  }
+
+  private List<Parameter> parseListOfParameter(Token separator, Token terminator) {
     var res = new ArrayList<Parameter>();
+    if (lookahead == terminator) {
+      return res;
+    }
```
HT154, line 1745: This extra check is needed because some callers parse the element of the list themselves (while determining what construct it is) so it's possible for this function to be reached while at `terminator`.

pkl-core/src/test/files/LanguageSnippetTests/input/classes/constraintsTrailingComma.pkl
```diff
@@ -0,0 +1,9 @@
+res1: Int(isEven, isNonZero, isBetween(-10, 10), ) = 6
```
bioball: Move all of these to `input/parser` dir

pkl-core/src/test/files/LanguageSnippetTests/input/lambdas/lambdaTrailingCommas.pkl
```diff
@@ -0,0 +1,58 @@
...
+lB2Call = lB2.apply(1, 2, 3)
+lB3Call = lB3.apply(1, 2, 3)
+lC1Call = new Dynamic {} |> lC1
+lC2Call = new Dynamic {} |> lC2
+lC3Call = new Dynamic {} |> lC3
+lC4Call = new Dynamic {} |> lC4
+lD1Call = new Dynamic {} |> lD1
+lD2Call = new Dynamic {} |> lD2
+lD3Call = new Dynamic {} |> lD3
+lD4Call = new Dynamic {} |> lD4
```
bioball:
```suggestion
```

You can remove all of these; it's okay if the output is an empty file

pkl-core/src/test/files/LanguageSnippetTests/input/types/trailingCommas.pkl
```diff
@@ -0,0 +1,22 @@
...
+    >(a: A, b: B,): Value? = "\(a):\(b) - \(test)"
+}
+
+res1: Foo<String, String,>(!test.isEmpty) = new Foo<String, String,> {
+  test {
+    ["hello"] = "world"
+  }
+}
+
+res2 = res1.bar(123, "foo")
```
bioball: Same here; remove because this line doesn't have to do with parsing trailing commas

```suggestion
```

pkl-core/src/test/files/LanguageSnippetTests/input/types/trailingCommas.pkl
```diff
@@ -0,0 +1,22 @@
+class Foo<
+  Key,
+  Value,
+> {
```
bioball: I think let's add a one-off test in `org.pkl.core.parser` for class/method type params, and remove this one so that our expected output isn't "Only standard library members can have type parameters."

And you can add a comment there explaining why the test is implemented as a junit test (see `MultiLineStringLiteralTest`)

bioball (changes requested):
Just a few nits, but overall this looks good!

Would also like @stackoverflow to take a look and review the parser changes.

PR #1138, opened 2025-07-22
Un-deprecate Resource.base64, compute bytes/base64 in terms of each other

bioball:
This makes changes to avoid a needless breaking change. It preserves code like `new Resource { base64 = someValue }.sha256`, and also code that renders `Resource` into static formats.


HT154 (approved):

pkl-core/src/test/files/LanguageSnippetTests/input/api/Resource.pkl
```diff
@@ -16,16 +16,40 @@ examples {
...
   }
 
   ["sha256Int"] {
     empty.sha256Int
+    empty.bytes.sha256Int
     fox.sha256Int
+    fox.bytes.sha256Int
+  }
+
+  ["in-langauge"] {
```
stackoverflow:
```suggestion
  ["in-language"] {
```

stackoverflow (approved):

PR #1143, opened 2025-07-23
Defer noproxy to settings.pkl or PklProject if not set explicitly

### LINKED ISSUE #1142: CLI: `noProxy` option in settings.pkl and PklProject are ignored
When using the CLI, the noProxy settings set in `~/.pkl/settings.pkl` and `PklProject` are ignored.

bioball:
Fixes an issue where `http.noProxy` settings are ignored.

Closes #1142


HT154 (approved):
Good catch! Easy fix

PR #1152, opened 2025-07-25
Fix encoding for mapping with local members

bioball:
Fixes an issue where local members are included in the mapping entry count.


pkl-server/src/test/files/SnippetTests/output/map.yaml
```diff
@@ -41,4 +41,28 @@
...
+              - foo
+        bar: 
+          - 1
+          - Dynamic
+          - pkl:base
+          - 
+            - 
+              - 16
+              - name
+              - foobar
```
HT154, line 68: What's the output from this snippet without the bug fix?

bioball: You get some error, like `org.msgpack.core.MessageInsufficientBufferException`

HT154 (approved):

sin-ack (approved):
Tested with original occurrence (private), solves the issue.

PR #1161, opened 2025-07-30
Fix shell completion for paths

gordonbondon:


stackoverflow (approved):
Looks good. But now we need to be aware to add `CompletionCandidate.Path` to every new command that accepts files.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -187,9 +189,10 @@ class BaseOptions : OptionGroup() {
       .splitAll(File.pathSeparator)
 
   val settings: URI? by
-    option(names = arrayOf("--settings"), help = "Pkl settings module to use.").single().convert {
-      parseModuleName(it)
-    }
+    option(names = arrayOf("--settings"), help = "Pkl settings module to use.")
+      .path()
+      .single()
+      .convert { parseModuleName(it.toString()) }
```
bioball: See comment above

```suggestion
  val settings: URI? by
    option(
        names = arrayOf("--settings"),
        help = "Pkl settings module to use.",
        completionCandidates = CompletionCandidates.Path,
      )
      .single()
      .convert { parseModuleName(it) }
```

pkl-cli/src/main/kotlin/org/pkl/cli/commands/TestCommand.kt
```diff
@@ -31,7 +32,8 @@ class TestCommand : BaseCommand(name = "test", helpLink = helpLink) {
 
   val modules: List<URI> by
     argument(name = "modules", help = "Module paths or URIs to evaluate.")
-      .convert { BaseOptions.parseModuleName(it) }
+      .path()
+      .convert { BaseOptions.parseModuleName(it.toString()) }
```
bioball: Using `.path()` tells Clikt to parse this into a `java.nio.file.Path`, which can throw in some cases, and also doesn't round-trip back to string in other cases (`foo://bar/baz` turns into `foo:/bar/baz`)

I think we should just use completionCandidates here.

```kt
    argument(
        name = "modules",
        help = "Module paths or URIs to evaluate.",
        completionCandidates = CompletionCandidates.Path,
      )
      .convert { BaseOptions.parseModuleName(it) }
      .multiple()
```

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -166,6 +167,7 @@ class BaseOptions : OptionGroup() {
         names = arrayOf("-f", "--format"),
         help = "Output format to generate. <${output.joinToString()}>",
       )
+      .choice(*output)
```
bioball: It's possible to pass in a custom format string, as long as your Pkl module supports it (you just need to override base module's renderer).

So, we don't want this `choice` rule.

However, I think we can still provide completions using `CompletionCandidates.Fixed`, e.g. `completionCandidates = CompletionCandidates.Fixed(output.toSet())`

bioball (changes requested):

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
bioball: Still need to remove `choice()` here

```suggestion
```

PR #1169, opened 2025-08-08
Introduce pkl-doc model version 2

bioball:
Implementation of https://github.com/apple/pkl-evolution/pull/20

Currently, in order to update a pkl-doc documentation site, almost the entire existing site is read in order to update metadata like known versions, known subtypes, and more.

For example, adding a new version of a package requires that the existing runtime data of all existing versions be updated. Eventually, this causes the required storage size to balloon exponentially to the number of versions.

This addresses these limitations by:

1. Updating the runtime data structure to move "known versions" metadata to the package level (the same JSON file is used for all versions).
2. Eliminating known subtype and known usage information at a cross-package level.
3. Generating the search index by consuming the previously generated search index.
4. Generating the main page by consuming the search index.

Because this changes how runtime data is stored, an existing docsite needs to be migrated.

This also introduces a new migration command, `pkl-doc --migrate`, which transforms an older version of the website into a newer version.

NOTE: most of the additions are due to new input/output tests. I split those off into a separate commit. To review this PR, just take a look at the first commit.

The generated output changes now need to be served via an HTTP server (because of ES6 module imports).
You can review the generated site using a command like `python3 -m http.server -d pkl-doc/src/test/files/DocGeneratorTest/output/run-1/`).


pkl-doc/src/main/kotlin/org/pkl/doc/DocGenerator.kt
```diff
@@ -94,37 +100,40 @@ class DocGenerator(
...
+      }
       val htmlGenerator =
         HtmlGenerator(docsiteInfo, docPackages, importResolver, outputDir, isTestMode)
       val searchIndexGenerator = SearchIndexGenerator(outputDir)
       val packageDataGenerator = PackageDataGenerator(outputDir)
       val runtimeDataGenerator = RuntimeDataGenerator(descendingVersionComparator, outputDir)
 
       for (docPackage in docPackages) {
-        if (docPackage.isUnlisted) continue
-
```
bioball, line 110: Unlisted filtering logic is moved up to line 104

pkl-doc/src/main/kotlin/org/pkl/doc/PageGenerator.kt
```diff
@@ -127,10 +159,13 @@ internal abstract class PageGenerator<out S>(
...
-          id = "search-icon"
-          classes = setOf("material-icons")
-          +"search"
+        label {
+          htmlFor = "search-input"
+          i {
+            id = "search-icon"
+            classes = setOf("material-icons")
+            +"search"
+          }
```
bioball, line 172: Unrelated: address a warning that an input is missing a label

pkl-doc/src/main/kotlin/org/pkl/doc/RuntimeDataGenerator.kt
```diff
@@ -65,153 +61,82 @@ internal class RuntimeDataGenerator(
...
-
-    array {
-      for (item in items) {
-        obj {
-          name("text").value(text(item))
-          name("href").value(href(item))
-          name("classes").value(classes(item))
-        }
-      }
-    }
```
bioball, line 212: The `JsonWriter` and `JsonReader` APIs too low level; since we are also parsing runtime data, it's much nicer to use kotlinx json.

We don't want to use kotlinx json in pkl-core (which is partially why JsonWriter/JsonReader exist), but we don't have any issues using this library inside the pkl-doc tool.

pkl-doc/src/main/kotlin/org/pkl/doc/SearchIndexGenerator.kt
```diff
@@ -15,166 +15,296 @@
...
+  )
+
+  private fun List<SearchIndexEntry>.writeTo(path: Path) {
+    path.outputStream().use { stream ->
+      stream.write(PREFIX.toByteArray(StandardCharsets.UTF_8))
+      json.encodeToStream(this, stream)
+      stream.write(POSTFIX.toByteArray(StandardCharsets.UTF_8))
+      stream.flush()
+    }
+  }
```
bioball, line 110: We are converting the runtime data files from `.js` to `.json`. I chose to avoid doing that for the search index because it's not actually a problem right now, and takes more work (more migration needed).

pkl-doc/src/main/kotlin/org/pkl/doc/PageGenerator.kt
```diff
@@ -47,33 +52,61 @@ internal abstract class PageGenerator<out S>(
...
+      is PackageScope -> version
+      else -> parent?.getVersion()
+    }
+  }
+
   protected fun HTML.renderHtmlHead() {
+    lang = "en-US"
+
     head {
+      meta { charset = "UTF-8" }
```
bioball, line 82: Unrelated: address warnings emitted by linters

pkl-doc/src/main/kotlin/org/pkl/doc/RuntimeDataGenerator.kt
```diff
@@ -15,225 +15,206 @@
...
-    for (pkg in packages) {
-      writePackageFile(pkg.ref)
-      for (mod in pkg.modules) {
-        writeModuleFile(mod.ref)
-        for (clazz in mod.classes) {
-          writeClassFile(clazz.ref)
+  val writtenFiles = mutableSetOf<Path>()
+
+  private suspend fun writeData(packages: List<PackageData>) {
+    coroutineScope {
```
bioball, line 81: A `coroutineScope` will block until any jobs spawned by `launch` are completed. And each of these `launch` jobs are executed in parallel.

pkl-doc/src/main/kotlin/org/pkl/doc/PackageDataGenerator.kt
```diff
@@ -138,20 +172,32 @@ internal data class TypeRef(
...
     }
   }
 
+  override val packageRelativeDataPath: String
+    get() {
+      // typealiases don't have their own runtime data.
+      require(!isTypeAlias) { "typealiases don't have runtime data" }
+      return when {
+        isModuleClass -> "$module/index.json".pathEncoded
+        else -> "$module/$type.json".pathEncoded
```
HT154, line 195: Would pkldoc have a bad time with `class index` in a module? If so, it's probably worth catching that and throwing an error or rewriting it to something else.

bioball: Yeah, good catch, I think it'll conflict. I won't address that in this PR; this is an existing issue and I'd rather not block this PR on it.

PR #1175, opened 2025-08-13
Fix missing resources in native pkldoc, and disable test mode

bioball:
This fixes two issues:

1. Test mode is enabled in pkldoc without the ability to turn it off
2. Native pkldoc is missing required resources

This also adds tests for both `jpkldoc` and `pkldoc`.


HT154 (approved):

pkl-commons-test/src/main/kotlin/org/pkl/commons/test/Executables.kt
```diff
@@ -0,0 +1,85 @@
...
+    val linuxAmd64: Path = executable("pkldoc-linux-amd64")
+    val alpineAmd64: Path = executable("pkldoc-alpine-linux-amd64")
+    val windowsAmd64: Path = executable("pkldoc-windows-amd64.exe")
+
+    val javaExecutable: Path by lazy { javaExecutable("jpkldoc") }
+
+    // order (aarch64 before amd64, linux before alpine) affects [firstExisting]
+    override val allNative: List<Path> =
+      listOf(macAarch64, macAmd64, linuxAarch64, linuxAmd64, alpineAmd64, windowsAmd64)
+  }
```
bioball, line 84: Out of scope for this PR, but, we should add some tests for pkl-codegen-java and pkl-codegen-kotlin too.

pkl-doc/src/main/kotlin/org/pkl/doc/Main.kt
```diff
@@ -78,7 +81,7 @@ class DocCommand : BaseCommand(name = "pkldoc", helpLink = helpLink) {
       CliDocGeneratorOptions(
         baseOptions.baseOptions(modules, projectOptions),
         outputDir,
-        true,
+        isTestMode,
```
bioball, line 84: Setting test mode to true disables some optimizations. For example, normally, we don't type-check properties or method return types on standard library (external) methods. However, we do type-check them when run with test mode on.

For normal CLI usages, there should be no reason to enable this flag.

PR #1180, opened 2025-08-20
Fix cyclic dependencies bug

stackoverflow:
In some very specific scenarios cyclic dependencies may lead to a null pointer due to the module not being properly initialized.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
@@ -288,6 +288,10 @@ public class AstBuilder extends AbstractAstBuilder<Object> {
     isMethodReturnTypeChecked = !isStdLibModule || IoUtils.isTestMode();
   }
 
+  public ModuleInfo getModuleInfo() {
+    return moduleInfo;
+  }
+
   public static AstBuilder create(
       Source source,
       VmLanguage language,
--- a/pkl-core/src/main/java/org/pkl/core/ast/member/ClassNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/member/ClassNode.java
@@ -99,7 +99,10 @@ public final class ClassNode extends ExpressionNode {
       // nodes
       // via static final fields without having to fear recursive field initialization.
       prototype = module;
-      prototype.setExtraStorage(moduleInfo);
+      // Only set ModuleInfo if it hasn't been set already (to handle cyclic dependencies)
+      if (!prototype.hasExtraStorage()) {
+        prototype.setExtraStorage(moduleInfo);
+      }
       prototype.addProperties(prototypeMembers);
     } else {
       prototype =
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmLanguage.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmLanguage.java
@@ -108,6 +108,11 @@ public final class VmLanguage extends TruffleLanguage<VmContext> {
     var builder =
         AstBuilder.create(
             source, this, moduleContext, moduleKey, resolvedModuleKey, moduleResolver);
+
+    // Set the ModuleInfo on the empty module immediately to handle cyclic dependencies.
+    // This ensures it works even if the module is accessed during initialization via imports
+    emptyModule.setExtraStorage(builder.getModuleInfo());
+
     var moduleNode = builder.visitModule(moduleContext);
     moduleNode.getCallTarget().call(emptyModule, emptyModule);
     MinPklVersionChecker.check(emptyModule, importNode);


HT154 (approved):

bioball (changes requested):
I'm not sure if this is the correct fix for this bug.

Pkl allows circular imports (except in cases where a module's type is defined in terms of itself, like two modules extending each other).

Also, amending/extending modules expose types defined on their parent.

I created a reproducer for this bug in this issue here: https://github.com/apple/pkl/issues/1183

I think either:
1. We throw a meaningful error about some circular dependency
1. We fix some deeper underlying issue so that this code works

BTW: can we add the reproducer as a test case here? You can add them to the `input-helpers` dir.

pkl-core/src/main/java/org/pkl/core/ast/type/ResolveDeclaredTypeNode.java
```diff
@@ -73,11 +74,36 @@ protected VmTyped getImport(
     var result = module.getCachedValue(importName);
     if (result == null) {
       result = callNode.call(member.getCallTarget(), module, module, importName);
+
+      var importedModule = (VmTyped) result;
+      if (importedModule.isNotInitialized()
+          && importedModule.isModuleObject()
```
bioball: This should always be a module object, right?

pkl-core/src/main/java/org/pkl/core/ast/type/ResolveDeclaredTypeNode.java
```diff
@@ -73,11 +74,36 @@ protected VmTyped getImport(
...
+          && importedModule.getModuleInfo().isAmend()) {
+        // this is an amending module. Try to find the prototype
+        var prototypeModule = findPrototypeModule(importedModule);
+        if (prototypeModule != null) {
+          result = prototypeModule;
+        }
+      }
+
       module.setCachedValue(importName, result);
```
bioball, line 92: We shouldn't be cacheing this under the same name as the import value. Otherwise, this snippet breaks:

```pkl
import "amendingFoo.pkl"

// here, `amendingFoo` should resolve to the prototype
res: amendingFoo.Foo

// here, `amendingFoo` should resolve to the import itself
bar = amendingFoo
```

pkl-core/src/main/java/org/pkl/core/ast/type/ResolveDeclaredTypeNode.java
```diff
@@ -73,11 +74,36 @@ protected VmTyped getImport(
...
     return (VmTyped) result;
   }
 
+  private @Nullable VmTyped findPrototypeModule(VmTyped amendingModule) {
+    try {
+      var moduleInfo = amendingModule.getModuleInfo();
+      var amendedModuleKey = moduleInfo.getAmendedModuleKey();
+
+      if (amendedModuleKey != null) {
+        return VmLanguage.get(null).loadModule(amendedModuleKey);
```
bioball: The amended module key itself may be amending another module; we need to keep going until we find the first non-amending module. Can we add a test for this?

Also, pass the node in when looking up the reference

```suggestion
        return VmLanguage.get(this).loadModule(amendedModuleKey, this);
```

pkl-core/src/main/java/org/pkl/core/runtime/ModuleInfo.java
```diff
@@ -179,4 +201,11 @@ public ModuleSchema getModuleSchema(VmTyped module) {
   public boolean isAmend() {
     return isAmend;
   }
+
+  /**
+   * Returns the ModuleKey of the module being amended, or null if this is not an amending module.
```
bioball:
```suggestion
   * Returns the {@link ModuleKey} of the module being amended, or null if this is not an amending module.
```

pkl-core/src/main/java/org/pkl/core/ast/type/ResolveDeclaredTypeNode.java
```diff
@@ -73,11 +74,36 @@ protected VmTyped getImport(
...
+      var importedModule = (VmTyped) result;
+      if (importedModule.isNotInitialized()
+          && importedModule.isModuleObject()
+          && importedModule.getModuleInfo().isAmend()) {
+        // this is an amending module. Try to find the prototype
+        var prototypeModule = findPrototypeModule(importedModule);
+        if (prototypeModule != null) {
+          result = prototypeModule;
+        }
+      }
```
bioball, line 90: The result of `getImport` should always be the prototype in all cases, even when the module has not yet been initialized.

pkl-core/src/main/java/org/pkl/core/ast/type/ResolveDeclaredTypeNode.java
```diff
@@ -73,11 +74,36 @@ protected VmTyped getImport(
...
+  private @Nullable VmTyped findPrototypeModule(VmTyped amendingModule) {
+    try {
+      var moduleInfo = amendingModule.getModuleInfo();
+      var amendedModuleKey = moduleInfo.getAmendedModuleKey();
+
+      if (amendedModuleKey != null) {
+        return VmLanguage.get(null).loadModule(amendedModuleKey);
+      }
+    } catch (Exception ignored) {
+    }
```
bioball, line 106: Why ignore all exceptions here?

bioball (changes requested):
Good start, but this needs to be refined some. See comments

pkl-core/src/main/java/org/pkl/core/ast/type/ResolveDeclaredTypeNode.java
stackoverflow: Indeed. Removed the check.

pkl-core/src/main/java/org/pkl/core/ast/type/ResolveDeclaredTypeNode.java
stackoverflow: Good catch. `cycles2.pkl` tests that (fails if you cache).

pkl-core/src/main/java/org/pkl/core/ast/type/ResolveDeclaredTypeNode.java
stackoverflow: Indeed! I changed the test to have another layer of indirection.

pkl-core/src/main/java/org/pkl/core/ast/type/ResolveDeclaredTypeNode.java
stackoverflow: Yeah, if there's an error here it should just bubble up. Removed the try.

PR #1196, opened 2025-08-31
Add support for customizing HTTP headers

### LINKED ISSUE #633: Pkl package using private repo (using authentication)
We were trying to host our pkl releases in a private gitlab repo. However according to this discussion 

https://github.com/apple/pkl/discussions/632 that's not possible.

Could I suggest adding a method to allow for using authentication with packages?

kyokuping:
This PR adds support for custom HTTP headers, introducing a `--http-header` CLI flag to accept `key=value` pairs. These headers can also be specified within the `setting.pkl` file.

Closes #633

SPICE: https://github.com/apple/pkl-evolution/pull/24

### DIFF
--- a/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
+++ b/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
@@ -33,6 +33,7 @@ import org.pkl.commons.cli.CliException
 import org.pkl.commons.shlex
 import org.pkl.core.Pair as PPair
 import org.pkl.core.evaluatorSettings.Color
+import org.pkl.core.evaluatorSettings.PklEvaluatorSettings
 import org.pkl.core.evaluatorSettings.PklEvaluatorSettings.ExternalReader
 import org.pkl.core.evaluatorSettings.TraceMode
 import org.pkl.core.runtime.VmUtils
@@ -304,11 +305,19 @@ class BaseOptions : OptionGroup() {
         try {
           val uri = URI(uriStr.trim())
 
+          val headerRegex = Regex("""^(.+?):[ \t]*(.+)$""")
           val headerPairs =
             headers.split(',').map { header ->
-              val headerParts = header.split(":", limit = 2)
-              require(headerParts.size == 2) { "Header '$header' is not in 'name:value' format. " }
-              PPair(headerParts[0], headerParts[1])
+              val (headerName, headerValue) =
+                headerRegex.find(header)?.destructured
+                  ?: fail("Header '$header' is not in 'name:value' format.")
+              require(PklEvaluatorSettings.HEADER_NAME_REGEX.matcher(headerName).matches()) {
+                "HTTP header name '$headerName' has invalid syntax."
+              }
+              require(PklEvaluatorSettings.HEADER_VALUE_REGEX.matcher(headerValue).matches()) {
+                "HTTP header value '$headerValue' has invalid syntax"
+              }
+              PPair(headerName, headerValue)
             }
           uri to headerPairs
         } catch (e: IllegalArgumentException) {
--- a/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
+++ b/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
@@ -285,6 +285,14 @@ class BaseOptions : OptionGroup() {
       .multiple()
       .toMap()
 
+  val httpHeaders: Map<String, String> by
+    option(
+        names = arrayOf("--http-headers"),
+        metavar = "key=value",
+        help = "HTTP header to add to the request.",
+      )
+      .associate()
+
   val externalModuleReaders: Map<String, ExternalReader> by
     option(
         names = arrayOf("--external-module-reader"),
@@ -351,6 +359,7 @@ class BaseOptions : OptionGroup() {
       httpProxy = proxy,
       httpNoProxy = noProxy,
       httpRewrites = httpRewrites.ifEmpty { null },
+      httpHeaders = httpHeaders.ifEmpty { null },
       externalModuleReaders = externalModuleReaders,
       externalResourceReaders = externalResourceReaders,
       traceMode = traceMode,
--- a/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
+++ b/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
@@ -31,6 +31,7 @@ import java.util.regex.Pattern
 import org.pkl.commons.cli.CliBaseOptions
 import org.pkl.commons.cli.CliException
 import org.pkl.commons.shlex
+import org.pkl.core.Pair as PPair
 import org.pkl.core.evaluatorSettings.Color
 import org.pkl.core.evaluatorSettings.PklEvaluatorSettings.ExternalReader
 import org.pkl.core.evaluatorSettings.TraceMode
@@ -285,13 +286,49 @@ class BaseOptions : OptionGroup() {
       .multiple()
       .toMap()
 
-  val httpHeaders: Map<String, String> by
+  val httpHeaders: Map<URI, List<PPair<String, String>>> by
     option(
         names = arrayOf("--http-headers"),
-        metavar = "key=value",
+        metavar = "<uri>=<header name>:<header value>[,<header name>:<header value>...]",
         help = "HTTP header to add to the request.",
       )
-      .associate()
+      .convert { it ->
+        val (uriStr, headers) =
+          it.split("=", limit = 2).let { parts ->
+            require(parts.size == 2) {
+              "Headers must be in the form of <prefix>=<header name>:<header value>"
+            }
+            parts[0] to parts[1]
+          }
+
+        try {
+          val uri = URI(uriStr.trim())
+
+          val headerPairs =
+            headers.split(',').map { header ->
+              val headerParts = header.split(":", limit = 2)
+              require(headerParts.size == 2) { "Header '$header' is not in 'name:value' format. " }
+              PPair(headerParts[0], headerParts[1])
+            }
+          uri to headerPairs
+        } catch (e: IllegalArgumentException) {
+          fail(e.message!!)
+        } catch (e: URISyntaxException) {
+          val message = buildString {
+            append("HTTP headers target `${e.input}` has invalid syntax (${e.reason}).")
+            if (e.index > -1) {
+              append("\n\n")
+              append(e.input)
+              append("\n")
+              append(" ".repeat(e.index))
+              append("^")
+            }
+          }
+          fail(message)
+        }
+      }
+      .multiple()
+      .toMap()
 
   val externalModuleReaders: Map<String, ExternalReader> by
     option(
--- a/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
+++ b/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
@@ -308,13 +308,9 @@ class BaseOptions : OptionGroup() {
                 ?: fail("Header '$header' is not in 'name:value' format.")
             IoUtils.validateHeaderName(headerName)
             IoUtils.validateHeaderValue(headerValue)
-            val headerPair = PPair(headerName, headerValue)
-            val headerPairList = headersMap[stringPattern]
-            if (headerPairList == null) {
-              headersMap[stringPattern] = mutableListOf(headerPair)
-            } else {
-              headerPairList.add(headerPair)
-            }
+            headersMap
+              .computeIfAbsent(stringPattern) { mutableListOf() }
+              .add(PPair(headerName, headerValue))
           }
 
           headersMap.entries.map { PPair(GlobResolver.toRegexPattern(it.key), it.value) }
--- a/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
+++ b/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
@@ -33,10 +33,10 @@ import org.pkl.commons.cli.CliException
 import org.pkl.commons.shlex
 import org.pkl.core.Pair as PPair
 import org.pkl.core.evaluatorSettings.Color
-import org.pkl.core.evaluatorSettings.PklEvaluatorSettings
 import org.pkl.core.evaluatorSettings.PklEvaluatorSettings.ExternalReader
 import org.pkl.core.evaluatorSettings.TraceMode
 import org.pkl.core.runtime.VmUtils
+import org.pkl.core.util.GlobResolver
 import org.pkl.core.util.IoUtils
 
 @Suppress("MemberVisibilityCanBePrivate")
@@ -95,6 +95,9 @@ class BaseOptions : OptionGroup() {
         Pair(it.first, ExternalReader(cmd.first(), cmd.drop(1)))
       }
     }
+
+    val HEADER_NAME_REGEX = Pattern.compile("^[a-zA-Z0-9!#$%&'*+-.^_`|~]+$")
+    val HEADER_VALUE_REGEX = Pattern.compile("^[\\t\\u0020-\\u007E\\u0080-\\u00FF]*$")
   }
 
   private val defaults = CliBaseOptions()
@@ -287,14 +290,14 @@ class BaseOptions : OptionGroup() {
       .multiple()
       .toMap()
 
-  val httpHeaders: Map<URI, List<PPair<String, String>>> by
+  val httpHeaders: List<PPair<Pattern, List<PPair<String, String>>>> by
     option(
         names = arrayOf("--http-headers"),
-        metavar = "<uri>=<header name>:<header value>[,<header name>:<header value>...]",
+        metavar = "<url-pattern>=<header name>:<header value>[,<header name>:<header value>...]",
         help = "HTTP header to add to the request.",
       )
       .convert { it ->
-        val (uriStr, headers) =
+        val (stringPattern, headers) =
           it.split("=", limit = 2).let { parts ->
             require(parts.size == 2) {
               "Headers must be in the form of <prefix>=<header name>:<header value>"
@@ -303,7 +306,7 @@ class BaseOptions : OptionGroup() {
           }
 
         try {
-          val uri = URI(uriStr.trim())
+          var pattern = GlobResolver.toRegexPattern(stringPattern)
 
           val headerRegex = Regex("""^(.+?):[ \t]*(.+)$""")
           val headerPairs =
@@ -311,15 +314,15 @@ class BaseOptions : OptionGroup() {
               val (headerName, headerValue) =
                 headerRegex.find(header)?.destructured
                   ?: fail("Header '$header' is not in 'name:value' format.")
-              require(PklEvaluatorSettings.HEADER_NAME_REGEX.matcher(headerName).matches()) {
+              require(HEADER_NAME_REGEX.matcher(headerName).matches()) {
                 "HTTP header name '$headerName' has invalid syntax."
               }
-              require(PklEvaluatorSettings.HEADER_VALUE_REGEX.matcher(headerValue).matches()) {
+              require(HEADER_VALUE_REGEX.matcher(headerValue).matches()) {
                 "HTTP header value '$headerValue' has invalid syntax"
               }
               PPair(headerName, headerValue)
             }
-          uri to headerPairs
+          PPair(pattern, headerPairs)
         } catch (e: IllegalArgumentException) {
           fail(e.message!!)
         } catch (e: URISyntaxException) {
@@ -337,7 +340,6 @@ class BaseOptions : OptionGroup() {
         }
       }
       .multiple()
-      .toMap()
 
   val externalModuleReaders: Map<String, ExternalReader> by
     option(
--- a/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
+++ b/pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
@@ -300,28 +300,26 @@ class BaseOptions : OptionGroup() {
       .transformAll { it ->
         val headersMap = mutableMapOf<String, MutableList<PPair<String, String>>>()
 
-        for ((stringPattern, header) in it) {
-          val headerRegex = Regex("""^(.+?):[ \t]*(.+)$""")
-          val (headerName, headerValue) =
-            headerRegex.find(header)?.destructured
-              ?: fail("Header '$header' is not in 'name:value' format.")
-          require(HEADER_NAME_REGEX.matcher(headerName).matches()) {
-            "HTTP header name '$headerName' has invalid syntax."
-          }
-          require(HEADER_VALUE_REGEX.matcher(headerValue).matches()) {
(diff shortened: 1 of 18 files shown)


HT154 (commented):
Thanks for the contribution! 🎉

I have a few primary concerns here:
* Although rare, it is valid to send a specific header more than once in a request. When there are headers with the same name their order may also matter.
* How does your implementation handle "reserved" headers that users should not override like `Origin` or `User-Agent`?
* It seems less than ideal to send the same headers to all hosts, eg. if host A requires auth but host B does not, Pkl should not send a configured `Authorization` header to both hosts. I think a prefix matching scheme like the one used for rewrites is necessary here to ensure the proper security controls are in place.

To this end, I think the Pkl API for this needs to be something along these lines:
```pkl
headers: Mapping<HttpRewrite, Listing<Pair<String, String>>>
```

This complicates the CLI a bit, of course, but maybe something like this is reasonable?
```
--http-header <prefix>='<header name>: <header value>'
```

stdlib/EvaluatorSettings.pkl
```diff
@@ -164,6 +164,9 @@ class Http {
   /// (not schematically enforced).
   @Since { version = "0.29.0" }
   rewrites: Mapping<HttpRewrite, HttpRewrite>?
+  
+  /// HTTP headers to add to outbound requests targeting specified URLs.
+  headers: Mapping<HttpRewrite, Listing<Pair<String, String>>>?
 }
```
HT154: We need the `@Since` annotation here for documentation purposes. It would also be good to add specificity and examples to the doc comment, similar to `rewrites` above.

```suggestion
  /// HTTP headers to add to outbound requests targeting specified URLs.
  @Since { version = "0.30.0" }
  headers: Mapping<HttpRewrite, Listing<Pair<String, String>>>?
}
```

HT154: I'm also curious if it makes sense to rename `HttpRewrite` to `HttpPrefix` (with typealias marked `@Deprecated` for backwards compatibility) and refactoring the corresponding Java APIs. Interested in input from @bioball and @stackoverflow on this.

HT154:
Just a heads up, we want reviews from the entire core team because this feature affects the core language and has security implications. One member of the team is currently out on leave so it's going to be a little bit before we can fully review this, but we haven't forgotten about it!

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -274,6 +275,50 @@ class BaseOptions : OptionGroup() {
...
+          }
+
+        try {
+          val uri = URI(uriStr.trim())
+
+          val headerPairs =
+            headers.split(',').map { header ->
+              val headerParts = header.split(":", limit = 2)
+              require(headerParts.size == 2) { "Header '$header' is not in 'name:value' format. " }
+              PPair(headerParts[0], headerParts[1])
```
HT154: `curl` accepts this same syntax, but also accepts `<name>:[ \t]+<value>` and appears to strip the leading (but not trailing) whitespace from header values before sending the request. To avoid surprise for users of Pkl, it might be good to do the same. In theory, HTTP servers should be ignoring this leading (and trailing) whitespace when parsing headers so it _shouldn't_ matter, but it might be best to avoid taking chances here.

I'm also wondering if it makes sense to do some validation of characters in headers here (and in the `EvaluatorSettings` API) to eagerly report invalid characters instead of lazily doing so when a matching request is made for the first time.

stackoverflow (commented):
The design sounds sensible at first glance, but due to the complexity and security issues it may cause it may be a good idea to create a SPICE for it.
What do you think @bioball, @HT154?

pkl-core/src/main/java/org/pkl/core/http/JdkHttpClient.java
```diff
@@ -77,21 +81,31 @@ final class JdkHttpClient implements HttpClient {
...
-      return underlying.send(request, responseBodyHandler);
+      var wrappedRequestBuilder = HttpRequest.newBuilder(request, (name, value) -> true);
+      for (var entry : headers.entrySet()) {
+        if (RequestRewritingClient.matchesRewriteRule(request.uri(), entry.getKey())) {
+          for (var value : entry.getValue()) {
+            wrappedRequestBuilder.header(value.getFirst(), value.getSecond());
+          }
+        }
+      }
+      return underlying.send(wrappedRequestBuilder.build(), responseBodyHandler);
```
bioball: The logic of adding headers should be handled in `RequestRewritingClient`

bioball (commented):
Hey, thanks for the PR! I agree with @stackoverflow; it'd be good to have this discussed in a SPICE.

If you'd like to drive this feature, can you write one up? It should be submitted as PR to [pkl-evolution](https://github.com/apple/pkl-evolution).

Your SPICE should start by copy/pasting the [template](https://github.com/apple/pkl-evolution/blob/main/spice-templates/SPICE-0000-pkl-template.adoc). Your SPICE number would be the next number after the latest proposed SPICE see (https://github.com/apple/pkl-evolution/pulls)

bioball:
@kyokuping was the close intentional?

kyokuping:
Sorry, accidentally closed the PR. 😢

bioball:
Gotcha, feel free to re-open a new PR!

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -285,6 +290,57 @@ class BaseOptions : OptionGroup() {
...
+            }
+            parts[0] to parts[1]
+          }
+
+        try {
+          var pattern = GlobResolver.toRegexPattern(stringPattern)
+
+          val headerRegex = Regex("""^(.+?):[ \t]*(.+)$""")
+          val headerPairs =
+            headers.split(',').map { header ->
```
HT154: Use `,` as the separator here is problematic for header values that can contain commas (e.g. `Cookie`). Some potential ways to account for this:
* A different separator: doesn't seem like there is one that is both header-value-unsafe and shell-safe on first glance
* Treat `\,` as an escape to avoid splitting: non-standard, has some undesirable traits around copy-pasting values
* Only accept one header per occurrence of the option and instead require repeating the pattern: this is closer to the way curl accepts headers on the CLI, but less DRY for users

pkl-core/src/main/java/org/pkl/core/http/JdkHttpClient.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
```
HT154: No other changes in this file so the date bump should be reverted.

stdlib/EvaluatorSettings.pkl
```diff
@@ -169,6 +172,10 @@ class Http {
   /// (not schematically enforced).
   @Since { version = "0.29.0" }
   rewrites: Mapping<HttpRewrite, HttpRewrite>?
+
+  /// HTTP headers to add to outbound requests targeting specified URLs.
+  @Since { version = "0.31.0" }
+  headers: Mapping<UrlPattern, Mapping<HttpHeaderName, *Listing<HttpHeaderValue> | HttpHeaderValue>>?
```
HT154: All the `@Since` annotations need a bump!

```suggestion
  @Since { version = "0.32.0" }
  headers: Mapping<UrlPattern, Mapping<HttpHeaderName, *Listing<HttpHeaderValue> | HttpHeaderValue>>?
```

stdlib/EvaluatorSettings.pkl
```diff
@@ -235,3 +242,54 @@ class ExternalReader {
...
+local const httpHeaderNameRegex = Regex("^[a-zA-Z0-9!#\\$%&'*+-.^_`|~]+$")
+local const hasValidHttpHeaderName = (header: String) ->
+  !httpHeaderNameRegex.findMatchesIn(header).isEmpty
+
+@Since { version = "0.31.0" }
+typealias HttpHeaderName =
+  String(
+    (it: String) -> it == toLowerCase(),
+    (it: String) -> !(it is ReservedHttpHeaderName),
+    (it: String) -> !hasReservedHttpHeaderPrefix.apply(it),
```
HT154:
```suggestion
    this == toLowerCase(),
    !(this is ReservedHttpHeaderName),
    !hasReservedHttpHeaderPrefix.apply(this),
```

stdlib/EvaluatorSettings.pkl
```diff
@@ -235,3 +242,54 @@ class ExternalReader {
...
+    hasValidHttpHeaderName,
+  )
+
+local const httpHeaderValueRegex = Regex("^[\\t\\u0020-\\u007E\\u0080-\\u00FF]*$")
+local const hasValidHttpHeaderValue = (value: String) ->
+  !httpHeaderValueRegex.findMatchesIn(value).isEmpty
+
+@Since { version = "0.31.0" }
+typealias HttpHeaderValue = String(hasValidHttpHeaderValue)
```
HT154: Not sure the lambda adds a ton of value here over putting the expression directly in the type constraint.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -285,6 +290,57 @@ class BaseOptions : OptionGroup() {
...
+              val (headerName, headerValue) =
+                headerRegex.find(header)?.destructured
+                  ?: fail("Header '$header' is not in 'name:value' format.")
+              require(HEADER_NAME_REGEX.matcher(headerName).matches()) {
+                "HTTP header name '$headerName' has invalid syntax."
+              }
```
HT154: Unlike the project/settings method of setting headers, the CLI doesn't exclude specific headers names and prefixes. It would be good to make these consistent.

HT154 (commented):
Hey @kyokuping would you mind rebasing this to resolve the conflict(s)?

I have some pretty minor feedback, but from a technical standpoint this looks pretty good. I'd like to try and ship this in 0.32 in a few months, so it'd be great if this was ready for another set of eyes by the time the core team's all back together in a few weeks!

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -285,6 +290,41 @@ class BaseOptions : OptionGroup() {
...
+                ?: fail("Header '$header' is not in 'name:value' format.")
+            IoUtils.validateHeaderName(headerName)
+            IoUtils.validateHeaderValue(headerValue)
+            val headerPair = PPair(headerName, headerValue)
+            val headerPairList = headersMap[stringPattern]
+            if (headerPairList == null) {
+              headersMap[stringPattern] = mutableListOf(headerPair)
+            } else {
+              headerPairList.add(headerPair)
+            }
```
HT154: Nit: In this scenario I tend to find that `computeIfAbsent` can help make this concise without losing clarity:
```suggestion
            headersMap
              .computeIfAbsent(stringPattern) { mutableListOf() }
              .add(PPair(headerName, headerValue))
```

pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
```diff
@@ -854,4 +884,41 @@ public static void validateRewriteRule(URI rewrite) {
...
+
+  private static boolean hasReservedHeaderPrefix(String headerName) {
+    return Arrays.stream(reservedHeaderPrefixs).anyMatch((prefix) -> headerName.startsWith(prefix));
+  }
+
+  public static void validateHeaderName(String headerName) {
+    if (!headerName.equals(headerName.toLowerCase())) {
+      throw new IllegalArgumentException(
+          "HTTP header '%s' should be all lowercase".formatted(headerName));
+    }
```
HT154: Is this validation actually necessary? Why would this be preferable to normalizing header names during CLI parsing?

kyokuping: Based on the previous review on the SPICE, I updated the settings to enforce lowercase values and therefore expected the CLI to match this behavior. If we want to change this in the CLI, we should probably align the EvaluatorSettings implementation too. Which way do you prefer to go with?

HT154: In the evaluator settings, "pre"-normalization is required to ensure the `Mapping` has unique keys. On the CLI, each instance of the option is parsed linearly, so the CLI could perform its own normalization without issue.

In general, I think it's best that we accept the broadest input possible for each input method (even if those are not consistent). It's very possible the other core folks could disagree, so I'm happy to put a pin in this until they're back over the next few weeks.

kyokuping: Hey, had any chance discussing around this?

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -285,6 +290,37 @@ class BaseOptions : OptionGroup() {
...
+        metavar = "<url-pattern>=<header name>:<header value>",
+        help = "HTTP header to add to the request.",
+      )
+      .splitPair()
+      .transformAll { it ->
+        val headersMap = mutableMapOf<String, MutableList<PPair<String, String>>>()
+
+        try {
+          for ((stringPattern, header) in it) {
+            val headerRegex = Regex("""^(.+?):[ \t]*(.+)$""")
```
stackoverflow:
```suggestion
          val headerRegex = Regex("""^(.+?):[ \t]*(.+)$""")
          for ((stringPattern, header) in it) {
```
Wasteful to recompile the Regex on every iteration. Even better to move it to the companion object.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -93,6 +95,9 @@ class BaseOptions : OptionGroup() {
         Pair(it.first, ExternalReader(cmd.first(), cmd.drop(1)))
       }
     }
+
+    val HEADER_NAME_REGEX = Pattern.compile("^[a-zA-Z0-9!#$%&'*+-.^_`|~]+$")
```
stackoverflow: This is duplicated verbatim from `IoUtils`. You can make it public there and use it here.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -93,6 +95,9 @@ class BaseOptions : OptionGroup() {
...
+    val HEADER_VALUE_REGEX = Pattern.compile("^[\\t\\u0020-\\u007E\\u0080-\\u00FF]*$")
```
stackoverflow: Same here (duplicated from IoUtils).

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -285,6 +290,37 @@ class BaseOptions : OptionGroup() {
       .multiple()
       .toMap()
 
+  val httpHeaders: List<PPair<Pattern, List<PPair<String, String>>>> by
```
stackoverflow, line 290: This is more of a nit, but: `List<PPair<Pattern, List<PPair<String, String>>>>` is a quite generic type that doesn't convey much information. It would be better to have a properly-named record that stores the URL glob and the header name-value pairs.

pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
```diff
@@ -52,6 +52,36 @@ public final class IoUtils {
...
+    "permissions-policy",
+    "referer",
+    "te",
+    "trailer",
+    "transfer-encoding",
+    "upgrade",
+    "via"
+  };
+
+  private static final String[] reservedHeaderPrefixs = {"proxy-", "sec-", "access-control-"};
```
stackoverflow:
```suggestion
  private static final String[] reservedHeaderPrefixes = {"proxy-", "sec-", "access-control-"};
```

pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
```diff
@@ -52,6 +52,36 @@ public final class IoUtils {
...
+  private static final Pattern headerNameLike = Pattern.compile("^[a-zA-Z0-9!#$%&'*+-.^_`|~]+$");
+
+  private static final Pattern headerValueLike =
+      Pattern.compile("^[\\t\\u0020-\\u007E\\u0080-\\u00FF]*$");
+
+  private static final String[] reservedHeaderNames = {
+    "accept-charset",
+    "accept-encoding",
+    "access-control-request-headers",
+    "access-control-request-method",
```
stackoverflow: Do you need these here if you already have `"access-control-"` in the `reservedHeaderPrefixs`?

pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
```diff
@@ -854,4 +884,41 @@ public static void validateRewriteRule(URI rewrite) {
...
+    }
+
+    if (!headerNameLike.matcher(headerName).matches()) {
+      throw new IllegalArgumentException(
+          "HTTP header name '%s' has an invalid syntax".formatted(headerName));
+    }
+  }
+
+  public static void validateHeaderValue(String headerValue) {
+    if (headerValueLike.matcher(headerValue).matches()) {
```
stackoverflow:
```suggestion
    if (!headerValueLike.matcher(headerValue).matches()) {
```
I'm pretty sure your `headerValueLike` is a positive regex, so **not** matching it should throw an error.

pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
bioball: We just chatted on this; we've settled on: we don't think it's necessary to require any special rules for header names when configured using CLI flags.

pkl-core/src/main/java/org/pkl/core/evaluatorSettings/PklEvaluatorSettings.java
```diff
@@ -149,8 +156,37 @@ public record Http(@Nullable Proxy proxy, @Nullable Map<URI, URI> rewrites) {
...
+                              return ((List<String>) value)
+                                  .stream().map(v -> new Pair(header.getKey(), v));
+                            } else {
+                              return Stream.of(new Pair(header.getKey(), value));
+                            }
+                          })
+                      .toList();
+              parsedHeaderDefs.add(new Pair(urlPattern, pairs));
+            } catch (InvalidGlobPatternException e) {
+              throw new PklException(ErrorMessages.create("invalidUri", stringPattern));
```
bioball, line 185:
```suggestion
              throw new PklException(ErrorMessages.create("invalidGlobPattern", stringPattern));
```

pkl-core/src/main/java/org/pkl/core/http/HttpClientBuilder.java
```diff
@@ -110,6 +113,12 @@ public Builder addRewrite(URI sourcePrefix, URI targetPrefix) {
     return this;
   }
 
+  @Override
+  public Builder setHeaders(List<Pair<Pattern, List<Pair<String, String>>>> headers) {
```
bioball, line 117: I think it would be better if this signature matched the in-language evaluator settings.

I think this should be something like:

```java
setHeaderRules(Map<String, Map<String, List<String>>> headerRules)
```

Where the top-level map key is the glob pattern.
This method can throw `IllegalArgumentException` if a glob pattern is invalid.

And, I think we should also have:

```java
addHeaderRule(String globPattern, Map<String, List<String>> headers)
```

It'd also be good to avoid using `Pair` here.

stdlib/EvaluatorSettings.pkl
```diff
@@ -235,3 +242,51 @@ class ExternalReader {
...
+    | "trailer"
+    | "transfer-encoding"
+    | "upgrade"
+    | "via"
+
+local const ReservedHttpHeaderPrefix = new Listing {
+  "proxy-"
+  "sec-"
+  "access-control-"
+}
```
bioball:
```suggestion
local const reservedHttpHeaderPrefixes = Set
  "proxy-",
  "sec-",
  "access-control-",
)
```

stdlib/EvaluatorSettings.pkl
```diff
@@ -128,6 +128,9 @@ local const hasNonEmptyHostname = (it: String) ->
 @Since { version = "0.29.0" }
 typealias HttpRewrite = String(startsWith(Regex("https?://")), endsWith("/"), hasNonEmptyHostname)
 
+@Since { version = "0.32.0" }
+typealias UrlPattern = String(endsWith(Regex("[/*]")))
```
HT154, line 132: I think this validation needs to be tightened:
```suggestion
typealias UrlPattern = String(endsWith(Regex(#"/\*{0,2}"#)))
```

The idea here is that attempts to match a prefix should always end on a path boundary (exact or prefix) to avoid accidentally matching unintended origins, e.g. pattern `*.example.com` matches `foo.example.company.com`, which could leak credentials.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -285,6 +287,37 @@ class BaseOptions : OptionGroup() {
...
+        try {
+          val headerRegex = Regex("""^(.+?):[ \t]*(.+)$""")
+          for ((stringPattern, header) in it) {
+            val (headerName, headerValue) =
+              headerRegex.find(header)?.destructured
+                ?: fail("Header '$header' is not in 'name:value' format.")
+            IoUtils.validateHeaderName(headerName)
+            IoUtils.validateHeaderValue(headerValue)
+            headersMap
+              .computeIfAbsent(stringPattern) { mutableListOf() }
```
HT154, line 309: The pattern here isn't validated by the same rule as the pattern values in `pkl:EvaluatorSettings`

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
HT154: I agree this type's gotten a bit out of hand and a record would be a welcome improvement. I'm not eager to block this PR more than necessary, so I'm fine leaving this as follow-up work.

HT154 (commented):
I think this is getting real close now. From my perspective, just a little bit to hash out with URI patterns.

stdlib/EvaluatorSettings.pkl
bioball: I actually think we don't need to tighten this. Unlike string prefix matching, glob expressions are unlikely to be accidentally misconfigured.

These two are clear enough, in my opinion:

`https://example.com**` -- match everything that starts with `example.com`
`https://example.com/**` --  match everything within the host `example.com`

I do wonder if we want a `external boolean isGlobPattern` on class String, so this can be:

```
String(isGlobPattern)
```

HT154: Hmm I still think it would be ideal to prevent prefix matching on hosts, but I do see what you mean. I like the idea of `isGlobPattern` to mirror `isRegex`.

bioball (approved):
This is in a good enough place to merge; there's some things that I plan on addressing in a follow-up PR!

HT154 (approved):
With @bioball on this one.

Thanks again for all the work on this, @kyokuping! We really appreciate the contribution.

PR #1203, opened 2025-09-12
SPICE-0021 Implement Pkl binary renderer and parser

### LINKED ISSUE #994: Binary renderer/parser for Pkl
Provide an in-language binary renderer/parser, which conforms to the [Pkl binary encoding](https://pkl-lang.org/main/current/bindings-specification/binary-encoding.html).

This would be a lossless way to convert to Pkl and back again.

This should be provided as a:

* In-language API
* Java API (renderer)
* Java API (deserialize into `PObject` and friends)

Need to consider how to represent binary encoding.

HT154:
This PR is composed of two commits: the first adds the new functionality and the second regenerates all changed snippet outputs. Please review just the first commit.

Implements [SPICE-0021](https://github.com/apple/pkl-evolution/pull/23)

Resolves #994


pkl-core/src/main/java/org/pkl/core/stdlib/AbstractRenderer.java
HT154, line 1: This PR extracts `String`-specific functionality to a new `AbstractStringRenderer` subclass so that binary renderers can still benefit from the converter logic here in `AbstractRenderer`. This is a breaking change for any out-of-tree inheritors from `AbstractRenderer` but the fix is as simple as `s/AbstractRenderer/AbstractStringRenderer/g`.

pkl-core/src/main/java/org/pkl/core/stdlib/encoding/PklBinaryEncodingParserNodes.java
```diff
@@ -0,0 +1,133 @@
...
+        @Cached("create()") IndirectCallNode callNode) {
+      var bytes = (VmBytes) VmUtils.readMember(resource, Identifier.BYTES, callNode);
+      return doParse(module, bytes.getBytes());
+    }
+
+    @TruffleBoundary
+    private Object doParse(VmTyped module, byte[] bytes) {
+      var unpacker = MessagePack.newDefaultUnpacker(bytes);
+      return new PklBinaryDecoder(
+              unpacker, new Importer(module.getVmClass().getPClassInfo().getModuleUri()))
```
HT154: This is the piece I'm least confident in. Currently, the parser API requires users to supply a module as context. This module value is used to determine the URI of the "importing" module passed to `SecurityManager.checkImportModule`. Ideally this would be derived automatically from the call stack, but I'm unsure how to do that. I also see some utility to allowing callers to set this explicitly, eg. in cases where the parse happens inside a module in some package (which would in turn need to accept the context parameter from _its_ caller).

sin-ack:
Tested this out, and hit an edge case where if the moduleUri for a Typed binary-encoded value is the empty string, the following exception is thrown:

```
java.lang.NullPointerException

–– Pkl Error ––
None (cause has no message)

42 | external function parse(source: Resource|Bytes, context: Module): Any
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
at pkl.encoding#PklBinaryEncodingParser.parse (https://github.com/apple/pkl/blob/40bcfb18/stdlib/encoding.pkl#L42)

-- (snip) --

Pkl 0.30.0-dev+40bcfb18 (Linux 6.16.3-cachyos, native)

java.lang.NullPointerException
        at org.pkl.core.SecurityManagers.getDefaultTrustLevel(SecurityManagers.java:80)
        at org.pkl.core.SecurityManagers$Standard.checkImportModule(SecurityManagers.java:165)
        at org.pkl.core.stdlib.encoding.PklBinaryEncodingParserNodes$parse$Importer.importModule(PklBinaryEncodingParserNodes.java:126)
        at org.pkl.core.stdlib.encoding.PklBinaryEncodingParserNodes$parse$Importer.importClass(PklBinaryEncodingParserNodes.java:80)
        at org.pkl.core.PklBinaryDecoder.decodeObject(PklBinaryDecoder.java:230)
        at org.pkl.core.PklBinaryDecoder.decodeNonPrimitive(PklBinaryDecoder.java:162)
        at org.pkl.core.PklBinaryDecoder.doDecode(PklBinaryDecoder.java:141)
        at org.pkl.core.PklBinaryDecoder.decode(PklBinaryDecoder.java:109)
        at org.pkl.core.stdlib.encoding.PklBinaryEncodingParserNodes$parse.doParse(PklBinaryEncodingParserNodes.java:62)
        ...
```

HT154:
@sin-ack How did you produce the encoded data with the empty module URI string? Was that something you edited by hand?

sin-ack:
@HT154 It's from outside Pkl itself. We have a custom binary encoder/Pkl types representation that previously encoded `Dynamic`s with an empty module URI string (although I'm pretty sure `pkl server` also returns `Dynamic`s that way, but don't quote me on that).

pkl-core/src/main/java/org/pkl/core/PklBinaryDecoder.java
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  /**
+   * Decode a value from the supplied {@link MessageUnpacker}.
+   *
+   * @return the encoded value
+   */
+  public static Object decode(MessageUnpacker unpacker) {
+    return new PklBinaryDecoder(unpacker).decode();
+  }
```
bioball, line 50: Remove this method; we don't want `msgpack-java` to be a part of our own API. The other two methods can call `new PklBinaryDecoder(...)` directly.

pkl-core/src/main/java/org/pkl/core/util/pklbinary/AbstractPklBinaryDecoder.java
```diff
@@ -0,0 +1,446 @@
...
+
+  protected final Object decode() {
+    currPath = new ArrayDeque<>();
+    try {
+      return doDecode();
+    } catch (IOException e) {
+      throw doIOFail(e);
+    } catch (MessagePackException | DecodeException e) {
+      var path = currPath.stream().map(Object::toString).collect(Collectors.toList());
+      Collections.reverse(path);
```
bioball, line 70: We prefer to avoid streaming APIs in code that gets called from within our interpreter; can you implement this using a simple for loop?

pkl-core/src/main/java/org/pkl/core/util/pklbinary/AbstractPklBinaryDecoder.java
```diff
@@ -0,0 +1,446 @@
...
+import org.msgpack.core.MessagePackException;
+import org.msgpack.core.MessageUnpacker;
+import org.pkl.core.util.LateInit;
+
+/**
+ * Base class for implementing a decoder/parser for the <a
+ * href="https://pkl-lang.org/main/current/bindings-specification/binary-encoding.html"><code>
+ * pkl-binary</code></a> encoding.
+ */
+public abstract class AbstractPklBinaryDecoder {
```
bioball, line 41: Let's move this to new package `org.pkl.core.util.pklbinary`, which is private. Any classes we add to `org.pkl.core` is considered part of our public API.

pkl-core/src/main/java/org/pkl/core/util/pklbinary/AbstractPklBinaryDecoder.java
```diff
@@ -0,0 +1,446 @@
...
+    currPath.push("'bytes");
+    var result = doDecodeBytes(unpacker.readPayload(unpacker.unpackBinaryHeader()));
+    unpacker.skipValue(len - 2);
+    currPath.pop();
+    return result;
+  }
+
+  // some silly iterator classes because next() needs to throw IOException
+
+  protected abstract class DecodeIterator<T> {
```
bioball: No need to make `next()` throw `IOException`?

```suggestion
  protected abstract class DecodeIterator<T> implements Iterator<T> {
```

And you can get rid of `throws IOException` in all the `doDecode()` methods.

pkl-core/src/main/java/org/pkl/core/util/pklbinary/AbstractPklBinaryDecoder.java
```diff
@@ -0,0 +1,446 @@
...
+      return idx < size;
+    }
+
+    public T next() throws IOException {
+      currPath.push(idx);
+      var val = getNext();
+      currPath.pop();
+      idx++;
+      return val;
+    }
```
bioball:
```suggestion
    public T next() {
      currPath.push(idx);
      T val;
      try {
        val = getNext();
      } catch (IOException e) {
        throw ioFailure(e);
      }
      currPath.pop();
      idx++;
      return val;
    }
```

pkl-core/src/main/java/org/pkl/core/util/pklbinary/AbstractPklBinaryDecoder.java
```diff
@@ -0,0 +1,446 @@
...
+      super(size);
+    }
+
+    @Override
+    Object getNext() throws IOException {
+      return doDecode();
+    }
+  }
+
+  protected class MapDecodeIterator extends DecodeIterator<SimpleEntry<Object, Object>> {
```
bioball: Use class `org.pkl.core.util.Pair` instead (already used in other places, smaller footprint)

```suggestion
  protected class MapDecodeIterator extends DecodeIterator<Pair<Object, Object>> {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmPklBinaryDecoder.java
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  /**
+   * Decode a value from the supplied {@link MessageUnpacker}.
+   *
+   * @return the encoded value
+   */
+  public static Object decode(MessageUnpacker unpacker, Importer importer) {
+    return new VmPklBinaryDecoder(unpacker, importer).decode();
+  }
```
bioball: We don't really need to offer these methods, since this is an internal API.

We can just keep the constructor (make it public), and call `new VmPklBinaryDecoder(...).decode()` inside the truffle node.

pkl-core/src/main/java/org/pkl/core/stdlib/encoding/PklBinaryEncodingParserNodes.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    @Specialization
+    protected Object eval(
+        VmTyped self, VmTyped resource, @Cached("create()") IndirectCallNode callNode) {
+      var bytes = (VmBytes) VmUtils.readMember(resource, Identifier.BYTES, callNode);
+      return doParse(bytes.getBytes());
+    }
+
+    @TruffleBoundary
```
bioball: Note: this is a lot of code that we're placing behind a truffle boundary. Not necessarily blocking, but, we should try to make this visible to truffle inlining. Can be something we improve upon in the future.

pkl-core/src/main/java/org/pkl/core/util/pklbinary/AbstractPklBinaryDecoder.java
```diff
@@ -0,0 +1,446 @@
...
+
+      // things we should never see outside a non-primitive
+      case BIN8, BIN16, BIN32 -> throw new DecodeException("Unexpected msgpack bin value");
+      case MAP16, MAP32, FIXMAP -> throw new DecodeException("Unexpected msgpack map value");
+      case EXT8, EXT16, EXT32, FIXEXT1, FIXEXT2, FIXEXT4, FIXEXT8, FIXEXT16 ->
+          throw new DecodeException("Unexpected msgpack ext value");
+      case NEVER_USED ->
+          throw PklBugException
+              .unreachableCode(); // getNextFormat throws for this but switch needs to be exhaustive
+    };
```
bioball:
```suggestion
    return switch (unpacker.getNextFormat().getValueType()) {
      // primitives
      case NIL -> {
        unpacker.unpackNil();
        yield doDecodeNull();
      }
      case STRING -> unpacker.unpackString();
      case INTEGER -> unpacker.unpackLong();
      case BOOLEAN -> unpacker.unpackBoolean();
      case FLOAT -> unpacker.unpackDouble();

      // non-primitive
      case ARRAY -> decodeNonPrimitive();

      // things we should never see outside a non-primitive
      case BINARY -> throw new DecodeException("Unexpected msgpack bin value");
      case MAP -> throw new DecodeException("Unexpected msgpack map value");
      case EXTENSION -> throw new DecodeException("Unexpected msgpack ext value");
    };
```

pkl-core/src/main/java/org/pkl/core/PklBinaryDecoder.java
```diff
@@ -0,0 +1,204 @@
...
+        String.format(
+            "Exception while decoding binary data at offset %d, path [%s]",
+            offset, String.join(", ", path)),
+        cause);
+  }
+
+  @Override
+  protected RuntimeException doIOFail(Exception cause) {
+    return new RuntimeException("IO exception during decoding", cause);
+  }
```
bioball:
```suggestion
  protected RuntimeException doIOFail(IOException cause) {
    return new UncheckedIOException("IO exception during decoding", cause);
  }
```

pkl-core/src/main/java/org/pkl/core/PklBinaryDecoder.java
```diff
@@ -0,0 +1,204 @@
...
+
+  @Override
+  protected Object doDecodePair(Object first, Object second) {
+    return new Pair<>(first, second);
+  }
+
+  @Override
+  protected Object doDecodeIntSeq(long start, long end, long step) {
+    throw new DecodeException("Cannot export IntSeq value");
```
bioball:
```suggestion
    throw new DecodeException("Cannot decode IntSeq value");
```

pkl-core/src/main/java/org/pkl/core/PklBinaryDecoder.java
```diff
@@ -0,0 +1,204 @@
...
+      if (className.equals(BaseModule.getDynamicClass().getDisplayName())) {
+        return new PObject(PClassInfo.Dynamic, properties);
+      }
+
+      // pkl:base typed
+      if (!className.equals(BaseModule.getModule().getVmClass().getDisplayName())) {
+        return new PObject(PClassInfo.get("pkl.base", className, moduleUri), properties);
+        //        return new PModule(
+        //          moduleUri, className, PClassInfo.get(className, "ModuleClass", moduleUri),
+        // properties);
```
bioball:
```suggestion
```

pkl-core/src/main/java/org/pkl/core/util/pklbinary/PklBinaryEncoding.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.core;
+
+public class PklBinaryEncoding {
```
bioball: Can we make this an enum instead?

Also: move to package `org.pkl.core.util.pklbinary`

pkl-core/src/main/java/org/pkl/core/util/pklbinary/AbstractPklBinaryDecoder.java
```diff
@@ -0,0 +1,446 @@
...
+  protected class ObjectDecodeIterator extends DecodeIterator<DecodedObjectMember> {
+    ObjectDecodeIterator(int size) {
+      super(size);
+    }
+
+    @Override
+    DecodedObjectMember getNext() throws IOException {
+      var memberLen = unpacker.unpackArrayHeader();
+      if (memberLen != 3) {
+        throw new DecodeException("Expected 3 fields in object member, found %d", memberLen);
```
bioball, line 412: Can you include tests where we parse malformed bytes?

pkl-core/src/main/java/org/pkl/core/runtime/VmPklBinaryEncoder.java
```diff
@@ -0,0 +1,344 @@
...
+    try {
+      visitor.visit();
+    } catch (IOException e) {
+      throw new VmExceptionBuilder().evalError("ioErrorEncodingToBinary").withCause(e).build();
+    }
+  }
+
+  @Override
+  public void visitString(String value) {
+    doEncode(() -> packer.packString(value));
```
bioball: Instead of this, we should do:

```java
try {
  packer.packString(value);
} catch (IOException e) {
  throw PklBugException.unreachableCode();
}
```

It's not possible for these methods to throw `IOException`, because we have a `MessageBufferPacker` that doesn't do any IO when packing.

pkl-core/src/main/java/org/pkl/core/runtime/VmPklBinaryDecoder.java
```diff
@@ -0,0 +1,236 @@
...
+    return new VmExceptionBuilder()
+        .evalError("errorDecodingFromBinary", offset, String.join(" ", path))
+        .withCause(cause)
+        .build();
+  }
+
+  @Override
+  protected RuntimeException doIOFail(Exception cause) {
+    return new VmExceptionBuilder().evalError("ioErrorDecodingFromBinary").withCause(cause).build();
+  }
```
bioball: This will never happen, because we started off with bytes in the first place, whose unpacker doesn't do any IO.

```suggestion
    throw PklBugException.unreachableCode();
```

pkl-server/src/main/kotlin/org/pkl/server/BinaryEvaluator.kt
```diff
@@ -17,15 +17,14 @@ package org.pkl.server
...
-import org.pkl.core.ast.member.ObjectMember
 import org.pkl.core.evaluatorSettings.TraceMode
 import org.pkl.core.http.HttpClient
 import org.pkl.core.module.ModuleKeyFactory
 import org.pkl.core.project.DeclaredDependencies
 import org.pkl.core.resource.ResourceReader
 import org.pkl.core.runtime.*
+import org.pkl.core.runtime.VmPklBinaryEncoder
 
 internal class BinaryEvaluator(
```
bioball: Not blocking this PR, but: we should remove this class.

I think we'll want an eval method on the core `Evaluator` that does what this does (the Java version of `evaluateExpressionRaw` in pkl-swift/pkl-go/etc. And `Server.kt` should just call into that.

pkl-commons-test/src/main/kotlin/org/pkl/commons/test/MessagePackDebugRenderer.kt
```diff
@@ -104,4 +107,29 @@ class MessagePackDebugRenderer(bytes: ByteArray) {
     renderValue()
     sb.toString().removePrefix("\n")
   }
+
+  // always single quote stirngs for simplicity
+  // adapted from org.pkl.core.util.yaml.YamlEmitter.emitSingleQuotedString
+  fun emitString(str: String) {
+    sb.append('\'')
```
bioball, line 122: This is incorrect for multiline strings.

Maybe we can literal-style blocks for multiline strings? The style that starts with `|`.

pkl-config-java/src/main/java/org/pkl/config/java/ConfigPklBinaryDecoder.java
```diff
@@ -0,0 +1,82 @@
...
+  public Config decode(ByteArrayInputStream inputStream) {
+    return makeConfig(PklBinaryDecoder.decode(inputStream));
+  }
+
+  /**
+   * Decode a config from the supplied {@link MessageUnpacker}.
+   *
+   * @return the encoded config
+   */
+  public Config decode(MessageUnpacker unpacker) {
```
bioball: As mentioned elsewhere: we don't want msgpack-java to be part of our own API, so, let's get rid of this.

pkl-config-java/src/main/java/org/pkl/config/java/ConfigPklBinaryDecoder.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.io.ByteArrayInputStream;
+import java.util.Map;
+import org.msgpack.core.MessageUnpacker;
+import org.pkl.config.java.mapper.ValueMapper;
+import org.pkl.config.java.mapper.ValueMapperBuilder;
+import org.pkl.core.Composite;
+import org.pkl.core.PklBinaryDecoder;
+
+public class ConfigPklBinaryDecoder {
```
bioball: This API feels a little too verbose to me; I'm thinking that we can simplify this to some static methods on class `Config`:

```java
  default Config from(byte[] bytes, ValueMapper valueMapper) {
    // impl
  }

  default Config from(byte[] bytes) {
    return from(bytes, ValueMapper.preconfigured());
  }
```

On the usage side, it would look like this:

```java
var bytes = getClass().getResourceAsSteam("/my-config.msgpack").readAllBytes()
var cfg = Config.from(bytes).as(MyConfig::class)
```

bioball (changes requested):
Did a first pass!

pkl-core/src/main/java/org/pkl/core/PklBinaryDecoder.java
```diff
@@ -0,0 +1,204 @@
...
+  public static Object decode(byte[] bytes) {
+    return decode(MessagePack.newDefaultUnpacker(bytes));
+  }
+
+  /**
+   * Decode a value from the supplied {@link ByteArrayInputStream}.
+   *
+   * @return the encoded value
+   */
+  public static Object decode(ByteArrayInputStream inputStream) {
```
bioball: Change the param to just `InputStream`

pkl-core/src/main/java/org/pkl/core/EvaluatorImpl.java
```diff
@@ -172,6 +194,18 @@ public Object evaluateOutputValue(ModuleSource moduleSource) {
...
 
+  @Override
+  public byte[] evaluateOutputValuePklBinary(ModuleSource moduleSource) {
+    return doEvaluate(
+        moduleSource,
+        (module) -> {
+          var output = readModuleOutput(module);
+          var value = VmUtils.readMember(output, Identifier.VALUE);
+          VmValue.force(value, false);
+          return renderBinary(module);
```
bioball: Assuming we keep this method, this is a bug:

```suggestion
          return renderBinary(value);
```

pkl-core/src/main/java/org/pkl/core/Evaluator.java
```diff
@@ -70,6 +80,16 @@ static Evaluator preconfigured() {
...
+  /**
+   * Evaluates a module's {@code output.value} property, returning a byte array of the <a
+   * href="https://pkl-lang.org/main/current/bindings-specification/binary-encoding.html"><code>
+   * pkl-binary</code></a>-encoded representation of the result.
+   *
+   * @throws PklException if an error occurs during evaluation
+   * @throws IllegalStateException if this evaluator has already been closed
+   */
+  byte[] evaluateOutputValuePklBinary(ModuleSource moduleSource);
+
```
bioball: I think let's just keep `evaluateExpressionPklBinary` and eliminate `evaluatePklBinary` and `evaluateOutputValuePklBinary`.

This is an intentionally low-level way to eval, and I think it's okay that we have just one method.
Users that are interested in evaluating the whole module can do `evaluateExpressionPklBinary(myModule, "module")`;.

pkl-core/src/main/java/org/pkl/core/EvaluatorImpl.java
```diff
@@ -67,6 +70,8 @@ public class EvaluatorImpl implements Evaluator {
   protected final BufferedLogger logger;
   protected final PackageResolver packageResolver;
   private final VmValueRenderer vmValueRenderer = VmValueRenderer.singleLine(1000);
+  private static ThreadLocal<MessageBufferPacker> threadLocalBufferPacker =
+      ThreadLocal.withInitial(MessagePack::newDefaultBufferPacker);
```
bioball: Make this an instance field, and remove `ThreadLocal` (an evaluator is not assumed to be thread-safe, so we don't need to guard for race conditions for instance members).

pkl-core/src/main/java/org/pkl/core/PklBinaryDecoder.java
```diff
@@ -0,0 +1,194 @@
...
+   * Decode a value from the supplied byte array.
+   *
+   * @return the encoded value
+   */
+  public static Object decode(byte[] bytes) {
+    return new PklBinaryDecoder(MessagePack.newDefaultUnpacker(bytes)).decode();
+  }
+
+  /**
+   * Decode a value from the supplied {@link ByteArrayInputStream}.
```
bioball:
```suggestion
   * Decode a value from the supplied {@link InputStream}.
```

pkl-core/src/main/java/org/pkl/core/PklBinaryDecoder.java
```diff
@@ -0,0 +1,194 @@
...
+import org.pkl.core.runtime.BaseModule;
+import org.pkl.core.util.CollectionUtils;
+import org.pkl.core.util.pklbinary.AbstractPklBinaryDecoder;
+
+/**
+ * A decoder/parser for the <a
+ * href="https://pkl-lang.org/main/current/bindings-specification/binary-encoding.html"><code>
+ * pkl-binary</code></a> encoding. Returns "exported" value types like PObject, Map, List, and
+ * primitives.
+ */
```
bioball, line 35: So far, we haven't linked to the website from our Java docs. For consistency sake, let's also avoid doing that here.
```suggestion
/**
 * A decoder for pkl-binary.
 *
 * <p>For how pkl-binary turns Java, see {@link Value}.
 */
```

pkl-core/src/main/java/org/pkl/core/PklBinaryDecoder.java
```diff
@@ -0,0 +1,194 @@
...
+
+  private PklBinaryDecoder(MessageUnpacker unpacker) {
+    super(unpacker);
+  }
+
+  /**
+   * Decode a value from the supplied byte array.
+   *
+   * @return the encoded value
+   */
```
bioball: We tend to not use `@return` if the method signature already makes it quite obvious (but we're not consistent everywhere)

```suggestion
  /** Decode a value from the supplied byte array. */
```

pkl-core/src/main/java/org/pkl/core/PklBinaryDecoder.java
```diff
@@ -0,0 +1,194 @@
...
+  public static Object decode(byte[] bytes) {
+    return new PklBinaryDecoder(MessagePack.newDefaultUnpacker(bytes)).decode();
+  }
+
+  /**
+   * Decode a value from the supplied {@link ByteArrayInputStream}.
+   *
+   * @return the encoded value
```
bioball: See comment above

```suggestion
```

pkl-core/src/main/java/org/pkl/core/runtime/VmTyped.java
```diff
@@ -130,6 +131,13 @@ public VmMap getClassMirrors() {
     return builder.build();
   }
 
+  public @Nullable VmClass getClass(Identifier name) {
+    assert isModuleObject();
+    if (getModuleInfo().isAmend()) return null;
+    return (VmClass)
+        getMember(it -> it.isClass() && it.getName() == name); // Note: includes local classes
+  }
+
```
bioball: This and `getTypeAlias(Identifier name)` are unused, remove

Note: the correct way to do this is to use `VmUtils.readMember`

pkl-core/src/main/java/org/pkl/core/util/pklbinary/PklBinaryTypes.java
```diff
@@ -0,0 +1,75 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.core.util.pklbinary;
+
+import org.pkl.core.util.Nullable;
+
+public enum PklBinaryTypes {
```
bioball: Let's rename this:

```suggestion
public enum PklBinaryCode {
```

And for the rest of these, remove the `CODE_` prefix. So, usage looks like `PklBinaryCode.OBJECT` etc.

pkl-core/src/test/kotlin/org/pkl/core/PklBinaryDecoderTest.kt
```diff
@@ -0,0 +1,237 @@
...
+                "indent" to "  ",
+                "omitNullProperties" to false,
+                "useCustomStringDelimiters" to false,
+              ),
+            ),
+        ),
+      )
+
+    assertThat(decoded).isEqualTo(expected)
+    assertThat((decoded.getProperty("regex") as Pattern).toString()).isEqualTo("foo.*")
```
bioball:
```suggestion
    assertThat((decoded.getProperty("regex") as Pattern).pattern()).isEqualTo("foo.*")
```

pkl-server/src/main/kotlin/org/pkl/server/Server.kt
```diff
@@ -210,8 +213,9 @@ class Server(private val transport: MessageTransport) : AutoCloseable {
           buildDeclaredDependencies(proj.projectFileUri, proj.dependencies, null)
         }
       log("Got dependencies: $dependencies")
-      return BinaryEvaluator(
+      return EvaluatorImpl(
```
bioball: [nit] Change this code to use `EvaluatorBuilder` to build. You can start with `EvaluatorBuilder.unconfigured()`, then add all of these options.

pkl-server/src/test/kotlin/org/pkl/server/BinaryEvaluatorSnippetTests.kt
```diff
@@ -44,8 +44,9 @@ class BinaryEvaluatorSnippetTestEngine : InputOutputTestEngine() {
   }
 
   private val evaluator =
-    BinaryEvaluator(
+    EvaluatorImpl(
```
bioball: This whole class can be deleted, and also `BinaryEvaluatorTest`

bioball (changes requested):
This is getting pretty close! Made another pass.

pkl-core/src/main/java/org/pkl/core/EvaluatorImpl.java
```diff
@@ -67,6 +70,8 @@ public class EvaluatorImpl implements Evaluator {
   protected final BufferedLogger logger;
   protected final PackageResolver packageResolver;
```
bioball: Now that we don't have a subclass, let's change all these fields to `private`, and add modifier `final` to class `EvaluatorImpl`

pkl-core/src/main/java/org/pkl/core/runtime/VmTyped.java
HT154: Ah, good catch. These were left over from the iteration that included the in-language decoder.

pkl-core/src/main/java/org/pkl/core/EvaluatorImpl.java
HT154: pkl-server's BinaryEvaluator used a `ThreadLocal` for the packer. Why is that no longer needed in this case? Wouldn't creating a packer per instance increase memory usage in long-lived or high-throughput server usage?

bioball: That was actually not needed either! The messages handled by the server are processed by a single-threaded executor (see `handleEvaluate` in `Server.kt`).

I believe the guard used to be meaningful at one point, but we've since refactored the code and the `ThreadLocal`  there was accidentally left behind.

> Wouldn't creating a packer per instance increase memory usage in long-lived or high-throughput server usage?

Yeah, but it's negligible memory in the context of Pkl evaluation (evaluators are sparsely created). Additionally, we're trading memory for CPU here; using a thread local has its own overhead when fetching the local that doesn't happen if it's simply an instance value.

We can optimize this a little bit more by only initializing the buffer packer when it's needed, like with this getter:

```java
private MessagePacker getMessagePacker() {
  if (this.messagePacker == null) {
    this.messagePacker = MessagePack.newDefaultBufferPacker();
  }
  return this.messagePacker;
}
```

pkl-core/src/main/java/org/pkl/core/Evaluator.java
```diff
@@ -172,10 +172,19 @@ static Evaluator preconfigured() {
...
+   * @throws PklException if an error occurs during evaluation
+   * @throws IllegalStateException if this evaluator has already been closed
+   */
+  byte[] evaluateExpressionPklBinary(ModuleSource moduleSource, String expression);
+
   /**
    * Evaluates the Pkl expression, returning the stringified result.
    *
-   * <p>This is equivalent to wrapping the expression with {@code .toString()}
+   * <p>This is equivalent to wrapping the expression with {@code .toString()} xs
```
bioball: Accidentally snuck into the commit?

pkl-core/src/main/java/org/pkl/core/runtime/VmPklBinaryEncoder.java
```diff
@@ -0,0 +1,380 @@
...
+import org.msgpack.core.MessageBufferPacker;
+import org.pkl.core.PklBugException;
+import org.pkl.core.stdlib.AbstractRenderer;
+import org.pkl.core.stdlib.PklConverter;
+import org.pkl.core.util.pklbinary.PklBinaryCode;
+
+/**
+ * An encoder/renderer for the <a
+ * href="https://pkl-lang.org/main/current/bindings-specification/binary-encoding.html"><code>
+ * pkl-binary</code></a> encoding.
```
bioball: [nit] also remove this link to the website

```suggestion
 * An encoder/renderer for <code>pkl-binary</code> encoding.
```

pkl-core/src/main/java/org/pkl/core/stdlib/AbstractRenderer.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2025 Apple Inc. and the Pkl project authors. All rights reserved.
```
bioball:
```suggestion
 * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
```

pkl-core/src/main/java/org/pkl/core/stdlib/pklbinary/RendererNodes.java
```diff
@@ -0,0 +1,57 @@
...
+import org.pkl.core.stdlib.PklConverter;
+
+public final class RendererNodes {
+  public abstract static class renderDocument extends ExternalMethod1Node {
+    @Specialization
+    @TruffleBoundary
+    protected VmBytes eval(VmTyped self, Object value) {
+      var packer = MessagePack.newDefaultBufferPacker();
+      createRenderer(self, packer).renderDocument(value);
+      return new VmBytes(packer.toByteArray());
```
bioball, line 52: We can add the same optimization as in `EvaluatorImpl` here, where we initialize and cache the same buffer packer.

stdlib/pklbinary.pkl
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+/// APIs for handling Pkl's [binary encoding format](https://pkl-lang.org/main/current/bindings-specification/binary-encoding.html).
+@ModuleInfo { minPklVersion = "0.30.0" }
+module pkl.pklbinary
+
+/// Render values as the [`pkl-binary` encoding format](https://pkl-lang.org/main/current/bindings-specification/binary-encoding.html).
+@Since { version = "0.30.0" }
```
bioball: This is implied by the annotation on the module!

```suggestion
```

stdlib/pklbinary.pkl
```diff
@@ -0,0 +1,29 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
```
bioball:
```suggestion
// Copyright © 2025 Apple Inc. and the Pkl project authors. All rights reserved.
```

bioball (approved):
Some minor comments. LGTM to merge after they're addressed!

PR #1211, opened 2025-09-18
Fix bug in format apply

stackoverflow:
This fixes a bug where the formatter was not doing anything on `pkl format apply`.

### DIFF
--- a/pkl-cli/src/main/kotlin/org/pkl/cli/commands/FormatterCommand.kt
+++ b/pkl-cli/src/main/kotlin/org/pkl/cli/commands/FormatterCommand.kt
@@ -19,17 +19,13 @@ import com.github.ajalt.clikt.core.Context
 import com.github.ajalt.clikt.core.NoOpCliktCommand
 import com.github.ajalt.clikt.core.subcommands
 import com.github.ajalt.clikt.parameters.arguments.argument
-import com.github.ajalt.clikt.parameters.options.convert
-import com.github.ajalt.clikt.parameters.options.default
 import com.github.ajalt.clikt.parameters.options.flag
 import com.github.ajalt.clikt.parameters.options.option
 import com.github.ajalt.clikt.parameters.types.path
 import java.nio.file.Path
 import org.pkl.cli.CliFormatterApply
 import org.pkl.cli.CliFormatterCheck
-import org.pkl.commons.cli.CliException
 import org.pkl.commons.cli.commands.BaseCommand
-import org.pkl.formatter.Compat
 
 class FormatterCommand : NoOpCliktCommand(name = "format") {
   override fun help(context: Context) = "Run commands related to formatting"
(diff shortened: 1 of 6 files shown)


pkl-cli/src/main/kotlin/org/pkl/cli/commands/FormatterCommand.kt
```diff
@@ -41,19 +45,37 @@ class FormatterCheckCommand : BaseCommand(name = "check", helpLink = helpLink) {
   override val helpString: String =
     "Check if the given files are properly formatted, printing the file name to stdout in case they are not. Returns non-zero in case of failure."
 
+  val compat: Compat by
+    option(
+        names = arrayOf("--compat", "-c"),
+        help = "Run compatibility mode for older versions. Default: latest",
```
srueg: Might make it more clear that default is the current CLI version.
```suggestion
        help = "Run compatibility mode for older versions. Default: current version",
```

pkl-cli/src/main/kotlin/org/pkl/cli/commands/FormatterCommand.kt
```diff
@@ -41,19 +45,37 @@ class FormatterCheckCommand : BaseCommand(name = "check", helpLink = helpLink) {
...
-    CliFormatterCheck(baseOptions.baseOptions(emptyList()), path).run()
+    CliFormatterCheck(baseOptions.baseOptions(emptyList()), path, compat).run()
   }
 }
 
 class FormatterApplyCommand : BaseCommand(name = "apply", helpLink = helpLink) {
   override val helpString: String =
     "Overwrite all the files in place with the formatted version. Returns non-zero in case of failure."
 
+  val compat: Compat by
```
srueg: Can we reuse the `val compat: Compat` from above?

HT154:
I have some concerns about shipping this as-is. I think this might warrant some deeper discussion once Dan is back. I think if a compatibility mechanism is introduced, it should take one of two forms:
* A hidden, undocumented flag introduced for this release only and removed in 0.31 allowing Pkl 0.29 users to test the formatter without forcing them to adopt trailing commas.
* A principled compatibility strategy that uses the immediately previous minor release's grammar and formatting rules. Without the compat flag the formatter would format for the current Pkl release / rules only.
  * This would allow users that enforce formatting to adopt new Pkl releases without forcing them to adopt all formatting changes at the same time. Instead, adoption could happen in two steps 1) upgrade pkl and enable compatibility in the formatter then 2) disable formatter compatibility and apply any formatting changes.
  * Having a strict rule like "compatibility with the preceding minor version only" avoids the need to maintain a bunch of branches in the formatter indefinitely.

stackoverflow:
@HT154 Yeah, agree having this option as-is is not a good idea. I'll remove this for now and just leave the other changes. We can add this in some form later.

PR #1215, opened 2025-09-19
Allow formatting multiple files/directories

srueg:
This makes it easier to work with tools which return a list of file to format, e.g. to only handle files which have changed in a PR.

### DIFF
--- a/pkl-cli/src/main/kotlin/org/pkl/cli/CliFormatterCommand.kt
+++ b/pkl-cli/src/main/kotlin/org/pkl/cli/CliFormatterCommand.kt
@@ -31,7 +31,7 @@ abstract class CliFormatterCommand
 @JvmOverloads
 constructor(
   options: CliBaseOptions,
-  protected val path: Path,
+  protected val paths: List<Path>,
   protected val consoleWriter: Writer = System.out.writer(),
 ) : CliCommand(options) {
   protected fun format(file: Path, contents: String): Pair<String, Int> {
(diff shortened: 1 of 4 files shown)


stackoverflow (approved):

pkl-cli/src/main/kotlin/org/pkl/cli/CliFormatterCommand.kt
```diff
@@ -46,9 +46,15 @@ constructor(
...
-    return if (path.isDirectory()) {
-      path.walk().filter { it.extension == "pkl" || it.name == "PklProject" }
-    } else sequenceOf(path)
+  protected fun paths(): Set<Path> {
+    val allPaths = mutableSetOf<Path>()
+    for (path in paths) {
+      if (path.isDirectory()) {
+        allPaths.addAll(path.walk().filter { it.extension == "pkl" || it.name == "PklProject" })
+      } else {
+        allPaths.add(path)
```
StefMa, line 55: Stupid question, shouldn't the filter be applied here as well? 🤔 And if it's not the case, failing with a good error message?

pkl-cli/src/main/kotlin/org/pkl/cli/CliFormatterCommand.kt
```diff
@@ -46,9 +46,15 @@ constructor(
   }
 
   @OptIn(ExperimentalPathApi::class)
-  protected fun paths(): Sequence<Path> {
-    return if (path.isDirectory()) {
-      path.walk().filter { it.extension == "pkl" || it.name == "PklProject" }
-    } else sequenceOf(path)
+  protected fun paths(): Set<Path> {
+    val allPaths = mutableSetOf<Path>()
```
StefMa, line 50: A bit more idiomatic kotlin would be either using `= buildSet` or `paths.map`. The current code looks a bit Java-ish 😅

pkl-cli/src/main/kotlin/org/pkl/cli/CliFormatterCommand.kt
stackoverflow: No. This was done on purpose. If you pass a directory to the formatter it assumes you only want to format pkl files in that dir. But if you pass a glob of files it assumes you know what you are doing and want to format all the input.
Otherwise `pkl format apply my-pkl-file.not-pkl` wouldn't work.

pkl-cli/src/main/kotlin/org/pkl/cli/CliFormatterCommand.kt
stackoverflow: Feel free to create a pr ;)

PR #1226, opened 2025-10-06
Bump GraalVM to 25.0.0

bioball:
Also:

* Migrate shadow plugin to org.gradleup to correctly bundle Truffle in a fat jar
* Switch to GraalVM Community Edition


buildSrc/src/main/kotlin/pklFatJar.gradle.kts
```diff
@@ -94,6 +93,8 @@ tasks.shadowJar {
 
   configurations = listOf(project.configurations.runtimeClasspath.get())
 
+  addMultiReleaseAttribute = true
```
bioball, line 96: This bumps the shadow plugin because Truffle failed to initialize with error like "multi-release lost" or something like that.

The `addMultiReleaseAttribute` is a new option available in the `org.gradleup` version of the plugin.

HT154 (approved):

stackoverflow (approved):

PR #1227, opened 2025-10-07
Follow up for trace pretty printing

HT154:
This PR addresses @stackoverflow's feedback in #1100, removes the `hidden` trace mode, and renames the `default` mode to `compact.


pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -298,13 +298,13 @@ class BaseOptions : OptionGroup() {
...
-        metavar = "when",
+        metavar = "style",
         help =
-          "Specifies how calls to trace() are formatted. Possible values of <when> are 'default', 'pretty' and 'hidden'.",
+          "Specifies how calls to trace() are formatted. Possible values of <style> are 'compact' and 'pretty'.",
       )
       .enum<TraceMode> { it.name.lowercase() }
       .single()
-      .default(TraceMode.DEFAULT)
+      .default(TraceMode.COMPACT)
```
bioball: This should be nullable; we want this to fall back to the setting in `pkl:Project` if available.

Take a look at `org.pkl.commons.cli.CliCommand` and see how many of those options are set, and how the evaluator is built based on this.

This should also fix the broken tests on main currently.

bioball (approved):
Approving to unblock; there's broken tests on main that we should fix that we can either do here or in a follow-up PR.

stackoverflow (approved):
LGTM

PR #1228, opened 2025-10-08
Bump Gradle to 9.1.0

bioball:
Also, remove support for kotlin gradle plugin less than 1.8.x, because:

* Class `org.gradle.api.plugins.Convention ` is no longer available in the classpath in Gradle
* Continued support for legacy plugin would require heavy reflection, which is brittle and hard to verify
* Kotlin 1.7 is 3 years old and no longer updated

### DIFF
--- a/pkl-gradle/src/test/kotlin/org/pkl/gradle/KotlinCodeGeneratorsTest.kt
+++ b/pkl-gradle/src/test/kotlin/org/pkl/gradle/KotlinCodeGeneratorsTest.kt
@@ -108,7 +108,7 @@ class KotlinCodeGeneratorsTest : AbstractTest() {
   }
 
   private fun writeBuildFile() {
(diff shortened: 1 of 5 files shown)


pkl-gradle/src/test/kotlin/org/pkl/gradle/KotlinCodeGeneratorsTest.kt
```diff
@@ -108,7 +108,7 @@ class KotlinCodeGeneratorsTest : AbstractTest() {
   }
 
   private fun writeBuildFile() {
-    val kotlinVersion = "1.6.0"
+    val kotlinVersion = "2.0.21"
```
bioball, line 111: kotlin-gradle-plugin version 1.6.0 is not compatible with Gradle 9

HT154 (approved):

PR #1236, opened 2025-10-09
Apply pkl formatter to codebase

bioball:
This applies the Pkl formatter to `stdlib/` and `.circleci/`

### DIFF
--- a/stdlib/yaml.pkl
+++ b/stdlib/yaml.pkl
@@ -50,7 +50,7 @@ class Parser {
   /// - `"compat"` - YAML 1.1 _or_ 1.2 (default)
   /// - `"1.1"` - YAML 1.1
   /// - `"1.2"` - YAML 1.2 (Core schema)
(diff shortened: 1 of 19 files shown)


stdlib/yaml.pkl
```diff
@@ -50,7 +50,7 @@ class Parser {
   /// - `"compat"` - YAML 1.1 _or_ 1.2 (default)
   /// - `"1.1"` - YAML 1.1
   /// - `"1.2"` - YAML 1.2 (Core schema)
-  mode: "compat"|"1.1"|"1.2" = "compat"
+  mode: "compat" | "1.1" | "1.2" = "compat"
```
HT154, line 53: I'm not in love with the extra spaces around the `|` in single line union types.

bioball: We can possibly change this. But, as far as I'm aware, no other language formats their union types the way we have.

PR #1249, opened 2025-10-21
Add grammar compatibility option to the formatter

stackoverflow:


pkl-cli/src/main/kotlin/org/pkl/cli/commands/FormatterCommand.kt
```diff
@@ -47,8 +50,25 @@ class FormatterCheckCommand : BaseCommand(name = "check", helpLink = helpLink) {
       .path(mustExist = true, canBeDir = true)
       .multiple()
 
+  val compatVersion: CompatVersion by
+    option(names = arrayOf("-v", "--version"), help = compatVersionHelp)
```
bioball: We shouldn't conflict with the built-in `--version` flag that already exists.

Names are hard, but, maybe let's call this "grammar version"?

```suggestion
    option(names = arrayOf("--grammar-version"), help = compatVersionHelp)
```

I also don't think we need to offer a short version of this flag; descriptive flags help with readability of scripts.

pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -33,7 +33,7 @@ import org.pkl.parser.syntax.Operator
 import org.pkl.parser.syntax.generic.Node
 import org.pkl.parser.syntax.generic.NodeType
 
-internal class Builder(sourceText: String) {
+internal class Builder(sourceText: String, private val version: CompatVersion) {
```
bioball: Whatever name we settle on, we should change this and the class name too.

```suggestion
internal class Builder(sourceText: String, private val grammarVersion: GrammarVersion) {
```

pkl-cli/src/main/kotlin/org/pkl/cli/commands/ProjectCommand.kt
```diff
@@ -33,7 +33,7 @@ import org.pkl.commons.cli.commands.BaseCommand
 import org.pkl.commons.cli.commands.TestOptions
 import org.pkl.commons.cli.commands.single
 
-private const val NEWLINE = '\u0085'
+internal const val NEWLINE = '\u0085'
```
bioball: Move to a shared place, maybe a `util.kt` or something?

pkl-cli/src/main/kotlin/org/pkl/cli/commands/FormatterCommand.kt
stackoverflow: Changed everything to be `grammar version` and removed the short version. The other good name would be `grammar compatibility version`, but that's too verbose.

pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
stackoverflow: Changed.

pkl-cli/src/main/kotlin/org/pkl/cli/commands/ProjectCommand.kt
stackoverflow: Moved to a utils file.

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -743,6 +743,16 @@ It returns a non-zero status code in case violations are found.
 
 If the path is a directory, recursively looks for files with a `.pkl` extension, or files named `PklProject`.
 
+==== Options
+
+.--grammar-version
+[%collapsible]
+====
+Default: `V2` (latest version) +
```
bioball: Two nits:

1. Can we just make this a number? The "V" seems unnecessary.
2. Can we provide a table of grammar version to language version? Something like:

| Grammar Version | Pkl Version |
| --- | --- |
| 1 | 0.25 - 0.29 |
| 2 | 0.30+ |

stackoverflow: The cli option just requires a number now. I actually just used a number in the beginning but changed it later.
The description for each version is the version span it supports.

```bash
--grammar-version=(1|2)    The grammar compatibility version to use.
                           1: 0.25 - 0.29
                           2: 0.30+ (default)
```

PR #1250, opened 2025-10-21
Polish documentation for pkl-binary

bioball:


stdlib/pklbinary.pkl
```diff
@@ -14,15 +14,30 @@
...
+/// Render values as `pkl-binary`.
 class Renderer extends BytesRenderer {
-  /// Render a Pkl value as `pkl-binary`.
+  /// Renders [value] as `pkl-binary`.
   external function renderValue(value: Any): Bytes
 
-  /// Render a Pkl document as `pkl-binary`.
+  /// Renders [value] as `pkl-binary`.
+  ///
+  /// Every Pkl value is a valid `pkl-binary` document.
```
HT154: This could potentially be misinterpreted to contradict the restriction on `Function` values.

stackoverflow: Agree that this is worth pointing out.

PR #1251, opened 2025-10-22
Fix Lexer EOF sentinel collision with valid Unicode code points

spyoungtech:
# Summary

A change [introduced](https://github.com/apple/pkl/pull/917) in 0.28.0 reimplemented the Lexer in a way that implements EOF detection using an in-band sentinel value `Short.MAX_VALUE` (0x7FFF), which collides with the valid Unicode codepoint U+7FFF. When this character appears in source code, the lexer incorrectly terminates parsing, treating all subsequent content as non-existent. 

Although this is probably a rare occurrence for typical use, this issue presents an EOF injection vulnerability, which is very similar to a null byte injection vulnerability. This vulnerability enables **steganographic configuration manipulation** with the following characteristics:

1. **Silent Failure**: No parse errors, warnings, or diagnostic output
2. **Code Review Evasion**: The attack character is visually innocuous and easily overlooked
3. **Policy Bypass**: Security controls defined after the injection point are never evaluated

This affects (at least) all versions since 0.28.0: [0.28.0, 0.28.1, 0.28.2, 0.29.0, 0.29.1]

This fix removes reliance on lookahead/sentinel comparisons to prevent early parser termination.

<details><summary><h3>Proof of Concept</h3></summary>

Consider a base network policy with subsequent hardening:

**base.pkl**:
  ```pkl
  allowedPorts = new Listing {
      80
      443
      22
  }
```

  production.pkl (legitimate):
```pkl
  amends "base.pkl"

  // Production security: remove SSH access
  allowedPorts = new Listing {
      80
      443
  }
```
  production.pkl (compromised):

```pkl
  amends "base.pkl"

  // Production security: remove port 22; SSH not allowed in China env  翿!
  allowedPorts = new Listing {
      80
      443
  }
```

 Result:
  - Expected: SSH port 22 blocked
  - Actual: SSH port 22 remains open (security bypass)

  The configuration appears syntactically valid but the hardening directive is silently ignored.

</details>

## Proposed Remediation

The fix implements out-of-band EOF signaling using a dedicated atEOF boolean flag, eliminating the sentinel value collision:

  1. EOF state tracked via explicit boolean flag rather than reserved character value
  2. All EOF detection logic migrated to flag-based checks
  3. EOF sentinel changed to Character.MAX_VALUE (0xFFFF) with proper flag handling

This ensures no valid Unicode codepoint can be misinterpreted as EOF.

### Verification

Verification is added to the existing test suite in the form of a test that validates the Unicode code points, including the boundary values (U+7FFF, U+FFFF), are correctly processed without premature termination.

## References

  - https://cwe.mitre.org/data/definitions/147.html
  - https://cwe.mitre.org/data/definitions/115.html
  - https://www.unicode.org/reports/tr55/

### DIFF
--- a/pkl-parser/src/main/java/org/pkl/parser/Lexer.java
+++ b/pkl-parser/src/main/java/org/pkl/parser/Lexer.java
@@ -30,14 +30,15 @@ public class Lexer {
   private int sLine = 1;
   private int col = 1;
   private int sCol = 1;
-  private char lookahead;
+  private int lookahead;
   private State state = State.DEFAULT;
   private final Deque<InterpolationScope> interpolationStack = new ArrayDeque<>();
   private boolean stringEnded = false;
   private boolean isEscape = false;
   // how many newlines exist between two subsequent tokens
   protected int newLinesBetween = 0;
-  private boolean atEOF = false;
+
+  private static final int EOF = -1;
 
   public Lexer(String input) {
     source = input.toCharArray();
@@ -45,8 +46,7 @@ public class Lexer {
     if (size > 0) {
       lookahead = source[cursor];
     } else {
-      lookahead = '\uffff';
-      atEOF = true;
+      lookahead = EOF;
     }
   }
 
@@ -96,14 +96,12 @@ public class Lexer {
       sCol = col;
       ch = nextChar();
     }
-    if (atEOF && ch == '\uffff') {
-      // when EOF is reached we overshot the span
-      // If ch is not '\uffff' it means we still have one character to process
-      // and this condition should not catch
-      cursor--;
-      return Token.EOF;
-    }
     return switch (ch) {
+      case EOF -> {
+        // when EOF is reached we overshot the span
+        cursor--;
+        yield Token.EOF;
+      }
       case ';' -> Token.SEMICOLON;
       case '(' -> {
         var scope = interpolationStack.peek();
@@ -250,7 +248,7 @@ public class Lexer {
           yield lexNumber(ch);
         } else if (isIdentifierStart(ch)) {
           yield lexIdentifier();
-        } else throw lexError(ErrorMessages.create("invalidCharacter", ch), cursor - 1, 1);
+        } else throw lexError(ErrorMessages.create("invalidCharacter", (char) ch), cursor - 1, 1);
       }
     };
   }
@@ -264,7 +262,7 @@ public class Lexer {
       state = State.DEFAULT;
       return Token.STRING_END;
     }
-    if (atEOF) return Token.EOF;
+    if (lookahead == EOF) return Token.EOF;
     if (isEscape) {
       isEscape = false;
       // consume the `\#*`
@@ -298,7 +296,7 @@ public class Lexer {
       nextChar();
       pounds++;
     }
-    if (atEOF) {
+    if (lookahead == EOF) {
       throw lexError(ErrorMessages.create("unexpectedEndOfFile"), span());
     }
     if (lookahead != '"') {
@@ -337,7 +335,7 @@ public class Lexer {
     var poundsInARow = 0;
     var foundQuote = false;
     var foundBackslash = false;
-    while (!atEOF) {
+    while (lookahead != EOF) {
       var ch = nextChar();
       switch (ch) {
         case '\n', '\r' ->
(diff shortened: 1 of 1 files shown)


pkl-parser/src/main/java/org/pkl/parser/Lexer.java
```diff
@@ -37,8 +37,9 @@ public class Lexer {
   private boolean isEscape = false;
   // how many newlines exist between two subsequent tokens
   protected int newLinesBetween = 0;
+  private boolean atEOF = false;
 
-  private static final char EOF = Short.MAX_VALUE;
+  private static final char EOF = Character.MAX_VALUE;
```
bioball: This still has a conflict, no? `\uffff` is an "undefined character": https://www.compart.com/en/unicode/U+FFFF

I can still have Pkl code that looks like:

```pkl
// Hello there ￿ uffff
```

I think we can fix this by representing EOF as `-1`, and change `char lookahead` to `int lookahead`; since `-1` is never a valid character code, this eliminates any chance of conflict. And, we can go back to using `if (lookahead == EOF) in this case.

spyoungtech: The change of the sentinel value to a noncharacter is mostly cosmetic at this point. Originally, I made the change because I thought this would be a one-line kinda thing and just that it'd be better to use a non-character [designated for internal application use](https://www.unicode.org/faq/private_use) instead of a 'real' character assigned a glyph. 

But as you point out, that change alone would just shift the problem to a different character and it could still potentially collide with other applications' legitimate use of the character and the vulnerability would remain.

However, combined with the other changes, it doesn't really matter what this sentinel is. I did add tests for `\uffff` though to ensure the case is properly covered. The primary thing is that solves the problem is that `atEOF` is never set based on the content of the source, so injecting EOF behavior is no longer possible.

The approach you describe would work. It is the approach I tried first, but it caused so many cascading behavior changes that I wasn't confident enough to go all the way down that route without understanding the possible performance/behavior impacts.

bioball: Ah, I see. that's why `atEOF` was introduced.

In that case, I think it makes more sense to just get rid of `private static final char EOF`, because it'd just add some confusion for future maintainers.

pkl-parser/src/main/java/org/pkl/parser/Lexer.java
```diff
@@ -98,8 +96,10 @@ private Token nextDefault() {
       sCol = col;
       ch = nextChar();
     }
-    if (ch == EOF && atEOF) {
+    if (atEOF && ch == '\uffff') {
```
bioball:
```suggestion
    if (atEOF) {
```

pkl-parser/src/main/java/org/pkl/parser/Lexer.java
```diff
@@ -732,7 +732,7 @@ private ParserError lexError(String msg, Span span) {
   }
 
   private ParserError unexpectedChar(char got, String didYouMean) {
-    if (got == EOF) {
+    if (got == '\uffff' && atEOF) {
```
bioball:
```suggestion
    if (atEOF) {
```

pkl-parser/src/main/java/org/pkl/parser/Lexer.java
```diff
@@ -690,7 +690,7 @@ private char nextChar() {
     var tmp = lookahead;
     cursor++;
     if (cursor >= size) {
-      lookahead = EOF;
+      lookahead = '\uffff';
```
bioball:
```suggestion
```

pkl-parser/src/main/java/org/pkl/parser/Lexer.java
```diff
@@ -39,15 +39,13 @@ public class Lexer {
...
-  private static final char EOF = Character.MAX_VALUE;
-
   public Lexer(String input) {
     source = input.toCharArray();
     size = source.length;
     if (size > 0) {
       lookahead = source[cursor];
     } else {
-      lookahead = EOF;
+      lookahead = '\uffff';
```
bioball:
```suggestion
```

PR #1252, opened 2025-10-22
Adjust formatting of qualified access chains

bioball:
This adjusts formatting of qualified access chains so that leading dot calls are kept in the same line if possible.

Formatting rules:

If there is a call chain that ends with the only method call, prefer to only wrap the argument list. Render this:

```pkl
prop =
  foo.baz(new {
    foo = 1
  })
```

Instead of:

```pkl
prop =
  foo
    .baz(new {
      foo = 1
    })
```

If the method call is interspersed within the chain, prefer to wrap before the first method call: Render this:

```pkl
prop =
  foo.bar
    .baz(new {
      foo = 1
    })
    .qux()
```

Instead of:

```pkl
prop =
  foo
    .bar
    .baz(new {
      foo = 1
    })
    .qux()
```

### DIFF
--- a/pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
+++ b/pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
@@ -194,7 +194,7 @@ internal class Builder(sourceText: String) {
     return if (prefixes.isEmpty()) {
       res
     } else {
-      val sep = getSeparator(prefixes.last(), nodes.first())
+      val sep = getSeparator(prefixes.last(), nodes.first(), spaceOrLine())
       Nodes(formatGeneric(prefixes, spaceOrLine()) + listOf(sep, res))
     }
   }
@@ -304,8 +304,8 @@ internal class Builder(sourceText: String) {
     val separator: (Node, Node) -> FormatNode? = { prev, next ->
       when {
         prev.type == NodeType.OPERATOR -> null
-        next.type == NodeType.OPERATOR -> if (lambdaCount > 1) ForceLine else Line
-        else -> SpaceOrLine
+        next.type == NodeType.OPERATOR -> if (lambdaCount > 1) forceLine() else line()
+        else -> spaceOrLine()
       }
     }
 
@@ -315,15 +315,20 @@ internal class Builder(sourceText: String) {
           // lift argument list into its own node
           val (callChain, argsList) = splitFunctionCallNode(flat)
           val leadingNodes = indentAfterFirstNewline(formatGeneric(callChain, separator), true)
-          val trailingNodes = format(argsList)
-          leadingNodes + trailingNodes
+          val trailingNodes = formatGeneric(argsList, separator)
+          val sep = getBaseSeparator(callChain.last(), argsList.first())
+          if (sep != null) {
+            leadingNodes + sep + trailingNodes
+          } else {
+            leadingNodes + trailingNodes
+          }
         }
         methodCallCount > 0 && indexBeforeFirstMethodCall > 0 -> {
           val leading = flat.subList(0, indexBeforeFirstMethodCall)
           val trailing = flat.subList(indexBeforeFirstMethodCall, flat.size)
           val leadingNodes = indentAfterFirstNewline(formatGeneric(leading, separator), true)
           val trailingNodes = formatGeneric(trailing, separator)
-          leadingNodes + Line + trailingNodes
+          leadingNodes + line() + trailingNodes
         }
         else -> formatGeneric(flat, separator)
       }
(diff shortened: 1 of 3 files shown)


pkl-parser/src/main/java/org/pkl/parser/GenericParser.java
```diff
@@ -679,6 +679,10 @@ private Node parseExpr(@Nullable String expectation, int minPrecedence) {
           ff(children);
           expect(Token.RBRACK, children, "unexpectedToken", "]");
         }
+        case DOT, QDOT -> {
+          nodeType = NodeType.QUALIFIED_ACCESS_EXPR;
+          children.add(parseUnqualifiedAccessExpr());
```
bioball, line 684: Bugfix and parser change: parse qualified access as its own node type, and don't permit any arbitrary expression to follow the dot.

pkl-parser/src/test/kotlin/org/pkl/parser/SexpRenderer.kt
```diff
@@ -532,6 +530,18 @@ class SexpRenderer {
...
   }
 
+  fun renderQualifiedAccessExpr(expr: QualifiedAccessExpr) {
+    buf.append(tab)
+    buf.append(if (expr.isNullable) "(nullableQualifiedAccessExpr" else "(qualifiedAccessExpr")
+    val oldTab = increaseTab()
+    buf.append('\n')
+    renderExpr(expr.expr)
+    buf.append('\n')
+    renderUnqualifiedAccessExprOfQualified(expr)
```
bioball, line 540: Note: `QualifiedAccessExpr` should probably have `UnqualifiedAccessExpr` as a child, since the grammar is:

```bnf
unqualifiedAccessExpr :: identifier argumentList?

qualifiedAccessExpr :: expr ('.' | '.?') unqualifiedAccessExpr
```

pkl-parser/src/main/java/org/pkl/parser/GenericParser.java
```diff
@@ -719,6 +723,15 @@ private Node parseExpr(@Nullable String expectation, int minPrecedence) {
     };
   }
 
+  private Node parseUnqualifiedAccessExpr() {
+    var children = new ArrayList<Node>();
+    children.add(parseIdentifier());
+    if (lookahead == Token.LPAREN && noSemicolonInbetween() && _lookahead.newLinesBetween == 0) {
+      children.add(parseArgumentList());
+    }
```
stackoverflow, line 732:
```suggestion
    children.add(parseIdentifier());
    if (lookahead() == Token.LPAREN && noSemicolonInbetween() && _lookahead.newLinesBetween == 0) {
      ff(children);
      children.add(parseArgumentList());
    }
```
Just found another bug looking at this code: we are not fast forwarding after parsing the identifier so code like `foo.bar/*asd*/()` is not parsing, even though it works fine for the normal parser.

pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -239,6 +240,117 @@ internal class Builder(sourceText: String) {
...
+  /**
+   * Split a function call node to extract its identifier into the leading group. For example,
+   * `foo.bar(5)` becomes: leading gets `foo.bar`, rest gets `(5)`.
+   */
+  private fun splitFunctionCallNode(nodes: List<Node>): Pair<List<Node>, Node> {
+    assert(nodes.isNotEmpty())
+    val lastLeading = nodes.last()
+    val identifier = lastLeading.firstProperChild()!!
+    val argList = lastLeading.children.find { it.type == NodeType.ARGUMENT_LIST }!!
+    return (nodes.subList(0, nodes.lastIndex) + identifier) to argList
```
stackoverflow: We are losing the comments between the identifier and the argument list. (A bug I just pointed out in the other comment).
We should have a test for it in `expr-chain-grouping.pkl`.

pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -239,6 +240,117 @@ internal class Builder(sourceText: String) {
...
+    }
+
+    val nodes =
+      when {
+        methodCallCount == 1 && isMethodCall(flat.lastProperNode()!!) -> {
+          // lift argument list into its own node
+          val (callChain, argsList) = splitFunctionCallNode(flat)
+          val leadingNodes = indentAfterFirstNewline(formatGeneric(callChain, separator), true)
+          val trailingNodes = format(argsList)
+          leadingNodes + trailingNodes
```
stackoverflow: We should never concat nodes together without a separator, as we don't know where an affix can appear.
Same for line 323.
```suggestion
          leadingNodes + getSeparator(leadingNodes.last, trailingNodes, null) + trailingNodes
```

pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
bioball: Sounds good, adding a test for this.

pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
bioball: Ah, gotcha!

bioball: For line 323: we're using `line()` as a separator, and we know the trailing group doesn't start with an affix (we split right before the first method call).

If that seems wrong, let me know and we can follow up with a fix.

PR #1253, opened 2025-10-23
Enforce Pkl formatting of stdlib with spotless

bioball:
This adds a spotless formatting step using the new pkl formatter.

This only formats Pkl sources in the stdlib, because other sources are possibly test input and not meant to be formatted.

TODO: it'd be good to also format the sources in `.circleci` (and `.github` once we migrate to GHA), but I'm running into some Gradle weirdness when trying to set that up; seeing this error:

```
A problem was found with the configuration of task ':spotlessPkl' (type 'SpotlessTaskImpl').
  - Gradle detected a problem with the following location: '/Users/danielchao/code/apple/pkl'.

    Reason: Task ':spotlessPkl' uses this output of task ':stdlib:spotlessKotlinGradle' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
```

### DIFF
--- a/buildSrc/src/main/kotlin/PklFormatterSpotless.kt
+++ b/buildSrc/src/main/kotlin/PklFormatterSpotless.kt
@@ -1,0 +1,56 @@
+/*
+ * Copyright © 2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/kotlin/PklFormatterSpotless.kt
```diff
@@ -0,0 +1,56 @@
...
+class PklFormatterFunc(@Transient private val configuration: Configuration) :
+  FormatterFunc, Serializable {
+  companion object {
+    @Serial private const val serialVersionUID: Long = 1L
+  }
+
+  private val formatterClass by lazy {
+    val urls = configuration.files.map { it.toURI().toURL() }
+    val classLoader = URLClassLoader(urls.toTypedArray())
+    classLoader.loadClass("org.pkl.formatter.Formatter")
```
stackoverflow, line 46: I wonder how does that even work?

bioball: This works by:

1. Build the jar files
2. Create a classloader that loads classes from those jar files

Pretty wonky, and can possibly break in the future if we bump our Kotlin faster than Gradle's own Kotlin. If that happens, we might need to switch to calling Pkl as a sub-process.

PR #1256, opened 2025-10-27
Adjust formatting of line comments in lets

bioball:
Fixes an issue where line comments preceding let bodies are not indented.

Turn this:

```pkl
bar3 =
  let (bar = new Dynamic {})
  // some coment
    (bar) {
      qux = 4
    }
```

Into this:

```pkl
bar3 =
  let (bar = new Dynamic {})
    // some coment
    (bar) {
      qux = 4
    }
```

### DIFF
--- a/pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
+++ b/pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
@@ -972,21 +972,21 @@ internal class Builder(sourceText: String, private val grammarVersion: GrammarVe
 
   private fun formatLetExpr(node: Node): FormatNode {
     val separator = if (node.isMultiline()) forceLine() else spaceOrLine()
+    val endsWithLet = node.children.last().type == NodeType.LET_EXPR
     val nodes =
       formatGenericWithGen(
         node.children,
         { _, next -> if (next.type == NodeType.LET_PARAMETER_DEFINITION) Space else separator },
-      ) { node, next ->
+      ) { node, _ ->
         when {
           node.type == NodeType.LET_EXPR -> {
             // unpack the lets
             val group = formatLetExpr(node) as Group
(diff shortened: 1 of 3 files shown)


pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -941,13 +941,17 @@ internal class Builder(sourceText: String) {
...
-        } else format(node)
+          }
+          node.type == NodeType.LET_PARAMETER_DEFINITION ||
+            node.isTerminal("let") ||
+            next?.type == NodeType.LET_EXPR -> format(node)
+          else -> indent(format(node))
+        }
       }
     return Group(newId(), nodes)
   }
```
stackoverflow, line 993: Almost there, but you can't rely on `next?.type == LET_EXPR` because you might have more than one comment (in fact you should add this case to the tests):
```pkl
let (x = 1)
  // comment1
  // comment2
  x
```
Probably easier to check if the expression is another let and just indent based on that:
```kotlin
private fun formatLetExpr(node: Node): FormatNode {
    val endsWithLet = node.children.last().type == NodeType.LET_EXPR
    val nodes =
      formatGenericWithGen(
        node.children,
        { _, next -> if (next.type == NodeType.LET_PARAMETER_DEFINITION) Space else spaceOrLine() },
      ) { node, _ ->
        when {
          node.type == NodeType.LET_EXPR -> {
            // unpack the lets
            val group = formatLetExpr(node) as Group
            Nodes(group.nodes)
          }
          endsWithLet -> format(node)
          node.type.isExpression || node.type.isAffix -> indent(format(node))
          else -> format(node)
        }
      }
    return Group(newId(), nodes)
  }
```

bioball: Ah, thanks!

PR #1257, opened 2025-10-27
Add `Object.toMixin()` method

Mikulas:
Adds a `toMixin()` method to the `Object` class that converts an `Object`
into a `Mixin` function applicable via the pipe operator (`|>`).

A generic `toMixin()` method cannot be implemented in user land, so this
implementation provides a native method that properly handles:
- Property merging and overriding
- Element appending with correct index offsetting
- Entry merging with proper key handling
- Nested object replacement vs amendment semantics

Implementation uses the source `Object`'s enclosing frame to ensure
proper module context for type resolution during member evaluation.

**edit:** updated from `Dynamic` to `Object`


pkl-core/src/main/java/org/pkl/core/stdlib/base/DynamicNodes.java
```diff
@@ -113,4 +116,102 @@ protected VmObjectLike eval(VmDynamic self, VmClass clazz) {
...
+      var result = org.pkl.core.util.EconomicMaps.<Object, ObjectMember>create(
+          org.pkl.core.util.EconomicMaps.size(members));
+
+      var cursor = members.getEntries();
+      while (cursor.advance()) {
+        var key = cursor.getKey();
+        var member = cursor.getValue();
+
+        // If the key is a Long (element index), offset it by parentLength
+        if (key instanceof Long) {
```
HT154: This is not a sound assumption to make. Mappings and Dynamics can have entries with Int (in Pkl, Long in Java) keys. Probably better to check `member.isElement()` here instead (and you can then assume the key is a Long).

Mikulas: This would be the internal field representation, not the userland defined keys. I will see if I can get the key to be the right type at compile time and replace this check

Mikulas: updated to `if (member.isElement()) {`

pkl-core/src/main/java/org/pkl/core/stdlib/base/ObjectNodes.java
```diff
@@ -0,0 +1,123 @@
...
+      return new VmFunction(
+          VmUtils.createEmptyMaterializedFrame(),
+          null,
+          1,
+          rootNode,
+          null);
+    }
+  }
+
+  private static final class ObjectToMixinRootNode extends org.pkl.core.ast.PklRootNode {
```
HT154: Nit: this is probably large enough to warrant moving into `org.pkl.core.ast` somewhere. Generally we don't include the `Root` as part of the subclass name.

pkl-core/src/main/java/org/pkl/core/stdlib/base/ObjectNodes.java
```diff
@@ -0,0 +1,123 @@
...
+    private final VmObject sourceObject;
+
+    public ObjectToMixinRootNode(VmObject sourceObject, FrameDescriptor descriptor) {
+      super(null, descriptor);
+      this.sourceObject = sourceObject;
+    }
+
+    @Override
+    public com.oracle.truffle.api.source.SourceSection getSourceSection() {
```
HT154: Please import types instead of using their FQCNs where possible.

pkl-core/src/test/files/LanguageSnippetTests/input/api/objectToMixin.pkl
HT154, line 1: Here are a few un-tested cases I can think of that I'd like to see covered:
* Applying mixins with properties/entries to Listings (and other variants of this wrong-member-type error)
* Apply `Mixin<Listing>` to a `Dynamic` (and other variants; except `Dynamic |> Mixin<some Typed>` which is covered)
* Applying an object's mixin to the original object (`obj |> obj.toMixin()`)
* Applying a mixin to an object that would induce a stack overflow (`new Dynamic { foo = bar } |> new Dynamic { bar = foo }.toMixin()`)

stdlib/base.pkl
```diff
@@ -1861,7 +1861,21 @@ external class DataSize extends Any {
...
+  /// to amend it with the properties, elements, and entries from this object.
+  ///
+  /// Example:
+  /// ```
+  /// obj = new { name = "Pigeon"; age = 42 }
+  /// person = new Person { name = "Original" } |> obj.toMixin()
+  /// // person.name == "Pigeon", person.age == 42
+  /// ```
+  @Since { version = "0.31.0" }
+  external function toMixin(): Mixin
```
HT154, line 1877: Would that we had a `Self` type... 😔

Mikulas: A limitation of this patch is that the resulting Mixin is untyped. It would be great to have a typed Mixin!

pkl-core/src/main/java/org/pkl/core/stdlib/base/ObjectNodes.java
Mikulas: 👍 moved

pkl-core/src/main/java/org/pkl/core/stdlib/base/ObjectNodes.java
Mikulas: imported

pkl-core/src/test/files/LanguageSnippetTests/input/api/objectToMixin.pkl
Mikulas: This is great! Actually caught an issue with Listings where it was incorrectly overwriting keys. Now it correctly amends keys, same as explicit Mixin declaration does.

pkl-core/src/main/java/org/pkl/core/ast/ObjectToMixinNode.java
```diff
@@ -0,0 +1,116 @@
...
+      var member = cursor.getValue();
+
+      // If this is an element (not an entry with an Int key), offset the index
+      if (member.isElement()) {
+        // Elements always have Long keys
+        var newKey = (Long) key + parentLength;
+        org.pkl.core.util.EconomicMaps.put(result, newKey, member);
+      } else {
+        // Properties and entries are not offset
+        org.pkl.core.util.EconomicMaps.put(result, key, member);
```
HT154, line 144: A few more FQCNs still hanging around:

```suggestion
  @TruffleBoundary
  private static UnmodifiableEconomicMap<Object, ObjectMember> adjustMemberIndices(
      UnmodifiableEconomicMap<Object, ObjectMember> members,
      long parentLength) {
    if (parentLength == 0) {
      return members;
    }

    var result = EconomicMaps.<Object, ObjectMember>create(EconomicMaps.size(members));

    var cursor = members.getEntries();
    while (cursor.advance()) {
      var key = cursor.getKey();
      var member = cursor.getValue();

      // If this is an element (not an entry with an Int key), offset the index
      if (member.isElement()) {
        // Elements always have Long keys
        var newKey = (Long) key + parentLength;
        EconomicMaps.put(result, newKey, member);
      } else {
        // Properties and entries are not offset
        EconomicMaps.put(result, key, member);
```

pkl-core/src/main/java/org/pkl/core/ast/ObjectToMixinNode.java
```diff
@@ -0,0 +1,116 @@
...
+          .typeMismatch(targetObject, BaseModule.getDynamicClass())
+          .build();
+    }
+
+    var parent = (VmObject) targetObject;
+    var parentLength = getObjectLength(parent);
+    var sourceLength = getObjectLength(sourceObject);
+    var adjustedMembers = adjustMemberIndices(sourceObject.getMembers(), parentLength);
+
+    return new VmDynamic(
```
HT154, line 68: This should return the same type as `sourceObject`, so you'll need to choose the right `VmObject` subclass to build here. We definitely want this invariant to hold:
```pkl
local base = new Bird { name = "Parrot"; age = 20 }
local birthday = new Bird { age = super.age + 1 }.toMixin()
local aged = base |> birthday
invariant = base.getClass() == aged.getClass()
```

bioball:
This behavior seems obviously wrong, and is what this implementation currently does.

```pkl
foo {
  a = 1
}

bar = (foo) {
  b = 2
}

res = new Dynamic {} |> bar.toMixin()
```

Produces:

```pkl
foo {
  a = 1
}
bar {
  a = 1
  b = 2
}
res {
  b = 2
}
```

However, respecting the amends chain is somewhat problematic for `Typed.toMixin()`:

```pkl
class MyClass {
  a: Int
  b: Int
}

foo: MyClass = new { a = 1 }

res: MyClass = new MyClass {} |> foo.toMixin() // would throw "cannot read property `a`" if respecting the amends chain
```

This feature sounds good in theory, but has some nuances that are hard to resolve. Also, this definitely needs a SPICE to guide the design!

stdlib/base.pkl
bioball: We've considered introducing `this` as a type; which would allow:

```pkl
external function toMixin(): Mixin<this>
```

We actually _do_ have a self-type right now, but only for modules:

```pkl
function foo(): module
```

PR #1259, opened 2025-10-27
Respect newlines in if/else and let expressions

bioball:
Change the formatter to prevent multiline if/else and let expressions from collapsing into a single line.

### DIFF
--- a/pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
+++ b/pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
@@ -840,16 +840,14 @@ internal class Builder(sourceText: String) {
   }
 
   private fun formatIf(node: Node): FormatNode {
+    val separator = if (node.isMultiline()) forceLine() else spaceOrLine()
     val nodes =
-      formatGeneric(node.children) { prev, next ->
-        when {
-          next.type == NodeType.IF_ELSE_EXPR && next.children[0].type == NodeType.IF_EXPR -> Space
-          next.type == NodeType.IF_THEN_EXPR ||
-            next.type == NodeType.IF_ELSE_EXPR ||
-            next.isTerminal("else") ->
-            if (prev.linesBetween(next) > 0) forceLine() else spaceOrLine()
-          else -> spaceOrLine()
-        }
+      formatGeneric(node.children) { _, next ->
+        // produce `else if` in the case of nested if.
+        // note: don't need to handle if `next.children[0]` is an affix because that can't be
+        // emitted as `else if` anyway.
+        if (next.type == NodeType.IF_ELSE_EXPR && next.children[0].type == NodeType.IF_EXPR) Space
+        else separator
       }
     return Group(newId(), nodes)
   }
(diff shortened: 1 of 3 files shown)


pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -841,10 +841,15 @@ internal class Builder(sourceText: String) {
 
   private fun formatIf(node: Node): FormatNode {
     val nodes =
-      formatGeneric(node.children) { _, next ->
-        if (next.type == NodeType.IF_ELSE_EXPR && next.children[0].type == NodeType.IF_EXPR) {
-          Space
-        } else spaceOrLine()
+      formatGeneric(node.children) { prev, next ->
+        when {
+          next.type == NodeType.IF_ELSE_EXPR && next.children[0].type == NodeType.IF_EXPR -> Space
```
stackoverflow: I thought I spotted a bug here: we can't use `next.children[0]` because the first child might be an affix. But then I remembered I did this on purpose: if you add a comment between the `else` and the next `if` we are not lining up the else/ifs as they can't be properly lined anyway. But I forgot to add a comment there telling why we are using `children[0]`.

pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -841,10 +841,15 @@ internal class Builder(sourceText: String) {
 
   private fun formatIf(node: Node): FormatNode {
     val nodes =
-      formatGeneric(node.children) { _, next ->
-        if (next.type == NodeType.IF_ELSE_EXPR && next.children[0].type == NodeType.IF_EXPR) {
-          Space
-        } else spaceOrLine()
+      formatGeneric(node.children) { prev, next ->
```
stackoverflow: This could be simplified by only checking once:
```kotlin
private fun formatIf(node: Node): FormatNode {
  val separator = if (node.span.lineBegin() < node.span.lineEnd) forceLine() else spaceOrLine()
  val nodes =
    formatGeneric(node.children) { _, next ->
      if (next.type == NodeType.IF_ELSE_EXPR && next.children[0].type == NodeType.IF_EXPR) {
        Space
      } else separator
    }
  return Group(newId(), nodes)
}
```
Probably a good idea to extract the checking if a node is multiline to a function:
```kotlin
private fun Node.isMultiline(): Boolean = span.lineBegin < span.lineEnd
```

pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -939,7 +944,16 @@ internal class Builder(sourceText: String) {
     val nodes =
       formatGenericWithGen(
         node.children,
-        { _, next -> if (next.type == NodeType.LET_PARAMETER_DEFINITION) Space else spaceOrLine() },
+        { prev, next ->
```
stackoverflow: Same comment as the if: you can just check once:
```kotlin
val separator = if (node.isMultiline()) forceLine() else spaceOrLine()
// ...
{ _, next -> if (next.type == NodeType.LET_PARAMETER_DEFINITION) Space else separator },
```

stackoverflow (approved):

PR #1260, opened 2025-10-27
Adjust formatting of argument lists

bioball:
This changes code so that multiple lambda arguments makes the whole argument list wrap.

Improves the readability of code like:

```
foo
  .toMap(
    (it) -> makeSomeKey(it),
    (it) -> makeSomeValue(it)
  )
```

### DIFF
--- a/pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
+++ b/pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
@@ -617,6 +617,17 @@ internal class Builder(sourceText: String) {
     }
   }
 
+  /**
+   * Only considered trailing lamdba if there is only one lambda/new expr/amends expr in the list.
+   *
+   * E.g. avoid formatting `toMap()` weirdly:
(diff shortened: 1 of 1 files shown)


pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -619,12 +619,21 @@ internal class Builder(sourceText: String) {
 
   private fun hasTrailingLambda(argList: Node): Boolean {
     val children = argList.firstProperChild()?.children ?: return false
+    var seenArg = false
+    var ret = false
     for (i in children.lastIndex downTo 0) {
       val child = children[i]
       if (!child.isProper()) continue
-      return child.type in SAME_LINE_EXPRS
+      if (child.type in SAME_LINE_EXPRS) {
```
stackoverflow, line 638: Looks good, but I think it's worth it adding a comment here explaining this rule for our future selves benefit.

stackoverflow (approved):

PR #1268, opened 2025-10-29
Respect line breaks in operator chains and argument lists

bioball:
If an operator chain or method call is multiline, keep those newlines in the formatted output.

Help preserve code like:

```pkl
foo
  |> (it) -> it + 2
  |> (it) -> it / 2
```


pkl-formatter/src/test/files/FormatterSnippetTests/output/line-breaks.pkl
```diff
@@ -49,7 +49,10 @@ funcParam2 =
...
 
 funcParam4 =
-  aFun(foo, 10 * 10, anotherVariable, if (true) 100000 else 200000, new Listing {
+  aFun(foo,
+  10 * 10,
+  anotherVariable,
+  if (true) 100000 else 200000, new Listing {
     1
     2
   })
```
HT154: This looks unintentional

pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -587,10 +596,13 @@ internal class Builder(sourceText: String, private val grammarVersion: GrammarVe
     twoBy2: Boolean = false,
   ): FormatNode {
     val children = node.children
+    val sep: (Node, Node) -> FormatNode = { prev, next ->
+      if (prev.isTerminal(",") && prev.linesBetween(next) > 0) forceSpaceyLine() else spaceOrLine()
+    }
```
stackoverflow, line 600: This logic is arbitrary and error prone.
It doesn't work for many different cases or works by chance:
```pkl
// doesn't multiline
p1 = foo(1
        ,2
        ,3)

// here it does
p2 = foo(1,
        2
        ,3)

// doesn't multiline
p3 = foo(1, 2, 3, it
        .asd)

// works by chance, because we respect newlines after comments
p4 = foo(1, 2, 3, /*foo*/
        it)
```
I think we should go back to the drawing board with this logic.
Maybe remove this part and merge the rest. This could be a subsequent PR.

bioball: It's intentional that `p3` doesn't multiline; per our formatting rules, `it\n.bar` should be single-lined, and this formatting rule shouldn't interfere with that. I also think it's fine that `p4` works.

It's definitely a problem that `p1` doesn't multiline; just pushed up a commit to fix that. Also, this also fixes an issue where this doesn't multiline:

```
1 +
  2
```

stackoverflow (approved):

PR #1273, opened 2025-10-30
Make formatter stable

stackoverflow:


pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -621,7 +621,7 @@ internal class Builder(sourceText: String, private val grammarVersion: GrammarVe
     return if (node.children.isNotEmpty()) {
       endsWithClosingBracket(node.children.last())
     } else {
-      node.isTerminal("}")
+      node.isTerminal("}") || node.type.isAffix
```
stackoverflow, line 624: This has the side effect of not adding a newline even if the last element is a single line block comment:
```pkl
items = allItems.filter((it) ->
            it /*foo*/
          )

// formats to
items =
  allItems.filter((it) -> it
  /*foo*/)
```

But this is such an edge case I don't think it worth writing special handling.

pkl-formatter/src/test/files/FormatterSnippetTests/output/line-breaks.pkl
```diff
@@ -74,11 +74,11 @@ local function render(currentIndent: String) =
...
-//     badItems.containsKey(item) // some line comment 
-//     || item.tags.toList().findOrNull((it) -> it.type == "bookmark") != null // some other line comment
-//   )
-//     .sortBy((item) -> item.name)
+items: List<Item> =
+  allItems
+    .filter((item) ->
+      badItems.containsKey(item) // some line comment 
+        || item.tags.toList().findOrNull((it) -> it.type == "bookmark") != null
+    // some other line comment
```
HT154: Is this really correct? It seems odd to move a line comment off its original line.

stackoverflow: Indeed. It shouldn't be on a newline. Probably a side effect of the trailing lambda rule. Will take a look.

stackoverflow: Fixed now.

PR #1276, opened 2025-10-30
Add support for rendering `Bytes` values with `YamlRenderer`

HT154:


pkl-core/src/main/java/org/pkl/core/util/yaml/YamlEmitter.java
```diff
@@ -210,6 +212,11 @@ public final void emit(boolean value) {
     builder.append(value);
   }
 
+  public final void emit(byte[] value) {
+    builder.append("!!binary ");
+    builder.append(Base64.getEncoder().encodeToString(value));
```
bioball: Hmm.. should we encode this in RFC 2045 style? E.g. wrap at 76 chars with CRLF line breaks?

Seems like this is the typical way that this gets encoded in YAML: https://yaml.org/type/binary.html

pkl-core/src/test/files/LanguageSnippetTests/input/api/yamlRenderer2.yml.pkl
```diff
@@ -18,6 +18,7 @@ res13 = new Dynamic { name = "pigeon"; age = 30 }
 res14 = new Person { name = "pigeon" }
 res15 = null
 res16 = Pair(1, 2)
+res17 = Bytes(1, 2, 3)
```
bioball: Add tests for empty bytes, and, assuming we use RFC 2045 style encoding, bytes that exceed that length

pkl-core/src/main/java/org/pkl/core/util/yaml/YamlEmitter.java
bioball: We should handle empty bytes; one way to do it is `!!binary ''`.

Perhaps a better approach is to append the tag, and then call `org.pkl.core.util.yaml.YamlEmitter#emit(java.lang.String, java.lang.StringBuilder, boolean)`. That will give you a well-formed string scalar.

PR #1277, opened 2025-10-30
Switch `yaml.Parser` to parse binary scalars as `Bytes`

HT154:
This is a breaking change! Parsing YAML data containing `!!binary` scalars now produces a different Pkl value.

Builds on #1276

### DIFF
--- a/docs/modules/ROOT/partials/component-attributes.adoc
+++ b/docs/modules/ROOT/partials/component-attributes.adoc
@@ -75,6 +75,8 @@ endif::[]
 :uri-stdlib-xmlModule: {uri-pkl-stdlib-docs}/xml
 :uri-stdlib-protobufModule: {uri-pkl-stdlib-docs}/protobuf
 :uri-stdlib-pklbinaryModule: {uri-pkl-stdlib-docs}/pklbinary
+:uri-stdlib-yamlModule: {uri-pkl-stdlib-docs}/yaml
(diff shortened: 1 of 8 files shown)


pkl-core/src/main/java/org/pkl/core/util/yaml/YamlEmitter.java
```diff
@@ -210,6 +212,11 @@ public final void emit(boolean value) {
     builder.append(value);
   }
 
+  public final void emit(byte[] value) {
+    builder.append("!!binary ");
+    builder.append(Base64.getEncoder().encodeToString(value));
```
bioball: Hmm.. should we encode this in RFC 2045 style? E.g. wrap at 76 chars with CRLF line breaks?

Seems like this is the typical way that this gets encoded in YAML: https://yaml.org/type/binary.html

HT154: This doesn't seem to be required and dealing with CRLFs could be sticky. It doesn't look like snakeyaml attempts to do this, so I'm unsure it's worth doing. Worst case, should we discover that it's important, we can revise later without it being a breaking change.

bioball: Yeah, agreed!

PR #1280, opened 2025-10-30
Don't force multiline interpolation into a single line

bioball:
Also, fixes an issue where forced single-line formatting would break for if/else and let expressions


pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -1460,6 +1457,8 @@ internal class Builder(sourceText: String, private val grammarVersion: GrammarVe
 
   private fun Node.isMultiline(): Boolean = span.lineBegin < span.lineEnd
 
+  private fun List<Node>.isMultiline(): Boolean = first().span.lineBegin < last().span.lineEnd
```
stackoverflow:
```suggestion
  private fun List<Node>.isMultiline(): Boolean =
    if (isEmpty()) false else first().span.lineBegin < last().span.lineEnd
```
Better safe than sorry.

stackoverflow (approved):
LGTM

PR #1283, opened 2025-11-03
Fix formatting of argument lists

bioball:
This fixes two issues:

1. Leading/trailing line comments surrounding a lambda should make that lambda not "trailing", because the formatting otherwise looks bad and also isn't stable
2. Fix incorrect algorithm for detecting trailing lambda (currently, any number of lambdas makes the alg return `true`)

### DIFF
--- a/pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
+++ b/pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
@@ -1289,7 +1289,7 @@ internal class Builder(sourceText: String, private val grammarVersion: GrammarVe
 
   private fun getBaseSeparator(prev: Node, next: Node): FormatNode? {
     return when {
-      prev.lastChild().type == NodeType.LINE_COMMENT -> {
+      prevNode?.type == NodeType.LINE_COMMENT -> {
         if (prev.linesBetween(next) > 1) {
           TWO_NEWLINES
         } else {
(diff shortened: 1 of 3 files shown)


pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -1485,6 +1486,9 @@ internal class Builder(sourceText: String, private val grammarVersion: GrammarVe
   private fun List<Node>.isMultiline(): Boolean =
     if (isEmpty()) false else first().span.lineBegin < last().span.lineEnd
 
+  private tailrec fun Node.lastChild(): Node =
```
stackoverflow: You are not using this function anywhere.

pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -1288,7 +1289,7 @@ internal class Builder(sourceText: String, private val grammarVersion: GrammarVe
 
   private fun getBaseSeparator(prev: Node, next: Node): FormatNode? {
     return when {
-      prevNode?.type == NodeType.LINE_COMMENT -> {
+      prev.type == NodeType.LINE_COMMENT -> {
```
stackoverflow: This is a bugfix you reverted and it's the reason the build is broken.

PR #1296, opened 2025-11-04
Improve plugin logic

bioball:
* Make configurations show up in the run manager (the box on the top right)
* Make "run all snippets" use class as a selector, instead of unique id

### DIFF
--- a/pkl-internal-intellij-plugin/src/main/kotlin/org/pkl/internal/intellij/SnippetTestSplitEditor.kt
+++ b/pkl-internal-intellij-plugin/src/main/kotlin/org/pkl/internal/intellij/SnippetTestSplitEditor.kt
@@ -40,6 +40,9 @@ import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.UserDataHolderBase
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.openapi.vfs.VirtualFileManager
+import com.intellij.psi.JavaPsiFacade
+import com.intellij.psi.PsiClass
+import com.intellij.psi.search.GlobalSearchScope
 import com.intellij.ui.components.JBPanel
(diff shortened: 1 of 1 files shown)


pkl-internal-intellij-plugin/src/main/kotlin/org/pkl/internal/intellij/SnippetTestSplitEditor.kt
```diff
@@ -328,11 +331,25 @@ class SnippetTestSplitEditor(
...
+    // Pattern: .../LanguageSnippetTests/input/lambdas/lambdaStackTrace2.pkl
+    val pattern = Regex(".*/([^/]+)/src/test/files/(\\w+)/input/(.+)$")
+    val match = pattern.find(path)!!
+    val folder = match.groupValues[2]
+    val className =
+      when (folder) {
+        "LanguageSnippetTests" -> "org.pkl.core.LanguageSnippetTests"
+        "FormatterSnippetTests" -> "org.pkl.formatter.FormatterSnippetTests"
+        // legacy; doesn't exist on main branch
+        "SnippetTests" -> "org.pkl.server.SnippetTests"
```
HT154, line 345: What's the advantage of using the test class? Does it out-weigh needing to hard code these names?

stackoverflow (approved):

pkl-internal-intellij-plugin/src/main/kotlin/org/pkl/internal/intellij/SnippetTestSplitEditor.kt
bioball: This is just so it matches what IntelliJ does when you click the "run" button in the gutter:
Screenshot 2025-11-04 at 7 33 24 AM

PR #1312, opened 2025-11-10
Fix Map formatting

### LINKED ISSUE #1307: Incorrect formatting following a line comment in `Map()` call
These are considered correctly formatted:

```pkl
foo =
  Map(
    1, 2,
    3, 4, // hi!
    5,
    6,
    7, 8,
  )

bar =
  Map(
    1, 2,
    3, 4,
    // hi!
    5,
    6,
    7, 8,
  )
```

The two arguments following the line comment should be on the same line

stackoverflow:
Fixes #1307

### DIFF
--- a/pkl-formatter/src/test/files/FormatterSnippetTests/input/map-function.pkl
+++ b/pkl-formatter/src/test/files/FormatterSnippetTests/input/map-function.pkl
@@ -8,6 +8,7 @@ bar =
     // leading
     1,
     2,
+    // between
     3,
     4, // trailing
     5,
--- a/pkl-formatter/src/test/files/FormatterSnippetTests/output/map-function.pkl
+++ b/pkl-formatter/src/test/files/FormatterSnippetTests/output/map-function.pkl
@@ -17,6 +17,7 @@ bar =
   Map(
     // leading
     1, 2,
+    // between
     3, 4, // trailing
     5, 6,
     7, // mid


pkl-formatter/src/main/kotlin/org/pkl/formatter/Builder.kt
```diff
@@ -229,9 +229,14 @@ internal class Builder(sourceText: String, private val grammarVersion: GrammarVe
     if (children.size == 1) return format(children[0])
     val firstNode = node.firstProperChild()!!
     return if (firstNode.text() == "Map") {
-      val nodes = mutableListOf<FormatNode>()
-      nodes += format(firstNode)
-      nodes += formatArgumentList(children[1], twoBy2 = true)
+      val nodes =
```
stackoverflow, line 232: This is a slightly related bugfix.

pkl-formatter/src/test/files/FormatterSnippetTests/input/map-function.pkl
```diff
@@ -2,3 +2,18 @@
...
 
 incorrect = Map("This has", 1000000, "an incorrect number", 2000000, "of parameters", 30000000, "passed to Map")
+
+bar =
+  Map(
+    // leading
+    1,
+    2,
+    3,
+    4, // trailing
```
HT154, line 13: Can we also test that line comments between pairs work as expected?
```suggestion
  Map(
    // leading
    1,
    2,
    // between
    3,
    4, // trailing
```

stackoverflow: Added.

PR #1315, opened 2025-11-12
Switch to GitHub Actions

bioball:
This switches our builds over to GitHub Actions!

TODO (to be done in follow-up PRs)

* Add macOS/amd64 native-image builds; this isn't working right now
* Patch musl with security patches
* Add benchmark jobs over time

As part of this build, PRBs will now only run `./gradlew check` on Linux, but other jobs can be run by adding slash commands in the PR description.

[native-pkl-doc-alpine-linux-amd64]

### DIFF
--- a/.github/index.pkl
+++ b/.github/index.pkl
@@ -64,7 +64,7 @@ local buildNativeJobs: Mapping<String, BuildNativeJob> = new {
           }
         }
       }
-      ["\(_project)-linux-alpine-amd64-\(_dist)"] {
+      ["\(_project)-alpine-linux-amd64-\(_dist)"] {
         arch = "amd64"
         os = "linux"
         musl = true
@@ -112,27 +112,30 @@ local releaseJobs: PklJobs = new {
 // By default, just run ./gradlew check on linux.
 // Trigger other checks based on GitHub PR description. Examples:
 //
-//    * /build windows                     -- Test on Windows
-//    * /build native                      -- Test all native builds
-//    * /build native-pkl-cli              -- Test all pkl-cli os/arch pairs
-//    * /build native-pkl-cli-macos        -- Test pkl-cli on macOS
+//    * [windows]                     -- Test on Windows
+//    * [native]                      -- Test all native builds
+//    * [native-pkl-cli]              -- Test all pkl-cli os/arch pairs
+//    * [native-pkl-cli-macos]        -- Test pkl-cli on macOS
 prb {
   local prbJobs: Mapping<String, GradleJob> = new {
     ["gradle-check"] = gradleCheck
     ["gradle-check-windows"] = (gradleCheckWindows) {
-      `if` = "contains(github.event.pull_request.body, '/build windows')"
+      `if` = "contains(github.event.pull_request.body, '[windows]')"
     }
     for (jobName, job in buildNativeSnapshots) {
       [jobName] = (job) {
-        local tags =
-          List(
-            "native",
-            "native-\(job.project)",
-            "native-\(job.project)-\(job.os)",
-            "native-\(job.project)-\(job.os)-\(job.arch)",
-          )
+        local tags = new Listing {
+          "[native]"
+          "[native-\(job.project)]"
+          "[native-\(job.project)-\(job.os)]"
+          "[native-\(job.project)-\(job.os)-\(job.arch)]"
+          "[native-\(job.project)-\(job.os)-\(job.arch)]"
+          when (job.musl) {
+            "[native-\(job.project)-alpine-\(job.os)-\(job.arch)]"
+          }
+        }
         `if` =
-          tags.map((it) -> "contains(github.event.pull_request.body, '/build \(it)')").join(" || ")
+          tags.toList().map((it) -> "contains(github.event.pull_request.body, '\(it)')").join(" || ")
       }
     }
   }
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -274,7 +274,7 @@ jobs:
         name: test-results-html-pkl-cli-linux-aarch64-snapshot
         path: '**/build/reports/tests/**/*'
         if-no-files-found: ignore
-  pkl-cli-linux-alpine-amd64-snapshot:
+  pkl-cli-alpine-linux-amd64-snapshot:
     runs-on: ubuntu-latest
     env:
       LANG: en_US.UTF-8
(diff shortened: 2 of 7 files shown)


.github/jobs/BuildNativeJob.pkl
```diff
@@ -0,0 +1,85 @@
...
+        fi
+        """#
+    }
+  }
+  new {
+    name = "gradle buildNative"
+    shell = "bash"
+    run =
+      """
+      export PATH=~/staticdeps/bin:$PATH
```
HT154: The previous step should probably modify the path itself, eg.
```bash
echo "${HOME}/staticdeps/bin" >> "$GITHUB_PATH"
```

.github/jobs/BuildNativeJob.pkl
```diff
@@ -0,0 +1,85 @@
...
+    new {
+      name = "Install musl and zlib"
+      // language=bash
+      run =
+        #"""
+        mkdir -p ~/staticdeps/
+
+        # install zlib
+        if [[ ! -f ~/staticdeps/include/zlib.h ]]; then
+          curl -Lf https://github.com/madler/zlib/releases/download/v\#(zlibVersion)/zlib-\#(zlibVersion).tar.gz -o /tmp/zlib.tar.gz
```
HT154: Can we add gpg signature verification for both zlib and musl? They both provide .asc files alongside their releases.

.github/jobs/DeployJob.pkl
```diff
@@ -0,0 +1,38 @@
...
+      ["ORG_GRADLE_PROJECT_signingKeyId"] = Context.secrets("ORG_GRADLE_PROJECT_SIGNINGKEYID")
+      ["ORG_GRADLE_PROJECT_signingKey"] = Context.secrets("ORG_GRADLE_PROJECT_SIGNINGKEY")
+      ["ORG_GRADLE_PROJECT_signingPassword"] =
+        Context.secrets("ORG_GRADLE_PROJECT_SIGNINGPASSWORD")
+      ["ORG_GRADLE_PROJECT_sonatypePassword"] =
+        Context.secrets("ORG_GRADLE_PROJECT_SONATYPEPASSWORD")
+      ["ORG_GRADLE_PROJECT_sonatypeUsername"] =
+        Context.secrets("ORG_GRADLE_PROJECT_SONATYPEUSERNAME")
+    }
+    run = "/gradlew \(self.gradleArgs) \(module.command)"
```
HT154:
```suggestion
    run = "./gradlew \(self.gradleArgs) \(module.command)"
```

.github/jobs/GithubRelease.pkl
```diff
@@ -0,0 +1,37 @@
...
+  needs = "deploy-release"
+  steps {
+    new Artifact.Download {}
+    new {
+      name = "Publish release on GitHub"
+      env {
+        ["GH_TOKEN"] = Context.github.token
+        ["TAG_NAME"] = Context.github.refName
+        ["GIT_SHA"] = Context.github.sha
+        ["GIT_REPO"] = Context.github.repository
```
HT154: Looks we can set this and avoid passing `--repo` to the CLI
```suggestion
        ["GH_REPO"] = Context.github.repository
```

.github/jobs/GithubRelease.pkl
```diff
@@ -0,0 +1,37 @@
+import "@gha/actions/Artifact.pkl"
+import "@gha/Context.pkl"
+import "@gha/Workflow.pkl"
+
+fixed job: Workflow.Job = new {
+  `runs-on` = "ubuntu-24.04"
```
HT154: Is there a reason not to use `ubuntu-latest` here? The specific version names are already "un-pinned" in that github can and does change them.

.github/jobs/GradleJob.pkl
```diff
@@ -0,0 +1,93 @@
...
+fixed jdkGitHubReleaseName =
+  let (
+    ver =
+    // 17.0.9+9 is missing some binaries (see https://github.com/adoptium/adoptium-support/issues/994)
+      if (javaVersionFull == "17.0.9+9" && os == "windows")
+        "jdk-17.0.9+9.1"
+      else
+        "jdk-\(javaVersionFull)"
+  )
+    URI.encodeComponent(ver)
```
HT154: These all look unused
```suggestion
```

.github/jobs/GradleJob.pkl
```diff
@@ -0,0 +1,93 @@
...
+isRelease: Boolean = false
+
+/// The architecture to use
+arch: "amd64" | "aarch64" = "amd64"
+
+/// The OS to run on.
+os: "macOS" | "linux" | "windows" = "linux"
+
+/// The version of Java to use.
+javaVersion: "17.0" | "21.0"
```
HT154:
```suggestion
javaVersion: "17" | "21"
```

.github/jobs/SimpleGradleJob.pkl
```diff
@@ -0,0 +1,20 @@
...
+command: String
+
+os = "linux"
+
+javaVersion = "21.0"
+
+steps {
+  new {
+    name = module.name
+    shell = "bash"
```
HT154, line 12: Do we specifically care about overriding the default "unspecified" behavior here? https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsshell

.github/jobs/GradleJob.pkl
```diff
@@ -0,0 +1,93 @@
...
+      "windows-latest"
+    else
+      new Listing { "self-hosted"; "macos" }
+  steps {
+    // full checkout (needed for spotless)
+    new Common.Checkout {
+      with {
+        `fetch-depth` = 0
+      }
+    }
```
HT154, line 78: Should we only add this on jobs that actually run spotless?

.github/workflows/build.yml
HT154, line 1: Might be nice to use .gitattributes to mark the workflow outputs as generated

.github/index.pkl
```diff
@@ -119,8 +119,7 @@ prb {
             "native-\(job.project)-\(job.os)",
             "native-\(job.project)-\(job.os)-\(job.arch)",
           )
-        `if` =
-          tags.map((it) -> "startsWith(github.event.comment.body, '/build \(it)')").join(" || ")
+        `if` = tags.map((it) -> "contains(github.event.pull_request.body, '/build \(it)')").join(" || ")
```
StefMa: [`Context.github.event("pull_request.body")`](https://github.com/StefMa/pkl-gha/blob/7df223000cacc1ca06c4a412a5808402298afc0a/Context.pkl#L60) 🤓 

```suggestion
        `if` = tags.map((it) -> "contains(\(Context.github.event("pull_request.body")), '/build \(it)')").join(" || ")
```

.github/jobs/GradleJob.pkl
```diff
@@ -0,0 +1,93 @@
...
+    when (os == "windows") {
+      ["JAVA_HOME"] = "/jdk"
+    }
+  }
+  when (os == "macOS") {
+    `if` = "github.repository_owner == 'apple'"
+  }
+  `runs-on` =
+    if (os == "linux" && arch == "amd64")
+      "ubuntu-24.04"
```
StefMa: Its recommended to use `-latest`.
Or any other good reasons to pin it to this version? 🤔

.github/index.pkl
HT154: I don't think this is right since the `Context` APIs produce template tags (eg. `${{ github.even.pull_request.body }}`) but `if` should omit the tags.

.github/jobs/DeployJob.pkl
```diff
@@ -0,0 +1,38 @@
...
+    env {
+      ["ORG_GRADLE_PROJECT_signingKeyId"] = Context.secrets("ORG_GRADLE_PROJECT_SIGNINGKEYID")
+      ["ORG_GRADLE_PROJECT_signingKey"] = Context.secrets("ORG_GRADLE_PROJECT_SIGNINGKEY")
+      ["ORG_GRADLE_PROJECT_signingPassword"] =
+        Context.secrets("ORG_GRADLE_PROJECT_SIGNINGPASSWORD")
+      ["ORG_GRADLE_PROJECT_sonatypePassword"] =
+        Context.secrets("ORG_GRADLE_PROJECT_SONATYPEPASSWORD")
+      ["ORG_GRADLE_PROJECT_sonatypeUsername"] =
+        Context.secrets("ORG_GRADLE_PROJECT_SONATYPEUSERNAME")
+    }
```
HT154: These should come from https://github.com/apple/pkl-project-commons/pull/28 once it's merged and released.

.github/jobs/GithubRelease.pkl
```diff
@@ -0,0 +1,37 @@
...
+  permissions {
+    contents = "write"
+  }
+  needs = "deploy-release"
+  steps {
+    new Artifact.Download {}
```
HT154: Should we be specific about which artifact(s) we're downloading here?

.github/jobs/SimpleGradleJob.pkl
bioball: Yeah, we need to ensure that Windows is using bash too

StefMa:
> As part of this build, PRBs will now only run ./gradlew check on Linux, but other jobs can be run by adding slash commands in the PR description. For example, /build windows to run gradle check on Windows.

Is there a specific reason for not running all the others by default? 🤔  
If so, you might want to consider adding a PR template to make sure this feature isn’t forgotten.  
The issue with the current approach is that if you forget to include the command in the PR description, you’ll need to update the body and push another commit to trigger it again.  
Alternatively, you could add the workflow trigger `on: issue_comment`, which would allow you to simply comment with the slash command instead. 🤷‍♂️  

Just some random thoughts on this.

.github/jobs/BuildNativeJob.pkl
bioball: TIL about $GITHUB_PATH!

.github/jobs/BuildNativeJob.pkl
bioball: Good idea!

.github/jobs/DeployJob.pkl
bioball: Good catch!

.github/jobs/GithubRelease.pkl
bioball: Yeah, good point!

.github/workflows/build.yml
bioball: Good idea, added

.github/jobs/GradleJob.pkl
bioball: Got rid of this; we have no jobs that use Java 17

.github/jobs/GradleJob.pkl
bioball: Good idea!

HT154 (approved):
LGTM overall! Still one pending question about more specific artifact downloads.

.github/jobs/GithubRelease.pkl
bioball: Added!

PR #1323, opened 2025-11-18
Fix empty parethesized type unexpected error

spyoungtech:
When a parethesized type atom is empty, an unexpected error occurs (due to an `IndexOutOfBoundsException`)

For example take `t.pkl` as a reproducible example:

```pkl
class Bird { name: String }

bird1: String|Bird = "Pigeon"
bird2: String|()|Bird = new Bird { name = "Pigeon" }
```

**Unexpected behavior:** an unspecified error message is shown without source information:
```
$ pkl eval -f yaml t.pkl
```

**Expected behavior:** the parser gracefully throws an error that provides details of the error in the source file.

This PR addresses this issue by first checking if the `children` array is empty, and if it is, throwing a parser error when it is empty.

After applying the fix proposed here, the expected behavior occurs: 

```
$ pkl eval -f yaml t.pkl
–– Pkl Error ––
Unexpected token `)`. Expected a type.

4 | bird2: String|()|Bird = new Bird { name = "Pigeon" }
                   ^
at t (file:///Users/spencer/scratch/pkl/t.pkl, line 4)
```

### DIFF
--- a/pkl-parser/src/main/java/org/pkl/parser/Parser.java
+++ b/pkl-parser/src/main/java/org/pkl/parser/Parser.java
@@ -1394,8 +1394,7 @@ public class Parser {
           typ = new Type.FunctionType(children, tk.span.endWith(ret.span()));
         } else {
           if (children.isEmpty()) {
-            backtrack();
-            throw parserError("unexpectedTokenForType", ")");
+            throw new ParserError(ErrorMessages.create("unexpectedTokenForType", ")"), end);
           }
           typ = new ParenthesizedType((Type) children.get(0), tk.span.endWith(end));
         }


pkl-parser/src/main/java/org/pkl/parser/Parser.java
```diff
@@ -1393,6 +1393,10 @@ private Type parseTypeAtom(@Nullable String expectation) {
           children.add(ret);
           typ = new Type.FunctionType(children, tk.span.endWith(ret.span()));
         } else {
+          if (children.isEmpty()) {
+            backtrack();
+            throw parserError("unexpectedTokenForType", ")");
+          }
```
stackoverflow, line 1398:
```suggestion
          if (children.isEmpty()) {
            throw parserError("unexpectedTokenForType", ")");
          }
```
You don't need to backtrack if you are going to throw an error.

spyoungtech: The main reason I thought to backtrack here was to get the error to point to the correct place, since by this point the closing RPAREN token has already been consumed. Unless there's a better way to do that or if it's okay for the error to point at the subsequent token.

Without the `backtrack()`, the error points at the next token:


```
–– Pkl Error ––
Unexpected token `)`. Expected a type.

4 | bird2: String|()|Bird = new Bird { name = "Pigeon" }
                    ^
at t (file:///Users/spencer/scratch/pkl/t.pkl, line 4)

Comment (stackoverflow) [reply pkl-parser/src/main/java/org/pkl/parser/Parser.java:1398]: Yeah, good point, the error should point to `)`, but the proper way to do it, is to pass the span directly to the error:
```java
if (children.isEmpty()) {
  throw new ParserError(ErrorMessages.create("unexpectedTokenForType", ")"), end);
}
```
`backtrack` should only be used for actual parser backtrack.

spyoungtech: Makes sense! Made that adjustment.

stackoverflow (approved):
Looks good!

PR #1332, opened 2025-11-25
Add compose and andThen function to Function1

stackoverflow:


bioball (changes requested):
Can you describe the motivation for this?
The language already provides the pipe operator, which provides similar functionality.
So, to add this, I think we should have some rationale for why an additional API makes sense.

```pkl
local addOne = (x) -> x + 1
local double = (x) -> x * 2

res1 = 5 |> addOne |> double
```

Might be good to have a SPICE.

PR #1333, opened 2025-11-25
SPICE-0024: `Annotation` converters

### LINKED ISSUE #576: Output converters should allow selection for properties by Annotation
Currently, annotations are a helpful language feature for modules using `pkl:reflect` and tooling like IDEs, but they're not broadly useful in most modules. One way they could be made more useful is through supporting them as a type of match for `ValueRenderer` converters.

## Background

Here's an example where current language features fail to fully meet a module's needs. Take a module that has been adapted to validate property availability in multiple versions of a service:
```pkl
module MyServiceConfig

hidden targetVersion: UInt = 2

function availableBefore(version: UInt): Boolean = if (targetVersion < version) true
  else throw("available before \(version), currently targeting \(targetVersion)")

function availableAfter(version: UInt): Boolean = if (version >= targetVersion) true
  else throw("available after \(version), currently targeting \(targetVersion)")

myOldProperty: String(availableBefore(2))?

myNewProperty: String(availableAfter(2))?
```

The above code works just fine, but problems arise the moment this needs to be used outside the module's top level:

```pkl
myNestedValue: MyNestedClass

class MyNestedClass {
  nestedOldProperty: String(availableBefore(2))?
}
```

This produces an error:
```
Cannot call method availableBefore from here because it is not const.  Classes, typealiases, and annotations can only reference const members of their enclosing module.
To fix, either make the accessed member const, or add a self-import of this module, and access this member off of the self import.
```

Neither of the proposed solutions work here.
* If `availableBefore`/`availableAfter` are `const`, then `targetVersion` must also be `const`, but this design specifically requires that amending modules be able to set a `targetVersion`.
* The self-import also doesn't work here since it always gets the default value and not the one from the amending module.

One workaround today might be to define an output converter to handle this:
```pkl
module MyServiceConfig

hidden targetVersion: UInt = 2

function availableBefore(version: UInt): Null = if (targetVersion < version) null
  else throw("available before \(version), currently targeting \(targetVersion)")

function availableAfter(version: UInt): Null = if (version >= targetVersion) null
  else throw("available after \(version), currently targeting \(targetVersion)")

myOldProperty: String?

myNewProperty: String?

myNestedValue: MyNestedClass

class MyNestedClass {
  nestedOldProperty: String?
}

output {
  renderer = new YamlRenderer {
    converters {
      ["myOldProperty"] = (it) -> availableBefore(2) ?? it
      ["myNewProperty"] = (it) -> availableAfter(2) ?? it
      ["myNestedValue.nestedOldProperty"] = (it) -> availableBefore(2) ?? it
    }
  }
}

```

The major downside here is that this separates the information about availability from the property definition and assumes that `MyServiceConfig` will always be the root module being evaluated. This is particularly fatal in cases like Kubernetes configurations where there are many fields in many modules, some of which may not be known to the parent module at time of writing.

## Proposal

My proposed solution here is to allow converters to match on annotation classes. The converter function would need to be passed both the annotation value and the underlying property value, so it would either need to be allowed to be a `Function2<«annotation value», «property value», Any>` or be passed a `Pair<«annotation value», «property value»>` (or possibly some new class eg. `ValueRendererConverterMatchAnnotation`).

This would make annotations broadly useful in a variety of scenarios, including the versioning example above. Here's how that example might look if this were implemented:

```pkl
module MyServiceConfig

hidden targetVersion: UInt = 2

class Available extends Annotation {
  before: UInt?

  after: UInt?

  function evaluate(target: UInt, value: Any): Any =
    if (before != null && target >= before) throw("available before \(before), currently targeting \(target)")
    else if (after != null && after < target) throw("available after \(after), currently targeting \(target)")
    else value
}

@Available { before = 2 }
myOldProperty: String?

@Available { after = 2 }
myNewProperty: String?

myNestedValue: MyNestedClass

class MyNestedClass {
  @Available { before = 2 }
  nestedOldProperty: String?
}

output {
  renderer = new YamlRenderer {
    converters {
      [Available] = (it: Pair<Available, Any>) -> it.key.evaluate(targetVersion, it.value)
      // OR
      [Available] = (available: Available, value: Any) -> available.evaluate(targetVersion, value)
    }
  }
}
```

Modules embedding this one would need to copy its converters to inherit this behavior, but that's very doable since this is portable and not dependent on property key paths like the workaround above.

HT154:
This has been banging around in my head for a while and I finally needed to get it out.

This enables defining declarative key and/or value transformations in cases where neither `Class`- nor path-based converters can be applied gracefully. It is also the only way to express transforming the resulting property names in `Typed` objects without applying a converter to the entire containing type, which is cumbersome at best.

SPICE: https://github.com/apple/pkl-evolution/pull/26

This also fixes a silly wart in the API for getting class property mirrors added in #1106.

Resolves #576

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/ast/member/ClassProperty.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/member/ClassProperty.java
@@ -54,7 +54,6 @@ public final class ClassProperty extends ClassMember {
     this.initializer = initializer;
   }
 
-  /** Returned annotations are ordered from base class to leaf class then from top to bottom */
   public List<VmTyped> getAllAnnotations(boolean ascending) {
     var annotations = new ArrayList<VmTyped>();
 
--- a/pkl-core/src/main/java/org/pkl/core/runtime/Identifier.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/Identifier.java
@@ -50,7 +50,7 @@ public final class Identifier implements Comparable<Identifier> {
   // members of pkl.base#Listing and pkl.base#Mapping
   public static final Identifier DEFAULT = get("default");
 
-  // members of pkl.base#ValueRenderer subclasses
+  // members of pkl.base#BaseValueRenderer subclasses
   public static final Identifier MODE = get("mode");
   public static final Identifier INDENT = get("indent");
   public static final Identifier INDENT_WIDTH = get("indentWidth");
@@ -62,8 +62,11 @@ public final class Identifier implements Comparable<Identifier> {
   public static final Identifier ROOT_ELEMENT_NAME = get("rootElementName");
   public static final Identifier ROOT_ELEMENT_ATTRIBUTES = get("rootElementAttributes");
   public static final Identifier CONVERTERS = get("converters");
-  public static final Identifier ANNOTATION_TRANSFORMERS = get("annotationTransformers");
+  public static final Identifier CONVERT_PROPERTY_TRANSFORMERS = get("convertPropertyTransformers");
   public static final Identifier USE_MAPPING = get("useMapping");
+
+  // members of pkl.base#ConvertProperty
+  public static final Identifier RENDER = get("render");
 
   // members of pkl.base#RegexMatch
   public static final Identifier VALUE = get("value");
@@ -143,9 +146,6 @@ public final class Identifier implements Comparable<Identifier> {
 
   // members of pkl.yaml
   public static final Identifier MAX_COLLECTION_ALIASES = get("maxCollectionAliases");
-
-  // members of pkl.convert
-  public static final Identifier RENDER = get("render");
 
   // common in lambdas etc
   public static final Identifier IT = get("it");
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmFunction.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmFunction.java
@@ -64,12 +64,6 @@ public final class VmFunction extends VmObjectLike {
     return getCallTarget().call(thisValue, this, arg1, arg2);
   }
 
-  // if call site is a node, use ApplyVmFunction3Node.execute() or DirectCallNode.call() instead of
-  // this method
-  public Object apply(Object arg1, Object arg2, Object arg3) {
-    return getCallTarget().call(thisValue, this, arg1, arg2, arg3);
-  }
-
   public VmFunction copy(
       int newParamCount, @Nullable PklRootNode newRootNode, @Nullable Object newExtraStorage) {
     return new VmFunction(
--- a/pkl-core/src/main/java/org/pkl/core/stdlib/AbstractRenderer.java
+++ b/pkl-core/src/main/java/org/pkl/core/stdlib/AbstractRenderer.java
@@ -163,9 +163,6 @@ public abstract class AbstractRenderer implements VmValueVisitor {
   /** Visits a property of a {@link VmDynamic} or {@link VmTyped}. */
   protected abstract void visitProperty(Identifier name, Object value, boolean isFirst);
 
-  /** Perform logic for rendering a render directive in place of a property */
-  protected abstract void visitPropertyRenderDirective(VmTyped value, boolean isFirst);
-
   protected abstract void endDynamic(VmDynamic value, boolean isEmpty);
 
   protected abstract void endTyped(VmTyped value, boolean isEmpty);
@@ -338,14 +335,14 @@ public abstract class AbstractRenderer implements VmValueVisitor {
   private void doVisitProperty(
       Identifier name,
       Object value,
-      @Nullable ClassProperty property,
+      @Nullable ClassProperty classProperty,
       SourceSection sourceSection,
       MutableBoolean isFirst) {
     var prevSourceSection = currSourceSection;
     currSourceSection = sourceSection;
     currPath.push(name);
-    if (property != null) {
-      var propVal = converter.convertProperty(property, value, currPath);
+    if (classProperty != null) {
+      var propVal = converter.convertProperty(classProperty, value, currPath);
       name = propVal.getFirst();
       value = propVal.getSecond();
     }
--- a/pkl-core/src/main/java/org/pkl/core/stdlib/PklConverter.java
+++ b/pkl-core/src/main/java/org/pkl/core/stdlib/PklConverter.java
@@ -23,7 +23,7 @@ import org.pkl.core.util.Pair;
 
 public final class PklConverter implements VmValueConverter<Object> {
   private final Map<VmClass, VmFunction> typeConverters;
-  private final Map<VmClass, VmFunction> annotationTransformers;
+  private final Map<VmClass, VmFunction> convertPropertyTransformers;
   private final Pair<Object[], VmFunction>[] pathConverters;
   private final Object rendererOrParser;
 
@@ -48,13 +48,12 @@ public final class PklConverter implements VmValueConverter<Object> {
   private final @Nullable VmFunction typeAliasConverter;
 
   private PklConverter(
-      VmMapping converters, VmMapping annotationTransformers, Object rendererOrParser) {
-    // As of 0.18, `converters` is forced by the mapping type check,
-    // but let's not rely on this implementation detail.
+      VmMapping converters, VmMapping convertPropertyTransformers, Object rendererOrParser) {
     converters.force(false, false);
-    annotationTransformers.force(false, false);
+    convertPropertyTransformers.force(false, false);
     typeConverters = createTypeConverters(converters);
-    this.annotationTransformers = createAnnotationTransformers(annotationTransformers);
+    this.convertPropertyTransformers =
+        createConvertPropertyTransformers(convertPropertyTransformers);
     pathConverters = createPathConverters(converters);
     this.rendererOrParser = rendererOrParser;
 
@@ -84,9 +83,9 @@ public final class PklConverter implements VmValueConverter<Object> {
 
   public static PklConverter fromRenderer(VmTyped renderer) {
     var converters = (VmMapping) VmUtils.readMember(renderer, Identifier.CONVERTERS);
-    var annotationTransformers =
-        (VmMapping) VmUtils.readMember(renderer, Identifier.ANNOTATION_TRANSFORMERS);
-    return new PklConverter(converters, annotationTransformers, renderer);
+    var convertPropertyTransformers =
+        (VmMapping) VmUtils.readMember(renderer, Identifier.CONVERT_PROPERTY_TRANSFORMERS);
+    return new PklConverter(converters, convertPropertyTransformers, renderer);
   }
 
   public static PklConverter fromParser(VmTyped parser) {
@@ -216,7 +215,7 @@ public final class PklConverter implements VmValueConverter<Object> {
         continue;
       }
 
-      var transformer = findAnnotationTransformer(annotation.getVmClass());
+      var transformer = findConvertPropertyTransformer(annotation.getVmClass());
       if (transformer != null) {
         annotation = (VmTyped) transformer.apply(annotation);
       }
@@ -241,9 +240,10 @@ public final class PklConverter implements VmValueConverter<Object> {
     return result;
   }
 
-  private Map<VmClass, VmFunction> createAnnotationTransformers(VmMapping annotationTransformers) {
+  private Map<VmClass, VmFunction> createConvertPropertyTransformers(
+      VmMapping convertPropertyTransformers) {
     var result = new HashMap<VmClass, VmFunction>();
-    annotationTransformers.iterateMemberValues(
+    convertPropertyTransformers.iterateMemberValues(
         (key, member, value) -> {
           assert value != null; // forced in ctor
           result.put((VmClass) key, ((VmFunction) value));
@@ -287,8 +287,8 @@ public final class PklConverter implements VmValueConverter<Object> {
     return findConverterByType(typeConverters, clazz);
   }
 
-  private @Nullable VmFunction findAnnotationTransformer(VmClass clazz) {
-    return findConverterByType(annotationTransformers, clazz);
+  private @Nullable VmFunction findConvertPropertyTransformer(VmClass clazz) {
+    return findConverterByType(convertPropertyTransformers, clazz);
   }
 
   private <T> @Nullable T findConverterByType(Map<VmClass, T> bag, VmClass clazz) {
--- a/pkl-core/src/test/files/LanguageSnippetTests/input/api/annotationConverters.pkl
+++ b/pkl-core/src/test/files/LanguageSnippetTests/input/api/annotationConverters.pkl
@@ -40,7 +40,11 @@ open class Foo {
 
   @SubtractValue { difference = 2 }
   @MultiplyValue { factor = 3 }
-  @SubtractValue { difference = 4 }
+  @ConvertProperty {
+    render = (prop, _) -> Pair(prop.key, prop.value as Number + 4)
+  }
+  @MultiplyValue { factor = 3 }
+  @SubtractValue { difference = 11 }
   in_order: Int = 5
 }
 
@@ -52,7 +56,7 @@ class Bar extends Foo {
 output {
   value = new Bar {}
   renderer = new PcfRenderer {
-    annotationTransformers {
+    convertPropertyTransformers {
       [Prefix] { prefix = "foo_" }
     }
   }
--- a/pkl-core/src/test/files/LanguageSnippetTests/input/api/pListRenderer8.plist.pkl
+++ b/pkl-core/src/test/files/LanguageSnippetTests/input/api/pListRenderer8.plist.pkl
@@ -16,7 +16,7 @@ class Nested {
 
 output {
   renderer = new PListRenderer {
-    annotationTransformers {
+    convertPropertyTransformers {
       [annotationConverter.LineComment] { prefix = "<!-- "; suffix = " -->" }
     }
   }
--- a/pkl-core/src/test/files/LanguageSnippetTests/output/errors/cannotFindStdLibModule.err
+++ b/pkl-core/src/test/files/LanguageSnippetTests/output/errors/cannotFindStdLibModule.err
@@ -9,7 +9,6 @@ Available standard library modules:
 pkl:analyze
 pkl:base
 pkl:Benchmark
-pkl:convert
 pkl:DocPackageInfo
 pkl:DocsiteInfo
 pkl:EvaluatorSettings
--- a/stdlib/base.pkl
+++ b/stdlib/base.pkl
@@ -351,7 +351,6 @@ abstract class BaseValueRenderer {
   /// Paths are matched against path specs component-wise in reverse order.
   /// For example, paths `server.timeout` and `racks[*].server.timeout`
   /// both match path spec `server.timeout`, whereas path `server.timeout.millis` does not.
-  ///
   converters: Mapping<Class | String, (unknown) -> Any>
 
   /// Customizations for [ConvertProperty] annotation behaviors.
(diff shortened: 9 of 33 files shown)


stdlib/base.pkl
```diff
@@ -20,10 +20,13 @@
 @ModuleInfo { minPklVersion = "0.31.0" }
 module pkl.base
 
+import "pkl:json"
 import "pkl:jsonnet"
+import "pkl:math"
 import "pkl:pklbinary"
 import "pkl:protobuf"
 import "pkl:xml"
+import "pkl:yaml"
```
HT154, line 30: Note that these modules are unused by functional code, but are required to correctly resolve doc comment references.

bioball: Do you mind submitting this in a separate PR?

pkl-core/src/main/java/org/pkl/core/ast/member/ClassProperty.java
```diff
@@ -54,6 +54,17 @@ public ClassProperty(
     this.initializer = initializer;
   }
 
+  public List<VmTyped> getAllAnnotations() {
+    var annotations = new ArrayList<VmTyped>();
+    for (var klazz = getDeclaringClass(); klazz != null; klazz = klazz.getSuperclass()) {
```
bioball: [nit] Can you change all of these to `clazz`? That's the name we use everywhere else.

pkl-core/src/test/files/LanguageSnippetTests/input/api/jsonRenderer9.json.pkl
```diff
@@ -0,0 +1,21 @@
+// test json.Property implicit annotation converter
```
bioball: Rename this to `jsonRenderer9.json.pkl`

stdlib/base.pkl
HT154: Only `math` is affected in main: https://github.com/apple/pkl/pull/1400

pkl-core/src/test/files/LanguageSnippetTests/output/api/jsonRenderer9.json
```diff
@@ -0,0 +1,6 @@
+{
+  "foo": {
+    "new-name": "hello world"
+  },
+  "unaltered": "{\n  \"oldName\": \"hello world\"\n}\n"
+}
```
bioball: Delete this file and re-generate (should rename the file to `jsonRenderer9.json`)

stdlib/base.pkl
```diff
@@ -464,6 +488,10 @@ class PcfRenderDirective {
 }
 
 /// Renders values as JSON.
+///
+/// An annotation converter for [json.Property] is implicitly added to allow renaming properties
+/// during JSON rendering.
+/// Add an explicit annotation converter for [json.Property] to override this behavior.
```
bioball: This comment seems hard to understand; users might not know what an annotation converter is, and it's also somewhat unnecessary information (using `json.Property` doesn't require knowing what annotation converters are)

I think it would be better to have an example instead, something like:

```suggestion
///
/// The [json.Property] annotation can be used to change how a property name renders into JSON.
///
/// Example:
/// ```
/// import "pkl:json"
///
/// @json.Property { name = "wing_span" }
/// wingSpan: Int
/// ```
```

I think we should also remove the part about overriding this behavior, because it seems like quite an edge case.

stdlib/base.pkl
```diff
@@ -347,8 +351,28 @@ abstract class BaseValueRenderer {
...
+  /// annotations (top to bottom) of the current class, then repeating for each parent class in the
+  /// hierarchy.
+  /// Multiple annotation converters can apply; converters for matching annotations are called in
+  /// order, with the result of one being passed to the next.
+  ///
+  /// At most one converter is applied per annotation. The most specific class (according to class
+  /// hierarchy) wins.
+  ///
+  /// Annotation converters can overlap with [converters] and apply first.
+  annotationConverters: Mapping<Class, (String, Annotation, unknown) -> Pair<String, Any>>
```
bioball: Here, and everywhere else that has a new API:

```suggestion
  @Since { version = "0.31.0" }
  annotationConverters: Mapping<Class, (String, Annotation, unknown) -> Pair<String, Any>>
```

stdlib/base.pkl
```diff
@@ -347,8 +351,28 @@ abstract class BaseValueRenderer {
...
+  /// converter function must return a [Pair] of the converted property name and value.
+  ///
+  /// Each of the property annotations of the enclosing class is checked in order, starting with the
+  /// annotations (top to bottom) of the current class, then repeating for each parent class in the
+  /// hierarchy.
+  /// Multiple annotation converters can apply; converters for matching annotations are called in
+  /// order, with the result of one being passed to the next.
+  ///
+  /// At most one converter is applied per annotation. The most specific class (according to class
+  /// hierarchy) wins.
```
bioball: This seems to conflict with the previous paragraph?

stdlib/base.pkl
```diff
@@ -347,8 +351,28 @@ abstract class BaseValueRenderer {
...
+  ///
+  /// Annotation converters can overlap with [converters] and apply first.
```
bioball:
```suggestion
  /// Annotation converters can coexist with [converters], and apply first.
```

HT154:
> Look good. One question: As annotation converters run subclass -> superclass, you may get into cases where you can't override a superclass converter right?

Yeah, that's true. I wonder if this approach is worth revising. There are two main alternatives I see here:
* Reversing the order, so annotations are walked from parent to child class.
* Continue walking child->parent, but only allow each `annotationConverters` entry to be matched once per property. Example:
```pkl
abstract class RenameConverter extends Annotation {
  abstract function rename(name: String): String
}

class SnakeCase extends RenameConverter {
  // ...
}

class CamelCase extends RenameConverter {
  // ...
}

output {
  renderer {
    annotationConverters {
      [RenameConverter] = (name, annotation, value) -> Pair(annotation.rename(name), value)
    }
  }
}
```

Here, only the `CamelCase` converter would run because it would be checked first and match against `RenameConverter`. Subsequently, `SnakeCase` would be checked and not match any unused converter entries so it would be ignored.

This approach is more complex in that it's more rules to keep in mind when working with annotation converters, but it's also simpler: the user does not have to keep track of how each annotation's implementation might interact.

Thoughts on which route might be preferable here?

stdlib/base.pkl
bioball: More specific comment on this API:

It seems strange that the annotation is put in between the property name and the value.
E.g.

```pkl
annotationConverters {
  [MyAnnotation] = (key, annotation, value) -> /* etc */
}
```

This feels more natural:

```pkl
annotationConverters {
  [MyAnnotation] = (annotation, key, value) -> /* etc */
}
```

Sidenote: we should improve the type inference on the pkl-intellij and pkl-lsp side, so that `annotation` is inferred to be of type `MyAnnotation`.
Our type system doesn't support this, but we can special-case the flow typing here (we already do this for `converters`).

stdlib/base.pkl
HT154: Updated this to
```pkl
  /// Multiple annotation converters can apply per property; converters for matching annotations are
  /// called in order, with the result of one being passed to the next.
  ///
  /// At most one converter is applied per annotation. The most specific class (according to class
  /// hierarchy) wins.
```

Is this clearer?

stdlib/base.pkl
HT154: Agreed on all fronts!

pkl-core/src/main/java/org/pkl/core/ast/member/ClassProperty.java
```diff
@@ -54,6 +54,45 @@ public ClassProperty(
     this.initializer = initializer;
   }
 
+  /** Returned annotations are ordered from base class to leaf class then from top to bottom */
```
bioball: [nit] incorrect comment; iteration order is specified according to the ascending parameter

pkl-core/src/main/java/org/pkl/core/runtime/Identifier.java
```diff
@@ -143,8 +144,8 @@ public final class Identifier implements Comparable<Identifier> {
   // members of pkl.yaml
   public static final Identifier MAX_COLLECTION_ALIASES = get("maxCollectionAliases");
 
-  // members of pkl.encoding
-  public static final Identifier IMPORTS = get("imports");
+  // members of pkl.convert
```
bioball: [nit] incorrect

pkl-core/src/main/java/org/pkl/core/stdlib/AbstractRenderer.java
```diff
@@ -162,6 +163,9 @@ public void visit(Object value) {
   /** Visits a property of a {@link VmDynamic} or {@link VmTyped}. */
   protected abstract void visitProperty(Identifier name, Object value, boolean isFirst);
 
+  /** Perform logic for rendering a render directive in place of a property */
+  protected abstract void visitPropertyRenderDirective(VmTyped value, boolean isFirst);
```
bioball: Unused; remove

pkl-core/src/main/java/org/pkl/core/stdlib/AbstractRenderer.java
```diff
@@ -327,10 +336,19 @@ public final void visitMap(VmMap value) {
   }
 
   private void doVisitProperty(
-      Identifier name, Object value, SourceSection sourceSection, MutableBoolean isFirst) {
+      Identifier name,
+      Object value,
+      @Nullable ClassProperty property,
```
bioball: [nit]

```suggestion
      @Nullable ClassProperty classProperty,
```

pkl-core/src/main/java/org/pkl/core/stdlib/PklConverter.java
```diff
@@ -44,12 +47,16 @@ public final class PklConverter implements VmValueConverter<Object> {
   private final @Nullable VmFunction classConverter;
   private final @Nullable VmFunction typeAliasConverter;
 
-  public PklConverter(VmMapping converters) {
+  private PklConverter(
+      VmMapping converters, VmMapping annotationTransformers, Object rendererOrParser) {
     // As of 0.18, `converters` is forced by the mapping type check,
     // but let's not rely on this implementation detail.
```
bioball: Ah, straggler comment: this is not true anymore. Let's get rid of this!

pkl-core/src/test/files/LanguageSnippetTests/input/api/pListRenderer8.plist.pkl
```diff
@@ -0,0 +1,23 @@
...
+  @annotationConverter.LineComment { text = "qux" }
+  qux: String = "c"
+
+  quux: String = "c"
+}
+
+output {
+  renderer = new PListRenderer {
+    annotationTransformers {
+      [annotationConverter.LineComment] { prefix = "<!-- "; suffix = " -->" }
```
bioball, line 20: Huuuuge nitpick, but, XML just calls these comments, not line comments.

pkl-core/src/test/files/LanguageSnippetTests/output/errors/cannotFindStdLibModule.err
```diff
@@ -9,6 +9,7 @@ Available standard library modules:
 pkl:analyze
 pkl:base
 pkl:Benchmark
+pkl:convert
```
bioball: This is out-of-date

pkl-core/src/main/java/org/pkl/core/runtime/VmFunction.java
```diff
@@ -64,6 +64,12 @@ public Object apply(Object arg1, Object arg2) {
     return getCallTarget().call(thisValue, this, arg1, arg2);
   }
 
+  // if call site is a node, use ApplyVmFunction3Node.execute() or DirectCallNode.call() instead of
+  // this method
+  public Object apply(Object arg1, Object arg2, Object arg3) {
+    return getCallTarget().call(thisValue, this, arg1, arg2, arg3);
+  }
```
bioball: Remove; this is unused now

stdlib/base.pkl
```diff
@@ -466,6 +501,16 @@ class PcfRenderDirective {
 }
 
 /// Renders values as JSON.
+///
+/// The [json.Property] annotation can be used to change how a property name renders into XML.
```
bioball:
```suggestion
/// The [json.Property] annotation can be used to change how a property name renders into JSON.
```

stdlib/base.pkl
```diff
@@ -486,6 +531,16 @@ class JsonRenderer extends ValueRenderer {
 /// Renders values as YAML.
 ///
 /// To render a YAML stream, set [isStream] to [true].
+///
+/// The [yaml.Property] annotation can be used to change how a property name renders into XML.
```
bioball:
```suggestion
/// The [yaml.Property] annotation can be used to change how a property name renders into YAML.
```

stdlib/jsonnet.pkl
```diff
@@ -68,6 +68,16 @@ function ExtVar(_name: String): ExtVar = new { name = _name }
 ///   }
 /// }
 /// ```
+///
+/// The [Property] annotation can be used to change how a property name renders into XML.
```
bioball:
```suggestion
/// The [Property] annotation can be used to change how a property name renders into Jsonnet.
```

stdlib/protobuf.pkl
```diff
@@ -25,6 +25,16 @@ import "pkl:reflect"
 /// Note: This class is _experimental_ and not ready for production use.
 ///
 /// As of this release, only Protocol Buffers' text format is supported.
+///
+/// The [Property] annotation can be used to change how a property name renders into XML.
```
bioball:
```suggestion
/// The [Property] annotation can be used to change how a property name renders into Protobuf.
```

stdlib/base.pkl
```diff
@@ -349,13 +351,46 @@ abstract class BaseValueRenderer {
...
+/// The function must return a [Pair] of the converted property name and value.
+///
+/// Each of the property annotations of the enclosing class is checked in order, starting with the
+/// annotations (top to bottom) of the base class, then repeating for each child class in the
+/// hierarchy.
+/// Multiple annotation converters can apply per property; the output of one converter is used as
+/// input to the next.
+///
+/// Annotation converters can coexist with [BaseValueRenderer.converters] and apply first.
+open class ConvertProperty extends Annotation {
```
bioball, line 414: Here, and everywhere else that adds a new class or property

```suggestion
@Since { version = "0.31.0" }
open class ConvertProperty extends Annotation {
```

stdlib/base.pkl
```diff
@@ -349,13 +351,46 @@ abstract class BaseValueRenderer {
...
+  /// Customizations for [ConvertProperty] annotation behaviors.
+  ///
+  /// This property is consulted to transform [ConvertProperty] annotation values.
+  /// This can be used to customize or override the conversion behavior for a specific renderer.
+  /// If multiple entries match the annotation's class, the most specific class (according to class
+  /// hierarchy) wins.
+  ///
+  /// See [ConvertProperty] for detailed information about annotation converters.
+  @Since { version = "0.31.0" }
+  annotationTransformers: Mapping<Class, Mixin<ConvertProperty>>
```
bioball: How about let's call this `convertPropertyTransformers`? Since this is really only about the `ConvertProperty` annotation.

stdlib/base.pkl
```diff
@@ -349,13 +351,46 @@ abstract class BaseValueRenderer {
   /// Paths are matched against path specs component-wise in reverse order.
   /// For example, paths `server.timeout` and `racks[*].server.timeout`
   /// both match path spec `server.timeout`, whereas path `server.timeout.millis` does not.
+  ///
```
bioball:
```suggestion
```

stdlib/base.pkl
```diff
@@ -349,13 +351,46 @@ abstract class BaseValueRenderer {
...
+/// annotations (top to bottom) of the base class, then repeating for each child class in the
+/// hierarchy.
+/// Multiple annotation converters can apply per property; the output of one converter is used as
+/// input to the next.
+///
+/// Annotation converters can coexist with [BaseValueRenderer.converters] and apply first.
+open class ConvertProperty extends Annotation {
+  /// Function called by [BaseValueRenderer] types during rendering to transform property
+  /// names and values.
+  render: (Pair<String, Any>, BaseValueRenderer) -> (Pair<String, Any>)
```
bioball:
```suggestion
  render: (Pair<String, Any>, BaseValueRenderer) -> Pair<String, Any>
```

stdlib/base.pkl
```diff
@@ -349,13 +351,46 @@ abstract class BaseValueRenderer {
...
+/// function is called.
+/// The function must return a [Pair] of the converted property name and value.
+///
+/// Each of the property annotations of the enclosing class is checked in order, starting with the
+/// annotations (top to bottom) of the base class, then repeating for each child class in the
+/// hierarchy.
+/// Multiple annotation converters can apply per property; the output of one converter is used as
+/// input to the next.
+///
+/// Annotation converters can coexist with [BaseValueRenderer.converters] and apply first.
```
bioball:
```suggestion
/// Annotation converters can coexist with [BaseValueRenderer.converters], and apply first.
```

pkl-core/src/test/files/LanguageSnippetTests/input/api/pListRenderer8.plist.pkl
HT154: I know and I'm gonna leave this one as-is. It _is_ a comment that lands on the same line as the property data (except properties, where comments must be a full line).

stdlib/base.pkl
```diff
@@ -351,11 +353,47 @@ abstract class BaseValueRenderer {
...
+/// The function must return a [Pair] of the converted property name and value.
+///
+/// Each of the property annotations of the enclosing class is checked in order, starting with the
+/// annotations (top to bottom) of the base class, then repeating for each child class in the
+/// hierarchy.
+/// Multiple annotation converters can apply per property; the output of one converter is used as
+/// input to the next.
+///
+/// Annotation converters can coexist with [BaseValueRenderer.converters].
+/// They apply prior to class- and path-based converters.
```
bioball: Some suggestions here:

Feel free to re-word this; the key changes that I'd like to make are:

* Focus on "class property" (no need to mention `Typed`)
* Reword "Annotation converter" to "ConvertProperty annotation"
* Don't go into too much detail about how `convertPropertyTransformers` works; this is already explained in that property's doc comments.
* Add examples for how this works
* No "top to bottom" terminology (we don't use this anywhere else). "Parent" is preferred (e.g. parent property).

```suggestion
/// Conversion to be applied to properties when rendered through [BaseValueRenderer].
///
/// During rendering, the annotation's [render] function is called.
/// The function must return a [Pair] of the converted property name and value.
///
/// Multiple [ConvertProperty] annotations can apply per property, and the output of one
/// annotation's [render] function is used as input to the next.
/// Annotations are applied the order as declared on the property.
/// If the annotated property is overriding a parent property, the parent property's annotations are
/// applied first.
///
/// These conversions can co-exist with [BaseValueRenderer.converters], and applies first.
///
/// These conversions only affect rendering of class properties.
/// Applying this to other types of members does not impact rendering.
///
/// These conversions can be overriden with [BaseValueRenderer.convertPropertyTransformers].
///
/// Example:
///
/// ```
/// // convert duration to the number of seconds
/// @ConvertProperty {
///   render = (property, _) -> Pair(property.key, property.value.toUnit("s"))
/// }
/// timeout: Duration
/// ```
///
/// [ConvertProperty] can be subclassed to define re-usable property converters.
/// The conversion defined in the previous example can be rewritten as:
///
/// ```
/// class ConvertDuration extends ConvertProperty {
///   unit: DurationUnit
///
///   render = (property: Pair<String, Duration>, _) -> Pair(property.key, property.value.toUnit(unit))
/// }
///
/// @ConvertDuration { unit = "s" }
/// timeout: Duration
/// ```
```

pkl-core/src/test/files/LanguageSnippetTests/input/api/annotationConverters.pkl
```diff
@@ -0,0 +1,59 @@
...
+}
+
+output {
+  value = new Bar {}
+  renderer = new PcfRenderer {
+    convertPropertyTransformers {
+      [Prefix] { prefix = "foo_" }
+    }
+  }
+}
```
bioball, line 63: Can you add a test that exercises `@ConvertProperty` directly? Rather than subclass?

PR #1335, opened 2025-12-01
Fix parsing of URLs with plus signs

bioball:
This fixes an issue where URLs with schemes that contain `+`, `-`, and `.` would not be parsed correctly.

For example, `foo+bar:///?baz.pkl` would turn into `foo+bar:///%3Fbaz.pkl`. The query param is lost, and turned into the path.

### DIFF
--- a/docs/modules/release-notes/pages/0.31.adoc
+++ b/docs/modules/release-notes/pages/0.31.adoc
@@ -42,7 +42,7 @@ Things to watch out for when upgrading.
 
 The following bugs have been fixed.
 
-* Fix parsing of URLs with plus signs (https://github.com/apple/pkl/pull/1335[#1335]).
+* XXX (https://github.com/apple/pkl/issues/XXX[XXX])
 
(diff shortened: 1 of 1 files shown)


HT154 (approved):
Ahh, good catch!

pkl-core/src/test/files/LanguageSnippetTests/output/api/Resource.pcf
```diff
@@ -36,4 +36,7 @@ examples {
     "AQIDBA=="
     Bytes(1, 2, 3, 4)
   }
+  ["absolute URI scheme with plus"] {
+    "Cannot find resource `foo+bar///?baz=buz`."
```
bioball, line 40: Prior to this fix, it was parsing the query param as the path segment. For example, this error message would look like:

```
Cannot find resource `foo+bar///%3Fbaz=buz`
```

And the query param would be lost.

PR #1336, opened 2025-12-02
Vendor paguro, fix an int overflow breaking large `List`s

HT154:
Paguro is unmaintained, so in lieu of replacing it entirely (which we'd like to do eventually) this PR vendors it to fix the bug.


pkl-core/src/main/java/org/pkl/core/util/paguro/collections/RrbTree.java
```diff
@@ -0,0 +1,2229 @@
...
+      //     subNodeIndex = cumulativeSizes.length * treeIndex / treeSize
+      //
+      // Now guess the sub-node index (quickly).
+
+      long guessLong = ((long) cumulativeSizes.length * treeIndex) / size();
+      if (guessLong >= cumulativeSizes.length) {
+        // Guessed beyond end length - returning last item.
+        return cumulativeSizes.length - 1;
+      }
+      int guess = (int) guessLong;
```
HT154, line 1549: This is the actual bugfix: do math with `long`s for the bounds check.

Original code: https://github.com/GlenKPeterson/Paguro/blob/963e8ed91833656236489fd5a92b20b54f47b4b9/src/main/java/org/organicdesign/fp/collections/RrbTree.java#L1477-L1481

PR #1340, opened 2025-12-03
Change `pkl format --write` to exit 0 when formatting violations are found

### LINKED ISSUE #1324: `pkl format` should not exit with 11 in case of violation found
I'm not an CLI nor an exit code expert.
But when I run `pkl format -w .` I realize that it exit with 11 (violation found).
This is also documented:

Image

However, I think a CLI should only return non zero in case of an **error in the process**.
It should not return non zero in case the CLI (format) tool runs successfully, but found problems.

Looking also into wikipedia about [Exit status](https://en.wikipedia.org/wiki/Exit_status):
> For the shell's purposes, a command which exits with a zero exit status has succeeded. A nonzero exit status indicates failure. 

For me it sounds like "a non exit code means the program doesn't behave correctly, failed for some reaasons".
But it is not! It just found violation errors. This is his job 🤷 😁 

What do you think?

HT154:
Does not change the behavior of invocations without `--write`.

Resolves #1324

### DIFF
--- a/pkl-cli/src/main/kotlin/org/pkl/cli/CliFormatterCommand.kt
+++ b/pkl-cli/src/main/kotlin/org/pkl/cli/CliFormatterCommand.kt
@@ -110,7 +110,6 @@ constructor(
 
         val formatted = format(contents)
         if (contents != formatted) {
-          status.update(FORMATTING_VIOLATION)
           if (diffNameOnly || overwrite) {
             // if `--diff-name-only` or `-w` is specified, only write file names
             writeLine(pathStr)
@@ -118,6 +117,9 @@ constructor(
 
           if (overwrite) {
             path.writeText(formatted, Charsets.UTF_8)
+          } else {
+            // only exit on violation for "check" operations, not when overwriting
+            status.update(FORMATTING_VIOLATION)
           }
         }
 
--- a/docs/modules/pkl-cli/pages/index.adoc
+++ b/docs/modules/pkl-cli/pages/index.adoc
@@ -741,9 +741,9 @@ pkl shell-completion zsh
 This command formats or checks formatting of Pkl files. +
 Exit codes:
 
-* 0: No violations found.
+* 0: No violations found or files were updated.
 * 1: An unexpected error happened (ex.: IO error)
-* 11: Violations were found.
+* 11: Violations were found (when running without `--write`).
 
 If the path is a directory, recursively looks for files with a `.pkl` extension, or files named `PklProject`.
 


pkl-cli/src/main/kotlin/org/pkl/cli/CliFormatterCommand.kt
```diff
@@ -82,9 +82,10 @@ constructor(
 
     when (status.status) {
       FORMATTING_VIOLATION -> {
+        // only exit non-zero on formatting violation for "check" operations, not when overwriting
+        if (!overwrite) throw CliTestException("", status.status)
```
StefMa: All logic is handled in the handlePath function. Does it make sense to move this case into it too? handlePath would simply not update the Status class in case overwrite is true? 🤔

StefMa (commented):
I would suggest to move this logic to the other function to have it consistent. Otherwise I think it's good 🙃.

BTW, isn't there an unit test for that?

pkl-cli/src/main/kotlin/org/pkl/cli/CliFormatterCommand.kt
HT154: I'd argue that the status "formatting violation" rather than "success" here is more semantically correct. I'm treating `doRun()` as a presentation layer that controls how status is communicated via the CLI, but happy to defer to @stackoverflow or @bioball on this one.

StefMa: 😉 I saw this comming.
I do agree that `doRun` should communicate the exit status.
The problem is that currently the `Status` class handles both:
The exit code as well as the "status" of the pkl files (formatted yes or no, or error).
IMHO these are two separated things.
One should just run and report like "pkl files were formatted or not, or error happen", while the other returns the exit code, based on that result.

I experimented a bit, and have something like this in mind:
```kotlin
  override fun doRun() {
    var status = Status(SUCCESS)

    handlePaths(
      afterFileFormatted = { original, formatted ->
        if (original != formatted) {
          if (!silent && (diffNameOnly || overwrite)) {
            // if `--diff-name-only` or `-w` is specified, only write file names
            write(pathStr)
          }

          if (overwrite) {
            path.writeText(formatted, Charsets.UTF_8)
          }

          if(!overwrite) {
            status = Status(FORMATTING_VIOLATION)
          }
        }

        if (!silent && !diffNameOnly && !overwrite) {
          write(formatted)
        }
      },
      onError = {
        status = Status(ERROR)
        when (it) {
          is GenericParserError -> {
            writeErr("Could not format `$pathStr`: $pe")
          }

          is IOException -> {
            writeErr("IO error while reading `$pathStr`: ${e.message}")
          }

          else -> {
            throw it
          }
        }
      }
    )

    when (status.status) {
      FORMATTING_VIOLATION -> {
        // using CliTestException instead of CliException because we want full control on how to
        // print errors
        throw CliTestException("", status.status)
      }

      ERROR -> {
        if (!silent) {
          writeErr("An error occurred during formatting.")
        }
        throw CliTestException("", status.status)
      }
    }
  }

  private fun handlePaths(
    afterFileFormatted: ((String, String) -> Unit),
    onError: (Exception) -> Unit,
  ) {
    for (path in allPaths()) {
      val pathStr = path.toString()
      try {
        val contents =
          when {
            pathStr == "-" -> IoUtils.readString(System.`in`)
            else -> Files.readString(path)
          }
        if (pathStr == "-" && overwrite) {
          throw CliException("Cannot write to stdin", ERROR)
        }

        val formatted = format(contents)
        afterFileFormatted(contents, formatted)
      } catch (pe: Exception) {
        onError(pe)
      }
    }
  }
```
Not perfect, still needs a few adjustments.
But the point is clear:
`handlePaths` will **only take care** of reading files and formatting (maybe this is already too much 🤔 ).
The logic if we write to `system.out` or `system.err` is handled by `doRun`. 
Thats their job!
Its job is also to determinate the exit code.
So I moved the setting of the `status` also in there... 

`handlePath` is just a small function that takes care of reading and formatting.

What do you think? 🤔 

---

Non of this should be handled in this PR 😁 
Its just a general suggestions because the `handlePaths` does way too much...

If needed, we can also introduce coroutines here in the future, to read pkl (and maybe write) files in parallel 👀

bioball: I agree with the original comment that the status updating logic should just be moved to `handlePaths`.

The `handlePaths` method can _probably_ be split up, but I actually think the current implementation is readable enough. 

I think we should just add an if clause around the status update within `handlePaths`:

```kt
          if (!overwrite) {
            status.update(FORMATTING_VIOLATION)
          }
```

bioball (changes requested):
Just one nit about where the logic should go.

Also, can we add a test for this?

bioball (approved):
LGTM

Issue #1348, opened 2025-12-05
Incorrect formatting of Pkl files that only contain whitespace


bioball:
This code:

```bash
echo " " | pkl format - | jq -Rs .
```

Produces:

```json
" \n\n"
```

The extra `\n` is due to #1346.
But there also shouldn't be an extra space here.

PR #1351, opened 2025-12-05
Fix formatting of nodes with no children

bioball:
For example, this fixes an issue where an empty module turns into ` \n`.

Closes https://github.com/apple/pkl/issues/1348


pkl-formatter/src/test/kotlin/org/pkl/formatter/FormatterTest.kt
```diff
@@ -113,4 +113,11 @@ class FormatterTest {
 
     walkDir(outputDir)
   }
+
+  @Test
+  fun `whiespace only`() {
```
HT154:
```suggestion
  fun `whitespace only`() {
```

PR #1352, opened 2025-12-05
Build linux jobs inside docker container

bioball:
Use a container that has glibc 2.28 installed

[native]

### DIFF
--- a/.github/jobs/BuildNativeJob.pkl
+++ b/.github/jobs/BuildNativeJob.pkl
@@ -19,11 +19,28 @@ extraGradleArgs {
   }
 }
 
+preSteps {
+  when (os == "linux" && !musl) {
+    new {
+      name = "Install deps"
+      run = "dnf install -y git binutils gcc glibc-devel zlib-devel libstdc++-static glibc-langpack-en"
+    }
+  }
+}
+
 steps {
   when (musl) {
     new {
       name = "Install musl and zlib"
       run = read("../scripts/install_musl.sh").text
+    }
+  }
+  // workaround for https://github.com/actions/checkout/issues/1048
+  when (os == "linux" && !musl) {
+    new {
+      name = "Fix git ownership"
+      // language=bash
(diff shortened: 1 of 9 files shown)


.github/jobs/BuildNativeJob.pkl
```diff
@@ -19,13 +19,38 @@ extraGradleArgs {
   }
 }
 
+preSteps {
+  when (os == "linux" && !musl) {
+    new {
+      name = "Install deps"
+      run = """
+        yum install -y git binutils gcc glibc-devel zlib-devel libstdc++-static
```
HT154: EL8+ systems use `dnf` natively. In fact, `yum` is just a symlink:
```console
# file /usr/bin/yum
/usr/bin/yum: symbolic link to dnf-3
```

```suggestion
        dnf install -y git binutils gcc glibc-devel zlib-devel libstdc++-static
```

bioball: Good to know!

buildSrc/src/main/kotlin/pklNativeExecutable.gradle.kts
```diff
@@ -167,13 +167,46 @@ val testStartNativeExecutable by
     }
   }
 
+val requiredGlibcVersion: Version = Version.parse("2.17")
```
HT154, line 170: It would be good to update our documentation to specify what the supported glibc version is for our dynamically linked linux executables. It's probably worth specify a support policy, eg. "Pkl will support the glibc version of the oldest RedHat Linux release that is under active maintenance support". RHEL seems to have the longest support timelines for both the OS and for glibc versions. RHEL 8 gives us support for glibc 2.28 until June 2029, followed by RHEL 9 supporting glibc 2.34 until June 2032.

bioball: Down to have some sort of support policy for glibc, but I think it's somewhat orthogonal; this PR is to fix an unintentional breakage in a patch version.

PR #1354, opened 2025-12-06
SPICE-0020: Deferred, type-safe references

### LINKED ISSUE #912: `Reference<T>`: Type safe "deferred" references
One common pattern in infrastructure-as-code systems like Terraform and Pulumi is "references". A reference provides a way to refer to a resource (or a property thereof) that exists at a time after Pkl evaluation completes.

Example: The pulumi [aws-yaml-static-website example](https://github.com/pulumi/examples/blob/master/aws-yaml-static-website/Pulumi.yaml) creates an S3 bucket and creates a policy that references 

There should be an API within the standard library for representing and producing such references.

This requires a SPICE to be written to iterate on the design.

HT154:
SPICE: https://github.com/apple/pkl-evolution/pull/27

Resolves #912

[native-pkl-cli]


Reyna8780 (approved):

stdlib/ref.pkl
```diff
@@ -0,0 +1,80 @@
...
+  external function getPath(): List<Access>
+
+  external function toString(): String
+}
+
+abstract class Domain {
+  function referenceToString(data: Any, path: List<Access>): String =
+    throw(
+      "Reference domain \(getClass()) must override referenceToString to enable Reference.toString()"
+    )
```
bioball: Should use `abstract` for this

```suggestion
  abstract function referenceToString(data: Any, path: List<Access>): String
```

stdlib/ref.pkl
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/// Represents a property or subscript access as part of a [Reference].
+external class Access {
```
bioball: [nit] "AccessPart" feels like a better name for this.

Also, I don't think this needs to be an external class; it's helpful if users can `new` up these values and play around with them.

```suggestion
class AccessPart {
```

docs/modules/bindings-specification/pages/binary-encoding.adoc
```diff
@@ -187,6 +187,24 @@ The array's length is the number of slots that are filled. For example, xref:{ur
...
+|Array of link:{uri-stdlib-ReferenceAccess}[property access] values
+|
+|
+
+|link:{uri-stdlib-ReferenceAccess}[ReferenceAccess]
+|`0x21`
+|link:{uri-messagepack-str}[str] (when a property access) or link:{uri-messagepack-str}[nil] (when a subscript access)
+|Property name
+|`<value>` or link:{uri-messagepack-str}[nil] (when a property access)
+|Key value
```
bioball: If we don't make `ReferenceAccess` an external class (see comment below), this can be encoded how any other `VmTyped` is encoded.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2124,6 +2125,105 @@ protected boolean isParametric() {
...
+
+    public ReferenceTypeNode(
+        SourceSection sourceSection, TypeNode domainTypeNode, TypeNode referentTypeNode) {
+      super(sourceSection);
+      this.domainTypeNode = domainTypeNode;
+      this.referentTypeNode = referentTypeNode;
+      this.getModuleNode = new GetModuleNode(sourceSection);
+    }
+
+    @SuppressWarnings("unused")
```
```suggestion
```

docs/modules/bindings-specification/pages/binary-encoding.adoc
```diff
@@ -187,6 +187,24 @@ The array's length is the number of slots that are filled. For example, xref:{ur
 |
 |
 |
+
+|link:{uri-stdlib-Reference}[Reference]
+|`0x20`
+|`<value>`
+|Root value
+|link:{uri-messagepack-array}[array]
+|Array of link:{uri-stdlib-ReferenceAccess}[property access] values
```
bioball: This is also missing domain.

Food for thought: should this also include the referent type? This might be really useful for the language binding use-case. If we do, I'm not totally sure how that would work.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2124,6 +2125,105 @@ protected boolean isParametric() {
     }
   }
 
+  @ImportStatic(RefModule.class)
+  public abstract static class ReferenceTypeNode extends ObjectSlotTypeNode {
+    private TypeNode domainTypeNode;
```
```suggestion
    @Child private TypeNode domainTypeNode;
```

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,473 @@
...
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj) return true;
+    if (!(obj instanceof VmReference other)) return false;
+
+    return domain.equals(other.getDomain())
+        && data.equals(other.getData())
+        && path.equals(other.getPath())
+        && Objects.equals(candidateTypes, other.getCandidateTypes());
+  }
```
bioball: Also need to implement `hashCode()` here

pkl-core/src/main/java/org/pkl/core/stdlib/protobuf/RendererNodes.java
```diff
@@ -554,6 +555,18 @@ public void visitNull(VmNull value) {
       builder.append(value);
     }
 
+    @Override
+    public void visitReference(VmReference value) {
+      writePropertyName();
+      builder.append(value);
```
bioball, line 561: This should also be doing a `referenceToString()` call, right? Either way, `builder.append(value)` doesn't seem right, because you get the default Java `toString()` on `VmReference`.

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,473 @@
...
+    @Override
+    public <T> T accept(VmValueConverter<T> converter, Iterable<Object> path) {
+      return converter.convertReferenceAccess(this, path);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      if (this == obj) return true;
+      if (!(obj instanceof Access other)) return false;
+      return Objects.equals(property, other.getProperty()) && Objects.equals(key, other.getKey());
```
bioball: Assuming we keep this class, this method will throw if `property` is null (because `getProperty()` has an assertion around non-null)

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,473 @@
...
+    visitor.visitReference(this);
+  }
+
+  @Override
+  public <T> T accept(VmValueConverter<T> converter, Iterable<Object> path) {
+    return converter.convertReference(this, path);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
```
bioball:
```suggestion
  public boolean equals(@Nullable Object obj) {
```

stdlib/ref.pkl
```diff
@@ -0,0 +1,80 @@
...
+external class Reference<out D, out T> {
+  // these are methods instead of properties to avoid collisions with referenced properties
+
+  /// The value from which the Reference is rooted. 
+  external function getDomain(): D
+
+  /// The value from which the Reference is rooted. 
+  external function getData(): Any
+
+  /// The path of access from [rootValue] to the referent.
```
bioball:
```suggestion
  /// The path of access from [getData()] to the referent.
```

stdlib/ref.pkl
```diff
@@ -0,0 +1,80 @@
...
+/// Reference provides a type-checked way to produce references to data that is not known at time of
+/// evaluation. Different configuration domains may ascribe different meanings to references and
+/// require that textual rendering is formatted in specific ways.
+///
+/// Limitations:
+/// * Type constraints are ignored (erased) when referenced.
+/// * Properties marked `external` may not be referenced.
+external class Reference<out D, out T> {
+  // these are methods instead of properties to avoid collisions with referenced properties
+
```
bioball: [nit] don't include notes meant just for us.

```suggestion
```

stdlib/ref.pkl
```diff
@@ -0,0 +1,80 @@
...
+  /// The value from which the Reference is rooted. 
+  external function getDomain(): D
+
+  /// The value from which the Reference is rooted. 
+  external function getData(): Any
+
+  /// The path of access from [rootValue] to the referent.
+  external function getPath(): List<Access>
+
+  external function toString(): String
```
bioball: I feel that we should just implement this as normal Pkl. This allows for errors to show up as helpful stack frames (for example, if `referenceToString` is not implemented).

```suggestion
  function toString(): String = getDomain().referenceToString(getData(), getPath())
```

pkl-core/src/main/java/org/pkl/core/stdlib/AbstractRenderer.java
```diff
@@ -397,6 +398,16 @@ private void doVisitElement(
...
 
+  @Override
+  public void visitReference(VmReference value) {
+    cannotRenderTypeAddConverter(value);
+  }
+
+  @Override
+  public void visitReferenceAccess(VmReference.Access value) {
+    cannotRenderTypeAddConverter(value);
+  }
```
bioball: I think it's fine to allow references to have a default rendering (by calling its `toString()`) in all renderers, since the domain (intrinsic to references) provides an implementation for this. Otherwise I think you'll end up with a lot of boilerplate like:

```pkl
output {
  converters {
    [Reference] = (it) -> getDomain().referenceToString(it.getData(), it.getPath())
  }
}
```

pkl-core/src/main/java/org/pkl/core/PType.java
```diff
@@ -59,6 +75,10 @@ public StringLiteral(String literal) {
     public String getLiteral() {
       return literal;
     }
+
+    public String toString() {
```
bioball, line 80:
```suggestion
    @Override
    public String toString() {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmClass.java
```diff
@@ -611,53 +616,66 @@ public VmMap getAllMethodMirrors() {
...
       }
 
-      return __pClass;
+      for (var method : EconomicMaps.getValues(declaredMethods)) {
+        if (method.isLocal()) continue;
+        methods.put(method.getName().toString(), method.export(__pClass));
+      }
     }
+
+    return __pClass;
```
bioball: Need to return within the synchronized block

```suggestion
      }
      return _pClass;
    }
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2124,6 +2125,105 @@ protected boolean isParametric() {
...
+      this.getModuleNode = new GetModuleNode(sourceSection);
+    }
+
+    @SuppressWarnings("unused")
+    @Specialization(guards = "value.getVmClass() == getReferenceClass()")
+    protected Object eval(VirtualFrame frame, VmReference value) {
+      // constraints may not be used in Reference type annotation referents
+      // walk the type and throw if any part of the referent is constrained
+      // perform this check at eval-time instead of init-time
+      // because type aliases can replace nodes between init and eval
```
bioball: I don't think this is good enough; we should validate this when the type node gets initialized, not when the type check gets executed.

The late-init case happens in `VmTypeAlias#initialize`. We should be able to call validation both inside this type node's constructor, and also within the initialize method.

E.g this diff (assumes the presence of a `validateTypeArguments()` method):

```diff
diff --git a/pkl-core/src/main/java/org/pkl/core/runtime/VmTypeAlias.java b/pkl-core/src/main/java/org/pkl/core/runtime/VmTypeAlias.java
index f6d6bd1e..cbc7dc85 100644
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmTypeAlias.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmTypeAlias.java
@@ -31,8 +31,10 @@ import org.pkl.core.TypeParameter;
 import org.pkl.core.ast.VmModifier;
 import org.pkl.core.ast.type.TypeNode;
 import org.pkl.core.ast.type.TypeNode.ConstrainedTypeNode;
+import org.pkl.core.ast.type.TypeNode.ReferenceTypeNode;
 import org.pkl.core.ast.type.TypeNode.TypeVariableNode;
 import org.pkl.core.ast.type.TypeNode.UnknownTypeNode;
+import org.pkl.core.ast.type.VmTypeMismatchException;
 import org.pkl.core.util.LateInit;
 import org.pkl.core.util.Nullable;
 
@@ -199,7 +201,12 @@ public final class VmTypeAlias extends VmValue {
           }
           return true;
         });
-
+    clone.accept(node -> {
+      if (node instanceof TypeNode.ReferenceTypeNode referenceTypeNode) {
+        referenceTypeNode.validateTypeArguments();
+      }
+      return true;
+    });
     return clone;
   }
```

docs/modules/bindings-specification/pages/binary-encoding.adoc
```diff
@@ -187,6 +187,24 @@ The array's length is the number of slots that are filled. For example, xref:{ur
 |
 |
 |
+
+|link:{uri-stdlib-Reference}[Reference]
+|`0x20`
+|`<value>`
+|Root value
```
bioball:
```suggestion
|Data
```

pkl-core/src/test/files/LanguageSnippetTests/input/api/reference2.pkl
```diff
@@ -0,0 +1,17 @@
...
+
+class D extends ref.Domain
+typealias Ref<T> = ref.Reference<D, T>
+local d: D = new {}
+
+// foo = test.catch(() -> ref.Reference(d, String, "") as ref.Reference<D, String(true)>)
+foo = test.catch(() -> ref.Reference(d, String, "") as Ref<String(true)>)
+bar = test.catch(() -> ref.Reference(d, String, "") as Ref<String(true) | Int>)
+baz = test.catch(() -> ref.Reference(d, String, "") as Ref<Listing<String(true)>>)
+qux = test.catch(() -> ref.Reference(d, String, "") as Ref<Alias1?>)
```
bioball: Can you make each of these its own test within `errors`? So we can see the full error message (and tailor it to be helpful to users).

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,473 @@
...
+  private final ImRrbt<Access> path;
+  // candidate types can only be: PType.Class, PType.Alias (only preservedAliasTypes),
+  // PType.StringLiteral, or PType.UNKNOWN
+  private final Set<PType> candidateTypes;
+
+  private boolean forced = false;
+
+  private static final PType nullType = new PType.Class(BaseModule.getNullClass().export());
+  private static final Set<TypeAlias> intAliasTypes = getIntAliasTypes();
+  private static final Set<TypeAlias> preservedAliasTypes = intAliasTypes;
```
bioball: These two are the same?

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,473 @@
...
+
+  private static Set<TypeAlias> getIntAliasTypes() {
+    var types = new HashSet<TypeAlias>();
+    for (var t : BaseModule.getIntTypeAliases()) {
+      types.add(t.export());
+    }
+    return types;
+  }
```
bioball: Typically we do something like:

```java
  private static final Set<TypeAlias> intAliasTypes = new HashSet<>();

  static {
    for (var t : BaseModule.getIntTypeAliases()) {
      intAliasTypes.add(t.export());
    }
  }
```

See the static block of `BaseModule`, for example

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,473 @@
...
+  private static Set<PType> simplifyType(PType type, @Nullable PClass moduleClass) {
+    var types = new HashSet<PType>();
+    simplifyType(type, moduleClass, types);
+    return types;
+  }
+
+  private static void simplifyType(PType type, @Nullable PClass moduleClass, Set<PType> result) {
+    if (type == PType.UNKNOWN || type instanceof PType.StringLiteral) {
+      result.add(type);
+    } else if (type instanceof PType.Class klass) {
```
bioball: [nit] here, and everywhere else: we call this "clazz"

```suggestion
    } else if (type instanceof PType.Class clazz) {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,473 @@
...
+    } else if (type instanceof PType.Nullable nullable) {
+      simplifyType(nullable.getBaseType(), moduleClass, result);
+      result.add(nullType);
+    } else if (type instanceof PType.Constrained constrained) {
+      simplifyType(constrained.getBaseType(), moduleClass, result);
+    } else if (type instanceof PType.Alias alias) {
+      if (preservedAliasTypes.contains(alias.getTypeAlias())) {
+        result.add(alias);
+      } else {
+        simplifyType(alias.getAliasedType(), alias.getTypeAlias().getModuleClass(), result);
```
bioball: I think this loses type arguments

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,473 @@
...
+  public boolean checkType(PType type, @Nullable PClass moduleClass) {
+    // fast path: if this could be unknown, any type is accepted
+    if (candidateTypes.contains(PType.UNKNOWN)) {
+      return true;
+    }
+
+    // all candidate types must be subtypes of at least one target type
+    candidate:
+    for (var c : candidateTypes) {
+      for (var t : simplifyType(type, moduleClass)) {
```
bioball: Does this need to call `simplifyType`? You already call `simplifyType` in `withPropertyAccess` and `withSubscriptAccess`. If you also call this in the `VmReference` constructor, you have a guarantee that -these types are already simplified.

Also, maybe call this "normalizedTypes".

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,473 @@
...
+    //   * UInt16 is Int32|UInt32|UInt
+    //   * UInt32 is UInt
+    // * Class -> Class: if same class or A is a subclass of B
+    //   * if type args are present, must have equal number of them
+    //   * for each pair of type args, check variance
+    //     * invariant: A_i must be identical to B_i
+    //     * covariant: A_i must be a subtype of B_i
+    //     * contravariant: B_i must be a subtype of A_i
+
+    if (a instanceof PType.StringLiteral aStr) {
```
bioball, line 315: Can short-circuit if they are the same object

```suggestion
    if (a == b) {
      return true;
    }
    if (a instanceof PType.StringLiteral aStr) {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,473 @@
...
+        getVmClass().getPClassInfo(),
+        Map.of(
+            "candidateTypes",
+            candidateTypes,
+            "domain",
+            domain.export(),
+            "data",
+            VmValue.export(data),
+            "path",
+            pathList));
```
bioball: We should create a `org.pkl.core.Reference`, which is the Java representation of `pkl.ref#Reference`, and extends `Value`.

This should then return an instance of `Reference`.

`PObject` is for values that are subclasses of `pkl.base#Object`; `Reference` is not.
For example, `Datasize`, `Duration`, `Pair`, all have their own `Value` classes too.

pkl-core/src/test/files/LanguageSnippetTests/input/api/reference3.pkl
```diff
@@ -0,0 +1,34 @@
...
+quux = test.catch(() -> ref.Reference(d2, A, a).b as Ref<A>)
+
+typealias Alias1 = Int | Alias2?
+typealias Alias2 = Boolean
+
+// property not found
+corge = test.catch(() -> ref.Reference(d, String, a).c)
+
+// incorrect subscript type
+grault = test.catch(() -> ref.Reference(d, List, List())["hi"] as Ref<String>)
```
bioball: Like comment above, make these their own `error` test

stdlib/ref.pkl
```diff
@@ -0,0 +1,80 @@
...
+  /// If the access is access to a property.
+  external isProperty: Boolean
+
+  /// The property that was accessed.
+  ///
+  /// If non-null, this is a property access. If `null` this is a subscript access.
+  external property: String(key == null)?
+
+  /// If the access is access via subscript.
+  external isSubscript: Boolean
```
bioball: This is a binary condition; it's either property or entry access. Do we need two booleans?

stdlib/ref.pkl
```diff
@@ -0,0 +1,80 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+/// Type-safe deferred references to values not known at evaluation time.
+///
+/// References are an advanced API design tool that allow library authors to express domain-specific
+/// references to values that may not be known during evaluation. They are particularly suited for
+/// configuring execution systems where tasks have well-typed inputs and output
```
bioball:
```suggestion
/// Type-safe deferred references to values not known at evaluation time.
///
/// References are an API design tool that allow library authors to express domain-specific
/// references to values that may not be known during evaluation.
/// They are particularly suited for configuring execution systems where tasks have well-typed
/// inputs and outputs.output
```

stdlib/ref.pkl
```diff
@@ -0,0 +1,80 @@
...
+@ModuleInfo { minPklVersion = "0.32.0" }
+module pkl.ref
+
+/// Creates a reference to the given value.
+external const function Reference<D, T>(
+  domain: D(this is Domain),
+  clazz: Class<T>,
```
bioball:
```suggestion
  `class`: Class<T>,
```

stdlib/ref.pkl
```diff
@@ -0,0 +1,80 @@
...
+
+/// Type-safe deferred references to values not known at evaluation time.
+///
+/// References are an advanced API design tool that allow library authors to express domain-specific
+/// references to values that may not be known during evaluation. They are particularly suited for
+/// configuring execution systems where tasks have well-typed inputs and output
+@ModuleInfo { minPklVersion = "0.32.0" }
+module pkl.ref
+
+/// Creates a reference to the given value.
```
bioball: Not sure what "the given value" is?

```suggestion
/// Creates a reference.
```

stdlib/ref.pkl
```diff
@@ -0,0 +1,80 @@
...
+
+/// A reference to a value that may not exist at evaluation time.
+///
+/// Reference provides a type-checked way to produce references to data that is not known at time of
+/// evaluation. Different configuration domains may ascribe different meanings to references and
+/// require that textual rendering is formatted in specific ways.
+///
+/// Limitations:
+/// * Type constraints are ignored (erased) when referenced.
+/// * Properties marked `external` may not be referenced.
```
bioball: I think this should explain how to use it, complete with some examples.

We tend to provide really long doc comments when the use-case is advanced. For example, the doc comments within `pkl:Project` are quite long.

bioball (changes requested):
Did a first pass!

stdlib/ref.pkl
HT154: Hmm I'm not convinced re: naming. Each `pkl.ref#Access` corresponds to a qualified access or subscript [access].

docs/modules/bindings-specification/pages/binary-encoding.adoc
HT154: If we make this `VmTyped`, the encoding is _significantly_ bulkier without including any extra information. The encoding specified here is optimized with that in mind, but if you feel strongly I can change this.

Also re: directly instantiating access. I don't see a ton of value there unless users can also append them to a `Reference`. Maybe something like this:
* Users can instantiate `Access`
  * If it remains `external`: add external `PropertyAccess(property: String)` and `SubscriptAccess(key: Any)` functions to the module for construction
  * If it becomes a regular `VmTyped`: users can instantiate directly, but there would need to be a constraint preventing setting of both `property` and `key`
* Add a method to `Reference` like `with(access: Access)`

docs/modules/bindings-specification/pages/binary-encoding.adoc
HT154: If we were to include this, it would mean adding another slot to the table (not a big deal). It would also probably mean encoding the referent as an array of `Class` and `TypeAlias` (stdlib `U?Int\d+` types only) values. Maybe a future enhancement?

pkl-core/src/main/java/org/pkl/core/stdlib/protobuf/RendererNodes.java
HT154: Hmm what's the right way to do this from outside an existing node/frame?

docs/modules/bindings-specification/pages/binary-encoding.adoc
bioball: I think that if we're concerned about encoding size, we should fix that in our pkl-binary spec rather than here; there's already many other cases where the encoding is quite large. But FWIW xz, gz and zip are all really good at compressing pkl-binary.

I think we also should describe things as normal classes, unless it doesn't fit the mold of a `Typed` value for some reason.

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
HT154: These are used in semantically different ways:
* `intAliasTypes` gates the Int typealias comparison branch in `isSubtype`
* `preservedAliasTypes` directs `simplifyType`/`normalizedTypes` to not dereference an alias

These are the same sets currently, but may not always be in the future, e.g. if this is expanded to handle `Char` values.

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
HT154: Does it? `PType.Alias.getAliasedType()` returns the exported `TypeAliasTypeNode.aliasedTypeNode` (which is the instantiated `TypeAlias`)

stdlib/ref.pkl
HT154: This is for ease/fluency of use: if a user is writing e.g. a function that transforms a Reference to a String, they may need to know whether each access is a subscript key or property; it may be necessary to check one way or the other and `access.isSubscript` is more meaningful than `!access.isProperty` (and vice versa).

Another alternative here is something like this:
```pkl
type: "property" | "subscript"
```
But that makes usage sites pretty verbose!

Curious if you have thoughts or alternative suggestions here.

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
HT154: This is normalizing the type to check against, not the candidates, so `type` here will never have been normalized (it's `ReferenceTypeNode.referentTypeNode.export()` directly). Normalization here reduces the cases that `isSubtype` needs to be aware of significantly.

I did update this to pull the call out of the loop since it only needs to run once.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2178,6 +2163,30 @@ protected Object eval(VirtualFrame frame, VmReference value) {
       throw new VmTypeMismatchException.Reference(sourceSection, value, domainType, referentType);
     }
 
+    public void validateTypeArguments(@Nullable SourceSection aliasSourceSection) {
+      // constraints may not be used in Reference type annotation referents
+      // walk the type and throw if any part of the referent is constrained
+
+      // TODO improve error message when this type node and/or referent constraint are behind type
+      // aliases
```
HT154, line 2181: As discussed offline, the errors here are _just okay_, but could be improved as follow-up work

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -226,7 +231,7 @@ private static void getCandidateSubscriptType(PType type, Object key, Set<PType>
       for (var kt : keyTypes) {
         if (kt == PType.UNKNOWN
             || (kt instanceof PType.Class klazz
-                && klazz.getPClass().getInfo() == PClassInfo.forValue(key))
+                && klazz.getPClass().getInfo() == PClassInfo.forValue(VmValue.export(key)))
```
HT154, line 247: This fixes a crash when a subscript key is a `VmValue` (`value != value.export()`)

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2124,6 +2125,113 @@ protected boolean isParametric() {
...
+      referentTypeNode.acceptTypeNode(
+          true,
+          (typeNode) -> {
+            if (typeNode instanceof ConstrainedTypeNode) {
+              CompilerDirectives.transferToInterpreter();
+              var err =
+                  exceptionBuilder()
+                      .evalError("invalidReferenceTypeAnnotationWithConstraint")
+                      .withLocation(this);
```
bioball: Redundant; `exceptionBuilder()` already sets the location to the node instance

```suggestion
```

stdlib/ref.pkl
bioball: Is two booleans helpful? I suspect most code will look something like this:

```pkl
if (access.isProperty) ".\(access.property)"
else "[\(access.key)]"
```

docs/modules/bindings-specification/pages/binary-encoding.adoc
bioball: Yeah, I'm fine with not encoding the referent here.

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
bioball: Oh, I see, never mind!

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,441 @@
...
+    for (var t : BaseModule.getIntTypeAliases()) {
+      intAliasTypes.add(t.export());
+      preservedAliasTypes.add(t.export());
+    }
+  }
+
+  private record Access(@Nullable String property, @Nullable Object key) {}
+
+  private static final VmObjectFactory<Access> accessFactory =
+      new VmObjectFactory<>(RefModule::getAccessClass);
```
bioball: Use the `VmObjectBuilder` API instead for this. `VmObjectFactory` adds some unnecessary indirection (extra storage, property nodes, etc).

This can be:

```java
  private static VmTyped newAccess(@Nullable String property, @Nullable Object key) {
    return new VmObjectBuilder()
        .addProperty(Identifier.PROPERTY, property == null ? VmNull.withoutDefault() : property)
        .addProperty(Identifier.KEY, key == null ? VmNull.withoutDefault() : key)
        .toTyped(RefModule.getAccessClass());
  }
``` 

This implies this new method on `VmObjectBuilder`:

```java
  public VmTyped toTyped(VmClass clazz) {
    return new VmTyped(
        VmUtils.createEmptyMaterializedFrame(), clazz.getPrototype(), clazz, members);
  }
```

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,441 @@
...
+    return result;
+  }
+
+  public String toPklString(@Nullable VirtualFrame frame, @Nullable SourceSection sourceSection) {
+    if (frame == null) {
+      frame = VmUtils.createEmptyMaterializedFrame();
+    }
+    if (sourceSection == null) {
+      sourceSection = VmUtils.unavailableSourceSection();
+    }
```
bioball: I think this can just be:

```suggestion
  public String toPklString() {
    var frame = VmUtils.createEmptyMaterializedFrame();
    var sourceSection = VmUtils.unavailableSourceSection();
```

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,441 @@
...
+    return (String)
+        InvokeMethodVirtualNodeGen.create(
+                sourceSection,
+                Identifier.TO_STRING,
+                new ExpressionNode[] {},
+                MemberLookupMode.EXPLICIT_RECEIVER,
+                null,
+                null)
+            .executeWith(frame, this, getVmClass());
+  }
```
bioball: We should use a `DirectCallNode` here; `Reference` is a closed class and its call target is always known.

We should also avoid creating this node for every `toPklString` call; one of the principles here is that the number of truffle nodes should be linear to the amount of code that we are evaluating.

Add these statics:

```java
  private static final DirectCallNode toStringCallNode = toStringCallNode();

  private static DirectCallNode toStringCallNode() {
    var toStringMethod = RefModule.getReferenceClass().getMethod(Identifier.TO_STRING);
    assert toStringMethod != null;
    return DirectCallNode.create(toStringMethod.getCallTarget());
  }
```

Then, this impl can simplify to:

```java
  public String toPklString() {
    return (String) toStringCallNode.call(this, getVmClass().getPrototype());
  }
```

pkl-core/src/main/java/org/pkl/core/ast/internal/ToStringNode.java
```diff
@@ -60,9 +61,17 @@ protected String evalBoolean(boolean value) {
...
           InvokeMethodVirtualNode invokeNode) {
+    return (String) invokeNode.executeWith(frame, value, value.getVmClass());
+  }
 
+  @Specialization
+  protected String evalReference(
+      VirtualFrame frame,
+      VmReference value,
+      @Shared("invokeToString") @Cached(value = "createInvokeNode()", neverDefault = true)
+          InvokeMethodVirtualNode invokeNode) {
```
bioball:
```suggestion
  @Specialization
  protected String evalReference(VmReference value) {
    return value.toPklString();
  }
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2124,6 +2125,113 @@ protected boolean isParametric() {
...
+      super(sourceSection);
+      this.domainTypeNode = domainTypeNode;
+      this.referentTypeNode = referentTypeNode;
+      this.getModuleNode = new GetModuleNode(sourceSection);
+      validateTypeArguments(sourceSection);
+    }
+
+    @Specialization(guards = "value.getVmClass() == getReferenceClass()")
+    protected Object eval(VirtualFrame frame, VmReference value) {
+      if (referentTypeNode.isNoopTypeCheck()) {
```
```suggestion
      if (domainTypeNode.isNoopTypeCheck() && referentTypeNode.isNoopTypeCheck()) {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,441 @@
...
+
+    // all candidate types must be subtypes of at least one target type
+    candidate:
+    for (var c : candidateTypes) {
+      for (var t : checkTypes) {
+        if (isSubtype(c, t)) break candidate;
+      }
+      return false;
+    }
+    return true;
```
bioball: Shouldn't this just return if this is a subtype of one of the candidate types?

```suggestion
    // at least one candidate type must be subtypes of at least one target type
    for (var c : candidateTypes) {
      for (var t : checkTypes) {
        if (isSubtype(c, t)) return true;;
      }
    }
    return false;
```

If we really wanted to check all types, this should be:

```suggestion
    // all candidate types must be subtypes of at least one target type
    candidate:
    for (var c : candidateTypes) {
      for (var t : checkTypes) {
        if (isSubtype(c, t)) continue candidate;
      }
      return false;
    }
    return true;
```

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,441 @@
...
+                && stringLiteral.getLiteral().equals(key))) {
+          normalizeTypes(typeArgs.get(1), clazz.getPClass().getModuleClass(), result);
+          return;
+        }
+      }
+    }
+  }
+
+  /** Checks type against ref's candidateTypes. Does not check domain. */
+  public boolean checkType(PType type, @Nullable PClass moduleClass) {
```
bioball: Two nits:
1. `moduleClass` is never null
2. Rename method to tell what this method is doing

```suggestion
  public boolean referentTypeIsSubtypeOf(PType type, PClass moduleClass) {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,441 @@
...
+            || (kt instanceof PType.StringLiteral stringLiteral
+                && stringLiteral.getLiteral().equals(key))) {
+          normalizeTypes(typeArgs.get(1), clazz.getPClass().getModuleClass(), result);
+          return;
+        }
+      }
+    }
+  }
+
+  /** Checks type against ref's candidateTypes. Does not check domain. */
```
bioball:
```suggestion
  /** Tells if this referent type is a subtype of {@code type}. Does not check domain. */
```

pkl-core/src/main/java/org/pkl/core/Reference.java
```diff
@@ -0,0 +1,101 @@
...
+   * Returns the access path of this reference.
+   *
+   * <p>All elements are exported {@code pkl.ref#Access} instances.
+   */
+  public List<Composite> getPath() {
+    return path;
+  }
+
+  /** Returns the referent type of this reference. */
+  public PType getReferentTypee() {
```
bioball:
```suggestion
  public PType getReferentType() {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,441 @@
...
+  static {
+    accessFactory
+        .addProperty(
+            "property",
+            access -> access.property == null ? VmNull.withoutDefault() : access.property)
+        .addProperty("key", access -> access.key == null ? VmNull.withoutDefault() : access.key);
+  }
+
+  public VmReference(VmTyped domain, VmClass clazz, Object data) {
+    this(domain, data, RrbTree.empty(), Set.of(new PType.Class(clazz.export())));
```
bioball: This should still go through the `normalizeTypes` path.

```suggestion
    this(domain, data, RrbTree.empty(), normalizeTypes(new PType.Class(clazz.export()), clazz.getModule().getVmClass().export()));
```

Also, `normalizeTypes` should handle just being passed a `List`, `Listing`, `Map`, `Mapping`, etc, e.g. this Pkl code:
And normalizing this to `List<unknown>`, `Map<unknown, unknown>`, etc.

```pkl
foo = ref.Reference(myDomain, List, null)[0]
```

And this:

```pkl
foo = ref.Reference(myDomain, Outputs, null).prop[0]

class Outputs {
  prop: List
}
```
Need tests for these

bioball (changes requested):
Did another pass!

stdlib/ref.pkl
HT154: Yes, in the sense that the conditional may need to check if it's a subscript instead and
```pkl
if (access.isSubscript)
```
is more legible than
```pkl
if (!access.isProperty)
```
and both are preferable to
```pkl
if (access.type == "subscript")
```
or 
```pkl
if (access.property == null)
```

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,453 @@
...
+  public VmReference(VmTyped domain, Object data, ImRrbt<VmTyped> path, Set<PType> candidateTypes) {
+    this.domain = domain;
+    this.data = data;
+    this.candidateTypes = candidateTypes;
+    this.path = path;
+  }
+
+  public Set<PType> getCandidateTypes() {
+    return candidateTypes;
+  }
```
bioball: [nit] this is unused

```suggestion
```

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
HT154: Oof good catch. This needs to be a `continue`. This deceptively fiddly bit of logic is what makes `Reference<D, Foo | Bar | Baz>` not a subtype of `Reference<D, Foo | Bar>`. Adding a test for this.

pkl-core/src/main/java/org/pkl/core/ast/internal/ToStringNode.java
HT154: Doesn't this result in less helpful stack traces since the frame is not passed through?

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
HT154: Would you object to normalizing unknown type args during `TypeNode.export` instead? It's a behavior change for that API, but is much more consistent imo.

HT154: Nope, that breaks pkldoc! Adding normalization in VmReference itself.

pkl-core/src/main/java/org/pkl/core/ast/internal/ToStringNode.java
HT154: Confirmed that making this change loses this stack trace line:
```
xx | testInterpolation = "test:\(test)"
                               ^^^^^^^
at reference20#testInterpolation (file:///$snippetsDir/input/errors/reference20.pkl)
```
From this code
```pkl
local test = ref.Reference(d, A, "").foo 
testInterpolation = "test:\(test)"
```
Assuming `d.referenceToString()` throws

bioball: Oh, I see! Then, do:

```suggestion
  @Specialization
  protected String evalReference(
      VirtualFrame frame,
      VmReference value,
      @Cached(value = "createCallNode(value)") DirectCallNode callNode) {
    return (String) callNode.call(value, value.getVmClass().getPrototype());
  }

  protected DirectCallNode createCallNode(VmReference reference) {
    var toStringMethod = reference.getVmClass().getDeclaredMethod(Identifier.TO_STRING);
    assert toStringMethod != null;
    return DirectCallNode.create(toStringMethod.getCallTarget());
  }
```

HT154: For my benefit, what's the difference here that resolves this? It looks like the primary change is caching the `DirectCallNode` in the truffle layer instead of it being static inside `VmReference`. Why does that make the difference?

stdlib/ref.pkl
bioball: Okay, makes sense; kind of in line with how we provide both `isEmpty` and `isNotEmpty`, etc.

pkl-core/src/main/java/org/pkl/core/stdlib/AbstractStringRenderer.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
```
stackoverflow: No changes here.

pkl-core/src/main/java/org/pkl/core/PClass.java
```diff
@@ -119,6 +122,11 @@ public Map<String, Method> getAllMethods() {
     return allMethods;
   }
 
+  /** Returns the class's containing module's class, or this class if it is a module class. */
+  public PClass getModuleClass() {
+    return moduleClass != null ? moduleClass : this;
```
stackoverflow, line 127: If we are returning `this` we should assert it's a module class.

stdlib/ref.pkl
```diff
@@ -0,0 +1,207 @@
...
+@ModuleInfo { minPklVersion = "0.32.0" }
+module pkl.ref
+
+/// Creates a deferred [Reference] to a value of type [class] in the given [domain].
+///
+/// WARNING: The API and semantics of references are subject to change in a future release.
+/// The Pkl team is soliciting feedback from authors of libraries considering adopting references.
+/// For questions and feedback, please reach out via
+/// [GitHub Discussions](https://github.com/apple/pkl/discussions) or
+/// [create an issue](https://github.com/apple/pkl/issues/new).
```
bioball: [nit] I think it's sufficient to have this on the module's doc comment; same goes for the class

```suggestion
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2124,6 +2125,126 @@ protected boolean isParametric() {
...
+    public ReferenceTypeNode(
+        SourceSection sourceSection, TypeNode domainTypeNode, TypeNode referentTypeNode) {
+      super(sourceSection);
+      this.domainTypeNode = domainTypeNode;
+      this.referentTypeNode = referentTypeNode;
+      this.getModuleNode = new GetModuleNode(sourceSection);
+      validateTypeArguments(sourceSection);
+    }
+
+    @Specialization(guards = "value.getVmClass() == getReferenceClass()")
```
bioball: This guard doesn't do anything; `value.getVmClass()` is guaranteed to be the reference class

```suggestion
```

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,473 @@
...
+import org.pkl.core.util.paguro.RrbTree;
+import org.pkl.core.util.paguro.RrbTree.ImRrbt;
+
+public final class VmReference extends VmValue {
+
+  private final VmTyped domain;
+  private final Object data;
+  private final ImRrbt<VmTyped> path;
+  // candidate types can only be: PType.Class, PType.Alias (only preservedAliasTypes),
+  // PType.StringLiteral, or PType.UNKNOWN
```
bioball:
```suggestion
  // PType.StringLiteral, PType.UNKNOWN, or PType.Union
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
bioball: BTW: we use a pattern similar to this to avoid instanceof checks in guards sometimes, see org.pkl.core.ast.expression.member.ReadPropertyNode#evalObject

stdlib/ref.pkl
```diff
@@ -0,0 +1,207 @@
...
+///
+/// Example:
+/// ```
+/// import "pkl:ref"
+///
+/// /// Define a domain class for this configuration system:
+/// class Domain extends ref.Domain {
+///   /// Define how this domain renders references to strings:
+///   function renderReference(reference: ref.Reference<D, Any>): String =
+///     (List(data.toString()) + path.map((it) -> it.property ?? it.key.toString())).join("/")
```
bioball: This is invalid code; needs to be updated given the new API

bioball (approved):
Approving to unblock; just some minor comments!

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,473 @@
...
+  public boolean equals(@Nullable Object o) {
+    if (o == this) return true;
+    if (!(o instanceof VmReference that)) {
+      return false;
+    }
+
+    return domain.equals(that.domain)
+        && data.equals(that.data)
+        && path.equals(that.path)
+        && referentType.equals(that.referentType);
```
stackoverflow, line 462: PType doesn't implement `equals` (or hashcode). So this would compare by reference. Not sure this would always be safe.

pkl-core/src/test/files/LanguageSnippetTests/output/api/reference.pcf
```diff
@@ -0,0 +1,60 @@
...
+aValuesJoined = """
+  org.pkl.core.runtime.VmReference@<addr>
```
stackoverflow, line 53: Is it intended that `VmReference` has no toString?

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -0,0 +1,473 @@
...
+    return referentType;
+  }
+
+  // simplifies a type by:
+  // * erasing constraints
+  // * transforming T? into T|Null
+  // * dereferencing aliases (except for well-known stdlib alias types)
+  // * flattening unions
+  // * when moduleClass is supplied, replace PType.MODULE with appropriate PType.Class
+  // * drop PType.NOTHING, PType.Function, and PType.TypeVariable
```
stackoverflow: You mentioned NOTHING is dropped, but `normalizeTypes` add it to the result.

PR #1356, opened 2025-12-07
Adjust native lifecycle builds to not throw during configuration phase

bioball:
Allow Gradle to run other tasks on unsupported machines, but throw when running native lifecycle tasks (e.g. `buildNative`).


HT154 (approved):
Ah very nice!

PR #1367, opened 2025-12-12
SPICE-0025: `pkl run` CLI framework

HT154:
SPICE: https://github.com/apple/pkl-evolution/pull/28

### DIFF
--- a/docs/modules/language-reference/pages/index.adoc
+++ b/docs/modules/language-reference/pages/index.adoc
@@ -3961,6 +3961,7 @@ emailList: List<EmailAddress> // <2>
 <1> equivalent to `email: String(contains("@"))` for type checking purposes
 <2> equivalent to `emailList: List<String(contains("@"))>` for type checking purposes
 
+[[nullable-types]]
 ==== Nullable Types
 
 Class types such as `Bird` (see above) do not admit `null` values.
--- a/docs/modules/pkl-cli/pages/index.adoc
+++ b/docs/modules/pkl-cli/pages/index.adoc
@@ -489,7 +489,9 @@ If these are the only failures, the command exits with exit code 10.
 Otherwise, failures result in exit code 1.
 
 <modules>::
-The absolute or relative URIs of the modules to test. Relative URIs are resolved against the working directory.
+The absolute or relative URIs of the modules to test.
+The module must extend `pkl:test`.
+Relative URIs are resolved against the working directory.
 
 ==== Options
 
@@ -545,6 +547,23 @@ Use `--no-power-assertions` to disable this feature if you prefer simpler output
 ====
 
 This command also takes <<common-options, common options>>.
+
+[[command-run]]
+=== `pkl run`
+
+*Synopsis:* `pkl run [<options>] [<module>] [<command options>]`
+
+Evaluate a <<cli-tools,CLI command>> defined by `<module>`.
+
+<module>::
+The absolute or relative URIs of the command module to run.
+The module must extend `pkl:Command`.
+Relative URIs are resolved against the working directory.
+
+<command options>::
+Additional CLI options and arguments defined by `<module>`.
+
+This command also takes <<common-options, common options>>, but they must be specified before `<module>`. 
 
 [[command-repl]]
 === `pkl repl`
@@ -800,7 +819,7 @@ Write the path of files with formatting violations to stdout.
 [[common-options]]
 === Common options
 
-The <<command-eval>>, <<command-test>>, <<command-repl>>, <<command-project-resolve>>, <<command-project-package>>, <<command-download-package>>, and <<command-analyze-imports>> commands support the following common options:
+The <<command-eval>>, <<command-test>>, <<command-run>>, <<command-repl>>, <<command-project-resolve>>, <<command-project-package>>, <<command-download-package>>, and <<command-analyze-imports>> commands support the following common options:
 
 include::../../pkl-cli/partials/cli-common-options.adoc[]
 
@@ -900,6 +919,301 @@ pkl eval --format=json --output-path=%{moduleDir}/%{moduleName}.json config1.pkl
 If multiple module outputs are written to the same file, or to standard output, their outputs are concatenated.
 By default, module outputs are separated with `---`, as in a YAML stream.
 The separator can be customized using the `--module-output-separator` option.
+
+[[cli-tools]]
+== Implementing CLI Tools
+
+CLI tools can be implemented in Pkl by modules extending the `pkl:Command` module.
+With `pkl:Command`, you can define a script in Pkl that is executed by your shell, providing a better CLI experience.
+
+Regular evaluation requires use of xref:language-reference:index.adoc#resources[resources] like properties and evironment variables to provide parameters:
+[source,bash]
+----
+$ pkl eval script.pkl -p username=me -p password=password
+----
+
+Commands provide a native, familiar CLI experience:
+[source,bash]
+----
+$ pkl run script.pkl --username=admin --password=hunter2
+$ ./script.pkl --username=admin --password=hunter2
+----
+
+Pkl commands have a few properties that distinguish them from standard module evaluation:
+
+* Users provide input to commands using familiar command line idioms, providing a better experience than deriving inputs from xref:language-reference:index.adoc#resources[resources] like external properties or environment variables.
+* Commands can dynamically import modules when they are specified as command line options.
+* Commands may write to standard output (via `output.text` or `output.bytes`) and the filesystem (via `output.files`) in the same evaluation.
+* Command file output may write to any absolute path (not only relative to the `--multiple-file-output-path` option).
+** Relative output paths are written relative to the current working directory (or `--working-dir`, if specified).
+** Paths of output file are printed to the command's standard error.
+
+IMPORTANT: Users of `pkl run` must be aware of the security implications of this behavior.
+Using `pkl eval` prevents accidental overwrites by not allowing absolute paths, but `pkl run` does not offer this protection.
+Commands may write to any path the invoking user has permissions to modify.
+
+Commands are implemented as regular modules and declare their supported command line flags and positional arguments using a class with annotated properties.
+
+=== Defining Commands
+
+Commands are defined by creating a module that extends `pkl:Command`:
+
+[source,pkl%tested]
+.my-tool.pkl
+----
+/// This doc comment becomes part of the command's CLI help!
+/// Markdown formatting is **allowed!**
+extends "pkl:Command"
+
+options: Options // <1>
+
+class Options {
+  // Define CLI flags/arguments...
+}
+
+// Regular module code...
+----
+<1> Re-declaration of the `options` property's type.
+
+Like `pkl eval`, when a command completes without an evaluation error the process exits successfully (exit code 0).
+Commands can return a failure using `throw` (exit code 1), but otherwise may not control the exit code.
+
+Other than the differences listed above, commands behave like any other Pkl module.
+For example, there is no way to execute other programs or make arbitrary HTTP requests.
+If additional functionality is desired, xref:language-reference:index.adoc#external-readers[external readers] may be used to extends Pkl's capabilities.
+
+=== Command Options
+
+Each property of a command's options class becomes a command line option.
+Properties with the `local`, `hidden`, `fixed`, and/or `const` modifiers are not parsed as options
+A property's doc comment, if present, becomes the corresponding option's CLI help description.
+Doc comments are interpreted as Markdown text and formatted nicely when displayed to users.
+Properties must have xref:language-reference:index.adoc#type-annotations[type annotations] to determine how they are parsed.
+
+Properties may be xref:language-reference:index.adoc#annotations[annotated] to influence how they behave:
+
+* Properties annotated with link:{uri-stdlib-Command-Flag}[`@Flag`] become CLI flags named `--<property name>` that accept a value.
+* `Boolean` properties annotated with link:{uri-stdlib-Command-BooleanFlag}[`@BooleanFlag`] become CLI flags named `--<property name>` and `--no-<property name>` that result in `true` and `false` values, respectively.
+* `Int` (and type aliases of `Int`) properties annotated with link:{uri-stdlib-Command-CountedFlag}[`@CountedFlag`] become CLI flags named `--<property name>` that produce a value equal to the number of times they are present on the command line.
+* Properties annotated with link:{uri-stdlib-Command-Argument}[`@Argument`] become positional CLI arguments and are parsed in the order they appear in the class.
+* Properties with no annotation are treated the same as `@Flag` with no further customization.
+
+Flag options may set a `shortName` property to define a single-character abbreviation (`-<short name>`).
+Flag abbreviations may be combined (e.g. `-a -b -v -v -f some-value` is equivalent to `-abvvf some-value`).
+
+A `@Flag` or `@Argument` property's type annotation determines how it is converted from the raw string value:
+
+|===
+|Type |Behavior
+
+|`String`
+|Value is used verbatim.
+
+|`Char`
+|Value is used verbatim but must be exactly one character.
+
+|`Boolean`
+|True values: `true`, `t`, `1`, `yes`, `y`, `on`
+
+False values: `false`, `f`, `0`, `no`, `n`, `off`
+
+|`Number`
+|Value is parsed as an `Int` if possible, otherwise parsed as `Float`.
+
+|`Float`
+|Value is parsed as a `Float`.
+
+|`Int`
+|Value is parsed as a `Int`.
+
+|`Int8`, `Int16`, `Int32`, `UInt`, `UInt8`, `UInt16`, `UInt32`
+|Value is parsed as a `Int` and must be within the type's range.
+
+|xref:language-reference:index.adoc#union-types[Union] of xref:language-reference:index.adoc#string-literal-types[string literals]
+|Value is used verbatim but must match a member of the union.
+
+|`List<Element>`, `Listing<Element>`, `Set<Element>`
+|Each occurrence of the option becomes an element of the final value.
+
+`Element` values are parsed based on the above primitive types.
+
+|`Map<Key, Value>`, `Mapping<Key, Value>`
+|Each occurrence of the option becomes an entry of the final value.
+
+Values are split on the first `"="` character; the first part is parsed as `Key` and the second as `Value`, both based on the above primitive types.
+
+|`Pair<First, Second>`
+|Value is split on the first `"="` character; the first part is parsed as `First` and the second as `Second`, both based on the above primitive types.
+
+|===
+
+If a flag that accepts only a single value is provided multiple times, the last occurrence becomes the final value.
+
+Only a single positional argument accepting multiple values is permitted per command.
+
+A property with a xref:language-reference:index.adoc#nullable-types[nullable type] is optional and, if not specified on the command line, will have value `null`.
+Properties with default values are also optional.
+Type constraints are evaluated when the command is executed, so additional restrictions on option values are enforced at runtime.
+
+==== Custom Option Conversion and Aggregation
+
+A property may be annotated with any type if its `@Flag` or `@Argument` annotation sets the `convert` or `transformAll` properties.
+The `convert` property is a xref:language-reference:index.adoc#anonymous-functions[function] that overrides how _each_ raw option value is interpreted.
+The `transformAll` property is a function that overrides how _all_ parsed option values become the final property value.
+
+The `convert` function may return an link:{uri-stdlib-Command-Import}[`Import`] value that is replaced during option parsing with the actual value of the module specified by its `uri` property.
+If `glob` is `true`, the replacement value is a `Mapping`; its keys are the _absolute_ URIs of the matched modules and its values are the actual module values.
+When specifying glob import options on the command line, it is often necessary to quote the value to avoid it being interpreted by the shell.
+If the return value of `convert` is a `List`, `Set`, `Map`, or `Pair`, each contained value (elements and entry keys/values) that are `Import` values are also replaced.
+
+[IMPORTANT]
+====
+If an option has type `Mapping<String, «some module type»>` and should accept a single glob pattern value, the option's annotation must also set `multiple = false` to override the default behavior of `Mapping` options accepting multiple values.
+Example:
+[source%parsed,{pkl}]
+----
+@Flag {
+  convert = (it) -> new Import { uri = it; glob = true }
+  multiple = false
+}
+birds: Mapping<String, Bird>
+----
+
+If multiple glob patterns values should be accepted and merged, `transformAll` may be used to merge every glob-imported `Mapping`:
+[source%parsed,{pkl}]
+----
+@Flag {
+  convert = (it) -> new Import { uri = it; glob = true }
+  transformAll =
+    (values) -> values.fold(new Mapping {}, (result, element) ->
+      (result) { ...element }
+    )
+}
+birds: Mapping<String, Bird>
+----
+====
+
+=== Subcommands
+
+Like many other command line libraries, `pkl:Command` allows building commands into a hierarchy with a root command and subcommands:
+
+[source,pkl%tested]
+.my-tool.pkl
+----
+extends "pkl:Command"
+
+command {
+  subcommands {
+    import("subcommand1.pkl")
+    import("subcommand2.pkl")
+    for (_, subcommand in import*("./subcommands/*.pkl")) {
+      subcommand
+    }
+  }
+}
+----
+
+[source,pkl%tested]
+.subcommand1.pkl
+----
+extends "pkl:Command"
+
+import "my-tool.pkl"
+
+parent: `my-tool` // <1>
+
+// Regular module code...
+----
+<1> Optional; asserts that this is a subcommand of `my-tool` and simplifies accessing properties and options of the parent command
+
+Each element of `subcommands` must have a unique value for `command.name`.
+
+=== Testing Commands
+
+Command modules are normal Pkl modules, so they may be imported and used like any other module.
+This is particularly helpful when testing commands, as the command's `options` and `parent` properties can be populated by test code.
+
+Testing the above command and subcommand might look like this:
+
+[source,pkl%tested]
+----
+amends "pkl:test"
+
+import "my-tool.pkl"
+import "subcommand1.pkl"
+
+examples {
+  ["Test my-tool"] {
+    (`my-tool`) {
+      options {
+        // Set my-tool options here...
+      }
+    }.output.text
+  }
+  ["Test subcommand1"] {
+    (subcommand1) {
+      parent { // this amends `my-tool`
+        options {
+          // Set my-tool options here...
+        }
+      }
+      options {
+        // Set subcommand options here...
+      }
+    }.output.text
+  }
+}
+----
+
+[[commands-as-standalone-scripts]]
+=== Commands as standalone scripts
+
+On *nix platforms, Pkl commands can be configured to run as standalone tools that can be invoked without the `pkl run` command.
+To achieve this, the command file must be marked executable (i.e. `chmod +x my-tool.pkl`) and a link:https://en.wikipedia.org/wiki/Shebang_(Unix)[shebang comment] must be added on the first line of the file:
+
+[source,pkl%parsed]
+----
+#!/usr/bin/env -S pkl run
+----
+
+NOTE: The `-S` flag for `env` is required on Linux systems due to a limitation of shebang handling in the Linux kernel.
+While not required on other *nix platforms like macOS, but it should be included for compatibility.
+
+==== Shell Completion
+
+Like with Pkl's own CLI, <<command-shell-completion, shell completions>> can be generated for standalone scripts.
+
+[source,shell]
+----
+# Generate shell completion script for bash
+./my-tool.pkl shell-completion bash
+
+# Generate shell completion script for zsh
+./my-tool.pkl shell-completion zsh
+----
+
+==== Customizing Completion Candidates
+
+`@Flag` and `@Argument` annotations may specify the `completionCandidates` to improve generated shell completions.
+
+Valid values include:
+
+* A `Listing<String>` of literal string values to offer for completion.
+* The literal string `"path"`, which offers local file paths for completion.
+
+Options with a string literal union type implicitly offer the members of the union as completion candidates.
+
+=== Flag name ambiguities
+
+It is possible for commands to define flags with names or short names that collide with Pkl's own command line options.
+To avoid ambiguity in parsing these options, all flags for Pkl itself (e.g. `--root-dir`) must be placed before the root command module's URI.
+Command authors are encouraged to avoid overlapping with Pkl's built-in flags, but this may not always be feasible, especially for single-character abbreviated names.
+
+This imposes a limitation around <<commands-as-standalone-scripts,standalone commands>> that prevents users from customizing Pkl evaluator options when they are invoked.
+There are two recommended workarounds for this limitation:
+
+* Use a `PklProject` to define evaluator settings instead of doing so on the command line.
+* If the command line must be used, switch to invoking via `pkl run [<flags>] [<root command module>]`.
 
 [[repl]]
 == Working with the REPL
--- a/pkl-cli/src/main/kotlin/org/pkl/cli/CliCommandRunner.kt
+++ b/pkl-cli/src/main/kotlin/org/pkl/cli/CliCommandRunner.kt
@@ -1,0 +1,263 @@
+/*
+ * Copyright © 2025-2026 Apple Inc. and the Pkl project authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.cli
+
+import com.github.ajalt.clikt.completion.CompletionCandidates
+import com.github.ajalt.clikt.completion.CompletionCommand
+import com.github.ajalt.clikt.core.*
+import com.github.ajalt.clikt.parameters.arguments.*
+import com.github.ajalt.clikt.parameters.options.*
+import com.github.ajalt.clikt.parameters.types.int
+import java.io.OutputStream
+import java.nio.file.Path
+import kotlin.io.path.createParentDirectories
+import kotlin.io.path.exists
+import kotlin.io.path.isDirectory
+import kotlin.io.path.writeBytes
+import org.pkl.commons.cli.CliBaseOptions
+import org.pkl.commons.cli.CliCommand
+import org.pkl.commons.cli.CliException
+import org.pkl.commons.cli.commands.installCommonOptions
+import org.pkl.commons.currentWorkingDir
+import org.pkl.core.Closeables
+import org.pkl.core.CommandSpec
+import org.pkl.core.EvaluatorBuilder
+import org.pkl.core.FileOutput
+import org.pkl.core.ModuleSource.uri
+import org.pkl.core.PklBugException
+import org.pkl.core.PklException
+import org.pkl.core.util.IoUtils
+
+class CliCommandRunner
+@JvmOverloads
+constructor(
+  private val options: CliBaseOptions,
+  private val args: List<String>,
+  private val outputStream: OutputStream = System.out,
+  private val errStream: OutputStream = System.err,
+) : CliCommand(options) {
+
+  private val normalizedSourceModule = options.normalizedSourceModules.first()
+
+  override fun doRun() {
+    val builder = evaluatorBuilder()
+    try {
+      evalCmd(builder)
+    } finally {
+      Closeables.closeQuietly(builder.moduleKeyFactories)
+      Closeables.closeQuietly(builder.resourceReaders)
+    }
+  }
+
+  private fun evalCmd(builder: EvaluatorBuilder) {
+    val evaluator = builder.build()
+    evaluator.use {
+      evaluator.evaluateCommand(uri(normalizedSourceModule)) { spec ->
+        try {
+          val root = SynthesizedRunCommand(spec, this, options.sourceModules.first().toString())
+          root.installCommonOptions(includeVersion = false)
+          root.subcommands(
+            CompletionCommand(
+              name = "shell-completion",
+              help = "Generate a completion script for the given shell",
+            )
+          )
+          root.parse(args)
+        } catch (e: PklException) {
+          throw e
+        } catch (e: Exception) {
+          throw e.message?.let { PklException(it, e) } ?: PklException(e)
+        }
+      }
+    }
+  }
+
+  /** Renders the comand's `output.bytes`, writing it to the standard output stream. */
+  fun writeOutput(outputBytes: ByteArray) {
+    if (outputBytes.isEmpty()) return
+    outputStream.write(outputBytes)
+    outputStream.flush()
+  }
+
+  /**
+   * Renders the command's `output.files`, writing each entry as a file.
+   *
+   * File paths are written to the standard error stream.
+   *
+   * Unlike CliEvaluator, command outputs write relative to --working-dir and may write files
+   * anywhere in the filesystem. This is intentionally less sandboxed than `pkl eval` and directly
+   * targets the capabilities of CLI tools written in general purpose languages. Pkl commands should
+   * therefore be treated as untrusted code the way that any other CLI tool would be.
+   */
+  fun writeMultipleFileOutput(outputFiles: Map<String, FileOutput>) {
+    if (outputFiles.isEmpty()) return
+
+    val writtenFiles = mutableMapOf<Path, String>()
+    val outputDir = options.normalizedWorkingDir
+    if (outputDir.exists() && !outputDir.isDirectory()) {
+      throw CliException("Output path `$outputDir` exists and is not a directory.")
+    }
+    for ((pathSpec, fileOutput) in outputFiles) {
+      checkPathSpec(pathSpec)
+      val resolvedPath = outputDir.resolve(pathSpec).normalize()
+      val realPath = if (resolvedPath.exists()) resolvedPath.toRealPath() else resolvedPath
+      val previousOutput = writtenFiles[realPath]
+      if (previousOutput != null) {
+        throw CliException(
+          "Output file conflict: `output.files` entries `\"${previousOutput}\"` and `\"$pathSpec\"` resolve to the same file path `$realPath`."
+        )
+      }
+      if (realPath.isDirectory()) {
+        throw CliException(
+          "Output file conflict: `output.files` entry `\"$pathSpec\"` resolves to file path `$realPath`, which is a directory."
+        )
+      }
+      writtenFiles[realPath] = pathSpec
+      realPath.createParentDirectories()
+      realPath.writeBytes(fileOutput.bytes)
+      val displayPath =
+        if (Path.of(pathSpec).isAbsolute) pathSpec
+        else IoUtils.relativize(resolvedPath, currentWorkingDir).toString()
+      errStream.writeText(displayPath + IoUtils.getLineSeparator())
+      errStream.flush()
+    }
+  }
+
+  class SynthesizedRunCommand(
+    private val spec: CommandSpec,
+    private val runner: CliCommandRunner,
+    name: String? = null,
+  ) : CliktCommand(name = name ?: spec.name) {
+    init {
+      spec.options.forEach { opt ->
+        when (opt) {
+          is CommandSpec.Flag ->
+            registerOption(
+              option(
+                  names = opt.names,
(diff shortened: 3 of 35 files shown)


pkl-cli/src/main/kotlin/org/pkl/cli/CliCommandRunner.kt
```diff
@@ -0,0 +1,401 @@
...
+
+    override val hiddenFromHelp: Boolean = spec.hide
+
+    override fun help(context: Context): String = spec.description ?: ""
+
+    private fun registerFlag(flag: CommandSpec.Flag) {
+      val names =
+        if (flag.shortName == null) arrayOf("--${flag.name}")
+        else arrayOf("--${flag.name}", "-${flag.shortName}")
+      val flag0 = option(names = names, help = spec.description ?: "", hidden = flag.hide)
```
bioball:
```suggestion
      val flag0 = option(names = names, help = flag.description ?: "", hidden = flag.hide)
```

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
```diff
@@ -0,0 +1,759 @@
...
+      if (!required) {
+        throw exceptionBuilder()
+            .withSourceSection(prop.getHeaderSection())
+            .evalError("commandOptionTypeNullableWithDefaultValue", prop.getName())
+            .build();
+      }
+      required = false;
+      type = nullable.getBaseType();
+    }
+    if (parseFunction != null) {
```
bioball:
```suggestion
    if (parseFunction == null) {
```

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
```diff
@@ -0,0 +1,759 @@
...
+      }
+    }
+
+    return Pair.of(flags, args);
+  }
+
+  // This sigil used to indicate that an option has a default value that is a non-constant expr.
+  // These values cause a flag to be marked optional, but are not shown in CLI help.
+  // Thus, they are not evaluated at spec parse time.
+  private static final Object COMPLEX_DEFAULT_EXPRESSION = new Object() {};
```
bioball:
```suggestion
  private static final Object COMPLEX_DEFAULT_EXPRESSION = new Object();
```

docs/modules/language-reference/pages/index.adoc
```diff
@@ -2736,6 +2736,10 @@ output {
 
 For more on path-based converters, see {uri-stdlib-PcfRenderer-converters}[PcfRenderer.converters].
 
+Special <<annotations,Annotations>> may also be used to influence how class properties are rendered.
```
stackoverflow, line 2739: Oh, we missed that from the annotation PR.
Can you separate this change into another PR?
I don't think we need to tie this change to pkl run. It's self contained.

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -881,6 +896,290 @@ If multiple module outputs are written to the same file, or to standard output,
...
+Pkl provides no facility for making arbitrary HTTP requests (i.e. non-`GET` methods or setting headers), executing other programs, or any other side effect.
+If additional functionality is desired, xref:language-reference:index.adoc#external-readers[external readers] may be used to extends Pkl's capabilities.
+
+=== Command Options
+
+Each property of the class of a command's `options` property becomes a command line option.
+Properties with the `local`, `hidden`, `fixed`, and/or `const` modifiers are not parsed as options
+A property's doc comment, if present, becomes the corresponding option's CLI help description.
+Doc comments are interpreted as Markdown text and formatted nicely when displayed to users.
+Properties must have xref:language-reference:index.adoc#type-annotations[type annotations] to determine how they are parsed.
```
stackoverflow, line 991: What if they are not? I think it's worth clarifying.

pkl-cli/src/main/kotlin/org/pkl/cli/CliCommandRunner.kt
```diff
@@ -0,0 +1,263 @@
...
+          throw e.message?.let { PklException(it, e) } ?: PklException(e)
+        }
+      }
+    }
+  }
+
+  /**
+   * Renders the comand's `output.bytes`, writing it to the standard output stream.
+   *
+   * Unlike CliEvaluator, there is no need to handle
```
stackoverflow: No need to handle what?

pkl-cli/src/main/kotlin/org/pkl/cli/CliCommandRunner.kt
```diff
@@ -0,0 +1,263 @@
...
+          "Output file conflict: `output.files` entry `\"$pathSpec\"` resolves to file path `$realPath`, which is a directory."
+        )
+      }
+      writtenFiles[realPath] = pathSpec
+      realPath.createParentDirectories()
+      realPath.writeBytes(fileOutput.bytes)
+      val displayPath =
+        if (Path.of(pathSpec).isAbsolute) pathSpec
+        else IoUtils.relativize(resolvedPath, currentWorkingDir).toString()
+      errStream.writeText(displayPath + IoUtils.getLineSeparator())
```
stackoverflow, line 134: Why are you writing the files to the error stream?

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
```diff
@@ -0,0 +1,1243 @@
...
+    checkFlagNames(prop, name, shortName);
+
+    // assert type is integral
+    var typeInfo = resolveType(prop);
+    if (!(typeInfo.getFirst() instanceof TypeNode.IntTypeNode
+            || typeInfo.getFirst() instanceof TypeNode.Int8TypeAliasTypeNode
+            || typeInfo.getFirst() instanceof TypeNode.Int16TypeAliasTypeNode
+            || typeInfo.getFirst() instanceof TypeNode.Int32TypeAliasTypeNode
+            || typeInfo.getFirst() instanceof TypeNode.UIntTypeAliasTypeNode // also UInt16, UInt32
+            || typeInfo.getFirst() instanceof TypeNode.UInt8TypeAliasTypeNode)
```
stackoverflow: Wouldn't it be easier to extract the underlying type from the alias and check if it's Int?

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
```diff
@@ -0,0 +1,1243 @@
...
+
+  /** Unwrap nullables, constraints, and aliases and return Pair(underlying type, is nullable) */
+  private Pair<TypeNode, Boolean> resolveType(TypeNode typeNode) {
+    var isNullable = false;
+    while (true) {
+      if (typeNode instanceof TypeNode.NullableTypeNode nullableTypeNode) {
+        isNullable = true;
+        typeNode = nullableTypeNode.getElementTypeNode();
+      } else if (typeNode instanceof TypeNode.ConstrainedTypeNode constrainedTypeNode) {
+        typeNode = constrainedTypeNode.getChildTypeNode();
```
stackoverflow, line 382: Shouldn't this function be recursive? What if your constrained type node contains another constrained type node inside, or a nullable type?

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
```diff
@@ -0,0 +1,1243 @@
...
+                  : new URI(null, null, IoUtils.toNormalizedPathString(modulePath), null);
+      uriString =
+          uri.isAbsolute() ? uri.toString() : IoUtils.resolve(workingDirUri, uri).toString();
+    } catch (URISyntaxException e) {
+      throw exceptionBuilder()
+          .evalError("invalidModuleUri", moduleName)
+          .withHint(e.getReason())
+          .build();
+    }
+    ;
```
stackoverflow:
```suggestion
```

pkl-core/src/main/java/org/pkl/core/util/GlobResolver.java
```diff
@@ -486,22 +488,32 @@ public static Map<String, ResolvedGlobElement> resolveGlob(
 
     var result = new LinkedHashMap<String, ResolvedGlobElement>();
     var hasAbsoluteGlob = globPattern.matches("\\w+:.*");
+    if ((enclosingModuleKey == null || enclosingUri == null) && !hasAbsoluteGlob) {
+      // callers must check that the glob is absolute if calling with null enclosing info
+      throw PklBugException.unreachableCode();
```
stackoverflow: This is not really unreachable, as someone could make a mistake while calling this function. Better to show an informative error message.

pkl-core/src/main/java/org/pkl/core/PClassInfo.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
```
stackoverflow: Can be reverted.

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1076,3 +1076,48 @@ invalidStringBase64=\
...
+Commands must not assign or amend property `{0}`.
+
+commandOptionNoTypeAnnotation=\
+No type annotation found for `{0}` property.
+
+commandOptionsTypeNotClass=\
+Type annotation `{0}` on `options` property in `pkl:Command` subclass must be a class type.
+
+commandOptionsTypeAbstractClass=\
+Class `{0}` for type annotation on `options` property in `pkl:Command` subclass must not be abstract.
```
stackoverflow: nit: it took me 3~4 reads of this error to understand exactly what it's about. But I'm not sure how to word it better.

pkl-core/src/test/kotlin/org/pkl/core/runtime/CommandSpecParserTest.kt
```diff
@@ -0,0 +1,838 @@
...
+    //          "foo",
+    //          "a",
+    //          CommandSpec.OptionType.Primitive(CommandSpec.OptionType.Primitive.Type.STRING,
+    // true),
+    //          null,
+    //          null,
+    //          "foo in BaseOptions",
+    //          false,
+    //        )
+    //      )
```
stackoverflow:
```suggestion
```

pkl-core/src/test/kotlin/org/pkl/core/runtime/CommandSpecParserTest.kt
```diff
@@ -0,0 +1,838 @@
...
+    //        CommandSpec.Flag(
+    //          "qux",
+    //          null,
+    //          CommandSpec.OptionType.Map(str, str, false),
+    //          null,
+    //          null,
+    //          null,
+    //          false,
+    //        )
+    //      )
```
stackoverflow:
```suggestion
```

stackoverflow (approved):
Looks good.
Left some comments, but nothing major.
Amazing work!

docs/modules/language-reference/pages/index.adoc
HT154: Extracted this out into https://github.com/apple/pkl/pull/1427

docs/modules/pkl-cli/pages/index.adoc
HT154: Omitting a type annotation on an option property results in a straightforward evaluation error. Not sure that requires clarification, but I can add it if you feel strongly.

pkl-cli/src/main/kotlin/org/pkl/cli/CliCommandRunner.kt
HT154: With `pkl eval -m` the written file paths are printed to stdout, but since `pkl run` can both write files and control stdout (`output.bytes`) I opted to direct this to stderr instead. I think it's probably important to retain the record that the CLI wrote files, but I'm open to feedback here.

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
HT154: This is in a `while (true)` loop and constrained and nullable type nodes do not break. That should be sufficient, right?

pkl-core/src/main/java/org/pkl/core/util/GlobResolver.java
HT154: Makes sense. Do we still use `PklBugException` in cases like this since it's a mistake in the Pkl code to hit this condition?

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
HT154: Yeah wow. Opted for "Command options class `{0}` may not be abstract."

pkl-cli/src/main/kotlin/org/pkl/cli/CliCommandRunner.kt
stackoverflow: That sounds sensible. Users can redirect streams in the shell if they want.

pkl-core/src/main/java/org/pkl/core/util/GlobResolver.java
stackoverflow: Yes, I think you should use `PklBugException`. This is an internal bug.

stdlib/Command.pkl
```diff
@@ -0,0 +1,276 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the Pkl project authors. All rights reserved.
```
holzensp: You'll wanna retrigger autoformatting before trying to land.

stdlib/Command.pkl
```diff
@@ -0,0 +1,276 @@
...
+import "pkl:Command"
+import "pkl:reflect"
+
+local commandClass = reflect.Class(getClass())
+
+/// Command configuration.
+hidden command: CommandInfo = new {
+  name = // choose the name of the module if the command is a module or the class otherwise
+    if (commandClass.name == "ModuleClass") reflect.Module(module).name else commandClass.name
+  description = commandClass.docComment // works for both clases and module classes
```
holzensp:
```suggestion
  description = commandClass.docComment // works for both classes and module classes
```

stdlib/Command.pkl
```diff
@@ -0,0 +1,276 @@
...
+}
+
+/// Command line options.
+///
+/// Set by the runtime during command execution.
+/// Must not be amended or overridden.
```
holzensp: Style [throughout]: I do think full sentences should be something to hold onto. There are some notable exceptions in the stdlib (e.g. `/// Throws when`, `/// [false] when`), but those are bugs if you ask me.

Content: `Must not be amended or overridden` is a little confusing; that's only true for _this_ `options` in `stdlib/Command.pkl`; after it is extended and redeclared, there might well be loads of reasons to amend or override. The implicit 'this' in the sentence is ambiguous.

stdlib/Command.pkl
```diff
@@ -0,0 +1,276 @@
...
+/// the command.
+///
+/// Example:
+/// ```
+/// extends "pkl:Command"
+///
+/// options: Options
+///
+/// class Options {
+///   // define flags and arguments here
```
holzensp: I'd argue this falls short in precisely the place where the example is supposed to show stuff. At the risk of being long-winded, maybe actually flash out a minumal example; maybe just `verbose: Boolean` and sumething with a numeric argument like `maxTries: UInt`?

stdlib/Command.pkl
```diff
@@ -0,0 +1,276 @@
...
+/// Set by the runtime during command execution.
+///
+/// `null` for root commands.
+hidden fixed root: Command? = parent?.root ?? parent
+
+class CommandInfo {
+  /// The name of the subcommand.
+  ///
+  /// Default value: the name of the module or class extending [Command].
+  /// Names that are `CamelCased` are transformed to `kebab-case`.
```
holzensp: Did you mean to `PascalCase` that use of `camelCased`?

stdlib/Command.pkl
```diff
@@ -0,0 +1,276 @@
...
+  ///
+  /// | Type                       | Behavior |
+  /// | -------------------------- | -------- |
+  /// | [String]                   | Value is used verbatim |
+  /// | [Char]                     | Value is used verbatim; must be exactly one character |
+  /// | [Boolean]                  | True values: `true`, `t`, `1`, `yes`, `y`, `on`; False values: `false`, `f`, `0`, `no`, `n`, `off` |
+  /// | [Number]                   | Value is parsed as an [Int] if possible, otherwise parsed as [Float] |
+  /// | [Float]                    | Value is parsed as a [Float] |
+  /// | [Int]                      | Value is parsed as an [Int] |
+  /// | [Int8], [Int16], [Int32], [UInt], [UInt8], [UInt16], [UInt32] | Value is parsed as an [Int] and must be wihtin the type's range |
```
holzensp:
```suggestion
  /// | [Int8], [Int16], [Int32], [UInt], [UInt8], [UInt16], [UInt32] | Value is parsed as an [Int] and must be within the type's range |
```

stdlib/Command.pkl
```diff
@@ -0,0 +1,276 @@
...
+  /// When the return value is an [Import] value or a [Pair] member, [List] or [Set] element
+  /// containing an [Import], the URI or glob URI specified by the value is imported and the value
+  /// is replaced with the value of the imported module(s).
+  ///
+  /// If no transform is provided, the raw flag value are parsed according to the option's type:
+  ///
+  /// | Type                       | Behavior |
+  /// | -------------------------- | -------- |
+  /// | [String]                   | Value is used verbatim |
+  /// | [Char]                     | Value is used verbatim; must be exactly one character |
```
holzensp, line 158: Because parsing... are we assuming some character encoding? In other words; I know what `Char` means _inside_ Pkl, but when parsing, the difference between e.g. byte or codepoint matters.

stdlib/Command.pkl
```diff
@@ -0,0 +1,276 @@
...
+  /// by default.
+  completionCandidates: "path" | *Listing<String>(isDistinct)
+}
+
+/// Annotates [Boolean] [options] properties to configure them as named CLI flags that may be
+/// specified zero or one times.
+///
+/// Boolean flags produce a pair of flags in the form `--<name>`/`--no-<name>`.
+///
+/// Annotating a property with a type other than [Boolean] is an error.
```
holzensp, line 204: You'd almost wonder whether it shouldn't just be the default behaviour for any `Boolean` typed property. I can come up with synthetic examples, but not (yet) real ones, of where you wouldn't.

stdlib/Command.pkl
```diff
@@ -0,0 +1,276 @@
...
+class BooleanFlag extends BaseFlag
+
+/// Annotates [Integer] [options] properties to configure them as named CLI flags that may be
+/// specified zero or more times.
+///
+/// Counted flags produce a value equal to the number of times a flag is specified.
+///
+/// Annotating a property with a type other than [Int], [Int8], [Int16], [Int32], [UInt], [UInt8],
+/// [UInt16], or [UInt32] is an error.
+class CountedFlag extends BaseFlag
```
holzensp, line 213: Does this mean the same as `a type other than [Int]`? All `IntN` types are just constrained `Int`, but I can't tell from this whether it matters if I defined my own `Int42` or not.

pkl-formatter/src/main/kotlin/org/pkl/formatter/Formatter.kt
```diff
@@ -31,6 +31,7 @@ class Formatter {
    * @return the formatted Pkl source code as a string
    * @throws java.io.IOException if the file cannot be read
    */
+  @JvmOverloads
```
holzensp: Are these annotations (a mild case of) PR-creep, or did you have other edits here that ultimately only turned into formatting applied with changed rules?

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1076,3 +1076,48 @@ invalidStringBase64=\
 
 characterCodingException=\
 Invalid bytes for charset "{0}".
+
+commandSubcommandConflict=\
+Command `{0}` has subcommands with conflicting name "{1}".
```
holzensp: What should a user do, in this case? Can the error help set a direction for the bug hunt?

pkl-core/src/main/java/org/pkl/core/PType.java
```diff
@@ -187,14 +187,20 @@ public static final class Union extends PType {
     @Serial private static final long serialVersionUID = 0L;
 
     private final List<PType> elementTypes;
+    private final @org.pkl.core.util.Nullable PType defaultElement;
```
holzensp: Adding a field to `PType` seems smelly. Finally, you don't appear to be using this elsewhere (`getDefaultElement` occurs once in the codebase); remnant of hacking? If so, revert.

pkl-core/src/main/java/org/pkl/core/Evaluator.java
```diff
@@ -224,6 +225,21 @@ static Evaluator preconfigured() {
    */
   TestResults evaluateTest(ModuleSource moduleSource, boolean overwrite);
 
+  /**
+   * Parses the command into a spec describing it and invokes the {@code run} argument.
```
holzensp: This `it` feels ambiguous. Also, the rest of the phrasing. What is a "spec" (concretely)?

pkl-core/src/main/java/org/pkl/core/Evaluator.java
```diff
@@ -224,6 +225,21 @@ static Evaluator preconfigured() {
...
+   *
+   * <p>This requires that the target module be a command module; it must extend module {@code
+   * "pkl:Command"}. Otherwise, a type mismatch error is thrown.
```
holzensp:
```suggestion
   * <p>This requires that the target module be a {@code
   * "pkl:Command"} instance.
```

Feels like explaining what a typecheck is.

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -546,6 +546,21 @@ Use `--no-power-assertions` to disable this feature if you prefer simpler output
 
 This command also takes <<common-options, common options>>.
 
+[[command-run]]
+=== `pkl run`
+
+*Synopsis:* `pkl run [<options>] [<module>] [<command options>]`
+
+Evaluate a CLI command defined by `<module>`.
```
holzensp: It feels warranted to include here that the module should be of type `pkl:Command` (or am I misreading code elsewhere?)

pkl-cli/src/main/kotlin/org/pkl/cli/CliCommandRunner.kt
holzensp: Warrants a `--quiet` flag, I'd say. I'd like to be able to take the existence of "real" `stderr` output (as opposed to progress) as a sign of something going wrong. Wouldn't call it blocking per se, but this is the sort of behaviour is that is far too hard to change later on if people depend on it.

pkl-cli/src/main/kotlin/org/pkl/cli/OutputUtils.kt
```diff
@@ -0,0 +1,35 @@
+/*
```
holzensp, line 1: YAUF - Yet Another Utils File
I see the point of putting this in its own file, to signal reuse more explicitly (as opposed to just dropping the `private` and keeping it in `CliEvaluator.kt`), but not being able to come up with a better name than "utils" suggests you might as well have it in a not-technically-as-descriptive-as-could-be place. Case in point; searching for `Utils` in `CliEvaluator.kt` bounces between `VmUtils` and `IoUtils` uses.
If you do want to keep this in its own file, does `OutputUtils` sound reasonable? Can we be more informative, yet (nothing springs to mind)?

pkl-core/src/main/java/org/pkl/core/ast/member/ClassMember.java
```diff
@@ -53,6 +54,33 @@ public final List<VmTyped> getAnnotations() {
     return annotations;
   }
 
+  public List<VmTyped> getAllAnnotations(boolean ascending) {
```
holzensp, line 57: This move feels like a noop (am I missing some use that wasn't available before?). Why move it up? And if because "also applicable here," why not up further?

pkl-core/src/main/java/org/pkl/core/CommandSpec.java
```diff
@@ -0,0 +1,158 @@
...
+    List<CommandSpec> subcommands,
+    ApplyFunction apply) {
+
+  public sealed interface Option {
+    String name();
+
+    String[] getNames();
+
+    class MisingOption extends RuntimeException {
+      public MisingOption() {}
```
holzensp: Nit; typo (not suggesting replacement, because better to use refactoring to catch other uses): `s/Mising/Missing/`

pkl-core/src/main/java/org/pkl/core/runtime/VmMap.java
```diff
@@ -52,6 +52,14 @@ public static VmMap create(ImMap<Object, Object> map, ImRrbt<Object> keyOrder) {
...
   }
 
+  @TruffleBoundary
+  public static VmMap create(Map<Object, Object> map) {
+    if (map.isEmpty()) return EMPTY;
+    return new VmMap(
+        PersistentHashMap.of(map.entrySet()),
+        RrbTree.emptyMutable().concat(map.keySet()).immutable());
+  }
+
```
holzensp: Is it me, or is this unused?

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -887,7 +895,8 @@ public boolean isNoopTypeCheck() {
     protected PType doExport() {
       var elementTypes =
           Arrays.stream(elementTypeNodes).map(TypeNode::export).collect(Collectors.toList());
-      return new PType.Union(elementTypes);
+      var defaultType = defaultIndex == -1 ? null : elementTypes.get(defaultIndex);
+      return new PType.Union(elementTypes, defaultType);
```
holzensp: ^^ Unused default in `PType` (see `PType.java` comment)

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -1061,8 +1070,16 @@ private boolean contains(Object value) {
...
+      PType defaultType = null;
+      var elementTypes = new ArrayList<PType>(stringLiterals.size());
+      for (var element : stringLiterals) {
+        var literal = new StringLiteral(element);
+        elementTypes.add(literal);
+        if (element.equals(unionDefault)) {
+          defaultType = literal;
+        }
+      }
+      return new PType.Union(elementTypes, defaultType);
```
holzensp: ^^ ditto unused

pkl-core/src/main/java/org/pkl/core/ast/VmModifier.java
```diff
@@ -146,6 +146,10 @@ public static boolean isLocalOrExternalOrAbstract(int modifiers) {
     return (modifiers & (LOCAL | EXTERNAL | ABSTRACT)) != 0;
   }
 
+  public static boolean isLocalOrExternalOrAbstractOrFixedOrConst(int modifiers) {
+    return (modifiers & (LOCAL | EXTERNAL | ABSTRACT | FIXED | CONST)) != 0;
+  }
+
```
holzensp, line 152: One might fearfully suggest that with names like these, leaking implementation detail might not be so much of a problem. JK

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -901,6 +916,291 @@ If multiple module outputs are written to the same file, or to standard output,
...
+== Implementing CLI Tools
+
+CLI tools can be implemented in Pkl by modules extending the `pkl:Command` module.
+These tools are executed by running `pkl run`.
+
+Pkl commands have a few properties that distinguish them from standard module evaluation:
+
+* Users provide input to commands using familiar command line idioms, providing a better experience than deriving inputs from xref:language-reference:index.adoc#resources[resources] like external properties or environment variables.
+* Commands can produce output to standard output (via `output.text` or `output.bytes`) and written files (via `output.files`) at the same time.
+* Like CLI tools written in other languages, command file output may write to any absolute path (not just relative to the `--multiple-file-output-path` option).
```
holzensp: Seems needlessly defensive phrasing. 

```suggestion
* Command file output may write to any absolute path (not only relative to the `--multiple-file-output-path` option).
```

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -901,6 +916,291 @@ If multiple module outputs are written to the same file, or to standard output,
...
+* Commands provide a way to dynamically import modules when they are specified as command line options.
```
holzensp:
```suggestion
* Commands can dynamically import modules when they are specified as command line options.
```

Took me a while to parse that. One fewer indirection, I think, means the same.

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -901,6 +916,291 @@ If multiple module outputs are written to the same file, or to standard output,
 By default, module outputs are separated with `---`, as in a YAML stream.
 The separator can be customized using the `--module-output-separator` option.
 
+[[cli-tools]]
+== Implementing CLI Tools
+
+CLI tools can be implemented in Pkl by modules extending the `pkl:Command` module.
+These tools are executed by running `pkl run`.
+
+Pkl commands have a few properties that distinguish them from standard module evaluation:
```
holzensp, line 942: I feel this is somewhat out of order. I would _first_ like to understand what "commands" are and _then_ we can discuss how they behave differently from standard modules.

I'm not sure whether I would move "defining commands" up, or just have quick example use case; smth like

> With `pkl:Command`, you can define a script in Pkl that is executed by your shell, giving a better CLI experience. It allows you to not have to say
> ```
> $ pkl eval script.pkl -p username=me -p password=password
> ```
> 
> and instead say
>
> ```
> $ ./script.pkl --username=me --password=password
> ```

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -901,6 +916,291 @@ If multiple module outputs are written to the same file, or to standard output,
...
+Commands are defined by creating a module that extends `pkl:Command`:
+
+[source,pkl%tested]
+.my-tool.pkl
+----
+/// This doc comment becomes part of the command's CLI help!
+/// Markdown formatting is **allowed!**
+extends "pkl:Command"
+
+options: Options
```
holzensp: A comment or call-out that you're re-declaring this type, maybe? Cause it's not the most conventional type mechanism and it's easy to miss.

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -901,6 +916,291 @@ If multiple module outputs are written to the same file, or to standard output,
...
+
+class Options {
+  // Define CLI flags/arguments...
+}
+
+// Regular module code...
+----
+
+Like `pkl eval`, when a command completes without an evaluation error the process exits successfully (exit code 0).
+Commands can return a failure using `throw` (exit code 1), but otherwise may not control the exit code.
```
holzensp, line 979: Then again, `pkl:Command` could also define `output.exitCode : Int` :D

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -901,6 +916,291 @@ If multiple module outputs are written to the same file, or to standard output,
...
+[[cli-tools]]
+== Implementing CLI Tools
+
+CLI tools can be implemented in Pkl by modules extending the `pkl:Command` module.
+These tools are executed by running `pkl run`.
+
+Pkl commands have a few properties that distinguish them from standard module evaluation:
+
+* Users provide input to commands using familiar command line idioms, providing a better experience than deriving inputs from xref:language-reference:index.adoc#resources[resources] like external properties or environment variables.
+* Commands can produce output to standard output (via `output.text` or `output.bytes`) and written files (via `output.files`) at the same time.
```
holzensp: > at the same time

doesn't sit right with me, but I may be over-sensitive to the word "time." It's "in the same evaluation" or "for a single module" or something, but I can't quite put my finger on it.

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -901,6 +916,291 @@ If multiple module outputs are written to the same file, or to standard output,
...
+----
+
+Like `pkl eval`, when a command completes without an evaluation error the process exits successfully (exit code 0).
+Commands can return a failure using `throw` (exit code 1), but otherwise may not control the exit code.
+
+Commands may set `output.text`/`output.bytes` and/or `output.files`.
+Unlike `pkl eval`, both outputs (stdout and files) are processed for all commands, so commands may write files and generate output.
+Additionally, keys in `output.files` are not "sandboxed" by `--multiple-file-output-path`.
+Instead, file entries with absolute paths are written to the specified path and entries with relative path are written relative to the current working directory (or `--working-dir` if specified).
+Written file paths are printed to the command's standard error.
```
holzensp: A lot of the meaning of this paragraph overlaps with the bullets on how commands evaluate differently than modules, but worded slightly differently. This makes me feel I need to understand the differences in wording... and then I almost miss the last sentence, which is entirely new.

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -901,6 +916,291 @@ If multiple module outputs are written to the same file, or to standard output,
...
+
+IMPORTANT: Users of `pkl run` must be aware of the security implications of this behavior.
+Command line tools implemented in Pkl will be able to write or overwrite any file that the user has permissions to modify.
+This matches the behavior of command line tools written in any other language.
```
holzensp: Overly defensive, especially the "others do it, too".... If you want to warn explicitly, I would not defend it by comparing to other languages, but by contrasting it with usual Pkl behaviour...


```suggestion
IMPORTANT: Users of `pkl run` must be aware of the security implications of this behavior. Whereas usual `pkl eval` prevents accidental overwrites, by not allowing absolute paths, `pkl run` does not offer this protection. As such, it writes the credentials of the invoking user.
```

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -901,6 +916,291 @@ If multiple module outputs are written to the same file, or to standard output,
...
+
+Beyond writing to files (via `output.files`) and the standard output (via `output.text` or `output.bytes`), commands may not trigger side effects not already available in Pkl.
+Pkl provides no facility for making arbitrary HTTP requests (i.e. non-`GET` methods or setting headers), executing other programs, or any other side effect.
```
holzensp:
```suggestion
Other than the differences listed above, commands behave like any other Pkl module.
For example, there are no ways to execute other programs or make arbitrary HTTP requests.
```

The build-up to the external readers plug is well warranted, but it seems a little... too much. Do with it what you will; we're in taste-country now.

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -901,6 +916,291 @@ If multiple module outputs are written to the same file, or to standard output,
...
+If additional functionality is desired, xref:language-reference:index.adoc#external-readers[external readers] may be used to extends Pkl's capabilities.
+
+=== Command Options
+
+Each property of the class of a command's `options` property becomes a command line option.
```
holzensp: property, class, command, option

There's something in this use of jargon that makes it feel circular... it isn't, but it feels that way

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
```diff
@@ -0,0 +1,1232 @@
...
+              .build();
+        } else if (!(memberNode.getBodyNode() instanceof DefaultPropertyBodyNode)) {
+          throw new VmExceptionBuilder()
+              .evalError("commandMustNotAssignOrAmendProperty", name)
+              .withSourceSection(memberNode.getSourceSection())
+              .build();
+        }
+      }
+      value = value.getParent();
+    }
```
holzensp: Is this doing something more special than `VmUtils` lookups? It's a little hidden in the repetitiveness of the two very subtly different exceptions, but I have to admit, Java isn't helping here.

```suggestion
      var member = VmUtils.findMember(value, name);
      if (member == null) return;

      var memberNode = member.getMemberNode();
      var sourceSection = memberNode == null
        ? member.getSourceSection()
        : memberNode.getBodyNode() instanceof DefaultPropertyBodyNode
          ? null
          : memberNode.getSourceSection();
      if (sourceSection != null) {
        throw new VmExceptionBuilder()
            .evalError("commandMustNotAssignOrAmendProperty", name)
            .withSourceSection(sourceSection)
            .build();
      }

```

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
```diff
@@ -0,0 +1,1232 @@
...
+    var valueClass = typed.getVmClass();
+    while (valueClass != clazz) {
+      valueClass = valueClass.getSuperclass();
+      if (valueClass == null) {
+        throw exceptionBuilder().typeMismatch(value, clazz).build();
+      }
+    }
+
+    return typed;
+  }
```
holzensp: Looks a *lot* like `TestRunner::checkAmends`; maybe roll up into `VmUtils` function?

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
```diff
@@ -0,0 +1,1232 @@
...
+    return result;
+  }
+
+  private Object handleImport(VmTyped mport, URI workingDirUri) {
+    var moduleName = (String) VmUtils.readMember(mport, Identifier.URI);
+    String uriString;
+    // Ported from org.pkl.cli.commons.cli.commands.BaseOptions:
+    try {
+      // Can't just use URI constructor, because URI(null, null, "C:/foo/bar", null) turns
+      // into `URI("C", null, "/foo/bar", null)`.
```
holzensp, line 1121: *sigh*

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
```diff
@@ -0,0 +1,1232 @@
...
+
+  public CommandSpec parse(VmTyped command) {
+    checkAmends(command, CommandModule.getModule().getVmClass());
+    checkPropertyIsUndefined(command, Identifier.OPTIONS);
+    checkPropertyIsUndefined(command, Identifier.PARENT);
+
+    var optionsClass = getOptionsClass(command);
+    var commandInfo =
+        checkAmends(
+            VmUtils.readMember(command, Identifier.COMMAND), CommandModule.getCommandInfoClass());
```
holzensp, line 93: Why do you have to check this so specifically? Isn't this already a given after you've asserted the module amends `pkl:Command`.... am I missing a step where you bypass the normal type check?

holzensp (approved):
Approving to unblock, because no hard blockers. Some points to consider, though.

stdlib/Command.pkl
HT154: Ah, this is vestigial. Names are used verbatim now. Removed.

stdlib/Command.pkl
HT154: We're delegating to Java here (`argv` is a `[]String`) which means it depends on the Java version or native image behavior, which in turn can depend on the `file.encoding` property or the OS and locale. I don't see CLI frameworks typically specify here. Clikt doesn't specify and we rely on it here so I'm not sure we can specify either.

stdlib/Command.pkl
HT154: Updated. The implementation is actually looser here (but wasn't before addressing Islon's feedback)

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
HT154: `VmUtils.readMember` doesn't run type checks and we (intentionally) haven't forced the module at this point, so there's nothing stopping users from doing something silly like this:
```pkl
extends "pkl:Command"
class Foo
command: Foo = new {}
// OR, even sillier:
command: Int = 0
```

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
HT154: Good call, updated

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
HT154: Much better! Thanks for the suggestion

pkl-core/src/main/java/org/pkl/core/ast/member/ClassMember.java
HT154: "also applicable here" indeed, but it can't go up further since `ClassMember` is where annotations are introduced.

pkl-formatter/src/main/kotlin/org/pkl/formatter/Formatter.kt
HT154: Hmm creep from real early in the PR's life. Extracted this out into https://github.com/apple/pkl/pull/1428.

docs/modules/pkl-cli/pages/index.adoc
HT154: I considered this, but to do that we'd need a new `CommandOutput` class that subclassed `ModuleOutput` and I'm hesitant to `open` that up without being able to make `FileOutput` a `sealed` class.

This is a possible avenue for future enhancements if there is sufficient demand.

pkl-cli/src/main/kotlin/org/pkl/cli/OutputUtils.kt
HT154: Opted for `OutputUtils` here. Not in love with it, but can't move it down into `IoUtils` without a) sacrificing the kt extensions (a little bit unappealing) or b) having `checkPathSpec` throw something other than `CliException` and complicating/un-DRYing (WETting??) its two call sites (even less appealing).

I can see the argument for moving the extensions to `IoUtils` as static methods and dropping `checkPathSpec` into `pkl-commons-cli` somewhere (same naming issue there though), but I can't convince myself that's actually an improvement over just renaming this file.

stdlib/Command.pkl
HT154: My main thoughts here are two-fold:
* Consistency: having `@Flag` always do the same thing ("accept a single value") provides a clearer experience
* Explicitness: A `@BooleanFlag` is up front that it accepts no value and synthesizes a `--no-<name>` inversion.

The reality is I think `@Flag` with `Boolean` will probably be somewhat rare. If I could, I'd have them accept zero or one value so both ways Just Work, but this introduces potential for [ambiguity that clikt isn't willing to take on](https://github.com/ajalt/clikt/issues/280#issuecomment-788154939).

stdlib/Command.pkl
HT154: I've revised these specific doc comments as well as some of the fragments throughout.

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1076,3 +1076,49 @@ invalidStringBase64=\
 
 characterCodingException=\
 Invalid bytes for charset "{0}".
+
+commandSubcommandConflict=\
+Command `{0}` has subcommands with conflicting name "{1}".\n\
```
holzensp, line 1081: Nit; it seems we've not done so in `invalidOpaqueFileUri`, but for multiline errors, we usually have the one 'headline' error on its own line, followed by a white line, followed by a hint. Best keep that [this and below messages]
```suggestion
Command `{0}` has subcommands with conflicting name "{1}".\n\
\n\
```

pkl-core/src/main/java/org/pkl/core/ast/member/ClassMember.java
holzensp: That was some sloppy reviewing on my part (soz, build was broken; spotless & JDK versions). Carry on.

stdlib/Command.pkl
holzensp: Fair point.

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
holzensp: Ah, this being unforced still... so obvious once you realise ;)

docs/modules/pkl-cli/pages/index.adoc
```diff
@@ -901,6 +920,301 @@ If multiple module outputs are written to the same file, or to standard output,
...
+* Commands may write to standard output (via `output.text` or `output.bytes`) and the filesystem (via `output.files`) in the same evaluation.
+* Command file output may write to any absolute path (not only relative to the `--multiple-file-output-path` option).
+** Relative output paths are written relative to the current working directory (or `--working-dir`, if specified).
+** Paths of output file are printed to the command's standard error.
+
+IMPORTANT: Users of `pkl run` must be aware of the security implications of this behavior.
+Using `pkl eval` prevents accidental overwrites by not allowing absolute paths, but `pkl run` does not offer this protection.
+Commands may write to any path the invoking user has permissions to modify.
+
+Commands are implemented as regular modules and declare their supported command line flags and positional arguments using a class with annotated properties.
```
holzensp, line 955: *Much* improved!

holzensp (approved):
One final nit on white lines in error messages, but looking good.

PR #1369, opened 2025-12-14
Add error hint when accessing the default value of a union type with no default

### LINKED ISSUE #1368: Amending an undefined property with a union type with no default member should produce a more specific error
When a field is typed as `Listing` or `String`, Pkl will fail to intuit the type of the value if it inherits from Dynamic. `Int` will also be intuited correctly, whereas `Mapping` is also not intuited.

**Example:**
```pkl
class ExampleTyped {
    aa: Listing<String> | String
    bb: String
}

typed: ExampleTyped = new {
    aa { "hello" } // aa = new Listing { "hello" } works fine
    bb = "hello"
}
```

**Pkl Error:**
```
– Pkl Error ––
Tried to read property `aa` but its value is undefined.

21 | aa: Listing<String> | String
     ^^
at example.aa (file:///path/to/example.pkl, line 21)

The above error occurred when rendering path `typed.aa` of module `file:///path/to/example.pkl`.

128 | renderer.renderDocument(value)
at pkl.base#Module.output.text (https://github.com/apple/pkl/blob/0.30.1/stdlib/base.pkl#L128)

132 | if (renderer is BytesRenderer) renderer.renderDocument(value) else text.encodeToBytes("UTF-8")
                                                                         ^^^^
at pkl.base#Module.output.bytes (https://github.com/apple/pkl/blob/0.30.1/stdlib/base.pkl#L132)
```

Version: `Pkl 0.30.1 (Linux 6.11.0-1018-azure, native)`

HT154:
Resolves #1368

### DIFF
--- a/pkl-core/src/test/files/LanguageSnippetTests/output/errors/noDefault3.err
+++ b/pkl-core/src/test/files/LanguageSnippetTests/output/errors/noDefault3.err
@@ -1,12 +1,12 @@
 –– Pkl Error ––
-Tried to read property `foo` but its value is undefined.
-Union type `Foo|"baz"` has no default member.
+Tried to read property `b` but its value is undefined.
+Union type `Listing<String> | String` (from type alias `B`) has no default member.
 
-x | foo: Foo|"baz"
-    ^^^
-at noDefault2#foo (file:///$snippetsDir/input/errors/noDefault2.pkl)
+x | b: B
+    ^
+at noDefault3#b (file:///$snippetsDir/input/errors/noDefault3.pkl)
 
-The above error occurred when rendering path `foo` of module `file:///$snippetsDir/input/errors/noDefault2.pkl`.
+The above error occurred when rendering path `b` of module `file:///$snippetsDir/input/errors/noDefault3.pkl`.
 
 xxx | renderer.renderDocument(value)
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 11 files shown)


pkl-core/src/test/files/LanguageSnippetTests/output/errors/noDefault3.err
```diff
@@ -0,0 +1,17 @@
+–– Pkl Error ––
+Tried to read property `b` but its value is undefined.
+Union type `Listing<String> | String` (from type alias `B`) has no default member.
```
bioball: This message doesn't explain what "Default member" is. If we want this as the error message, probably should have a suggestion of how to mark a union default member.

But, this error message feels a little strange to me for a couple reasons. This throws too:

```pkl
foo: Int | Boolean | String
```

But having a default member doesn't help, because none of these types have default values.

Also, this error message feels oddly specific; the issue here is that the property doesn't have a default value.
There's no default value because:

1. The class property has no declared default value
2. The property's declared type has no default value
3. The object instance did not assign to this property

So there's three possible resolutions:

1. Add an explicit default in the class property (only if this is modifiable code)
2. Change the type to have an explicit default (only if one of the member types has a default value, and this is modifiable code)
3. Define the property in your instance

I think this current error message would probably be misleading too often.

HT154: I updated the condition here to check that the union has no default member and updated the message a bit. How does it look now?

bioball: It still feels odd to me:

1. Adding a default marker doesn't do anything for union types like `Int | Boolean | String`
2. The code might not be modifiable, so we shouldn't have verbage that suggests updating the type as a fix.

To address https://github.com/apple/pkl/issues/1368, it feels like the error message should be something like this?

```
Cannot instantiate type `Listing<String> | String` because it has no default value.
```

But we'd really only want to show this error message when we're trying to amend it with an object body.

PR #1383, opened 2025-12-22
Fix super method call inside let expression

### LINKED ISSUE #1309: Method supercall doesn't work inside let expression
```pkl
open class A {
  function foo() = "a"
}

class B extends A {
  function foo() =
    let (bar = "b")
      super.foo() + bar
}
```
Result
```
–– Pkl Error ––
Cannot invoke a supermethod from here.

8 | super.foo() + bar
    ^^^^^^^^^^^
at test#B.foo.<function#1> (file://<dir>/test.pkl, line 8)
```

This works as expected:
```pkl
class B extends A {
  function foo() =
    super.foo() + "b"
}
```

cipheraxat:
Fixes #1309

The issue was that super calls were blocked inside let expressions because:
1. The compiler's isClassMemberScope() check didn't skip over lambda scopes created by let expressions
2. The runtime's findSupermethod() didn't traverse past VmFunction owners to find the actual class prototype

Changes:
- SymbolTable.java: Updated isClassMemberScope() to skip lambda scopes before checking if the parent is a class or module scope
- InvokeSuperMethodNode.java: Updated findSupermethod() to skip VmFunction owners when looking for the class prototype

Added regression tests covering:
- Super method calls inside let expressions
- Super property access inside let expressions
- Nested let expressions with super calls

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/ast/expression/member/InvokeSuperMethodNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/expression/member/InvokeSuperMethodNode.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


pkl-core/src/main/java/org/pkl/core/ast/expression/member/InvokeSuperMethodNode.java
```diff
@@ -66,6 +66,9 @@ protected Object eval(
 
   protected ClassMethod findSupermethod(VirtualFrame frame) {
     var owner = VmUtils.getOwner(frame);
+    while (owner instanceof org.pkl.core.runtime.VmFunction) {
+      owner = owner.getEnclosingOwner();
+    }
```
bioball, line 72:
```suggestion
    }
    assert owner != null : "VmFunction always has a parent";
```

PR #1384, opened 2025-12-23
Implement power assertions

bioball:
This adds power assertions to Pkl!

This implements the SPICE described in https://github.com/apple/pkl-evolution/pull/29

This follows the power assertions style of reporting also found in
Groovy, Kotlin, and others.

* Literal values are not emitted in the diagram
* Stdlib constructors of literals like `List(1, 2)` are also considered
  literals

Power assertions are added to:

* Failing type constraints
* Failing test facts

Power assertions are implemented as a truffle instrument to observe execution.
When an assertion fails, the instrument is created and the assertion is run again to observe facts.
This incurs runtime overhead to collect facts, but has no impact on code in the non-error case.

Samples:

```
module test
  facts
    ✘ math
       myList.fold(0, (a, b) -> a + b) == 5 (/Users/danielchao/code/apple/pkl/.dan-scripts/test.pkl:6)
       |      |                        |
       |      6                        false
       List(1, 2, 3)

0.0% tests pass [1/1 failed], 0.0% asserts pass [1/1 failed]
```

```
–– Pkl Error ––
Type constraint `this == this.capitalize()` violated.
Value: "fooey"

    this == this.capitalize()
    |    |  |    |
    |    |  |    "Fooey"
    |    |  "fooey"
    |    false
    "fooey"

1 | typealias CapitalizedString = String(this == this.capitalize())
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^
at test#foo (/Users/danielchao/code/apple/pkl/.dan-scripts/test.pkl:1)

3 | foo: CapitalizedString = "fooey"
                             ^^^^^^^
at test#foo (/Users/danielchao/code/apple/pkl/.dan-scripts/test.pkl:3)
```

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/runtime/PowerAssertions.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/PowerAssertions.java
@@ -169,6 +169,7 @@ public class PowerAssertions {
 
   private static boolean shouldHide(
       Node truffleNode, org.pkl.parser.syntax.Node parserNode, Object value, Expr rootExpr) {
+    // literal values are self-evident in their source code
     if (isLiteral(truffleNode, parserNode)) {
       return true;
     }
--- a/docs/modules/release-notes/pages/0.31.adoc
+++ b/docs/modules/release-notes/pages/0.31.adoc
@@ -30,7 +30,7 @@ These are:
 * Type constraint expressions
 * Test facts
 
-Currently, when these assertions fail, the error message tells prints the assertion's source section.
+Currently, when these assertions fail, the error message prints the assertion's source section.
 However, this information can sometimes be lacking.
 It tells you what the mechanics of the assertion is, but doesn't tell you _why_ the assertion is failing.
 
--- a/pkl-core/src/main/java/org/pkl/core/ast/expression/binary/ComparatorNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/expression/binary/ComparatorNode.java
@@ -19,7 +19,7 @@ import com.oracle.truffle.api.frame.VirtualFrame;
 import com.oracle.truffle.api.source.SourceSection;
 
 public abstract class ComparatorNode extends BinaryExpressionNode {
-  public ComparatorNode(SourceSection sourceSection) {
+  protected ComparatorNode(SourceSection sourceSection) {
     super(sourceSection);
   }
 


pkl-core/src/main/java/org/pkl/core/ast/type/VmTypeMismatchException.java
```diff
@@ -26,16 +27,18 @@
...
 import org.pkl.core.runtime.VmException.ProgramValue;
+import org.pkl.core.util.AnsiStringBuilder;
 import org.pkl.core.util.ErrorMessages;
 import org.pkl.core.util.Nullable;
 
 /**
  * Indicates that a type check failed. [TypeNode]s use this exception instead of [VmException] to
  * make type checking of union types efficient. Note that any [TruffleBoundary] between throw and
- * catch location of this exception must set `transferToInterpreterOnException = false`. (Currently
+ * catch location of this exception must a `transferToInterpreterOnException = false`. (Currently
```
HT154: What's the intent here?

pkl-core/src/main/java/org/pkl/core/runtime/PowerAssertions.java
```diff
@@ -0,0 +1,496 @@
...
+
+  private record MarkerEntry(int startLine, int startColumn) implements LayerEntry {
+    @Override
+    public int length() {
+      return 1;
+    }
+
+    @Override
+    public void appendTo(AnsiStringBuilder builder) {
+      builder.append(AnsiCode.FAINT, "|");
```
HT154: Nit: style 😎

```suggestion
      builder.append(AnsiCode.FAINT, "│");
```

This matches [swift-power-assert's styling](https://github.com/kishikawakatsumi/swift-power-assert/blob/da46b4a95ae3b80dbe6375c4a5f7e5245a6db04e/Sources/PowerAssert/PowerAssert.swift#L165) and I think presents a slightly clearer visual.

pkl-core/src/main/java/org/pkl/core/runtime/PowerAssertions.java
```diff
@@ -0,0 +1,496 @@
...
+    var count = 0;
+
+    while (start < str.length() && count < n && Character.isWhitespace(str.charAt(start))) {
+      start++;
+      count++;
+    }
+
+    return str.substring(start);
+  }
+
```
HT154, line 352: Why two separate variables here? They're always set to the same value.

HT154 (approved):
Very nice!! Just a few minor things.

pkl-core/src/main/java/org/pkl/core/ast/type/VmTypeMismatchException.java
bioball: Bad find+replace 🤦

pkl-core/src/main/java/org/pkl/core/runtime/PowerAssertions.java
bioball: Ah, nice!

pkl-core/src/main/java/org/pkl/core/runtime/PowerAssertions.java
bioball: Good catch!

pkl-parser/src/main/java/org/pkl/parser/syntax/Expr.java
```diff
@@ -608,12 +624,12 @@ public String toString() {
     @SuppressWarnings("ConstantValue")
     @Override
     public boolean equals(Object o) {
-      if (this == o) {
-        return true;
-      }
       if (o == null || getClass() != o.getClass()) {
         return false;
       }
+      if (this == o) {
```
stackoverflow, line 686: Why is this changed?

pkl-parser/src/main/java/org/pkl/parser/GenericParser.java
```diff
@@ -48,6 +48,11 @@ private void init(String source) {
     spanLookahead = _lookahead.span;
   }
 
+  public Node parseExpression(String source) {
+    init(source);
+    return parseExpr();
```
stackoverflow: You should expect EOF after the expression (look at parseExpressionInput in Parser.java) unless you are partially parsing a expression and ignoring the rest of the file.

pkl-parser/src/main/java/org/pkl/parser/Span.java
```diff
@@ -42,4 +46,14 @@ public Span move(int amount) {
...
+
+  /** Tells if {@code other} is entirely within this span. */
+  public boolean contains(Span other) {
+    var offset = other.charIndex - charIndex;
+    if (offset < 0) {
+      return false;
+    }
+    var maxLength = length - offset;
+    return other.length <= maxLength;
+  }
```
stackoverflow, line 53:
```suggestion
  public boolean contains(Span other) {
    return charIndex <= other.charIndex
        && other.charIndex + other.length <= charIndex + length;
  }
```
nit

pkl-cli/src/main/kotlin/org/pkl/cli/CliTestRunner.kt
```diff
@@ -106,6 +106,7 @@ constructor(
         } catch (ex: Exception) {
           errWriter.appendLine("Error evaluating module ${moduleUri.path}:")
           errWriter.write(ex.message ?: "")
+          errWriter.write(ex.stackTraceToString())
```
stackoverflow: Is this intended to be here, or is it a debug leftover?

pkl-core/src/main/java/org/pkl/core/runtime/PowerAssertions.java
```diff
@@ -0,0 +1,495 @@
...
+    if (isLiteral(parserNode)) {
+      return true;
+    }
+    // assumption: only calls to methods within the base module are direct method calls.
+    if (!(truffleNode instanceof InvokeMethodDirectNode)) {
+      return false;
+    }
+    var accessExpr = (UnqualifiedAccessExpr) parserNode;
+    // assumption: "literal" values are method calls that are uppercased. e.g. IntSeq(..), Pair(..)
+    if (!Character.isUpperCase(accessExpr.getIdentifier().getValue().charAt(0))) {
```
stackoverflow, line 162: So, user created functions that start with an uppercase char will be treated as literals if they only receive literals. Is this on purpose?
We could check the identifier is `List`, `Pair`, etc.

pkl-core/src/main/java/org/pkl/core/runtime/PowerAssertions.java
```diff
@@ -0,0 +1,495 @@
...
+        skip = "super".length();
+      } else if (parserNode instanceof QualifiedAccessExpr qualifiedAccessExpr) {
+        skip =
+            qualifiedAccessExpr.getExpr().span().stopIndexExclusive()
+                - parserNode.span().charIndex();
+      }
+      if (skip == 0) {
+        return originalSource.getCharIndex();
+      }
+      // need to lex the source again because the parse tree doesn't carry the operator's position
```
stackoverflow, line 468: Using the GenericParser would give you the full source position. And as a bonus, line and column information in the spans.

pkl-cli/src/main/kotlin/org/pkl/cli/CliTestRunner.kt
bioball: Intentional, but this is a bugfix unrelated to this feature. I'll pull this into a separate PR

pkl-core/src/main/java/org/pkl/core/runtime/PowerAssertions.java
bioball: No, this is always a method off `pkl:base` (we only create `InvokeMethodDirectNode` when calling into methods defined on the base class)

pkl-core/src/main/java/org/pkl/core/runtime/PowerAssertions.java
bioball: I tried it, but it doesn't provide the span for the operator still.

pkl-parser/src/main/java/org/pkl/parser/syntax/Expr.java
bioball: This is a different change that snuck into this PR (see #1395)!

pkl-parser/src/main/java/org/pkl/parser/GenericParser.java
bioball: Actually, this method isn't used; will remove

pkl-core/src/main/java/org/pkl/core/runtime/PowerAssertions.java
stackoverflow: This seems like a bug in the GenericParser. But anyway, this was just a suggestion, we can keep using the regular parser.

stackoverflow (approved):
LGTM!

PR #1390, opened 2026-01-05
feat: add pkl bom module 

eddie4941:
This adds a module that will publish a bom file that will help consumers
align versions for all pkl artifacts using the single bom.


pkl-bom/pkl-bom.gradle.kts
```diff
@@ -0,0 +1,56 @@
...
+  `java-platform`
+  pklAllProjects
+  pklPublishBom
+}
+
+description = "Pkl BOM that includes all modules"
+
+dependencies {
+  constraints {
+    api(projects.pklCli)
```
eddie4941, line 27: @bioball Ive just added all modules. But it could very well make sense to only include those modules that are commonly consumed by users (pkl-core, pkl-executor, pkl-config-*, etc). Let me know what you think?

buildSrc/src/main/kotlin/PklPublishing.kt
```diff
@@ -0,0 +1,147 @@
...
+import org.gradle.api.Project
+import org.gradle.api.publish.PublishingExtension
+import org.gradle.api.publish.maven.MavenPublication
+import org.gradle.api.publish.maven.tasks.AbstractPublishToMaven
+import org.gradle.api.publish.maven.tasks.GenerateMavenPom
+import org.gradle.kotlin.dsl.*
+import org.gradle.plugins.signing.SigningExtension
+
+/** Configures common POM metadata (licenses, developers, SCM, etc.) for all Pkl publications. */
+fun Project.configurePklPomMetadata() {
```
eddie4941, line 27: This is a totally optional commit and can be dropped if you prefer.

bioball: Makes sense to have these!

buildSrc/src/main/kotlin/pklPublishBom.gradle.kts
```diff
@@ -0,0 +1,33 @@
...
+      create<MavenPublication>("library") { from(platformComponent) }
+    }
+  }
+}
+
+configurePklPomMetadata()
+
+configurePomValidation()
+
+configurePklSigning()
```
bioball: Since this is only used in one module, let's just move all of this directly into `pkl-bom.gradle.kts` and get rid of this file.

pkl-bom/pkl-bom.gradle.kts
bioball: Yeah, some of these modules aren't meant to be public APIs.

Let's get rid of:

* pklCommonsTest
* pklCommonsCli
* pklCommons
* pklServer

I think the rest of this is fine to keep.

buildSrc/src/main/kotlin/pklPublishBom.gradle.kts
eddie4941: done in: https://github.com/apple/pkl/pull/1390/commits/f35d52220413702404ac43809648d5bbaa448222

pkl-bom/pkl-bom.gradle.kts
eddie4941: done in: https://github.com/apple/pkl/pull/1390/commits/cf1c8b9feceb46c5f7c3f1331ad0ae8113ffe3d8

PR #1394, opened 2026-01-06
Change loading of external readers

bioball:
This introduces breaking changes for external readers are loaded:

1. In PklProject, relative paths are resolved relative to the enclosing PklProject file (make behavior consistent with how other settings work)
1. Make CLI flags blow away any settings set on a PklProject
1. Introduce a new `workingDir` property, which defaults to the PklProject dir

The overall goal is to make this behavior consistent with how other settings work.
For example, relative paths for other evaluator settings are already relative to the project directory.
Additionally, in every other case, CLI flags will overwrite any setting set within PklProject.

[windows]


pkl-core/src/main/java/org/pkl/core/evaluatorSettings/PklEvaluatorSettings.java
```diff
@@ -190,10 +193,15 @@ public static Proxy create(@Nullable String address, @Nullable List<String> noPr
...
-    public static ExternalReader parse(Value input) {
+    public static ExternalReader parse(
+        Value input, BiFunction<String, String, Path> pathNormalizer) {
       if (input instanceof PObject externalReader) {
         var executable = (String) externalReader.getProperty("executable");
+        var executablePath = pathNormalizer.apply(executable, "executable");
         var arguments = (List<String>) externalReader.get("arguments");
+        if (Files.exists(executablePath)) {
+          return new ExternalReader(executablePath.toString(), arguments);
+        }
```
HT154: Hmm this is actually the opposite behavior of the way most shells work. Typically if you have a file `foo.sh` in your PATH and in the current directory, executing `foo.sh` uses the one from PATH. Running `./foo.sh` disambiguates, asking the shell to run the one in the working dir. Maybe we should treat this similarly? I wonder if—provided the executable contains no `/`—we should try resolving against PATH at this layer (and in the CLI and in the language bindings) and fall back to relative-to-project-dir and then require absolute paths only in `ExternalReaderProcess`.

stdlib/EvaluatorSettings.pkl
```diff
@@ -118,6 +120,70 @@ externalResourceReaders: Mapping<String, ExternalReader>?
 @Since { version = "0.30.0" }
 traceMode: ("compact" | "pretty")?
 
+/// The enclosing URI used to resolve relative paths.
+@Since { version = "0.31.0" }
+hidden enclosingUri: String
```
HT154: Is this used anywhere?

stdlib/Project.pkl
```diff
@@ -441,3 +468,54 @@ typealias CommonSpdxLicenseIdentifier =
 @Unlisted
 @Since { version = "0.27.0" }
 fixed annotations: List<Annotation> = reflect.moduleOf(this).annotations
+
+/// Machiney for URI percent decoding.
```
HT154:
```suggestion
/// Machinery for URI percent decoding.
```

HT154 (approved):
LGTM modulo the above comments, approving to unblock

stdlib/EvaluatorSettings.pkl
bioball: Ah, nope, good catch!

stackoverflow (approved):
LGTM

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/CliCommand.kt
```diff
@@ -109,8 +109,7 @@ abstract class CliCommand(protected val cliOptions: CliBaseOptions) {
   }
 
   private val evaluatorSettings: PklEvaluatorSettings? by lazy {
-    @Suppress("PklCliDirectProjectEvaluatorSettingsAccess")
-    if (cliOptions.omitProjectSettings) null else project?.evaluatorSettings
```
HT154, line 113: The annotation should be preserved and the rule in .idea/inspectionProfiles/Project_Default.xml should be updated to additionally catch `project.resolvedEvaluatorSettings` as well to prevent regressing on #1459.

stdlib/EvaluatorSettings.pkl
```diff
@@ -118,6 +123,105 @@ externalResourceReaders: Mapping<String, ExternalReader>?
...
+        [readerName] {
+          executable = resolveExecutable(enclosingUri, reader.executable, forWindows)
+          workingDirectory =
+            let (dir = reader.workingDirectory)
+              if (dir == null)
+                resolvePath(enclosingUri, "./", forWindows)
+              else
+                resolvePath(enclosingUri, dir, forWindows)
+        }
+      }
```
HT154: This could be significantly more concise, no?
```suggestion
      [[true]] {
        executable = resolveExecutable(enclosingUri, super.executable, forWindows)
        workingDirectory = resolvePath(enclosingUri, super.workingDirectory ?? "./", forWindows)
      }
```

stdlib/EvaluatorSettings.pkl
```diff
@@ -118,6 +123,105 @@ externalResourceReaders: Mapping<String, ExternalReader>?
...
+    hasSep =
+      if (forWindows)
+        path.contains(Regex(#"[/\\]"#))
+      else
+        path.contains("/")
+  )
+    if (hasSep)
+      resolvePath(base, path, forWindows)
+    else
+      path
```
HT154: Nit: concision/readability
```suggestion
local function resolveExecutable(base: String, path: String, forWindows: Boolean) =
  if (path.contains(if (forWindows) Regex(#"[/\\]"#) else "/"))
    resolvePath(base, path, forWindows)
  else
    path
```

stdlib/EvaluatorSettings.pkl
```diff
@@ -118,6 +123,105 @@ externalResourceReaders: Mapping<String, ExternalReader>?
...
+
+external local function resolvePath(
+  baseUri: String(startsWith("file:/")),
+  path: String,
+  forWindows: Boolean,
+): String
```
HT154, line 207: Am I correct in assuming the intent here is to replace this with a first-class path API at some point?

pkl-core/src/main/java/org/pkl/core/util/PathResolver.java
```diff
@@ -0,0 +1,194 @@
...
+        return getRoot(base) + path;
+      }
+      if (base.endsWith("\\")) {
+        return base + path;
+      }
+      return base + '\\' + path;
+    }
+
+    private boolean isDriveLetter(String path) {
+      return path.length() >= 2 && Character.isLetter(path.charAt(0)) && path.charAt(1) == ':';
```
HT154: `IoUtils.windowsDriveLetterLike` specifically looks for capital letters, but this appears to accept lowercase as well. Is that intentional?

HT154 (changes requested):

stdlib/EvaluatorSettings.pkl
bioball: Yeah, that's why this a local!

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/CliCommand.kt
bioball: For some reason, this inspection is not doing anything in my IJ. But, I went ahead and added a new rule for this.

pkl-core/src/main/java/org/pkl/core/util/PathResolver.java
bioball: Good catch!

stdlib/EvaluatorSettings.pkl
bioball: I think inlining the let expr reduces readability. But, we can definitely make this more concise by making the let expr a single line.

stdlib/EvaluatorSettings.pkl
bioball: Ah, good catch!

stdlib/EvaluatorSettings.pkl
```diff
@@ -118,6 +123,92 @@ externalResourceReaders: Mapping<String, ExternalReader>?
...
+    enclosingUri,
+    platform.current.operatingSystem,
+  )
+
+local function resolveExecutable(base: String, path: String, forWindows: Boolean) =
+  let (hasSep = if (forWindows) path.contains(Regex(#"[/\\]"#)) else path.contains("/"))
+    if (hasSep)
+      resolvePath(base, path, forWindows)
+    else
+      path
```
HT154: Not a blocker:
```suggestion
    if (hasSep) resolvePath(base, path, forWindows) else path
```

HT154 (approved):
LGTM after a rebase!

pkl-core/src/test/files/LanguageSnippetTests/output/api/project1.pcf
```diff
@@ -0,0 +1,25 @@
...
+        ["path"] {
+          executable = "/path/to/foo/executable"
+          workingDirectory = "/path/to"
+        }
+        ["command name"] {
+          executable = "foo"
+          workingDirectory = "/path/to"
+        }
+        ["file URI"] {
+          executable = "/path/to/file:/path/to/executable"
```
stackoverflow: This seems mangled. Shouldn't the output be `path/to/executable`?

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
```diff
@@ -55,7 +55,8 @@ class BaseOptions : OptionGroup() {
             if (IoUtils.isUriLike(moduleName)) URI(moduleName)
             // Can't just use URI constructor, because URI(null, null, "C:/foo/bar", null) turns
             // into `URI("C", null, "/foo/bar", null)`.
-            else if (IoUtils.isWindowsAbsolutePath(moduleName)) Path.of(moduleName).toUri()
+            else if (IoUtils.isWindows() && IoUtils.isWindowsAbsolutePath(moduleName))
```
stackoverflow, line 58: This looks correct. But `CommandSpecParser.java:1162` still calls the old `IoUtils.isWindowsAbsolutePath(moduleName)`. I imagine it should also be updated.

pkl-core/src/main/java/org/pkl/core/project/Project.java
```diff
@@ -301,9 +302,14 @@ public static Project parseProject(PObject module) throws URISyntaxException {
         getProperty(
             module,
             "evaluatorSettings",
-            (settings) ->
-                PklEvaluatorSettings.parse(
-                    (Value) settings, (it, name) -> resolveNullablePath(it, projectBaseUri, name)));
+            (settings) -> PklEvaluatorSettings.parse((Value) settings));
```
stackoverflow, line 305: This is a breaking change, right? We should add it to the changelog, because this is a public api.

docs/modules/release-notes/pages/0.32.adoc
```diff
@@ -21,11 +21,21 @@ XXX
...
-====
+=== Standard Library Changes
+
+==== `pkl:EvaluatorSettings`
+
+**Additions**
+
+* New method: link:{uri-stdlib-evaluatorSettingsModule}#resolve()[`EvaluatorSettings.resolve()`]
+* New method: link:{uri-stdlib-evaluatorSettingsModule}#resolveForOs()[`EvaluatorSettings.resolveForOs()`]
+* New property: link:{uri-stdlib-evaluatorSettingsExternalReaderClass}#workingDirectory[`EvaluatorSettings#ExternalReader.workingDir`]
```
stackoverflow: The uri is `workingDirectory` but the label is `workingDir`. It should be one or the other.
Noticed also that gradle is using `workingDir` but the rest of the code is using `workingDirectory`.

pkl-gradle/src/main/java/org/pkl/gradle/spec/ExternalReaderSpec.java
```diff
@@ -45,8 +46,14 @@ public String getName() {
   @Input
   public abstract ListProperty<String> getArguments();
 
+  @Input
+  @Optional
+  public abstract Property<String> getWorkingDir();
+
   public ExternalReader toExternalReader() {
-    return new ExternalReader(getExecutable().get(), getArguments().get());
+    var systemWorkingDir = System.getProperty("user.dir");
```
stackoverflow: nit: `user.dir` (I think) is gradle's cwd and may not be what the user expect. Also differs from the cli/project where working dir means something else. Though, not sure what to do here. Maybe just comment the discrepancy.

pkl-core/src/main/java/org/pkl/core/evaluatorSettings/PklEvaluatorSettings.java
```diff
@@ -127,6 +124,17 @@ public static PklEvaluatorSettings parse(
         traceMode == null ? null : TraceMode.valueOf(traceMode.toUpperCase()));
   }
 
+  private static Path parsePathOrUri(String input) {
+    if (IoUtils.isUriLike(input)) {
+      try {
+        return Path.of(new URI(input));
```
stackoverflow: nit: `Path.of` can throw IllegalArgumentException/FileSystemNotFoundException which won't be wrapped in a PklException.

stackoverflow (approved):
Some small stuff. Approving to unblock.

pkl-core/src/test/files/LanguageSnippetTests/output/api/project1.pcf
bioball: Actually this is an old test that I should get rid of!

This is output is expected; `file:///path/to/executable` is interpreted as a relative path with segments `file:`, `path`, `to`, `executable`

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseOptions.kt
bioball: Good catch!

pkl-core/src/main/java/org/pkl/core/evaluatorSettings/PklEvaluatorSettings.java
bioball: Ah, this is actually incorrect code; these settings don't accept URIs!

docs/modules/release-notes/pages/0.32.adoc
bioball: Great catch! We use "dir" everywhere else (`rootDir`, `projectDir`, etc). I'll change this to `workingDir`.

pkl-gradle/src/main/java/org/pkl/gradle/spec/ExternalReaderSpec.java
bioball: Good point; this should probably default to the Gradle project dir, not the PWD.

I think it's okay that this diverges from the CLI; it diverges but improves ergonomics for Gradle users.

PR #1411, opened 2026-01-21
Fix VmFunction toString

### LINKED ISSUE #1410: Incorrect Function.toString()
The default Java `toString()` is leaking through:

```pkl
local myFunc = () -> null

res = myFunc.toString()
```

Produces:

```
res = "org.pkl.core.runtime.VmFunction@3cab13d4"
```

bioball:
Fixes an issue where the underlying default Java toString() is leaking through

Closes #1410

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmValueRenderer.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmValueRenderer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


pkl-core/src/main/java/org/pkl/core/runtime/VmValueRenderer.java
```diff
@@ -34,8 +34,13 @@ public final class VmValueRenderer {
   private final String interiorNewline;
   private final String indent;
   private String currIndent = "";
+  private static final VmValueRenderer maxSingleLine =
+      new VmValueRenderer(Integer.MAX_VALUE, " ", "; ", "");
```
stackoverflow, line 38: Why do you need this special case renderer?

bioball: Optimization; a bunch of different nodes use this, so we can just have them share the same instance here.

PR #1412, opened 2026-01-21
Change how VmValueRenderer renders Typed values

bioball:
This addresses an issue where rendered values can be ambiguous. For example, a module gets rendered as `new ModuleClass {}`. Also, a class gets rendered as `new Person {}`, without specifying where that class comes from.

This changes it to render the qualified name instead of the simple name.

* Module classes render the module name
* Classes within module render the module name and the class name joined by `#`
* Rendering of members of `pkl:base` does not change

For example:

* `new k8s.core.v1.Service {}` instead of `new ModuleClass {}`.
* `new k8s.core.v1.Service#ServiceSpec {}` instead of `new ServiceSpec {}`
* `new pkl.Project {}` instead of `new ModuleClass {}`

This affects the following:

* `toString()`
* `trace()`
* Error messages
* REPL


HT154 (approved):
😍

stackoverflow (approved):
Nice!

PR #1419, opened 2026-01-30
Add flag to turn power assertions on/off

stackoverflow:
Power assertions have a big impact in the runtime performance of Pkl (~2x). This is negligible in most scenarios, but in some cases the performance lost is unacceptable.
This PR adds a flag to `pkl eval` and `pkl test`: `--power-assertions`.
~This flag is enabled by default for `pkl test` and disabled by default for `pkl eval`.~
Users can decide by themselves if they want it or not.

Edit: the flag is enabled by default, for both eval and test command.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/EvaluatorBuilder.java
+++ b/pkl-core/src/main/java/org/pkl/core/EvaluatorBuilder.java
@@ -70,7 +70,7 @@ public final class EvaluatorBuilder {
 
   private TraceMode traceMode = TraceMode.COMPACT;
 
-  private boolean powerAssertions = false;
+  private boolean powerAssertionsEnabled = false;
 
   private EvaluatorBuilder() {}
 
@@ -472,13 +472,13 @@ public final class EvaluatorBuilder {
 
   /** Sets whether power assertions are enabled. */
   public EvaluatorBuilder setPowerAssertionsEnabled(boolean powerAssertions) {
-    this.powerAssertions = powerAssertions;
+    this.powerAssertionsEnabled = powerAssertions;
     return this;
   }
 
   /** Returns whether power assertions are enabled. */
   public boolean getPowerAssertionsEnabled() {
-    return powerAssertions;
+    return powerAssertionsEnabled;
   }
 
   /**
@@ -592,6 +592,6 @@ public final class EvaluatorBuilder {
         dependencies,
         outputFormat,
         traceMode,
-        powerAssertions);
+        powerAssertionsEnabled);
   }
 }
--- a/pkl-core/src/main/java/org/pkl/core/project/Project.java
+++ b/pkl-core/src/main/java/org/pkl/core/project/Project.java
@@ -82,7 +82,7 @@ public final class Project {
       @Nullable java.time.Duration timeout,
       StackFrameTransformer stackFrameTransformer,
       Map<String, String> envVars,
-      boolean powerAssertions) {
+      boolean powerAssertionsEnabled) {
     try (var evaluator =
         EvaluatorBuilder.unconfigured()
             .setSecurityManager(securityManager)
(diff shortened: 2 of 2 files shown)


pkl-cli/src/main/kotlin/org/pkl/cli/commands/EvalCommand.kt
```diff
@@ -74,10 +74,23 @@ class EvalCommand : ModulesCommand(name = "eval", helpLink = helpLink) {
   private val testMode: Boolean by
     option(names = arrayOf("--test-mode"), help = "Internal test mode", hidden = true).flag()
 
+  private val powerAssertions: Boolean by
+    option(
+        names = arrayOf("--power-assertions"),
+        help = "Enable power assertions for detailed assertion failure messages.",
+      )
+      .flag()
```
HT154: To ease copy/paste of options between test and eval commands, it might be nice to also support `--no-power-assertions` here

```suggestion
      .flag("--no-power-assertions", default = false, defaultForHelp = "disabled")
```

pkl-core/src/main/java/org/pkl/core/runtime/VmContext.java
```diff
@@ -159,6 +162,10 @@ public TraceMode getTraceMode() {
     return holder.traceMode;
   }
 
+  public boolean getPowerAssertions() {
```
HT154: Nit: naming?
```suggestion
  public boolean getPowerAssertionsEnabled() {
```

HT154:
> Don't we better figure out why the performance drops and fix it? 🤔

Natch. We understand why it happens and are looking into how to best resolve it.

Either way, I think it's still valuable to have an option for this so that users with large test suites can disable power assertions for a performance boost.

pkl-cli/src/main/kotlin/org/pkl/cli/commands/EvalCommand.kt
```diff
@@ -74,10 +74,23 @@ class EvalCommand : ModulesCommand(name = "eval", helpLink = helpLink) {
   private val testMode: Boolean by
     option(names = arrayOf("--test-mode"), help = "Internal test mode", hidden = true).flag()
 
+  private val powerAssertions: Boolean by
```
HT154: Thoughts on keeping the var name consistent through the stack?

```suggestion
  private val powerAssertionsEnabled: Boolean by
```

HT154 (approved):
One comment re: naming, but approving to unblock

pkl-core/src/main/java/org/pkl/core/project/Project.java
```diff
@@ -81,7 +81,8 @@ public static Project loadFromPath(
       SecurityManager securityManager,
       @Nullable java.time.Duration timeout,
       StackFrameTransformer stackFrameTransformer,
-      Map<String, String> envVars) {
+      Map<String, String> envVars,
+      boolean powerAssertions) {
```
HT154: Nit: naming. Less critical since this isn't API, so feel free to ignore

pkl-core/src/main/java/org/pkl/core/runtime/VmLocalContext.java
```diff
@@ -19,6 +19,15 @@
 public class VmLocalContext {
   private boolean shouldEagerTypecheck = false;
 
+  /** Whether we are currently inside a type test ({@code is} check). */
+  private boolean inTypeTest = false;
```
HT154, line 23: Same here

pkl-core/src/main/java/org/pkl/core/EvaluatorBuilder.java
```diff
@@ -70,6 +70,8 @@ public final class EvaluatorBuilder {
 
   private TraceMode traceMode = TraceMode.COMPACT;
 
+  private boolean powerAssertions = false;
```
HT154: And here

pkl-core/src/main/java/org/pkl/core/runtime/VmLocalContext.java
stackoverflow: What's you issue with `inTypeTest`? The names already gives the idea of it being a boolean.

HT154: Uhh I think I mis-clicked on this line, but I can't figure out what my original intent was any more. Ignore this one 😓

PR #1426, opened 2026-02-09
Fix possible race condition during module and resource reading

stackoverflow:


pkl-core/src/main/java/org/pkl/core/module/ResolvedModuleKeys.java
```diff
@@ -86,6 +99,11 @@ public URI getUri() {
     @Override
     public String loadSource() throws IOException {
       try {
+        if (nofollow) {
+          try (InputStream in = Files.newInputStream(path, LinkOption.NOFOLLOW_LINKS)) {
```
HT154:
```suggestion
          try (var in = Files.newInputStream(path, LinkOption.NOFOLLOW_LINKS)) {
```

HT154 (approved):
One small thing, approving to unblock

PR #1429, opened 2026-02-18
Resolve variables at parse time

### LINKED ISSUE #1154: Resolve names at parse time
`AstBuilder` should perform the logic of resolving variables, rather than creating `ResolveVariableNode` and `ResolveMethodNode`.

This enables features such as:

* Flat member syntax (#423)
* User-facing varargs
* Local for-generator properties

stackoverflow:
This replaces `ResolveVariableNode` and `ResolveMethodNode` with their resolution. When we build the truffle node tree, we determine whether names resolve to:

* lexical scope
* base module
* implicit this

Then, we use this information to directly construct the underlying nodes (`ReadPropertyNode`, `ReadLocalPropertyNode`, etc).

Additionally, `AstBuilder` determines whether the property access must be const or not.

This introduces a `BaseModuleMembers` registry, which gets generated as part of Java compilation.

Closes https://github.com/apple/pkl/issues/1154

## Breakages

This introduces two known breakages.

### REPL

```
pkl0> foo { bar = baz }
pkl1> local baz = 3
pkl2> :f foo
new Dynamic {
  bar = 3
}
```

Now, `:f foo` produces an error, because `baz` is resolved to implicit this.

### Generator members

The following code will currently dynamically resolve to either `"outer"` or `"inner"` depending on whether `cond` evals to true or not. At parse time, we can't tell which level this should resolve to.

```pkl
qux = "outer"

foo {
  when (cond) {
    qux = "inner"
  }
  res = qux
}
```

In this PR, this will always resolve to the inner qux. This will throw if `cond` evals to false.


pkl-core/src/main/java/org/pkl/core/ast/builder/SymbolTable.java
```diff
@@ -394,42 +574,231 @@ public TypeParameterizableScope(
...
+        Identifier name, int levelUp, Function<MethodResolution, Object> callback) {
+      return null;
+    }
+  }
+
+  // A scope used only for variable resolution
+  public static final class ForEagerScope extends Scope {
+    private ForEagerScope(@Nullable Scope parent, String qualifiedName) {
+      super(parent, null, qualifiedName, ConstLevel.NONE, FrameDescriptor.newBuilder());
     }
```
HT154: Apologies if I'm being dense, but what does the name of this scope mean?

docs/modules/style-guide/pages/index.adoc
```diff
@@ -193,7 +193,7 @@ Too many lines separate `foo` and `baz`.
 Properties that override an existing property shouldn't have doc comments nor type annotations,
 unless the type is intentionally overridden via `extends`.
 
-[source%tested,{pkl}]
+[source%parsed,{pkl}]
```
bioball, line 196: Currently, `amends "..."` REPL statements don't do anything. In this PR, the header _still_ mostly does nothing, but it will throw if the amends has a relative path.

This is because our REPL logic has changed; before the AstBuilder can visit an expression, it first needs to visit the module to collect names. In the process of visiting the module, it will validate this module header.

pkl-core/src/main/java/org/pkl/core/ast/frame/ReadFrameSlotNode.java
```diff
@@ -15,12 +15,14 @@
...
 package org.pkl.core.ast.frame;
 
+import com.oracle.truffle.api.dsl.NodeChild;
 import com.oracle.truffle.api.dsl.Specialization;
 import com.oracle.truffle.api.frame.FrameSlotTypeException;
 import com.oracle.truffle.api.frame.VirtualFrame;
 import com.oracle.truffle.api.source.SourceSection;
 import org.pkl.core.ast.ExpressionNode;
 
+@NodeChild(value = "enclosingFrame", type = GetEnclosingFrameNode.class)
```
bioball, line 25: Optimization: move logic around finding the enclosing frame into its own node so that we don't keep re-executing the traversal logic every time we get a `FrameSlotTypeException`

pkl-core/src/main/java/org/pkl/core/repl/ReplServer.java
```diff
@@ -178,7 +183,132 @@ private List<ReplResponse> handleEval(Eval request) {
...
+      } else if (adjustedNode instanceof org.pkl.parser.syntax.ClassMethod classMethod) {
+        addModuleMethodDef(builder.visitClassMethod(classMethod));
+      } else if (adjustedNode instanceof ModuleDecl) {
+        // do nothing for now
+      } else {
+        results.add(
+            new ReplResponse.InternalError(new IllegalStateException("Unexpected parse result")));
+      }
+    } catch (VmException e) {
+      // TODO: patch stack trace for constants
```
bioball, line 286: This is an existing TODO, not newly introduced by this PR (GitHub isn't able to track that the code just got shuffled around).

pkl-core/src/main/java/org/pkl/core/runtime/FrameDescriptorBuilder.java
```diff
@@ -0,0 +1,83 @@
...
+import com.oracle.truffle.api.frame.FrameDescriptor;
+import com.oracle.truffle.api.frame.FrameSlotKind;
+import java.util.Arrays;
+import org.pkl.core.util.Nullable;
+
+/**
+ * A wrapper for Truffle's {@link FrameDescriptor.Builder}, but also lets us find the slot of a
+ * given {@link Identifier}.
+ */
+public class FrameDescriptorBuilder {
```
bioball, line 27: When we resolve frame slot variables, we need to know the names of each slot. The `FrameDescriptor.Builder` API keeps these names inside private fields. So, we use our own builder abstraction that exposes the slot names.

pkl-core/src/generator/java/org/pkl/core/generator/BaseModuleMemberRegistryGenerator.java
```diff
@@ -0,0 +1,131 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import javax.lang.model.element.Modifier;
+import org.pkl.parser.Parser;
+import org.pkl.parser.syntax.Modifier.ModifierValue;
+
+public final class BaseModuleMemberRegistryGenerator {
```
bioball: This generates a class called `BaseModuleMembers`, and gets called as part of java compilation.

This generates a class that looks something like:

```java
public final class BaseModuleMembers {
  private BaseModuleMembers() {
  }

  public static boolean hasProperty(String name) {
    return switch (name) {
      case "AlsoKnownAs",
        "Annotation",
        "Any",
        "BaseValueRenderer",
        "Boolean",
        "Bytes",
        "BytesRenderer",
        "Char",
        "Charset",
        "Class",
        "Collection",
        "Comparable",
        "ConvertProperty",
        // etc
        "YamlRenderer" -> true;
      default -> false;
    }
  }
}
```

This is used by `SymbolTable` to figure out if some name is defined on the base module or not.
This is much faster than looking up members on `BaseModule.getModule()`.

pkl-core/src/main/java/org/pkl/core/ast/expression/member/ReadAmbiguousLocalityPropertyNode.java
```diff
@@ -0,0 +1,83 @@
...
+import com.oracle.truffle.api.frame.VirtualFrame;
+import com.oracle.truffle.api.source.SourceSection;
+import org.pkl.core.ast.ExpressionNode;
+import org.pkl.core.ast.MemberLookupMode;
+import org.pkl.core.ast.expression.primary.GetEnclosingReceiverNode;
+import org.pkl.core.ast.expression.primary.GetReceiverNode;
+import org.pkl.core.runtime.Identifier;
+import org.pkl.core.runtime.VmUtils;
+
+public class ReadAmbiguousLocalityPropertyNode extends ExpressionNode {
```
stackoverflow: It would be nice to have some doc comments here explaining why this node exists.

pkl-core/src/main/java/org/pkl/core/runtime/VmExceptionRenderer.java
```diff
@@ -60,6 +60,7 @@ private void renderBugException(VmBugException exception, AnsiStringBuilder out)
     // if a cause exists, it's more useful to report just that
     var exceptionToReport = exception.getCause() != null ? exception.getCause() : exception;
     var exceptionUrl = URLEncoder.encode(exceptionToReport.toString(), StandardCharsets.UTF_8);
+    System.out.println(exception.getCause().toString());
```
stackoverflow: Leftover.
```suggestion
```

pkl-core/src/main/java/org/pkl/core/runtime/VmObjectLike.java
```diff
@@ -82,6 +82,42 @@ public boolean isModuleObject() {
   /** Returns the declared members of this object. */
   public abstract UnmodifiableEconomicMap<Object, ObjectMember> getMembers();
 
+  /**
+   * Tells if this it's impossible to determine if this is object describes a scope during parse
```
stackoverflow:
```suggestion
   * Tells if it's impossible to determine if this is object describes a scope during parse
```

pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
```diff
@@ -159,6 +163,40 @@ public static VmObjectLike getOwner(Frame frame) {
...
 
+  public static VmObjectLike getOwner(VirtualFrame frame, int levelsUp) {
+    return getOwner(getFrame(frame, levelsUp));
+  }
+
+  public static Object getReceiver(VirtualFrame frame, int levelsUp) {
+    return getReceiver(getFrame(frame, levelsUp));
+  }
+
+  @ExplodeLoop
```
stackoverflow: nit: I don't think truffle can explode a dynamic loop (`isParseTimeInvisibleScope` makes the loop dynamic). It's not wrong to leave it there, but truffle will probably give up and do nothing, so the annotation is misleading.

stackoverflow (commented):
@bioball Great work!

pkl-core/src/main/java/org/pkl/core/runtime/VmExceptionRenderer.java
bioball: Whoops!

pkl-core/src/main/java/org/pkl/core/ast/expression/primary/GetEnclosingOwnerNode.java
```diff
@@ -31,13 +31,6 @@ public GetEnclosingOwnerNode(int levelsUp) {
 
   @ExplodeLoop
   public Object executeGeneric(VirtualFrame frame) {
```
HT154: There's no longer a loop here to explode!
```suggestion
public Object executeGeneric(VirtualFrame frame) {
```

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -1024,9 +1161,18 @@ public ExpressionNode visitFunctionLiteralExpr(FunctionLiteralExpr expr) {
           .build();
     }
 
+    for (var par : params.getParameters()) {
+      if (par instanceof TypedIdentifier ti) {
+        bindings.add(ti.getIdentifier().getValue());
+      } else {
+        bindings.add("_");
+      }
+    }
```
HT154: This pattern gets repeated a few times, maybe a helper method is in order? Something like this:

```java
    private static void bindParameter(List<String> bindings, Parameter param) {
        bindings.add(param instanceof Parameter.TypedIdentifier id ? d.getIdentifier().getValue() : "_");
    }

    private static void bindParameters(List<String> bindings, List<Parameter> params) {
        for (var param : params) {
             addParameter(bindings, param)
        }
    }
```

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -1511,31 +1674,52 @@ public ObjectMember visitImportClause(ImportClause imp) {
...
     return symbolTable.enterClass(
         className,
+        modifiers,
         typeParameters,
         scope -> {
+          var bodyNode = clazz.getBody();
+
+          List<ClassProperty> properties = bodyNode != null ? bodyNode.getProperties() : List.of();
+          List<ClassMethod> methods = bodyNode != null ? bodyNode.getMethods() : List.of();
+          registerClassScopeNames(scope, properties, methods);
```
HT154, line 1715: Nit: readability

```suggestion
          if (bodyNode != null) {
            registerClassScopeNames(scope, bodyNode.getProperties(), bodyNode.getMethods());
          } else {
            registerClassScopeNames(scope, List.of(), List.of());
          }
```

HT154 (approved):
Very nice!! Just a few small nits from me.

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
bioball: `properties` and `methods` are used elsewhere in this method, alas

PR #1432, opened 2026-02-19
Command flag behavior improvements

HT154:
* Forbid overlap of built-in and command-defined flag names 
* Allow interleaving built-in and command-defined flags on the command line
* List abbreviated flag names first, matching the behavior of built-in flags


pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/BaseCommand.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
```
stackoverflow: Revert?

PR #1434, opened 2026-02-20
Allow referring to remote project dependencies on the CLI with @-notation

HT154:
* Disable Clikt [`@argfiles`](https://ajalt.github.io/clikt/advanced/#command-line-argument-files-argfiles) for all CLI commands.
  * This is a breaking change, but imo it's okay since we don't document this behavior.
  * Without this, all `@`-notation usage would require escaping, e.g. `@@foo/bar.pkl`
* Add `CliCommand.resolvedSourceModules`, which resolves any `@dep` URIs to their underlying package asset URI.
  * Only remote deps are supported since local deps would require more invasive project handling.
  * Dependency lock state is ignored, only the package URI present in the PklProject itself is used.
* Switch usage of `CliBaseOptions.normalizedSourceModules` in commands to use `CliCommand.resolvedSourceModules`.

### DIFF
--- a/pkl-cli/src/main/kotlin/org/pkl/cli/CliCommandRunner.kt
+++ b/pkl-cli/src/main/kotlin/org/pkl/cli/CliCommandRunner.kt
@@ -52,8 +52,6 @@ constructor(
   private val errStream: OutputStream = System.err,
 ) : CliCommand(options) {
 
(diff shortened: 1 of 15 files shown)


Rijal95:
Y

PR #1445, opened 2026-02-26
Add `starlark` output format

fmeum:
TODO:

- [x] Add support for `Pair` (tuple)
- [x] Insert empty lines around top-level statements
- [x] Sort keyword arguments
- [x] Find a way to annotate classes with a Bazel `load` label, collect the labels for all classes in the module and emit them at the top


HT154 (commented):
Hi there! We generally encourage renderer development to be done in Pkl code to avoid increasing the API surface and maintenance cost of the core language. Pkl code may subclass `ValueRenderer` and directly implement its methods. This is how the TOML renderer is implemented: https://github.com/apple/pkl-pantry/tree/main/packages/pkl.toml. It would be much easier for us to accept a contribution of that nature.

PR #1460, opened 2026-03-02
Bump pkl.impl.ghactions to version 1.5.0

HT154:
Updates pkl.impl.ghactions package to version 1.5.0

### DIFF
--- a/.github/jobs/BuildNativeJob.pkl
+++ b/.github/jobs/BuildNativeJob.pkl
@@ -12,7 +12,7 @@ project: String
 extraGradleArgs {
   when (os == "macOS" && arch == "amd64") {
     "-Dpkl.targetArch=\(module.arch)"
-    "-Dpkl.native--native-compiler-path=\(context.github.workspace)/.github/scripts/cc_macos_amd64.sh"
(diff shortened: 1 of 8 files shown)


.github/jobs/BuildNativeJob.pkl
```diff
@@ -46,6 +46,9 @@ steps {
   new {
     name = "gradle buildNative"
     shell = "bash"
+    env {
```
madrob: I don't think we need this, I think it is automatically set for us by GitHub. https://docs.github.com/en/actions/reference/workflows-and-actions/variables

.github/jobs/BuildNativeJob.pkl
```diff
@@ -12,7 +12,7 @@ project: String
 extraGradleArgs {
   when (os == "macOS" && arch == "amd64") {
     "-Dpkl.targetArch=\(module.arch)"
-    "-Dpkl.native--native-compiler-path=\(context.github.workspace)/.github/scripts/cc_macos_amd64.sh"
+    "-Dpkl.native--native-compiler-path=${GITHUB_WORKSPACE}/.github/scripts/cc_macos_amd64.sh"
```
madrob:
```suggestion
    #"-Dpkl.native--native-compiler-path="${GITHUB_WORKSPACE}/.github/scripts/cc_macos_amd64.sh""#
```

PR #1462, opened 2026-03-05
Do not activate power assertions when a single union member containing a type constraint fails

HT154:
Prior to this change, this code would activate powers assertions / instrumentation permanently:
```pkl
foo: String(contains("a")) | String(contains("b")) = "boo"
```

This is because the `contains("a")` constraint would fail, triggering power assertions, but the subsequent check of the union's `contains("b")` branch would succeed.
As observed in #1419, once instrumentation is enabled, all subsequent evaluation slows significantly.
As with #1419, the fix here is to disable power assertions via `VmLocalContext` until we know that all union members failed type checking; then, each member is re-executed with power assertions allowed to provide the improved user-facing error.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/EvaluatorImpl.java
+++ b/pkl-core/src/main/java/org/pkl/core/EvaluatorImpl.java
@@ -337,10 +337,10 @@ public final class EvaluatorImpl implements Evaluator {
   }
 
   // for use in tests to determine whether an evaluator ever triggered instrumentation
-  boolean wasInstrumentationUsed() {
+  boolean isInstrumentationEverUsed() {
     polyglotContext.enter();
     try {
-      return VmLanguage.get(null).localContext.get().getInstrumentationUsed();
+      return VmLanguage.get(null).localContext.get().isInstrumentationEverUsed();
     } finally {
       polyglotContext.leave();
     }
--- a/pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
@@ -987,7 +987,7 @@ public abstract class TypeNode extends PklNode {
 
       // disallow power assertions from triggering in case one union member checks successfully
       var localContext = VmLanguage.get(this).localContext.get();
-      boolean wasInTypeTest = localContext.isInTypeTest();
+      var wasInTypeTest = localContext.isInTypeTest();
       localContext.setInTypeTest(true);
 
       // Do eager checks (shallow-force) if there are two listings or two mappings represented.
@@ -1039,7 +1039,7 @@ public abstract class TypeNode extends PklNode {
 
       // disallow power assertions from triggering in case one union member checks successfully
       var localContext = VmLanguage.get(this).localContext.get();
-      boolean wasInTypeTest = localContext.isInTypeTest();
+      var wasInTypeTest = localContext.isInTypeTest();
       localContext.setInTypeTest(true);
 
       for (var i = 0; i < elementTypeNodes.length; i++) {
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmLocalContext.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmLocalContext.java
@@ -28,7 +28,7 @@ public class VmLocalContext {
    */
   private int activeTrackerDepth = 0;
 
-  private boolean instrumentationUsed = false;
+  private boolean instrumentationEverUsed = false;
 
   public VmLocalContext() {}
 
@@ -50,7 +50,7 @@ public class VmLocalContext {
 
   public void enterTracker() {
     activeTrackerDepth++;
-    instrumentationUsed = true;
+    instrumentationEverUsed = true;
   }
 
   public void exitTracker() {
@@ -61,7 +61,7 @@ public class VmLocalContext {
     return activeTrackerDepth > 0;
   }
 
-  public boolean getInstrumentationUsed() {
-    return instrumentationUsed;
+  public boolean isInstrumentationEverUsed() {
+    return instrumentationEverUsed;
   }
 }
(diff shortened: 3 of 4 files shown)


pkl-core/src/main/java/org/pkl/core/EvaluatorImpl.java
```diff
@@ -336,6 +336,17 @@ public void close() {
     }
   }
 
+  // for use in tests to determine whether an evaluator ever triggered instrumentation
+  // public here so it's accessible in tests, but not in the Evaluator interface
+  public boolean getInstrumentationUsed() {
```
HT154: Not in love with the need for this method to be able to test this. Open to suggestions for alternatives!

holzensp: Java's default visibility (no modifiers) is package-private. The test lives in the same package `org.pkl.core`, so I'd...
```suggestion
// for use in tests to determine whether an evaluator ever triggered instrumentation
boolean getInstrumentationUsed() {
```

I'm a big proponent of making things that shouldn't be used awkward to use. Longer names do that. I'd consider documenting entirely in the name; `wasInstrumentationUsedDuringTest()` maybe?

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -1011,14 +1037,36 @@ public Object executeEagerly(VirtualFrame frame, Object value) {
...
+      if (VmContext.get(this).getPowerAssertionsEnabled()
+          && (!wasInTypeTest || localContext.hasActiveTracker())) {
+        for (var i = 0; i < elementTypeNodes.length; i++) {
+          try {
+            elementTypeNodes[i].executeEagerly(frame, value);
+          } catch (VmTypeMismatchException e) {
+            typeMismatches[i] = e;
+          }
+        }
+      }
```
holzensp, line 1068: My spidey senses tingle at the lack of DRYness. I don't want to insist on DRYness per se, but would like to ask the question out loud; worth a `private static` helper or similar.

HT154: This is certainly possible, but I'm hesitant to do it without more context here, i.e. why does `executeLazily` use `@ExplodeLoop` but `executeEagerly` does not? Was this not already DRY'd up because of some optimization?

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -985,22 +985,48 @@ protected Object executeLazily(VirtualFrame frame, Object value) {
       // escape analysis should remove this allocation in compiled code
       var typeMismatches = new VmTypeMismatchException[elementTypeNodes.length];
 
+      // disallow power assertions from triggering in case one union member checks successfully
+      var localContext = VmLanguage.get(this).localContext.get();
+      boolean wasInTypeTest = localContext.isInTypeTest();
```
stackoverflow:
```suggestion
      var wasInTypeTest = localContext.isInTypeTest();
```
nit

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -985,22 +985,48 @@ protected Object executeLazily(VirtualFrame frame, Object value) {
...
-          }
+          var result =
+              shouldEagerCheck
+                  ? elementTypeNode.executeEagerly(frame, value)
+                  : elementTypeNode.executeLazily(frame, value);
+          localContext.setInTypeTest(wasInTypeTest);
+          return result;
         } catch (VmTypeMismatchException e) {
           typeMismatches[i] = e;
         }
```
stackoverflow, line 1008:
```suggestion
          var result =
              shouldEagerCheck
                  ? elementTypeNode.executeEagerly(frame, value)
                  : elementTypeNode.executeLazily(frame, value);
          return result;
       } catch (VmTypeMismatchException e) {
         typeMismatches[i] = e;
       } finally {
         localContext.setInTypeTest(wasInTypeTest);
      }
```
I know it's the same thing in practice, but I think using a finally conveys the intent better.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -1011,14 +1037,36 @@ public Object executeEagerly(VirtualFrame frame, Object value) {
       // escape analysis should remove this allocation in compiled code
       var typeMismatches = new VmTypeMismatchException[elementTypeNodes.length];
 
+      // disallow power assertions from triggering in case one union member checks successfully
+      var localContext = VmLanguage.get(this).localContext.get();
+      boolean wasInTypeTest = localContext.isInTypeTest();
```
stackoverflow:
```suggestion
      var wasInTypeTest = localContext.isInTypeTest();
```
nit

pkl-core/src/main/java/org/pkl/core/runtime/VmLocalContext.java
```diff
@@ -57,4 +60,8 @@ public void exitTracker() {
   public boolean hasActiveTracker() {
     return activeTrackerDepth > 0;
   }
+
+  public boolean getInstrumentationUsed() {
```
stackoverflow:
```suggestion
  public boolean isInstrumentationUsed() {
```
Normally we use boolean prefixes like is/should/has for boolean getters.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
HT154: This is meaningfully different (and does not work) because the `finally` block runs after the `catch` too. In this case, we only want to run it in the case that the member type check succeeds. If it were in the `finally` block, `inTypeTest` would be reverted after just the first member type of the union is checked and the bug is not fixed.

I've improved the tests to exercise this scenario to show that making this suggested change does activate instrumentation erroneously.

PR #1466, opened 2026-03-09
Prevent I/O when checking UNC paths against `--root-dir`

HT154:
Test on [windows] please

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/SecurityManager.java
+++ b/pkl-core/src/main/java/org/pkl/core/SecurityManager.java
@@ -52,6 +52,7 @@ public interface SecurityManager {
    * <p>Returns {@code null} for non-{@code file:} URIs or if no root directory is configured.
    *
    * @param uri the URI to resolve
+   * @param isResource denotes if uri belongs to a resource (otherwise, a module)
    * @return the resolved, symlink-free path under root directory, or {@code null}
    * @throws SecurityManagerException if the resolved path is not within the root directory
    * @throws IOException if the path cannot be resolved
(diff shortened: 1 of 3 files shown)


pkl-core/src/main/java/org/pkl/core/SecurityManagers.java
```diff
@@ -173,14 +173,21 @@ public void checkImportModule(URI importingModule, URI importedModule)
...
+          || !uri.getAuthority().isEmpty()) {
         return null;
       }
       var path = Path.of(uri);
       var realPath = path.toRealPath();
       if (!realPath.startsWith(rootDir)) {
-        throw new SecurityManagerException(ErrorMessages.create("modulePastRootDir", uri, rootDir));
+        var errorMessageKey = isResource ? "resourcePastRootDir" : "modulePastRootDir";
+        var message = ErrorMessages.create(errorMessageKey, uri, rootDir);
+        throw new SecurityManagerException(message);
```
HT154, line 189: This fixes an incorrect error message (referencing "module") when reading a resource.

pkl-core/src/main/java/org/pkl/core/SecurityManager.java
stackoverflow, line 56: We should document `isResource` here.

pkl-core/src/main/java/org/pkl/core/SecurityManager.java
```diff
@@ -56,7 +56,8 @@ public interface SecurityManager {
    * @throws SecurityManagerException if the resolved path is not within the root directory
    * @throws IOException if the path cannot be resolved
    */
-  default @Nullable Path resolveSecurePath(URI uri) throws SecurityManagerException, IOException {
+  default @Nullable Path resolveSecurePath(URI uri, boolean isResource)
```
stackoverflow, line 60: This is a breaking change in the API. We could leave the old single parameter function around and just call the new one, preventing the breakage.

HT154: Added a new default method that calls this one with `isResource` set to false

PR #1467, opened 2026-03-11
Prevent `--multiple-file-output-path` writes from following symlinks outside the target directory

HT154:


pkl-cli/src/main/kotlin/org/pkl/cli/CliEvaluator.kt
```diff
@@ -269,4 +268,22 @@ constructor(
       }
     }
   }
+
+  /**
+   * Resolves [rel] against this Path name-by-name. At each step, the real path is resolved if the
+   * file exists. The normalized real path and normalized resolved path are returned. This has the
+   * same effect as `this.resolve(rel).toRealPath().normalize()`, except that the real paths account
```
stackoverflow: The doc is a bit misleading:
```
This has the same effect as `this.resolve(rel).toRealPath().normalize()`
```
`toRealpath` would throw if the resolved path doesn't exist.

PR #1471, opened 2026-03-20
Allow custom/external resources to be "not found"

HT154:
This allows custom/external resources to produce `null` values for nullable reads (`read?`)

Ref: https://github.com/apple/pkl-go/issues/157


pkl-core/src/main/java/org/pkl/core/externalreader/ExternalResourceResolverImpl.java
```diff
@@ -48,15 +49,14 @@ final class ExternalResourceResolverImpl implements ExternalResourceResolver {
 
   public Optional<Object> read(URI uri) throws IOException {
     var result = doRead(uri);
-    return Optional.of(new Resource(uri, result));
+    return result == null ? Optional.empty() : Optional.of(new Resource(uri, result));
```
HT154, line 52: The `contents` field is nullable in the underlying `ReadResourceResponse` message, but when it was null we previously substituted an empty result. This exposes the that null value to the resource reader API as an empty result, allowing nullable reads to work.

This meaningfully changes existing behavior. Bound languages may still send an empty array instead of null in the message to yield an empty resource.

PR #1486, opened 2026-04-01
Fix data race in MessagePack encoder for concurrent server sends

ldaley:
## Summary

- The pkl server's main thread (handling `CreateEvaluatorRequest`) and executor thread (sending `EvaluateResponse`/`ReadModuleRequest`) both call `AbstractMessagePackEncoder.encode()`
concurrently.
  Without synchronization their writes to the shared `MessagePacker` interleave, corrupting the MessagePack stream on stdout.
- Add `synchronized(packer)` around the encode-and-flush sequence in `AbstractMessagePackEncoder.encode()`.
- Add a regression test that exercises the race directly: two threads write different message types through the same encoder into a pipe while a reader thread decodes every message.
  Without the fix the test fails reliably with decode errors; with the fix all 4000 messages decode cleanly.

## How the race manifests

Clients that create new evaluators while existing evaluations are in progress (the bidirectional `ReadModuleRequest`/`ReadModuleResponse` callback flow) see corrupted MessagePack frames on
the server's stdout — wrong array headers, truncated maps, or bytes from one message spliced into another.

This was diagnosed by driving `pkl server` directly from a Python stress test that registers a custom module reader and continuously creates evaluators while evaluations with module-read
callbacks are active.
Unidirectional evaluations (inline `moduleText`, no callbacks) are unaffected because the executor thread is the only writer.

## Test plan

- [ ] `./gradlew pkl-server:test --tests "org.pkl.server.JvmServerTest.concurrent encoding*"` passes with the fix
- [ ] Same test fails reliably when the `synchronized` block is removed

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/messaging/AbstractMessagePackEncoder.java
+++ b/pkl-core/src/main/java/org/pkl/core/messaging/AbstractMessagePackEncoder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


pkl-server/src/test/kotlin/org/pkl/server/AbstractServerTest.kt
```diff
@@ -839,6 +841,101 @@ abstract class AbstractServerTest {
...
+    val pipeIn = PipedInputStream(1 shl 20) // 1 MB buffer
+    val pipeOut = PipedOutputStream(pipeIn)
+    val encoder = ServerMessagePackEncoder(pipeOut)
+    val decoder = ServerMessagePackDecoder(pipeIn)
+
+    val iterations = 2000
+    val padding = ByteArray(8192) // large payload to widen the race window
+    val errors = mutableListOf<Throwable>()
+    val decoded = java.util.concurrent.atomic.AtomicInteger(0)
+    val done = java.util.concurrent.CountDownLatch(2)
```
HT154: These should be imported directly:

```suggestion
    val decoded = AtomicInteger(0)
    val done = CountDownLatch(2)
```

HT154 (approved):
LGTM once CI is passing

pkl-server/src/test/kotlin/org/pkl/server/AbstractServerTest.kt
ldaley: Done.

PR #1500, opened 2026-04-06
feat(pkl-gradle): Implement gradle configuration cache support

### LINKED ISSUE #1425: Support the Gradle Configuration Cache with `pkl-gradle`
## Problem

The `pkl-gradle` plugin is incompatible with the [Gradle configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html). Running any pkl task with `--configuration-cache` fails for a variety of reasons. As a team adopting PKL in a large multi-project build, we'd like to fully utilize the configuration cache to reduce the amount of configuration time (currently 3-4 min for some builds!) we're working against.

### Reproduction

```bash
./gradlew evalTest --configuration-cache
```

Fails with:
> Configuration cache state could not be cached: field `__project__` of task `:evalTest` of type `BasePklTask`: cannot serialize object of type `org.gradle.api.internal.project.DefaultProject`, as these are not supported with the configuration cache.
> ...

## Proposed Solution

Address the set of core issues related to the plugin that are incompatible today with the configuration cache

1. **Tasks access `Project` at execution time.** `BasePklTask.getCliBaseOptions()` and `ModulesTask.getCliBaseOptions()` call `getProject().getProjectDir()` inside `@TaskAction` methods. The `Project` object is not serializable and must not be referenced during task execution when the configuration cache is active.
2. **Tasks hold non-serializable mutable fields.** `BasePklTask` stores a `CliBaseOptions cachedOptions` field and `ModulesTask` stores a `HashMap parsedSourceModulesCache` — neither is serializable by the configuration cache.
3. **`JavaCodeGenTask` and `KotlinCodeGenTask` call `getProject().file()`** inside `doRunTask()` to resolve the output directory.
4. **`ModulesTask.getSourceModuleFiles()` calls `getProject().files()`** to create a `FileCollection`, embedding a project reference in the task's input tracking.
5. **`PklPlugin` uses `project.afterEvaluate()`**, which is disallowed by the configuration cache.
6. **`PklPlugin` stores `Project` as an instance field**, which would be captured if the plugin class were ever serialized.

We can also implement a `ConfigurationCacheTest` class to provide regression coverage. It will use `GradleRunner` in forked mode with `--configuration-cache`, runs the task twice, and assert the cache is stored on the first run and reused on the second. Any future change that introduces a configuration cache incompatibility will cause these tests to fail.

ffluk3:
Fixes #1425 

Gradle's [configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html) is one of the more impactful performance features in modern Gradle. It serializes the task graph after the first configuration phase so subsequent builds can skip it entirely. For large projects this can mean the difference between a multi-second configuration overhead and near-instant task scheduling.

The root cause preventing `pkl-gradle` from adopting the configuration cache was a handful of patterns the configuration cache explicitly prohibits: holding a reference to the `Project` object in a plugin field, accessing `getProject()` inside task action code at execution time, and caching CLI option objects that close over non-serializable state. These patterns work fine in a standard Gradle build, but they prevent Gradle from safely serializing the task graph to disk.

### PKL Plugin rework

The most visible structural change in `PklPlugin` is the removal of the `@LateInit private Project project` field. Rather than storing the project on the plugin instance, it's now passed as a parameter through the configure call chain. This is a mechanical but pervasive refactor, since every `configure*` method gains a `project` argument and passes via property drilling.

### Per-task changes

- `BasePklTask` captures `project.getLayout().getProjectDirectory()` at configuration time into a `getWorkingDir()` `DirectoryProperty`, rather than calling `getProject().getProjectDir()` at execution time. `CliBaseOptions` is now constructed fresh each time it's needed inside a task action, which is a trade-off between the (potentially negligible) cost of construction, and caching it, which would require holding a cross-boundary reference that defeats the configuration cache.
- `EvalTask`'s lazily-cached `Provider<CliEvaluator>` is similarly replaced with a factory method, and `JavaCodeGenTask` and `KotlinCodeGenTask` leverage the output dir.
- In `ModulesTask`, `getTransitiveModules()` changes from a `ListProperty<File>` to a `ConfigurableFileCollection` (enabling `.from()` wiring rather than `.set()`), and gains a `@PathSensitive(ABSOLUTE)` annotation so the task doesn't unnecessarily re-run when files are moved between machines but their content and paths are unchanged. The `parsedSourceModulesCache` map is also removed — it was there to avoid redundant parsing but is incompatible with the cache since it holds mutable state across invocations.

### Cleanup

The `AnalyzeImportsTask` no longer wraps `CliImportAnalyzer` construction in a lazily-cached `Provider`; instead it constructs the analyzer inline inside `doRunTask()`, consistent with the other task types. The transitive-import parsing logic in `PklPlugin` and `PluginUtils` is unchanged.

### Tests

Each task type gets a new `is configuration cache compatible` test that runs the task twice with `--configuration-cache` and asserts that the first run stores a cache entry and the second reuses it. The test infrastructure for this lives in a new `runTaskWithConfigurationCache` helper on `AbstractTest`. NOTE: these tests run slightly slower than the rest of the suite because they need the forked Gradle process to execute twice to prove out configuraion cache usage.

### DIFF
--- a/pkl-gradle/src/main/java/org/pkl/gradle/PklPlugin.java
+++ b/pkl-gradle/src/main/java/org/pkl/gradle/PklPlugin.java
@@ -16,8 +16,6 @@
 package org.pkl.gradle;
 
 import java.io.File;
-import java.nio.file.Files;
-import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
@@ -38,7 +36,6 @@ import org.gradle.plugins.ide.idea.model.IdeaModel;
 import org.gradle.util.GradleVersion;
 import org.jspecify.annotations.Nullable;
 import org.pkl.cli.CliEvaluatorOptions;
-import org.pkl.core.ImportGraph;
 import org.pkl.core.OutputFormat;
 import org.pkl.core.util.IoUtils;
 import org.pkl.gradle.spec.AnalyzeImportsSpec;
@@ -70,45 +67,38 @@ public class PklPlugin implements Plugin<Project> {
 
   private static final String MIN_GRADLE_VERSION = "8.2";
 
-  private @Nullable Project __project;
-
   @Override
   public void apply(Project project) {
-    __project = project;
 
     if (GradleVersion.current().compareTo(GradleVersion.version(MIN_GRADLE_VERSION)) < 0) {
       throw new GradleException(
           String.format("Plugin `org.pkl` requires Gradle %s or higher.", MIN_GRADLE_VERSION));
     }
 
-    var extension = project().getExtensions().create("pkl", PklExtension.class);
-    configureExtension(extension);
+    var extension = project.getExtensions().create("pkl", PklExtension.class);
+    configureExtension(project, extension);
   }
 
-  private Project project() {
-    assert __project != null;
-    return __project;
+  private void configureExtension(Project project, PklExtension extension) {
+    configureEvalTasks(project, extension.getEvaluators());
+    configureJavaCodeGenTasks(project, extension.getJavaCodeGenerators());
+    configureKotlinCodeGenTasks(project, extension.getKotlinCodeGenerators());
+    configurePkldocTasks(project, extension.getPkldocGenerators());
+    configureTestTasks(project, extension.getTests());
+    configureProjectPackageTasks(project, extension.getProject().getPackagers());
+    configureProjectResolveTasks(project, extension.getProject().getResolvers());
+    configureAnalyzeImportsTasks(project, extension.getAnalyzers().getImports());
   }
 
-  private void configureExtension(PklExtension extension) {
-    configureEvalTasks(extension.getEvaluators());
-    configureJavaCodeGenTasks(extension.getJavaCodeGenerators());
-    configureKotlinCodeGenTasks(extension.getKotlinCodeGenerators());
-    configurePkldocTasks(extension.getPkldocGenerators());
-    configureTestTasks(extension.getTests());
-    configureProjectPackageTasks(extension.getProject().getPackagers());
-    configureProjectResolveTasks(extension.getProject().getResolvers());
-    configureAnalyzeImportsTasks(extension.getAnalyzers().getImports());
-  }
-
-  private void configureProjectPackageTasks(NamedDomainObjectContainer<ProjectPackageSpec> specs) {
+  private void configureProjectPackageTasks(
+      Project project, NamedDomainObjectContainer<ProjectPackageSpec> specs) {
     specs.all(
         spec -> {
-          configureBaseSpec(spec);
+          configureBaseSpec(project, spec);
           spec.getOutputPath()
-              .convention(project().getLayout().getBuildDirectory().dir("generated/pkl/packages"));
+              .convention(project.getLayout().getBuildDirectory().dir("generated/pkl/packages"));
           spec.getOverwrite().convention(false);
-          var packageTask = createTask(ProjectPackageTask.class, spec);
+          var packageTask = createTask(project, ProjectPackageTask.class, spec);
           packageTask.configure(
               task -> {
                 task.getProjectDirectories().from(spec.getProjectDirectories());
@@ -117,12 +107,12 @@ public class PklPlugin implements Plugin<Project> {
                 task.getJunitReportsDir().set(spec.getJunitReportsDir());
                 task.getOverwrite().set(spec.getOverwrite());
               });
-          project()
+          project
               .getPluginManager()
               .withPlugin(
                   "base",
                   appliedPlugin ->
-                      project()
+                      project
                           .getTasks()
                           .named(
                               LifecycleBasePlugin.BUILD_TASK_NAME,
@@ -130,38 +120,40 @@ public class PklPlugin implements Plugin<Project> {
         });
   }
 
-  private void configureProjectResolveTasks(NamedDomainObjectContainer<ProjectResolveSpec> specs) {
+  private void configureProjectResolveTasks(
+      Project project, NamedDomainObjectContainer<ProjectResolveSpec> specs) {
     specs.all(
         spec -> {
-          configureBaseSpec(spec);
-          var resolveTask = createTask(ProjectResolveTask.class, spec);
+          configureBaseSpec(project, spec);
+          var resolveTask = createTask(project, ProjectResolveTask.class, spec);
           resolveTask.configure(
               task -> task.getProjectDirectories().from(spec.getProjectDirectories()));
         });
   }
 
-  private void configureAnalyzeImportsTasks(NamedDomainObjectContainer<AnalyzeImportsSpec> specs) {
+  private void configureAnalyzeImportsTasks(
+      Project project, NamedDomainObjectContainer<AnalyzeImportsSpec> specs) {
     specs.all(
         spec -> {
-          configureBaseSpec(spec);
+          configureBaseSpec(project, spec);
           spec.getOutputFormat().convention(OutputFormat.PCF.toString());
-          var analyzeImportsTask = createTask(AnalyzeImportsTask.class, spec);
+          var analyzeImportsTask = createTask(project, AnalyzeImportsTask.class, spec);
           analyzeImportsTask.configure(
               task -> {
                 task.getOutputFormat().set(spec.getOutputFormat());
                 task.getOutputFile().set(spec.getOutputFile());
-                configureModulesTask(task, spec, null);
+                configureModulesTask(project, task, spec, null);
               });
         });
   }
 
-  private void configureEvalTasks(NamedDomainObjectContainer<EvalSpec> specs) {
+  private void configureEvalTasks(Project project, NamedDomainObjectContainer<EvalSpec> specs) {
     specs.all(
         spec -> {
-          configureBaseSpec(spec);
+          configureBaseSpec(project, spec);
           spec.getOutputFile()
               .convention(
-                  project()
+                  project
                       .getLayout()
                       .getProjectDirectory()
                       // %{moduleDir} is resolved relatively to the working directory,
@@ -174,7 +166,7 @@ public class PklPlugin implements Plugin<Project> {
           spec.getExpression()
               .convention(CliEvaluatorOptions.Companion.getDefaults().getExpression());
 
-          createModulesTask(EvalTask.class, spec)
+          createModulesTask(project, EvalTask.class, spec)
               .configure(
                   task -> {
                     task.getOutputFile().set(spec.getOutputFile());
@@ -186,11 +178,12 @@ public class PklPlugin implements Plugin<Project> {
         });
   }
 
-  private void configureJavaCodeGenTasks(NamedDomainObjectContainer<JavaCodeGenSpec> specs) {
+  private void configureJavaCodeGenTasks(
+      Project project, NamedDomainObjectContainer<JavaCodeGenSpec> specs) {
     specs.all(
         spec -> {
-          configureBaseSpec(spec);
-          configureCodeGenSpec(spec);
+          configureBaseSpec(project, spec);
+          configureCodeGenSpec(project, spec);
 
           spec.getGenerateGetters().convention(false);
           spec.getGenerateJavadoc().convention(false);
@@ -198,14 +191,13 @@ public class PklPlugin implements Plugin<Project> {
           // constructor parameters annotations by setting this property to `null`.
           spec.getParamsAnnotation()
               .set(
-                  project()
-                      .provider(
-                          () ->
-                              spec.getGenerateSpringBootConfig().get()
-                                  ? null
-                                  : "org.pkl.config.java.mapper.Named"));
+                  project.provider(
+                      () ->
+                          spec.getGenerateSpringBootConfig().get()
+                              ? null
+                              : "org.pkl.config.java.mapper.Named"));
 
-          createModulesTask(JavaCodeGenTask.class, spec)
+          createModulesTask(project, JavaCodeGenTask.class, spec)
               .configure(
                   task -> {
                     configureCodeGenTask(task, spec);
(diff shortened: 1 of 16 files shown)


pkl-gradle/src/main/java/org/pkl/gradle/task/ModulesTask.java
```diff
@@ -49,22 +49,20 @@ public abstract class ModulesTask extends BasePklTask {
   public abstract ListProperty<Object> getSourceModules();
 
   @InputFiles
-  public abstract ListProperty<File> getTransitiveModules();
-
-  private final Map<List<Object>, Pair<List<File>, List<URI>>> parsedSourceModulesCache =
-      new HashMap<>();
+  @org.gradle.api.tasks.PathSensitive(org.gradle.api.tasks.PathSensitivity.RELATIVE)
```
HT154: We tend to prefer importing the types here instead of the fq name at the usage site.

```suggestion
  @PathSensitive(PathSensitivity.RELATIVE)
```

HT154 (commented):
Thanks! Just one little nit.

This seems reasonable to me, but I'm going to defer to @bioball's greater Gradle expertise for final approval.

pkl-gradle/src/main/java/org/pkl/gradle/task/ModulesTask.java
ffluk3: @HT154 addressed!

pkl-gradle/src/main/java/org/pkl/gradle/task/ModulesTask.java
```diff
@@ -53,24 +53,21 @@ public abstract class ModulesTask extends BasePklTask {
   public abstract ListProperty<Object> getSourceModules();
 
   @InputFiles
-  @PathSensitive(PathSensitivity.ABSOLUTE)
-  public abstract ListProperty<File> getTransitiveModules();
-
-  private final Map<List<Object>, Pair<List<File>, List<URI>>> parsedSourceModulesCache =
-      new HashMap<>();
+  @PathSensitive(PathSensitivity.RELATIVE)
```
HT154: Looks like this got lost in the rebase
```suggestion
  @PathSensitive(PathSensitivity.ABSOLUTE)
```

ffluk3: Sorry about that. Fixed!

ffluk3:
Good morning @bioball! Please let me know if there are other requests or questions around getting this verified or merged in. Happy to help out where I can!

bioball:
Hi! This is on my list of TODOs. I haven't forgotten about it! Apologies, my availability is quite limited right now so I haven't gotten to a review yet.

ffluk3:
No problem at all! Appreciate both of your time managing and maintaining the project 😄

pkl-gradle/src/main/java/org/pkl/gradle/task/AnalyzeImportsTask.java
```diff
@@ -35,20 +38,45 @@ public abstract class AnalyzeImportsTask extends ModulesTask {
...
+   * @return the list of file-based transitive import paths
+   */
+  @SuppressWarnings("unused") // called as a method reference in PklPlugin
+  public static List<File> parseTransitiveFiles(RegularFile outputFile) {
+    if (!outputFile.getAsFile().exists()) {
+      return Collections.emptyList();
+    }
+    try {
+      var contents = java.nio.file.Files.readString(outputFile.getAsFile().toPath());
+      ImportGraph importGraph = ImportGraph.parseFromJson(contents);
```
bioball: Two notes:

1. Import the class rather than use fully qualified name
2. Prefer `var` over declaring the type

```suggestion
      var contents = Files.readString(outputFile.getAsFile().toPath());
      var importGraph = ImportGraph.parseFromJson(contents);
```

pkl-gradle/src/main/java/org/pkl/gradle/task/AnalyzeImportsTask.java
```diff
@@ -35,20 +38,45 @@ public abstract class AnalyzeImportsTask extends ModulesTask {
...
+   * task. Returns an empty list if the file does not exist. Throws a {@link RuntimeException} if
+   * the file exists but cannot be read or parsed, so that upstream errors are not silently lost.
+   *
+   * <p>The automatically-created {@code GatherImports} tasks always write JSON, so this method
+   * assumes JSON format and should only be called for those tasks.
+   *
+   * @param outputFile the output file produced by this task
+   * @return the list of file-based transitive import paths
+   */
+  @SuppressWarnings("unused") // called as a method reference in PklPlugin
```
bioball: I'm not seeing a warning here regardless?

```suggestion
```

pkl-gradle/src/main/java/org/pkl/gradle/task/AnalyzeImportsTask.java
```diff
@@ -35,20 +38,45 @@ public abstract class AnalyzeImportsTask extends ModulesTask {
...
+  public static List<File> parseTransitiveFiles(RegularFile outputFile) {
```
bioball: Can you move this method into an internal util class? Something like `PluginUtils` or something like that should work.

`AnalyzeImportsTask` is a public API, and this method is somewhat of an implementation detail.

pkl-gradle/src/main/java/org/pkl/gradle/PklPlugin.java
```diff
@@ -209,48 +206,39 @@ private void configureJavaCodeGenTasks(NamedDomainObjectContainer<JavaCodeGenSpe
                     task.getParamsAnnotation().set(spec.getParamsAnnotation());
                     task.getNonNullAnnotation().set(spec.getNonNullAnnotation());
                   });
-        });
 
-    project.afterEvaluate(
```
bioball: We need to keep these in `project.afterEvaluate`, because `configureCodeGenSpecSourceDirectories` calls `.getSourceSet().getOrNull()`; it doesn't stay lazy.

Without this code, this test breaks (add this to `JavaCodeGeneratorsTest`):

```java
  @Test
  fun `source set configured after pkl block`() {
    writeFile(
      "build.gradle",
      """
      plugins {
        id "java"
        id "org.pkl-lang"
      }

      sourceSets {
        integTest {}
      }

      repositories {
        mavenCentral()
      }

      dependencies {
        integTestImplementation "javax.inject:javax.inject:1"
        integTestImplementation "com.google.code.findbugs:jsr305:3.0.2"
      }

      pkl {
        javaCodeGenerators {
          configClasses {
            sourceModules = ["mod.pkl"]
            outputDir = file("build/generated")
            paramsAnnotation = "javax.inject.Named"
            nonNullAnnotation = "javax.annotation.Nonnull"
            settingsModule = "pkl:settings"
            renames = ['org': 'foo.bar']
          }
        }
      }

      pkl.javaCodeGenerators.configClasses.sourceSet = sourceSets.integTest
      """,
    )
    writePklFile()

    runTask("compileIntegTestJava")

    val classesDir = testProjectDir.resolve("build/classes/java/integTest")
    assertThat(classesDir.resolve("foo/bar/Mod.class")).exists()
  }
```

pkl-gradle/src/main/java/org/pkl/gradle/task/ModulesTask.java
```diff
@@ -54,23 +54,20 @@ public abstract class ModulesTask extends BasePklTask {
 
   @InputFiles
   @PathSensitive(PathSensitivity.ABSOLUTE)
-  public abstract ListProperty<File> getTransitiveModules();
-
-  private final Map<List<Object>, Pair<List<File>, List<URI>>> parsedSourceModulesCache =
-      new HashMap<>();
+  @Optional
+  public abstract ConfigurableFileCollection getTransitiveModules();
```
bioball: Is this change related to enabling Gradle configuration cache support? If not, I'd prefer changing this back to `Property<List<File>>` to avoid the breaking change.

Also:
1. your PR comment states that this gains `PathSensitivity.RELATIVE`, but your PR doesn't support this
2. `PathSensitivity.RELATIVE` doesn't seem correct; a Pkl module can have an import that is machine dependent (e.g. `import "/foo/bar.pkl"`)

pkl-gradle/src/test/kotlin/org/pkl/gradle/AbstractTest.kt
```diff
@@ -56,6 +70,51 @@ abstract class AbstractTest {
...
+      try {
+        createRunner().build()
+      } catch (e: UnexpectedBuildFailure) {
+        throw AssertionError(
+          "Configuration cache: second run (reuse) failed\n${e.buildResult.output}"
+        )
+      }
+
+    return Pair(firstRun, secondRun)
+  }
```
bioball, line 116: Very nice!

bioball (changes requested):
Thanks for this! This mostly looks great. I have just a few comments.

Also, would love some eyes from @netvl, who wrote much of the code in our Gradle plugin.

pkl-gradle/src/main/java/org/pkl/gradle/task/ModulesTask.java
ffluk3: Reverted to ListProperty<File> to avoid the breaking change. You're right that this wasn't needed for configuration cache support. Dropped the `PathSensitivity.RELATIVE` claim as well, `ABSOLUTE` is correct and I think that was simply from a bad rebase...

pkl-gradle/src/main/java/org/pkl/gradle/task/AnalyzeImportsTask.java
ffluk3: done!

pkl-gradle/src/main/java/org/pkl/gradle/PklPlugin.java
ffluk3: Thanks for the keen eye! I added this test in and reverted the functionality.

ffluk3:
Thanks @bioball! All feedback should be addressed now 😄

pkl-gradle/src/main/java/org/pkl/gradle/PklPlugin.java
```diff
@@ -216,59 +208,58 @@ private void configureJavaCodeGenTasks(NamedDomainObjectContainer<JavaCodeGenSpe
...
-                      configureCodeGenSpecSourceDirectories(
-                          spec, "java", s -> Optional.of(s.getJava()));
-                    }));
+    project.afterEvaluate(
+        prj ->
+            specs.all(
+                spec -> {
+                  configureIdeaModule(project, spec);
+                  configureCodeGenSpecSourceDirectories(
+                      project, spec, "java", s -> Optional.of(s.getJava()));
```
bioball:
```suggestion
                  configureIdeaModule(prj, spec);
                  configureCodeGenSpecSourceDirectories(
                      prj, spec, "java", s -> Optional.of(s.getJava()));
```

pkl-gradle/src/main/java/org/pkl/gradle/task/AnalyzeImportsTask.java
```diff
@@ -35,20 +34,27 @@ public abstract class AnalyzeImportsTask extends ModulesTask {
...
+   * Parses the list of file-scheme transitive imports from the JSON output file produced by this
+   * task. Returns an empty list if the file does not exist. Throws a {@link RuntimeException} if
+   * the file exists but cannot be read or parsed, so that upstream errors are not silently lost.
+   *
+   * <p>The automatically-created {@code GatherImports} tasks always write JSON, so this method
+   * assumes JSON format and should only be called for those tasks.
+   *
+   * @param outputFile the output file produced by this task
+   * @return the list of file-based transitive import paths
+   */
```
bioball:
```suggestion
```

bioball (approved):
LGTM, thanks! This looks great! I'll wait to see if @netvl wants to provide some comments before merging this.

Some small nits to fix.

ffluk3:
> LGTM, thanks! This looks great! I'll wait to see if @netvl wants to provide some comments before merging this.
> 
> Some small nits to fix.

Thanks! Feedback incorporated and I squashed to a single commit.

pkl-gradle/src/main/java/org/pkl/gradle/PklPlugin.java
```diff
@@ -412,21 +402,23 @@ private Provider<Set<File>> getResourceSourceDirectoriesExceptSpecOutput(CodeGen
...
       String languageName,
       Function<? super SourceSet, ? extends Optional<SourceDirectorySet>>
           extractSourceDirectorySet) {
-    var task = project().getTasks().named(spec.getName(), CodeGenTask.class);
-    var sourceSet = spec.getSourceSet().get();
+    var task = project.getTasks().named(spec.getName(), CodeGenTask.class);
+    var sourceSet = spec.getSourceSet().getOrNull();
+    if (sourceSet == null) {
+      return;
+    }
```
netvl, line 418: Why is this change needed? I think it would make sense to throw if the source set is not defined.

pkl-gradle/src/main/java/org/pkl/gradle/PklPlugin.java
```diff
@@ -436,18 +428,17 @@ private void configureCodeGenSpecSourceDirectories(
...
         .withPlugin(
             "idea",
             plugin -> {
-              var module = project().getExtensions().getByType(IdeaModel.class).getModule();
+              var module = project.getExtensions().getByType(IdeaModel.class).getModule();
               var outputDir = spec.getOutputDir().get().getAsFile();
               module.getGeneratedSourceDirs().add(outputDir);
-              if (spec.getSourceSet().get().getName().toLowerCase().contains("test")) {
+              var sourceSet = spec.getSourceSet().getOrNull();
+              if (sourceSet != null && sourceSet.getName().toLowerCase().contains("test")) {
```
netvl: Same here, I think

netvl (commented):
Just a couple of questions, overall looks awesome!

pkl-gradle/src/main/java/org/pkl/gradle/PklPlugin.java
ffluk3: I was getting some IDE warnings for this particular issue, but I don't think this is required for the configuration cache improvements. What I'll do is throw an explicit `GradleException` when the `sourceSet` can't be fetched, so it's clear to the end user.

pkl-gradle/src/main/java/org/pkl/gradle/PklPlugin.java
ffluk3: same as above!

ffluk3:
> Just a couple of questions, overall looks awesome!

@netvl thanks for your review! I was able to roll back my null checks, plus some more helpful exception messages on missing sourcesets.

ffluk3:
Hey @bioball thanks for merging this! It was great to collaborate to get this update in.

Quick question on the release cadence: I know your minor revisions will be frequently deployed Feb/Jun/Oct per your roadmap. Having said that, will there be any plans to cut this change (and the existing merged) as a patch before June? Our team is excited to make use of this enhancement!

Thanks again.

bioball:
We generally don't put behavior changes into patch versions. So, alas, this change will be part of the 0.32 release cycle!

PR #1507, opened 2026-04-07
SPICE-0028: Add support for multi-line string line continuations

HT154:
SPICE: https://github.com/apple/pkl-evolution/pull/31


docs/modules/language-reference/pages/index.adoc
```diff
@@ -305,7 +304,8 @@ To write a string that spans multiple lines, use a multiline string literal:
 ----
 """
 Although the Dodo is extinct,
-the species will be remembered.
+the species \
+will be remembered.
```
bioball: This is the first time multiline strings are introduced. I think it's better to talk about continuations elsewhere.

I think we can have another paragraph at the bottom of this section, something like:

```asciidoc
To prevent line breaks from becoming part of the string's value, use a backslash (`\`) to end those lines.

[source%tested,{pkl}]
----
str = """
  Although the Dodo is extinct, \
  the species will be remembered.
  """
----

This multiline string is equivalent to the following single-line string:

[source%parsed,{pkl-expr}]
----
"Although the Dodo is extinct, the species will be remembered."
----
```

bioball (changes requested):
Looks good! Just some minor comments.

Can we also add a parser error test for whitespace after line continuation?

pkl-core/src/test/files/LanguageSnippetTests/input/basic/stringMultiline.pkl
```diff
@@ -25,6 +25,22 @@ examples {
...
+    #"""
+    hello \#
+    world
+    """#
+  
+    """
+    hello \
+    \
+    world
+    """
```
bioball: Move these into a new entry called "string continuation"

pkl-core/src/test/files/LanguageSnippetTests/output/errors/parser21.err
```diff
@@ -0,0 +1,8 @@
+–– Pkl Error ––
+Invalid character escape sequence `\ `.
+
+Valid character escape sequences are:  \n  \r  \t  \"  \\
```
bioball: Can we add some sort of hint that string continuation is also valid?

bioball (approved):
LGTM!

Also looking for a thumb from @stackoverflow

pkl-core/src/test/files/LanguageSnippetTests/output/errors/parser21.err
HT154: Hmm given that `\n` is already present here, any ideas on how to communicate this? `\«newline»`?

HT154: Went with `\<newline>`, but open to input here

pkl-formatter/src/main/java/org/pkl/formatter/Builder.java
```diff
@@ -922,7 +927,9 @@ private List<FormatNode> formatStringParts(List<Node> nodes) {
       if (elem.type == NodeType.TERMINAL && text(elem).endsWith("(")) {
         isInStringInterpolation = true;
       }
-      result.add(format(elem));
+      var formatted = format(elem);
+      if (formatted instanceof Nodes formattedNodes) result.addAll(formattedNodes.nodes());
+      else result.add(format(elem));
```
stackoverflow:
```suggestion
      else result.add(formatted);
```

stackoverflow (approved):
LGTM.

pkl-formatter/src/main/java/org/pkl/formatter/Builder.java
HT154: Ah good catch, thanks! Must've missed this one with the kt->java translation.

PR #1515, opened 2026-04-09
pkl-parser: Migrate nullness to JSpecify

odenix:
based on https://github.com/apple/pkl/pull/1516

Motivation:
Increase null safety of pkl-parser and its clients by replacing custom nullness annotations with JSpecify and enabling build-time nullness checks.

Changes:
- Replace pkl-parser's custom nullness annotations with JSpecify
- Add missing `@Nullable` annotations
- Use a start-of-file sentinel token instead of a nullable ParserImpl.prev
- Remove support for ParserVisitors with nullable result types
  - The added complexity is not worth it; visitor implementations can use java.util.Optional if needed
- Replace assertion-based null narrowing with Objects.requireNonNull where required by NullAway
- Add the pklJSpecify Gradle convention plugin and apply it to pkl-parser
  - The plugin adds the JSpecify dependency and enables build-time nullness checking with NullAway
- Include JSpecify in fat JARs without relocation

Result:
- pkl-parser now uses JSpecify annotations
- No NullAway errors in Gradle
- No nullness warnings in IntelliJ

Open questions (to be addressed in this PR):
- Should JSpecify be included in fat JARs, and if so, should it be relocated?
  - Current approach: yes for inclusion, no for relocation.
- Should AST methods such as `org.pkl.parser.syntax.Class.getName()` use `Objects.requireNonNull()` instead of assertions for return-value narrowing?
  - NullAway does not currently require `Objects.requireNonNull()` in these cases, possibly because the value is first cast to a non-nullable type.
  - Even so, switching to `Objects.requireNonNull()` would make these methods fail reliably if parser invariants are violated.

JSpecify/NullAway learnings:
- Migration to JSpecify and NullAway has gained momentum in the Java OSS ecosystem.
  Projects such as Spring Framework and Spring Boot have already completed the transition.
- NullAway recommends compiling with JDK 25 for best results, even when targeting older Java versions.
  For pkl-parser, compiling with JDK 25 instead of JDK 21 does not currently surface additional findings, so a JDK upgrade can be postponed.
- NullAway deliberately ignores assertions because they are not reliably enforced at runtime.
  This is why some assertion-based null narrowing had to be replaced with `Objects.requireNonNull()`.
- IntelliJ's nullness analysis is currently more sensitive in some cases.
  There is ongoing work to close the gap between IntelliJ and NullAway.
  
Further reading:
- https://docs.spring.io/spring-framework/reference/core/null-safety.html
- https://blog.jetbrains.com/idea/2025/11/one-could-simply-add-nullability-check-support-without-even-noticing-it/


bioball (approved):
This looks great, thanks!

Looking for one more review from @stackoverflow or @HT154 before merging this.

Responses to open Qs:

> Should JSpecify be included in fat JARs, and if so, should it be relocated?
> Current approach: yes for inclusion, no for relocation.

Makes sense to me. We can revisit this later if it causes problems for downstream users (which seems unlikely)

> Should AST methods such as org.pkl.parser.syntax.Class.getName() use Objects.requireNonNull() instead of assertions for return-value narrowing?
> * NullAway does not currently require Objects.requireNonNull() in these cases, possibly because the value is first cast to a non-nullable type.
> * Even so, switching to Objects.requireNonNull() would make these methods fail reliably if parser invariants are violated.

I think it makes sense to keep using `assert` in cases where NullAway doesn't complain; it's better that Java can compile away these assertions. We can rely on unit tests to catch these cases if our invariants are violated.

odenix:
> I think it makes sense to keep using assert in cases where NullAway doesn't complain; it's better that Java can compile away these assertions. We can rely on unit tests to catch these cases if our invariants are violated.

Less work for me! For comparison, `Objects.requireNonNull()` should have roughly the same cost as Kotlin's `!!`. For an AST traversed once and lowered into a Truffle AST, the cost is probably negligible.

Edit: I just discovered`-XepOpt:NullAway:AssertsEnabled=true`, which instructs NullAway to honor assertions. It's off by default because it's less safe. I'm starting to think we should enable this option and rely on non-null assertions during the migration, since that’s the style currently used in the codebase. "Should we sometimes use `Objects.requireNonNull` instead of an assertion?" then becomes a separate question/task.

bioball:
Ah, yeah, then keeping the current code style is preferred!

odenix:
> Ah, yeah, then keeping the current code style is preferred!

Done and ready for review. Also rebased onto the JVM toolchain PR.

pkl-parser/src/main/java/org/pkl/parser/ParserVisitor.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
```
stackoverflow: This file doesn't seem to be touched.

stackoverflow (approved):
LGTM.

pkl-parser/src/main/java/org/pkl/parser/ParserVisitor.java
odenix: fixed

HT154 (approved):

PR #1518, opened 2026-04-14
Add spotless formatter step to revert copyright year only changes

bioball:
This avoids an issue where, during the course of development, a file is touched, thus modifying the copyright year.
Then, undoing the previous change does not undo the copyright year change.


HT154 (approved):
Slick!!

stackoverflow (approved):
Nice

PR #1521, opened 2026-04-15
Fix gradle build issue

stackoverflow:
The build was broken for me because gradle couldn't resolve some classes while trying to format Pkl.


bioball (approved):
let

PR #1523, opened 2026-04-15
Add dependency submission

bioball:
This adds jobs to add Gradle dependencies to [GitHub's dependency submission API](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/using-the-dependency-submission-api), and to review when these dependencies change.

### DIFF
--- a/.github/index.pkl
+++ b/.github/index.pkl
@@ -22,6 +22,8 @@ testReports {
   excludeJobs {
     "bench"
     "github-release"
+    "dependency-submission"
+    "dependency-review"
     Regex("deploy-.*")
   }
 }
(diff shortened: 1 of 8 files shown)


.github/index.pkl
```diff
@@ -114,6 +116,23 @@ local releaseJobs: PklJobs = new {
...
+local dependencySubmission: Workflow.Job = new {
+  `runs-on` = "ubuntu-latest"
+  permissions {
+    contents = "write"
+  }
+  steps {
+    module.catalog.`actions/checkout@v6`
+    (module.catalog.`actions/setup-java@v5`) {
+      with {
+        `java-version` = "25"
```
HT154: Probably best to pull this out somewhere so it's not in multiple places in our GHA configs

.github/index.pkl
```diff
@@ -174,6 +202,10 @@ main {
         needs = buildAndTestJobs.keys.toListing()
       }
     } |> toWorkflowJobs
+
+  jobs = (_jobs) {
+    ["dependency-submission"] = dependencySubmission
+  }
```
HT154: This is a little awkward. Can `toWorkflowJobs` just pass `Workflow.Job` instances through instead?

HT154 (approved):
Approving to unblock, but a couple small things

HT154:
Looks like dependency submission needs
```yaml
permissions:
  contents: write
```

Do we really need to be doing this on PRs?

bioball:
I have `contents: write` set on the job level, but it looks like it needs to be on the workflow level in order for it to work.

The goal was to have dependency changes reviewed by [dependency-review-action](https://github.com/actions/dependency-review-action), but I think we can try tackling that later (if at all). I'll remove that from the PRB builds.

.github/index.pkl
bioball: I can't think of an easy place to put this off the bat, so I'm just going to keep this as-is. Maybe we can refactor things, but I'll consider this as out-of-scope for now.

.github/index.pkl
bioball: Good idea!

PR #1524, opened 2026-04-15
Gradle: Replace legacy buildSrc mechanism with included build

odenix:
Motivation:
buildSrc is a special-case legacy mechanism.
Gradle recommends using an included build named build-logic instead:
https://docs.gradle.org/current/userguide/best_practices_structuring_builds.html#favor_composite_builds

Changes:
- Rename buildSrc/ to build-logic/
  - triggers reformatting
- Replace occurrences of "buildSrc" with "build-logic"
- Include the build-logic build in the main build (via settings.gradle.kts)
- Apply convention plugins via plugin IDs instead of type-safe accessors
  - small tradeoff compared to buildSrc

Result:
- Faster and more isolated builds
- Build logic behaves like a normal build, making it easier to evolve and reason about


HT154 (approved):
Seems reasonable to me, but the rest of the team may have input.

build-logic/src/main/kotlin/pklAllProjects.gradle.kts
```diff
@@ -136,18 +136,18 @@ spotless {
   val revertYearOnlyChangesStep =
     RevertYearOnlyChangesStep(rootProject.rootDir, ratchetFrom!!).create()
 
-  // When building root project, format buildSrc files too.
-  // We need this because buildSrc is not a subproject of the root project, so a top-level
-  // `spotlessApply` will not trigger `buildSrc:spotlessApply`.
+  // When building root project, format build-logic files too.
+  // We need this because build-logic is not a subproject of the root project, so a top-level
+  // `spotlessApply` will not trigger `build-logic:spotlessApply`.
   if (project === rootProject) {
```
bioball: Not sure how this got lost; I thought I had changed this line already. But as of Gradle 9.3, we shouldn't depend on referential equality for `Project`

```suggestion
  if (project.path == rootProject.path) {
```

bioball (approved):
TIL! Thanks! This looks good to me too.

PR #1531, opened 2026-04-17
Remove Config.makeConfig

bioball:
This doesn't really make sense as part of the `Config` API.

We can maybe make class `Utils` public, but, I don't know how useful it is anyways; it's more of an implementation detail.

### DIFF
--- a/pkl-config-java/src/main/java/org/pkl/config/java/Utils.java
+++ b/pkl-config-java/src/main/java/org/pkl/config/java/Utils.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright © 2026 Apple Inc. and the Pkl project authors. All rights reserved.
+ *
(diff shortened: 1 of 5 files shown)


pkl-config-java/src/main/java/org/pkl/config/java/Utils.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.config.java;
+
+import java.util.Map;
+import org.pkl.config.java.mapper.ValueMapper;
+import org.pkl.core.Composite;
+
+class Utils {
```
odenix: Could be `ConfigUtils` to make it a bit more unique and easier to search for / open.

HT154 (approved):

PR #1532, opened 2026-04-17
Enable codeql scanning

bioball:
This enables security vulnerability scanning using CodeQL.

### DIFF
--- a/.github/codeql.pkl
+++ b/.github/codeql.pkl
@@ -1,0 +1,66 @@
+amends "@gha/Workflow.pkl"
+
+import "@gha/catalog.pkl"
+
+on {
+  push {
+    branches {
+      "main"
+    }
+  }
+  schedule {
+    // Run at 01:38 on Saturday
+    new { cron = "38 1 * * 6" }
+  }
+}
+
+local class CodeQLScan {
+  language: String
+
+  `build-mode`: String
+}
+
+local scans: Listing<CodeQLScan> = new {
+  new {
+    language = "actions"
+    `build-mode` = "none"
+  }
+  new {
+    language = "java-kotlin"
+    `build-mode` = "autobuild"
+  }
+  new {
+    language = "javascript-typescript"
+    `build-mode` = "none"
+  }
+}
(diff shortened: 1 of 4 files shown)


.github/codeql.pkl
```diff
@@ -0,0 +1,66 @@
...
+  }
+  new {
+    language = "javascript-typescript"
+    `build-mode` = "none"
+  }
+}
+
+jobs {
+  for (scan in scans) {
+    ["analyze-\(scan.language)"] {
```
StefMa, line 40: Isn't it better/easier to go with a matrix build instead of multiple jobs? 🤔 🤷‍♂️
It's the same outout, so it actually doesn't matter. Just my initial thought about it 😂

bioball: Matrices feel like a poor man's for loop that we can express much more easily with Pkl. And, you get actual Pkl dot access this way, rather than a stringy typed value (`"${{ matrix.language }}"` vs. `scan.language`). I think this ends up as basically the same thing? Each turns into its own job execution regardless.

.github/codeql.pkl
```diff
@@ -0,0 +1,66 @@
...
+
+import "@gha/catalog.pkl"
+
+on {
+  push {
+    branches {
+      "main"
+    }
+  }
+  schedule {
```
HT154, line 12: Is there a reason not to run this on push or PR? Having this able to block PRs would be very handy. I'd be interested in adding at least one custom query to block a problem pattern in CliCommand.

bioball: I was concerned about how long these things take to run. But, we can play around with it and see what happens!

HT154: Analysis for java-kotlin finished within a minute of the gradle-check job, so I think this would probably be okay to enable on PRs. A <1m delay in exchange for better security enforcement on PRs seems okay to me.

bioball: Added PRs

HT154 (approved):

PR #1536, opened 2026-04-20
use strict equality in r

tejas-ae:
The path through r looks mostly fine, but The loose equality here can coerce values in ways that are hard to spot later. this patch tightens it to strict equality. Happy to adjust if you’d prefer a different approach.
Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.


tejasae-afk:
Closing — this change targets a vendored minified file, which is not the right place to apply source-level fixes.

PR #1544, opened 2026-04-20
pkl-config-java: Refine nullness handling in Config and JavaType

odenix:
This PR may need further discussion, but I wanted to start with a concrete proposal.

Motivation:
Config.as() causes nullness warnings when its result is intentionally assigned to a non-null variable

Changes:
- Introduce Config.asNullable() to handle nullable values explicitly
- Change the signature of Config.as() back to `<T> T as(Class<T> type)` despite @NullMarked
  - This changes the semantics of as() to return a non-null value while retaining source, binary, and runtime compatibility
  - Users who perform nullness checks at build time (for example, using NullAway in error mode) will need to replace as() with asNullable() where appropriate
  - Other users will be guided by IDE warnings to do so
  - In the future, as() could enforce a non-null return value at runtime, aligning with the Kotlin extension function to() (this would be a breaking change)
- Introduce ofNullable variants for most JavaType factory methods (e.g., JavaType.listOfNullable) to simplify constructing nullable types
- Overhaul Javadoc of Config and JavaType

Result:
- Clear separation between accessing nullable and non-null values
- No spurious nullness warnings after replacing as() with asNullable() where appropriate
- The most ergonomic method name, as(), is used for the common non-null case
- More ergonomic construction of nullable types
- More detailed and consistent documentation

### DIFF
--- a/pkl-config-java/src/main/java/org/pkl/config/java/Config.java
+++ b/pkl-config-java/src/main/java/org/pkl/config/java/Config.java
@@ -17,60 +17,130 @@ package org.pkl.config.java;
 
 import java.io.InputStream;
 import java.lang.reflect.Type;
+import org.jspecify.annotations.NullUnmarked;
 import org.jspecify.annotations.Nullable;
 import org.pkl.config.java.mapper.ConversionException;
 import org.pkl.config.java.mapper.ValueMapper;
-import org.pkl.core.Evaluator;
 
 /**
- * A root, intermediate, or leaf node in a configuration tree. Child nodes can be obtained by name
- * using {@link #get(String)}. To consume the node's composite or scalar value, convert the value to
- * the desired Java type, using one of the provided {@link #as} methods.
+ * A root, intermediate, or leaf node in a configuration tree.
+ *
+ * <p>To navigate to a child node, use {@link #get(String)} with the child's unqualified name.
+ *
+ * <p>To retrieve this node's value, use:
+ *
+ * <ul>
+ *   <li>{@link #as(Class)} for non-null types.
+ *   <li>{@link #asNullable(Class)} for nullable types.
+ *   <li>{@link #as(JavaType)} for parameterized types, such as {@code List<@Nullable String>}.
+ * </ul>
+ *
+ * <p>Whether a method can return null depends on the method and target type used. For example,
+ * {@code asNullable(String.class)} can return {@code null}, while {@code
+ * as(JavaType.listOfNullable(String.class))} can return a {@code List<String>} with nullable
+ * elements. These nullness rules are for static analysis tools such as IntelliJ IDEA and NullAway
+ * and are not enforced at runtime.
  */
 @SuppressWarnings({"DeprecatedIsStillUsed"})
 public interface Config {
   /**
-   * The dot-separated name of this node. For example, the node reached using {@code
-   * rootNode.get("foo").get("bar")} has qualified name {@code foo.bar}. Returns the empty String
-   * for the root node itself.
+   * Returns the qualified name of this node, or the empty string if this is the root node.
+   *
+   * <p>The qualified name is formed by joining child names using {@code '.'}. For example, {@code
+   * rootNode.get("foo").get("bar").getQualifiedName()} returns {@code "foo.bar"}.
    */
   String getQualifiedName();
 
   /**
-   * The raw value of this node, as provided by {@link Evaluator}. Typically, a node's value is not
-   * consumed directly, but converted to the desired Java type using {@link #as}.
+   * Returns the underlying value of this node.
+   *
+   * <p>This value is typically accessed indirectly via {@link #as(Class)}, {@link
+   * #asNullable(Class)}, or {@link #as(JavaType)}.
    */
   Object getRawValue();
 
   /**
    * Returns the child node with the given unqualified name.
    *
+   * <p>For example, {@code get("foo").get("bar")} returns the child named {@code "bar"} of the
+   * child named {@code "foo"}. Passing a qualified name to this method, as in {@code
+   * get("foo.bar")}, is not supported.
+   *
    * @throws NoSuchChildException if a child with the given name does not exist
    */
   Config get(String childName);
 
   /**
-   * Converts this node's value to the given {@link Class}.
+   * Returns this node's value as a non-null value of the given {@link Class}.
+   *
+   * <p>If this node's value may be {@code null}, use {@link #asNullable(Class)} instead. In a
+   * future version, this method will perform a non-null check.
    *
    * @throws ConversionException if the value cannot be converted to the given type
    */
-  <T extends @Nullable Object> T as(Class<T> type);
+  @NullUnmarked
+  <T> T as(Class<T> type);
 
   /**
-   * Converts this node's value to the given {@link Type}.
+   * Returns this node's value as a non-null value of the given {@link JavaType}.
    *
-   * <p>Note that usages of this method are not type safe.
+   * <p>If this node's value may be {@code null}, use {@link #asNullable(JavaType)} instead. In a
+   * future version, this method will perform a non-null check.
+   *
+   * <p>Use this method when converting to a parameterized type, such as {@code List<@Nullable
+   * String>}.
    *
    * @throws ConversionException if the value cannot be converted to the given type
    */
-  <T extends @Nullable Object> T as(Type type);
+  @NullUnmarked
+  <T> T as(JavaType<T> type);
 
   /**
-   * Converts this node's value to the given {@link JavaType}.
+   * Returns this node's value as a non-null value of the given {@link Type}.
+   *
+   * <p>If this node's value may be {@code null}, use {@link #asNullable(Type)} instead. In a future
+   * version, this method will perform a non-null check.
+   *
+   * <p>Use this method when the target type is already available as a {@link Type}; otherwise,
+   * prefer {@link #as(Class)} or {@link #as(JavaType)}.
    *
    * @throws ConversionException if the value cannot be converted to the given type
    */
-  <T extends @Nullable Object> T as(JavaType<T> type);
+  @NullUnmarked
+  <T> T as(Type type);
+
+  /**
+   * Returns this node's value as a nullable value of the given {@link Class}.
+   *
+   * <p>If this node's value cannot be {@code null}, use {@link #as(Class)} instead.
+   *
+   * @throws ConversionException if the value cannot be converted to the given type
+   */
+  <T> @Nullable T asNullable(Class<T> type);
+
+  /**
+   * Returns this node's value as a nullable value of the given {@link Class}.
+   *
+   * <p>If this node's value cannot be {@code null}, use {@link #as(JavaType)} instead.
+   *
+   * <p>Use this method when converting to a parameterized type, such as {@code List<@Nullable
+   * String>}.
+   *
+   * @throws ConversionException if the value cannot be converted to the given type
+   */
+  <T> @Nullable T asNullable(JavaType<T> type);
+
+  /**
+   * Returns this node's value as a nullable value of the given {@link Type}.
+   *
+   * <p>If this node's value cannot be {@code null}, use {@link #as(Type)} instead.
+   *
+   * <p>Use this method when the target type is already available as a {@link Type}; otherwise,
+   * prefer {@link #asNullable(Class)} or {@link #as(JavaType)}.
+   *
+   * @throws ConversionException if the value cannot be converted to the given type
+   */
+  <T extends @Nullable Object> T asNullable(Type type);
 
   /**
    * Decodes a config from the supplied byte array.
--- a/pkl-config-java/src/main/java/org/pkl/config/java/JavaType.java
+++ b/pkl-config-java/src/main/java/org/pkl/config/java/JavaType.java
@@ -23,30 +23,77 @@ import org.pkl.config.java.mapper.Types;
 import org.pkl.core.Pair;
 
 /**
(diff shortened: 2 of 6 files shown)


pkl-config-java/src/main/java/org/pkl/config/java/Config.java
```diff
@@ -20,53 +20,106 @@
...
+   * Returns this node's value as a nullable value of the given {@link Class}.
    *
-   * <p>Note that usages of this method are not type safe.
+   * <p>If this node's value cannot be {@code null}, use {@link #as(Class)} instead.
    *
    * @throws ConversionException if the value cannot be converted to the given type
    */
-  <T extends @Nullable Object> T as(Type type);
+  default <T> @Nullable T asNullable(Class<T> type) {
+    return as(type); // currently no difference at runtime
```
bioball, line 89: Is this right? Seems like `as` should be running a non-null check before returning, whereas `as` can freely return the result of the mapper.

Otherwise, this will throw NPE:

```java
  @Test
  public void testAsWithNull() {
    var cfg = loadConfig("foo = null");
    var str = cfg.get("foo").as(String.class);
    System.out.println(str.toLowerCase());
  }
```

pkl-config-java/src/test/java/org/pkl/config/java/ConfigEvaluatorTest.java
```diff
@@ -50,4 +52,31 @@ public void evaluateExpression() {
     var address = addressConfig.as(Address.class);
     assertThat(address.street).isEqualTo("Fuzzy St.");
   }
+
+  @Test
+  public void asNullableWithClass() {
+    var mod = evaluator.evaluate(ModuleSource.text("nullValue = null; strValue = \"Bob\""));
+    @Nullable String nullValue = mod.get("nullValue").asNullable(String.class);
+    @Nullable String strValue = mod.get("strValue").asNullable(String.class);
```
bioball, line 60: Seems like nullaway can catch issues if you simply use `var `here.

pkl-config-java/src/main/java/org/pkl/config/java/Config.java
odenix: > Seems like as should be running a non-null check before returning

Performing a runtime check in 0.32 would be a major breaking change that could affect many users. The most compatible alternative would be to deprecate as() and introduce asNonNull() and asNullable(). However, this would make the common non-null case more verbose and symmetric with the less common nullable case.

This PR proposes a middle ground:

Repurpose as() for the common non-null case. To preserve runtime compatibility, do not add a runtime check yet. Instead, guide users to replace as() with asNullable() where appropriate via static analysis warnings (IDEs, NullAway).
At a later stage (e.g., 0.35), consider introducing a runtime check. This would improve safety and align with Kotlin’s to(). It would still be a breaking change, but users would have had time to migrate.

pkl-config-java/src/test/java/org/pkl/config/java/ConfigEvaluatorTest.java
odenix: Are you saying that NullAway reports a false positive when using `var` here?

pkl-config-java/src/main/java/org/pkl/config/java/Config.java
bioball: What about just introducing `asNonNull`? And `as` retains its current behavior?

odenix: What do you mean by "current"? Which signature, and `@NullMarked` or `@NullUnmarked`?

pkl-config-java/src/test/java/org/pkl/config/java/ConfigEvaluatorTest.java
bioball: No, the other way around; e.g. this will correctly error:

```java
var strValue = mod.get("strValue").asNullable(String.class);
strValue.toLowerCase();
```

pkl-config-java/src/main/java/org/pkl/config/java/Config.java
bioball: `@NullMarked`, and `Config#as` behaves like `asNullable` does in this PR (it gives you a `@Nullable T`)

odenix: `as()` + `asNonNull()` is very backward-compatible because it preserves the meaning of `as()`. The downside is that it makes nullable the default, which runs counter to JSpecify’s non-null default. It also makes the common non-null case more verbose, and users without strong nullness tooling (IntelliJ, NullAway) may never switch from `as()` to `asNonNull()` where appropriate.

odenix: It’s a trade-off between preserving existing static and runtime behavior and moving toward clearer null handling. Since this is still a 0.x release, I'd lean towards the latter. Alternatively, null handling could be improved in future Java bindings. (I think pkl-config-java has a decent API, but it's reflection-based and not aligned with other language bindings.)

bioball: I feel like `as()` returning a `@Nullable` type is totally reasonable. We make the assumption, for better or worse, that users of our libraries set up tooling to provide errors/warnings around nullability (we make this assumption with pkl-codegen-java too).

But, I could also be the crazy one here. Curious what @stackoverflow, @holzensp or @HT154 think here.

odenix: `as()` + `asNonNull()` would be a design smell in a new Java API, but it’s workable. If we go this route, I’d also replace all `JavaType.*OfNullable` methods with `*OfNonNull` equivalents.

stackoverflow: As we are still in 0.*, I think we should strive for the better API, not the better compatibility. And changing `as` to be the non-null is IMO the better API.
We should clearly mark in the docs that this is going to break if you are expecting nulls, and point to the new method.
I'm also fine with not adding a new method and leaving `as` as `@Nullable`. We can't make sure the value is non-null anyway, as it comes from Pkl. And if users want to assert it's not null they can add an assert/runtime check.

HT154: I'm leaning toward making the breaking change in pursuit of the more idiomatic, non-null-by-default approach.

bioball: Okay, ya'll win :)

In that case, I think it makes sense to just introduce a breaking change now, with a runtime null check. It seems odd to me that JSpecify tells you this can't be null, when it can actually be null in practice.

odenix: Why not give users time to migrate? What we’ve done so far is the closest equivalent to deprecating an API for future removal. Since this change affects runtime compatibility, a staged rollout seems even more important.

bioball: I think we'd be leaving users more vulnerable to NPE's during this rollout, because our JSpecify annotation there is lying to you.

Maybe a middle path would be: mark both methods as `@Nullable`, but add some note that "`as` will be receiving a runtime null-check in 0.XX".

odenix: > I think we'd be leaving users more vulnerable to NPE's during this rollout, because our JSpecify annotation there is lying to you.

Good point. Here is my updated proposal:

Step 1 (now):

```java
@NullUnmarked
<T> T as(Class<T> type); // unchanged signature and behavior; document intended usage

<T> @Nullable T asNullable(Class<T> type);
````

Step 2 (later, before 1.0):

```java
<T> T as(Class<T> type); // changed signature and behavior (non-null check)
```

A downside is that users won’t be guided by warnings to replace `as()` where appropriate. However, I now believe this is unavoidable without deprecating `as()`.

PS: After step 2, we still won't have exhaustive non-null checks for more complex cases, e.g., `as(JavaType.listOf(JavaType.listOf(String.class)))`.

odenix: @bioball how would you like to proceed?

bioball: I'm trying to understand `NullUnmarked` but am struggling; how does it help our users migrate? In our case, we know that `as` can produce `null`, so `@Nullable` seems like the correct annotation?

odenix: `@Nullable` changes semantics compared to 0.31, whereas `@NullUnmarked` doesn't. I believe neither helps users migrate, but `@Nullable` can introduce new warnings for properties known to be non-null in Pkl.

bioball: Gotcha. If neither approach is helpful for migration purposes, I think I'd rather either just break this API now, or annotate this as `@Nullable`.

My sense is that `@NullUnmarked` is for users that are migrating existing codebases to JSpecify and wish to retain semantics of non-migrated code temporarily. I feel like this isn't appropriate for us here, because we know the method can produce `null` and should mark it as such.

odenix: In my mind, the essence of this API is to trust the user on nullness at the Pkl->Java boundary and keep them honest thereafter.

Adding a non-null runtime check will replace "trust" with "verify" for the returned value itself.
However, values nested inside the returned value will remain "trust" unless their accessors perform their own runtime checks, which is generally impossible for collections. In other words, "verify" will remain shallow, probably forever.

Adding `@Nullable` risks introducing false warnings for properties users trust to be non-null in Pkl. These warnings will force users to add either casts or suppressions. Once we remove `@Nullable` in the future, those casts or suppressions will become redundant. I'm not convinced any of this is helpful to users.

My ranking, from most to least preferred:

1. Add `@NullUnmarked` now; remove it and add a non-null check later. "I trust you on this because I can't tell whether this particular `as` call can return null" is not lying to users.
1. Add a non-null check now.
1. Add `@Nullable` now; remove it and add a non-null check later.

bioball: Okay, I think that's a reasonable explanation actually. `@NullUnmarked` sounds fine to me here.

For the nested case: another solution here is to require non-null values (everywhere), and map Pkl null into `Optional` (but that's a whole another discussion).

odenix: Added `@NullUnmarked` as discussed.
Also revisited `JavaType` and turned it into a non-null top-level type. See new commit message for details.
Also removed interface default impls of `Config.asNullable()` because delegating to `Config.as()` is confusing and won't work in the long run. Adding `asNullable()` w/o default impl is a breaking change (only) for implementers.

pkl-config-kotlin/src/main/kotlin/org/pkl/config/kotlin/ConfigExtensions.kt
```diff
@@ -15,7 +15,7 @@
  */
 package org.pkl.config.kotlin
 
-import kotlin.reflect.jvm.javaType
+import kotlin.reflect.javaType
```
bioball: Why switch to the experimental API?
```suggestion
import kotlin.reflect.jvm.javaType
```

pkl-config-java/src/test/java/org/pkl/config/java/ConfigEvaluatorTest.java
bioball, line 60: Are we intentionally testing something by using `@Nullable String varName` here? If not, let's just use `var`.

```suggestion
    var nullValue = mod.get("nullValue").asNullable(String.class);
    var strValue = mod.get("strValue").asNullable(String.class);
```

Same goes for the rest of the methods

pkl-config-java/src/main/java/org/pkl/config/java/JavaType.java
```diff
@@ -59,101 +106,382 @@ private JavaType(Type type) {
...
+    return JavaType.of(Types.mapOf(keyType, valueType));
+  }
+
+  /** Creates a {@link Map} type with the given nullable key and nullable value types. */
+  public static <K, V> JavaType<Map<@Nullable K, @Nullable V>> mapOfNullableKeysAndValues(
+      JavaType<K> keyType, JavaType<V> valueType) {
+    return JavaType.of(Types.mapOf(keyType.type, valueType.type));
+  }
+
   public Type getType() {
```
bioball, line 486: This doc comment got dropped

```suggestion
  /** Returns the underlying {@link Type}. */
  public Type getType() {
```

bioball (changes requested):
Mostly looks great! Just some minor comments.

pkl-config-kotlin/src/main/kotlin/org/pkl/config/kotlin/ConfigExtensions.kt
odenix: fixed

pkl-config-java/src/test/java/org/pkl/config/java/ConfigEvaluatorTest.java
odenix: Yes, because `var` cannot be annotated.

pkl-config-java/src/main/java/org/pkl/config/java/JavaType.java
odenix: fixed

odenix:
Ready from my side.
To sum up the most important change compared to early versions of this PR,
I believe that making `JavaType` a non-null top-level type is more practical despite being less elegant.

bioball (approved):

PR #1547, opened 2026-04-22
Added `pkg` module key factory and resource reader to project loading

netvl:
This change allows `PklProject` files, usually loaded via the `Project` static methods, to have references to external packages via `package://` URIs.

This is helpful for centralizing and sharing common package configuration via packages.

I don't think there is an infrastructure in tests to test `package:` URIs since they require HTTP calls, and this PR doesn't seem like a good place to add it, but if you think it would make sense to do it, I can try to come up with something.


odenix:
> I don't think there is an infrastructure in tests to test package: URIs since they require HTTP calls

In case it helps, pkl-commons-test has `PackageServer`.

bioball (approved):
Yeah, this change makes sense to me. Thoughts, @HT154 or @stackoverflow?

HT154 (approved):
No objections from me!

PR #1557, opened 2026-04-24
Only set `--stacktrace --info` if verbose logging is enabled

bioball:
This is a quality-of-life improvement; make our build logs more easy to read through for the default case.

If we need more information, we can click on the "Enable debug logging" checkbox when re-running a job, which then populates the `runner.debug` context variable.

### DIFF
--- a/.github/jobs/GradleJob.pkl
+++ b/.github/jobs/GradleJob.pkl
@@ -29,8 +29,7 @@ fetchDepth: Int?
 
 fixed gradleArgs =
   new Listing {
-    "--info"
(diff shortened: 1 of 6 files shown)


StefMa:
We should probably add the runner context to pkl-gha 👀 🙃

HT154 (approved):

bioball:
@StefMa It actually exists! But sometimes it's actually easier to use `// language=GithubExpressionLanguage`, which provides richer language support. It breaks if you mix in interpolation into the string.

PR #1560, opened 2026-04-25
Reformat Kotlin code

bioball:
ktfmt has much improved how it formats Kotlin code. Unfortunately, this means that whenever we touch a single line in a Kotlin file, we get a _lot_ more changes thanks to ratcheting now picking up this file for formatting.

This PR just reformats every single Kotlin file so we don't have to deal with this churn in future PRs that touch Kotlin code.


HT154 (approved):
It would be good to add the merged commit's sha to .git-blame-ignore-revs after this merges to keep blame views usable.

PR #1562, opened 2026-04-27
Follow symbolic links during glob expansion

nkakouros:
FileResolver.listElements() unconditionally skipped all symlinks to prevent cyclical globs from directory symlinks. As a side effect, import*() and read*() silently dropped symlinks to regular files from their results, returning partial output with no warning, and skipped symlinks to directories even when no cycle existed.

Stop filtering symlinks. Switch the directory check from Files.isDirectory(), which silently returns false on errors, to Files.readAttributes(), which lets I/O errors propagate. This way the existing OS-level symlink-resolution limit surfaces a clear "Too many levels of symbolic links" error on cycles instead of a silent truncation. Broken symlinks are still surfaced as non-directory entries via a NoSuchFileException catch.

Test on [windows] and [native]


docs/modules/release-notes/pages/changelog.adoc
```diff
@@ -4,6 +4,11 @@ include::ROOT:partial$component-attributes.adoc[]
 [[release-0.32.0]]
 == 0.32.0 (UNRELEASED)
 
+=== Fixes
+
+* Follow symbolic links during `import*` and `read*` glob expansion; symlink cycles now surface as a clear I/O error
+(pr:https://github.com/apple/pkl/pull/XXXX[]).
```
HT154, line 10: No action needed here, but just a note for us when prepping the release: this is a breaking change and we'll need to ensure it's marked accordingly.

HT154 (approved):
LGTM, thanks!

bioball:
Filling in from an external conversation: we want this, but we want to first introduce this behind a feature flag. We don't have a feature flagging system yet, so we'd like to introduce one before merging this.

PR #1563, opened 2026-04-28
Introduce "minimal" test reporter

stackoverflow:


pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/TestOptions.kt
```diff
@@ -51,7 +51,16 @@ class TestOptions : OptionGroup() {
   private val overwrite: Boolean by
     option(names = arrayOf("--overwrite"), help = "Force generation of expected examples.").flag()
 
+  private val showOnlyFailed: Boolean by
+    option(names = arrayOf("--show-only-failed"), help = "Only show failed test output.").flag()
```
HT154: It might be nice to specify that this only affects the output written to stdout, not the junit report.

pkl-core/src/main/java/org/pkl/core/stdlib/test/report/SimpleReport.java
```diff
@@ -92,10 +94,13 @@ public void summarize(List<TestResults> allTestResults, Writer writer) throws IO
...
+            ? section.results().stream().filter(TestResult::isFailure).toList()
+            : section.results();
+    if (!results.isEmpty()) {
       builder.append("  ").append(section.name()).append('\n');
-      StringUtils.joinToStringBuilder(
-          builder, section.results(), "\n", res -> reportResult(res, builder));
+      StringUtils.joinToStringBuilder(builder, results, "\n", res -> reportResult(res, builder));
       builder.append('\n');
     }
   }
```
HT154: (Not a blocker) This result in output that includes the test module name but no further details:
```console
❯ ../pkl/pkl-cli/build/executable/jpkl test --project-dir packages/com.github.dependabot --show-only-failed
module com.github.dependabot.tests.Dependabot

module com.github.dependabot.tests.examples

100.0% tests pass [2 passed], 100.0% asserts pass [2 passed]
```

Is this really desired? It seems like it might be nice to either omit module names with only passing tests or indicate somehow that all tests in the module passed.

HT154 (approved):
Just a couple nits, approving to unblock

pkl-core/src/main/java/org/pkl/core/stdlib/test/report/SimpleReport.java
stackoverflow: > Is this really desired?

Yes. The idea is to show all modules being tests. I think hiding the module names might be confusing.
We could show the green passed checkmark after the tests if we are hiding all the results and there's no failure. Open to ideas.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/TestOptions.kt
```diff
@@ -51,7 +51,21 @@ class TestOptions : OptionGroup() {
   private val overwrite: Boolean by
     option(names = arrayOf("--overwrite"), help = "Force generation of expected examples.").flag()
 
+  private val showOnlyFailed: Boolean by
+    option(
+        names = arrayOf("--show-only-failed"),
+        help =
+          "Only show failed test output. JUnit reports are not affect by this option, only CLI.",
```
bioball: IMO, mentioning JUnit here is overly specific.

```suggestion
          "Only show failed test output.",
```

pkl-core/src/main/java/org/pkl/core/stdlib/test/report/SimpleReport.java
bioball: This does seem strange to me too. Folks that use this option likely want to minimize their terminal output. But this will still give them potentially quite long test output.

I think it should be fine to only show the summary if everything has passed.

bioball (changes requested):
It makes sense to have this functionality, although I'm thinking that this should be introduced as a test reporter.

Something like:

```
pkl test --test-reporter=minimal
```

We considered introducing a test reporter API before, and now is probably the right time to add it. If we introduce test reports later, the `--show-only-failed` is likely to be deprecated.

If we have a test reporter option, it also allow us to introduce reporters like "json", "dot", "tap", which are all somewhat common amongst test frameworks. And, our current reporter can be called the "spec" reporter (our test reports are modeled after Mocha, and that's what Mocha calls theirs).

pkl-core/src/main/java/org/pkl/core/stdlib/test/report/SimpleReport.java
stackoverflow: The new minimal reporter doesn't show module names for modules with only passing tests/examples.
So in case of no errors/failures, only the summary is shown.

pkl-core/src/main/java/org/pkl/core/stdlib/test/report/MinimalReport.java
```diff
@@ -0,0 +1,172 @@
...
+import org.pkl.core.TestResults;
+import org.pkl.core.TestResults.TestResult;
+import org.pkl.core.TestResults.TestSectionResults;
+import org.pkl.core.util.AnsiStringBuilder;
+import org.pkl.core.util.AnsiStringBuilder.AnsiCode;
+import org.pkl.core.util.AnsiTheme;
+import org.pkl.core.util.StringUtils;
+
+/** Minimal reporter. Only reports failures and errors. */
+public final class MinimalReport implements TestReport {
```
stackoverflow: There's a lot of duplication with SimpleReport here. But I don't think it's worth DRYing it because future reports (like JSON, for example) might not share code with them.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/commands/TestOptions.kt
```diff
@@ -51,21 +53,19 @@ class TestOptions : OptionGroup() {
...
 
-  private val showOnlyFailed: Boolean by
-    option(
-        names = arrayOf("--show-only-failed"),
-        help =
-          "Only show failed test output. JUnit reports are not affect by this option, only CLI.",
-      )
-      .flag()
+  private val reporter: TestReporter by
+    option(names = arrayOf("--reporter"), help = "Which test reporter to use for CLI output.")
```
bioball: To dis-ambiguate test flags in commands like `pkl project package`, etc, I think we should call this `--test-reporter`

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/TestReporter.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.commons.cli
+
+enum class TestReporter {
+  SIMPLE,
```
bioball: Since this reporter follows Mocha's "spec" reporter, I think let's also call this one "SPEC"

pkl-core/src/main/java/org/pkl/core/stdlib/test/report/MinimalReport.java
bioball: Can we move this into a base class that's shared by MInimalReport and SimpleReport?

Also, nit: should call this "MinimalReporter", and change `TestReport` to `TestReporter`

bioball (changes requested):
One main comment here is: can we de-duplicate the logic here by moving things into a base class?

Also, I'm seeing that testing multiple modules writes a newline even the module passes; there's quite a bit of empty space here:

```
$ pkl test test1.pkl test2.pkl
module test
  facts
    ✘ qux
       false (/Users/danielchao/code/apple/pkl/.dan-scripts/test.pkl:8)



75.0% tests pass [1/4 failed], 0.0% asserts pass [1/1 failed]
```

pkl-core/src/main/java/org/pkl/core/stdlib/test/report/MinimalReport.java
stackoverflow: Moved.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/TestReporters.kt
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.commons.cli
+
+enum class TestReporters {
```
bioball:
```suggestion
enum class TestReporter {
```

pkl-core/src/main/java/org/pkl/core/stdlib/test/report/SimpleReporter.java
```diff
@@ -0,0 +1,57 @@
...
+package org.pkl.core.stdlib.test.report;
+
+import java.io.IOException;
+import java.io.Writer;
+import org.pkl.core.TestResults;
+import org.pkl.core.TestResults.TestSectionResults;
+import org.pkl.core.util.AnsiStringBuilder;
+import org.pkl.core.util.StringUtils;
+
+public final class SimpleReporter extends BaseReporter {
```
bioball:
```suggestion
public final class SpecReporter extends BaseReporter {
```

pkl-gradle/src/main/java/org/pkl/gradle/task/ProjectPackageTask.java
```diff
@@ -60,6 +61,10 @@ public abstract class ProjectPackageTask extends BasePklTask {
   @Optional
   public abstract Property<Boolean> getSkipPublishCheck();
 
+  @Input
+  @Optional
+  public abstract Property<String> getReporter();
```
bioball:
```suggestion
  public abstract Property<String> getTestReporter();
```

pkl-gradle/src/main/java/org/pkl/gradle/spec/TestSpec.java
```diff
@@ -22,4 +22,6 @@ public interface TestSpec extends ModulesSpec {
   DirectoryProperty getJunitReportsDir();
 
   Property<Boolean> getOverwrite();
+
+  Property<String> getReporter();
```
bioball:
```suggestion
  Property<String> getTestReporter();
```

docs/modules/pkl-gradle/pages/index.adoc
```diff
@@ -322,6 +322,15 @@ Default: `false` +
 Whether to ignore expected example files and generate them again.
 ====
 
+[[reporter]]
+.reporter: Property<String>
+[%collapsible]
+====
+Default: `"simple"` +
+Example: `reporter = "minimal"` +
+Which test reporter to use for CLI output. Possible values are `"simple"` and `"minimal"`.
```
bioball:
```suggestion
.testReporter: Property<String>
[%collapsible]
====
Default: `"spec"` +
Example: `reporter = "minimal"` +
Which test reporter to use for CLI output. Possible values are `"spec"` and `"minimal"`.
```

docs/modules/pkl-gradle/pages/index.adoc
```diff
@@ -677,6 +686,14 @@ Default: `false` +
 Whether to ignore expected example files and generate them again.
 ====
 
+.reporter: Property<String>
+[%collapsible]
+====
+Default: `"simple"` +
+Example: `reporter = "minimal"` +
+Which test reporter to use for CLI output. Possible values are `"simple"` and `"minimal"`.
```
bioball:
```suggestion
.testReporter: Property<String>
[%collapsible]
====
Default: `"spec"` +
Example: `reporter = "minimal"` +
Which test reporter to use for CLI output. Possible values are `"spec"` and `"minimal"`.
```

pkl-gradle/src/main/java/org/pkl/gradle/task/TestTask.java
```diff
@@ -41,6 +42,10 @@ public abstract class TestTask extends ModulesTask {
   @Input
   public abstract Property<Boolean> getOverwrite();
 
+  @Input
+  @Optional
+  public abstract Property<String> getReporter();
```
bioball:
```suggestion
  public abstract Property<String> getTestReporter();
```

pkl-gradle/src/main/java/org/pkl/gradle/task/TestTask.java
```diff
@@ -54,7 +59,8 @@ protected void doRunTask() {
                 mapAndGetOrNull(getJunitReportsDir(), it -> it.getAsFile().toPath()),
                 getOverwrite().get(),
                 getJunitAggregateReports().getOrElse(false),
-                getJunitAggregateSuiteName().get()),
+                getJunitAggregateSuiteName().get(),
+                TestReporters.valueOf(getReporter().getOrElse("SPEC").toUpperCase())),
```
bioball: We should convert `IllegalArgumentException` into a `InvalidUserDataException`, e.g. something like:

```java
    TestReporters testReporter;
    try {
      testReporter = TestReporters.valueOf(getReporter().getOrElse("SPEC").toUpperCase());
    } catch (IllegalArgumentException e) {
      throw new InvalidUserDataException(
          "Invalid reporter: '%s'. Valid reporter options: %s"
              .formatted(
                  getReporter().get(),
                  TestReporters.getEntries().stream()
                      .map((it) -> it.name().toLowerCase())
                      .collect(Collectors.joining(", "))));
    }
```

Same comment applies to `ProjectPackageTask`

bioball (approved):
Some cleanup needed, but, approving to unblock!

Can you also add docs for the `--test-reporter` CLI flag?

PR #1571, opened 2026-05-04
Fix `pkl:test` fact power assertions when member source section is unavailable

HT154:
Power assertions only work when the source section is available. If it is unavailable, power assertions throw a ParserError (unexpected EOF on an empty input) when re-parsing the expression for presentation.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/runtime/PowerAssertions.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/PowerAssertions.java
@@ -30,6 +30,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
+import org.pkl.core.PklBugException;
 import org.pkl.core.ast.ConstantValueNode;
 import org.pkl.core.ast.expression.member.InferParentWithinMethodNode;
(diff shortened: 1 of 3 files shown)


bioball (approved):

pkl-core/src/main/java/org/pkl/core/runtime/PowerAssertions.java
```diff
@@ -83,6 +84,10 @@ public static void render(
       SourceSection sourceSection,
       Map<Node, List<Object>> trackedValues,
       @Nullable Consumer<AnsiStringBuilder> firstFrameSuffix) {
+    if (!sourceSection.isAvailable()) {
+      throw new PklBugException("Power assertions require an available source section");
+    }
```
HT154, line 89: Added this check to simplify debugging similar issues in the future.

bioball: Use an assertion here instead, e.g.

```java
assert sourceSection.isAvailable() : "Power assertions require an available source section"
```

Assertions are there for conditions that we guarantee to be true, and (ideally) surface via our tests. They get compiled away when we build our executable.

PR #1578, opened 2026-05-11
Added support for external readers in Gradle plugins

netvl:
This PR adds support for configuring external module and resource readers in the Gradle plugin, for all tasks which may need it.

Currently, no external readers are passed to the CliBaseOptions instance which is constructed by every task. Therefore, the only way external readers can be used with Gradle is via PklProject definitions, which sometimes may be limiting. After this change, it will be possible to configure external readers for any tasks, including those which operate on individual modules and not the entire projects.

I'm using NamedDomainObjectContainer to store reader definitions. The name of each reader's definition is passed through into the CLI classes as the map key. Therefore, you can define an external reader like this:

```kotlin
val myPklTask by pkl.operationKind.creating {
    val foo by externalModuleReaders.creating {
        executable = "external-executable"
        arguments = listOf("--bar", "baz")
    }
}
```

and then use `foo` as the URI schema in the Pkl code:

```pkl
import "foo:a/b/c"
```

Properties of each external reader spec are defined as proper properties and are propagated as input dependencies to the task, so the caching and task dependency resolution behaves as expected.

I completely forgot that https://github.com/apple/pkl/pull/1067 exists and reimplemented this from scratch. But that one was closed, my approach to the interface is slightly different and more in line with Gradle recommendations, IMO, and I really need this functionality so I intend to see this through. As in that PR, there are no tests which validate the actual invocation of the external command because creating one requires writing a command which does msgpack, which I'm unsure of how to do properly within this project. If you think this is necessary, I can work on this too, but it would be great if there are some suggestions on how to better tackle this, as I'm not sure how.

Test on [windows]

### DIFF
--- a/pkl-gradle/src/main/java/org/pkl/gradle/spec/BasePklSpec.java
+++ b/pkl-gradle/src/main/java/org/pkl/gradle/spec/BasePklSpec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@ package org.pkl.gradle.spec;
 
 import java.net.URI;
 import java.time.Duration;
+import org.gradle.api.NamedDomainObjectContainer;
 import org.gradle.api.file.ConfigurableFileCollection;
 import org.gradle.api.file.DirectoryProperty;
 import org.gradle.api.provider.ListProperty;
@@ -59,4 +60,8 @@ public interface BasePklSpec {
   ListProperty<String> getHttpNoProxy();
 
   MapProperty<URI, URI> getHttpRewrites();
+
+  NamedDomainObjectContainer<ExternalReaderSpec> getExternalModuleReaders();
+
+  NamedDomainObjectContainer<ExternalReaderSpec> getExternalResourceReaders();
 }
--- a/pkl-gradle/src/test/kotlin/org/pkl/gradle/ExternalReadersTest.kt
+++ b/pkl-gradle/src/test/kotlin/org/pkl/gradle/ExternalReadersTest.kt
@@ -1,0 +1,356 @@
+/*
+ * Copyright © 2026 Apple Inc. and the Pkl project authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.gradle
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.condition.DisabledIf
+
+class ExternalReadersTest : AbstractTest() {
+  companion object {
+    // Adjust paths on Windows to prevent unexpected character escapes.
+    private fun getPathSafeSystemProperty(name: String): String? =
+      System.getProperty(name)?.replace('\\', '/')
+
+    private val externalReaderJar: String? by lazy {
+      getPathSafeSystemProperty("pklGradle.externalReaderJar")
+    }
+
+    private val javaExecutable: String? by lazy {
+      getPathSafeSystemProperty("pklGradle.javaExecutable")
+    }
+
+    @JvmStatic
+    fun systemPropertiesAreNotSet(): Boolean {
+      return externalReaderJar == null || javaExecutable == null
+    }
+  }
+
+  @Test
+  fun `external module readers DSL is accepted`() {
+    writeBuildFile(
+      externalModuleReaders =
+        """
+        externalModuleReaders {
+          myscheme {
+            executable = "/nonexistent/my-reader"
+            arguments = ["--arg1", "--arg2"]
+          }
+        }
+        """
+          .trimIndent()
+    )
+    writePklFile()
+    runTask("evalTest")
+  }
+
+  @Test
+  fun `external resource readers DSL is accepted`() {
+    writeBuildFile(
+      externalResourceReaders =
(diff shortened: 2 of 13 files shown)


netvl:
Pushed a change which builds a jar file which serves as a test external reader, so in tests the external reader executable is configured to the `java` command and the jar file as one of its arguments. Seems to be working :)

netvl:
Fixed windows paths in tests causing issues, please run CI again. Thanks!

HT154 (approved):

pkl-gradle/src/test/kotlin/org/pkl/gradle/ExternalReadersTest.kt
```diff
@@ -0,0 +1,345 @@
+/*
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
```
bioball:
```suggestion
 * Copyright © 2026 Apple Inc. and the Pkl project authors. All rights reserved.
```

pkl-gradle/src/test/kotlin/org/pkl/gradle/ExternalReadersTest.kt
```diff
@@ -0,0 +1,345 @@
...
+class ExternalReadersTest : AbstractTest() {
+  private fun getSystemProperty(name: String): String =
+    System.getProperty(name)
+      .also { assumeTrue(!it.isNullOrBlank(), "$name is not defined") }
+      // Adjust paths on Windows to prevent unexpected character escapes.
+      .replace('\\', '/')
+
+  private val externalReaderJar: String by lazy { getSystemProperty("pklGradle.externalReaderJar") }
+
+  private val javaExecutable: String by lazy { getSystemProperty("pklGradle.javaExecutable") }
```
bioball: The `getSystemProperty` helper doesn't do anything helpful

Also, move these into a companion object and add `systemPropertiesAreNotSet`:

```kt
  companion object {
    private val externalReaderJar: String? by lazy { System.getProperty("pklGradle.externalReaderJar") }

    private val javaExecutable: String? by lazy { System.getProperty("pklGradle.javaExecutable") }

    @JvmStatic
    fun systemPropertiesAreNotSet(): Boolean {
      return externalReaderJar == null || javaExecutable == null
    }
  }
```

Then, disable the failing tests based on this, e.g.

```kt
  @DisabledIf("systemPropertiesAreNotSet")
  @Test
  fun `external resource reader reads and uppercases content`() {
```

The two failing tests are "external resource reader reads and uppercases content", and "external resource reader reads and uppercases content"

pkl-gradle/src/main/java/org/pkl/gradle/spec/BasePklSpec.java
```diff
@@ -59,4 +60,8 @@ public interface BasePklSpec {
   ListProperty<String> getHttpNoProxy();
 
   MapProperty<URI, URI> getHttpRewrites();
+
+  NamedDomainObjectContainer<ExternalReaderSpec> getExternalModuleReaders();
+
+  NamedDomainObjectContainer<ExternalReaderSpec> getExternalResourceReaders();
```
bioball, line 66: This actually feels a little counter-intuitive to me. Everything else is `ListProperty`, `MapProperty`; why make this one a `NamedDomainObjectContainer`?

pkl-gradle/src/main/java/org/pkl/gradle/utils/ProviderUtils.java
```diff
@@ -0,0 +1,36 @@
...
+   * Equivalent to {@code provider.map(it -> f.apply(it)).getOrNull()}.
+   *
+   * <p>This function is necessary because in some cases doing {@code
+   * someProvider.map(...).getOrNull()} may trigger validation errors inside Gradle, when {@code
+   * someProvider} is derived from a property.
+   */
+  public static <T, U> @Nullable U mapAndGetOrNull(Provider<T> provider, Function<T, U> f) {
+    @Nullable T value = provider.getOrNull();
+    return value == null ? null : f.apply(value);
+  }
```
bioball: Move this to `PluginUtils` (no need for a new utilities class)

bioball (changes requested):
Some minor nits, but also: not totally sure about using `NamedDomainObjectContainer `?

pkl-gradle/src/main/java/org/pkl/gradle/spec/BasePklSpec.java
netvl: If you take a closer look at all of these list properties, map properties, etc you will see that they all contain "simple" values, primarily strings (and URIs which are basically strings). For such values, it is idiomatic to use collection properties.

For more "complex" objects, it is idiomatic to use nested managed objects, and the primary way to do this is via domain object containers. Since in this case the name of the external reader is important, relying on the named variant of the domain object container is basically a natural conclusion.

pkl-gradle/src/main/java/org/pkl/gradle/utils/ProviderUtils.java
netvl: good point, fixed!

pkl-gradle/src/test/kotlin/org/pkl/gradle/ExternalReadersTest.kt
netvl: Fixed

pkl-gradle/src/test/kotlin/org/pkl/gradle/ExternalReadersTest.kt
netvl: It does fix paths on windows (there were failing tests because of that).
Anyway, fixed according to your suggestions (while keeping the path fixing logic), please take a look :)

pkl-gradle/src/main/java/org/pkl/gradle/spec/BasePklSpec.java
netvl: Also, just in case - `NamedDomainObjectContainer` is an equally valid piece of Gradle API which is supposed to be used for defining extension and task properties, it is not some kind of special API for certain purposes, or for internal use. So it is totally correct to use it when the situation needs it, and this is such a case, according to my Gradle knowledge and experience :)

pkl-gradle/src/test/kotlin/org/pkl/gradle/ExternalReadersTest.kt
bioball: Oh, my bad!

pkl-gradle/src/main/java/org/pkl/gradle/spec/BasePklSpec.java
bioball: Okay, gotcha!

bioball (approved):

netvl:
Fixed formatting, should be green now

PR #1581, opened 2026-05-11
Add resource readers from service providers in CLI

netvl:
This omission, in particular, prevents Gradle plugins (which rely on CLI classes) from adding custom resource readers via the service loading mechanism. This change seems benign, especially since this is already done for module key factories.


HT154 (approved):

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/CliCommand.kt
```diff
@@ -310,6 +310,7 @@ abstract class CliCommand(protected val cliOptions: CliBaseOptions) {
       add(ResourceReaders.file())
       add(ResourceReaders.http())
       add(ResourceReaders.https())
+      addAll(ResourceReaders.fromServiceProviders())
```
bioball: Move this one to right about `add(ResourceReaders.file()))` to match the precedence rules that we use for module keys

netvl: Fixed

bioball (approved):

PR #1584, opened 2026-05-13
Improve HTTP headers logic

bioball:
* Relax forbidden headers constraints
  - remove restriction on browser-related headers
  - allow any glob pattern (no need to end with `/` or `*`, because glob patterns already require users to explicitly declare prefix matches if that's the intention)
* Replace `List<Pair<, ...>>`; use `Map<String, ...>` instead
* Use glob pattern strings as an API throughout, instead of `Pattern` (e.g. in `HttpClientBuilder`)
* Add HTTP headers to message passing API
* Add HTTP headers to executor API (introduces `ExecutorSpiOptions4`)
* Add tests for Gradle, CLI, and pkl-executor invocations
* Improve documentation
* Add `isGlobPattern` API to class `String` for in-language validation of http headers
* Behavior change: make sure explicitly configured `User-Agent` in `HttpClientBuilder` can be shadowed by headers (allows users to set `--http-header "**=User-Agent: My User Agent"` and for this to be the only user agent).

CC @kyokuping

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/http/HttpClient.java
+++ b/pkl-core/src/main/java/org/pkl/core/http/HttpClient.java
@@ -158,6 +158,19 @@ public interface HttpClient extends AutoCloseable {
      * <p>This method clears all existing headers and replaces them with the contents of the
      * provided map.
      *
+     * <p>{@code headerRules} is a map whose keys are <a
+     * href="https://pkl-lang.org/main/current/language-reference/index.html#glob-patterns">glob
+     * patterns</a>, and values is a map of header names and values. Multiple header values turn
+     * into multiple individual headers in the HTTP request.
+     *
+     * <p>To add headers to all requests, use {@code **} as the glob pattern.
+     *
+     * <p>To describe a prefix match, add {@code **} to the glob pattern (e.g. {@code
+     * https://example.com/**}).
+     *
+     * <p>Before an HTTP request is made, each key is matched against the request URL. If any
+     * matches are found, each of their headers are added to the request.
+     *
      * @throws IllegalArgumentException if any of the keys are invalid glob patterns, or if any of
      *     the header names or values are invalid.
      * @since 0.32.0
@@ -170,6 +183,7 @@ public interface HttpClient extends AutoCloseable {
      * @throws IllegalArgumentException if {@code globPattern} is an invalid glob pattern, or if any
      *     of the header names or values are invalid.
      * @since 0.32.0
+     * @see Builder#setHeaders
      */
     Builder addHeaders(String globPattern, Map<String, List<String>> headers);
 
--- a/pkl-core/src/main/java/org/pkl/core/http/HttpClient.java
+++ b/pkl-core/src/main/java/org/pkl/core/http/HttpClient.java
@@ -23,10 +23,8 @@ import java.net.http.HttpTimeoutException;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
-import java.util.regex.Pattern;
 import javax.net.ssl.SSLContext;
 import org.jspecify.annotations.Nullable;
-import org.pkl.core.Pair;
 
 /**
  * An HTTP client.
@@ -47,6 +45,9 @@ public interface HttpClient extends AutoCloseable {
      * Sets the {@code User-Agent} header.
      *
      * <p>Defaults to {@code "Pkl/$version ($os; $flavor)"}.
+     *
+     * <p>An existing "User-Agent" from {@link Builder#setHeaders} and {@link Builder#addHeaders}
+     * takes precedence over this field.
      */
     Builder setUserAgent(String userAgent);
 
@@ -146,7 +147,6 @@ public interface HttpClient extends AutoCloseable {
     /**
      * Adds a rewrite rule.
      *
-     * @see Builder#setRewrites(Map)
      * @throws IllegalArgumentException if {@code sourcePrefix} or {@code targetPrefix} is invalid.
      * @since 0.29.0
      */
@@ -157,8 +157,21 @@ public interface HttpClient extends AutoCloseable {
      *
      * <p>This method clears all existing headers and replaces them with the contents of the
      * provided map.
+     *
+     * @throws IllegalArgumentException if any of the keys are invalid glob patterns, or if any of
+     *     the header names or values are invalid.
+     * @since 0.32.0
      */
-    Builder setHeaders(List<Pair<Pattern, List<Pair<String, String>>>> headers);
+    Builder setHeaders(Map<String, Map<String, List<String>>> headerRules);
+
+    /**
+     * Adds HTTP headers for URL requests that match {@code globPattern}.
+     *
+     * @throws IllegalArgumentException if {@code globPattern} is an invalid glob pattern, or if any
+     *     of the header names or values are invalid.
+     * @since 0.32.0
+     */
+    Builder addHeaders(String globPattern, Map<String, List<String>> headers);
 
     /**
      * Creates a new {@code HttpClient} from the current state of this builder.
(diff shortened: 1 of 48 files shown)


pkl-core/src/main/java/org/pkl/core/http/HttpClient.java
```diff
@@ -157,8 +157,17 @@ interface Builder {
...
+     */
+    Builder setHeaders(Map<String, Map<String, List<String>>> headerRules);
+
+    /**
+     * Adds a header rule.
+     *
+     * @throws IllegalArgumentException if {@code globPattern} is invalid.
      */
-    Builder setHeaders(List<Pair<Pattern, List<Pair<String, String>>>> headers);
+    Builder addHeader(String globPattern, Map<String, List<String>> headers);
```
HT154: Naming is a little odd here given that this adds a single rule, not a single header. Maybe something closer to "add one header" would be better? Not a blocker, but curious about your thoughts.

```suggestion
    Builder addHeader(String globPattern, String name, String value);
```

pkl-server/src/test/kotlin/org/pkl/server/AbstractServerTest.kt
```diff
@@ -1100,6 +1112,37 @@ abstract class AbstractServerTest {
       .contains("Rewrite rule must end with '/', but was 'https://example.com'")
   }
 
+  @Nested
+  @WireMockTest
+  inner class HttpTests {
```
bioball, line 1117: Use nested class to prevent spinning up sockets for every test here

pkl-core/src/main/java/org/pkl/core/messaging/AbstractMessagePackEncoder.java
```diff
@@ -70,6 +70,21 @@ protected void packMapHeader(
         size + (value1 != null ? 1 : 0) + (value2 != null ? 1 : 0) + (value3 != null ? 1 : 0));
   }
 
+  protected void packMapHeader(
```
HT154, line 65: The 3-value method above this one is now unused and should be removed

HT154 (approved):
Nice!! A couple small things; approving to unblock.

pkl-core/src/main/java/org/pkl/core/http/HttpClient.java
bioball: Yeah, the names are a bit odd, since we call the enclosing 2D map "headers" and also call these `headers`.

I renamed this to `addHeaders`; does that make more sense?

HT154: Yeah, that seems reasonable to me

pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
```diff
@@ -225,14 +235,28 @@ private URI rewriteUri(URI uri) {
     return ret;
   }
 
+  private boolean matches(Pattern pattern, URI uri) {
+    // optimization: "**" always matches, no need to execute regex
+    if (pattern == IoUtils.doubleStarGlob) {
```
stackoverflow, line 242: nit: I'd add a comment here explaining why the `==`. For someone who just happen upon this code, it seems like a bug.

pkl-executor/src/main/java/org/pkl/executor/ExecutorOptions.java
```diff
@@ -215,6 +219,18 @@ public Builder httpRewrites(Map<URI, URI> httpRewrites) {
       return this;
     }
 
+    /**
+     * API equivalent of the {@code --http-header} CLI option.
+     *
+     * <p>This option is ignored on Pkl 0.32 and older.
```
stackoverflow:
```suggestion
     * <p>This option is ignored on Pkl 0.31 and older.
```

pkl-core/src/main/java/org/pkl/core/http/HttpClientBuilder.java
```diff
@@ -41,7 +43,9 @@ final class HttpClientBuilder implements HttpClient.Builder {
   private int testPort = -1;
   private ProxySelector proxySelector;
   private Map<URI, URI> rewrites = new HashMap<>();
-  private List<Pair<Pattern, List<Pair<String, String>>>> headers = new ArrayList<>();
+  // okay to use Pattern as a map key here because `GlobResolver.toRegexPattern()` caches and
+  // gives the same `Pattern` instance for an existing glob pattern.
+  private Map<Pattern, Map<String, List<String>>> headers = new HashMap<>();
```
stackoverflow: nit: the previous code used a List, which preserved header order. `HashMap` won't.

stdlib/EvaluatorSettings.pkl
```diff
@@ -243,50 +276,63 @@ class ExternalReader {
...
-local const httpHeaderValueRegex = Regex("^[\\t\\u0020-\\u007E\\u0080-\\u00FF]*$")
-
+/// An HTTP header value.
+///
+/// It consists of ASCII characters between the range of 0x20 - 0x7E, and 0x80 - 0xFF (all visible
+/// ASCII characters), and also the tab character.
 @Since { version = "0.32.0" }
-typealias HttpHeaderValue = String(!httpHeaderValueRegex.findMatchesIn(this).isEmpty)
+typealias HttpHeaderValue =
+  String(matches(Regex(#"[\t\u0020-\u007E\u0080-\u00FF]*"#)), length < 4096)
```
stackoverflow, line 342: nit: we enforce `length < 4096` on the Pkl side, but the Java side is unconstrained. One can pass a 10MB header via CLI.

stdlib/EvaluatorSettings.pkl
```diff
@@ -243,50 +276,63 @@ class ExternalReader {
...
-  ReservedHttpHeaderPrefix.any((it) -> header.startsWith(it))
+local const hasValidHeaderNameSyntax = (header: String) ->
+  header.matches(Regex(#"[a-zA-Z0-9!#$%&'*+-.^_`|~]+"#))
 
-local const httpHeaderNameRegex = Regex("^[a-zA-Z0-9!#\\$%&'*+-.^_`|~]+$")
-local const hasValidHttpHeaderName = (header: String) ->
-  !httpHeaderNameRegex.findMatchesIn(header).isEmpty
+/// A mapping of headers name to header value(s).
+///
+/// The mapping must have distinct header keys (case insensitive).
```
stackoverflow, line 311: nit: the Java side has no such limitation (case insensitive), right?

stdlib/EvaluatorSettings.pkl
```diff
@@ -243,50 +276,63 @@ class ExternalReader {
...
 
-local const hasReservedHttpHeaderPrefix = (header: String) ->
-  ReservedHttpHeaderPrefix.any((it) -> header.startsWith(it))
+local const hasValidHeaderNameSyntax = (header: String) ->
+  header.matches(Regex(#"[a-zA-Z0-9!#$%&'*+-.^_`|~]+"#))
 
-local const httpHeaderNameRegex = Regex("^[a-zA-Z0-9!#\\$%&'*+-.^_`|~]+$")
-local const hasValidHttpHeaderName = (header: String) ->
-  !httpHeaderNameRegex.findMatchesIn(header).isEmpty
+/// A mapping of headers name to header value(s).
```
stackoverflow:
```suggestion
/// A mapping of header names to header values.
```

stackoverflow (approved):
Looks good. Some small things.

pkl-core/src/main/java/org/pkl/core/http/HttpClientBuilder.java
bioball: Ah, good point, we might as well keep this the same. I'll switch this to `LinkedHashMap`.

stdlib/EvaluatorSettings.pkl
bioball: Yeah, this is one case where we don't enforce the same rules.

We're stricter in Pkl because we want to design the type to work well with object amends; e.g. this should throw if you are trying to reassign the user-agent header, but it already exists with different casing:

```pkl
headers = (baseHeaders) {
  ["user-agent"] = "Bob"
}
```

It's probably not appropriate to enforce this in our lower level API because:

1. It means users of pkl-config-java, pkl-swift, pkl-go, etc, need to verify this when setting headers
2. It falls through to setting the header multiple times, which is how Java's HttpClient already works.

PR #1590, opened 2026-05-14
pkl-doc: Fix/improve Executor handling in DocGenerator

### LINKED ISSUE #1583: pkl-doc: DocGenerator never shuts down its thread pool
Every time `DocGenerator.run()` is called, a new thread pool (`Executor`) is created, but it never seems to be shut down.
The created thread pool also cannot be shut down by the caller because it is not accessible.
The docs say "the default executor", but there does not seem to be a way to replace it.

To be a well-behaved library, `DocGenerator` or its `run()` method should probably accept an `Executor`.
Any `Executor` implicitly created by `run()` should probably be shut down before returning.

`CliDocGenerator` probably suffers from the same problem.

odenix:
run() now creates and closes a default Executor per call. This is fine because there is no good reason to call this method multiple times.

run(Executor) now lets callers provide their own Executor, which is customary for a well-behaved library.

Also: Fix IntelliJ warning by calling toSet()

Closes #1583

### DIFF
--- a/pkl-doc/src/main/kotlin/org/pkl/doc/DocGenerator.kt
+++ b/pkl-doc/src/main/kotlin/org/pkl/doc/DocGenerator.kt
@@ -22,7 +22,9 @@ import java.net.URI
 import java.nio.file.Files
 import java.nio.file.Path
 import java.util.concurrent.Executor
+import java.util.concurrent.ExecutorService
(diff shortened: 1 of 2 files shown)


HT154 (approved):
Thanks!

PR #1592, opened 2026-05-14
pkl-doc: Support single-package docsite mode

odenix:
When a docsite has only one package name and no DocsiteInfo.overview, treat it like Javadoc's single-module output: redirect the top-level index to the package page and omit the site-title breadcrumb segment from generated pages.

Add src/test/files/SinglePackageTest fixtures to cover multiple package versions, redirect behavior, breadcrumb behavior, and unchanged site structure.

Also:
- Shut down Executor used in test.
- Declare expected output fixtures of DocGenerator as test inputs, not outputs.
- Fix IntelliJ warning by using a Set for the right-hand side of collection subtraction.


HT154 (approved):
This looks good to me! I'd like @bioball or @stackoverflow take a look before merging.

bioball (changes requested):
Seems a little strange to render a package index page just because there's doc comments in the `docsite-info.pkl`.

How about we just render the overview on the package page if an overview exists?

I think this should do the trick:

```diff
diff --git a/pkl-doc/src/main/kotlin/org/pkl/doc/PackagePageGenerator.kt b/pkl-doc/src/main/kotlin/org/pkl/doc/PackagePageGenerator.kt
index 5b66c5f9..ee83f3f4 100644
--- a/pkl-doc/src/main/kotlin/org/pkl/doc/PackagePageGenerator.kt
+++ b/pkl-doc/src/main/kotlin/org/pkl/doc/PackagePageGenerator.kt
@@ -67,6 +67,21 @@ internal class PackagePageGenerator(
           Triple("Modules", "#_modules", docPackage.hasListedModule),
         )
 
+        if (isSinglePackageSite && docsiteInfo.overview != null) {
+          renderAnchor("_overview")
+
+          val memberDocs =
+            MemberDocs(docsiteInfo.overview, pageScope, listOf(), isDeclaration = true)
+
+          div {
+            id = "_declaration"
+            classes = setOf("member")
+
+            memberDocs.renderExpandIcon(this)
+            memberDocs.renderDocComment(this)
+          }
+        }
+
         renderAnchor("_overview")
         div {
           id = "_declaration"

```

odenix:
> How about we just render the overview on the package page if an overview exists?

What's the upside? Right now, users have full control:

* no or only package docs -> single-package docsite
* add docsite overview -> multi-package docsite

bioball:
It feels like quite hidden behavior; if you add a doc comment, you get a package index page. If not, you get a redirect to that one package.

If you just have one package, it always feels like the better experience to just be redirected to that one package page; why would you ever want the index page?

odenix:
> If you just have one package, it always feels like the better experience to just be redirected to that one package page; why would you ever want the index page?

Because you're starting a larger project and don't want to conflate site-wide docs with package docs. Otherwise, why would you add `docsite-info.pkl`?

Perhaps I'm missing something, but having docsite info implicitly turn into package info until a second package is added feels more magical to me (replace/prepend/append?).

bioball:
Oh, I forgot that `docsite-info.pkl` is an optional input. Sidenote: our website incorrectly states that it's required: https://github.com/apple/pkl/blob/a7a64acbac16392a5e280828532b021139e80872/docs/modules/pkl-doc/pages/index.adoc?plain=1#L207 (https://github.com/apple/pkl/pull/1598 to close).

This makes sense to me, then!

bioball (approved):

PR #1595, opened 2026-05-15
Improve handling of evaling dependency notation URIs

bioball:
* Move logic of resolving these URIs to `EvaluatorImpl`
  - Extends support for evaluting these to message passing API, Java API, etc
* Remove limitation where only remote dependencies can be evaluated
* Treat `@`-prefixed inputs as dependency notation unambiguously

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/EvaluatorImpl.java
+++ b/pkl-core/src/main/java/org/pkl/core/EvaluatorImpl.java
@@ -16,6 +16,8 @@
 package org.pkl.core;
 
 import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.file.Path;
 import java.time.Duration;
 import java.util.Collection;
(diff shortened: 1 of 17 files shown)


pkl-core/src/main/java/org/pkl/core/EvaluatorImpl.java
```diff
@@ -424,10 +435,37 @@ private <T> T doEvaluate(Supplier<T> supplier) {
     return evalResult;
   }
 
+  /** Resolve dependency notation URIs (e.g. `@foo/bar.pkl`) to its resolved absolute URI. */
+  private ModuleSource normalizeModuleSource(ModuleSource moduleSource) {
+    if (moduleSource.getContents() == null
+        && !moduleSource.getUri().isAbsolute()
+        && moduleSource.getUri().getPath().startsWith("@")) {
```
HT154: Inverting this condition for an early return would make this a little easier to scan

HT154 (approved):
Very nice!!

PR #1597, opened 2026-05-19
Test reporter fixes

bioball:
* Fix error message when an invalid test reporter is supplied in Gradle
* Fix Gradle property name in docs
* Fix Gradle property name in tasks
* Introduce `TestReporter.default`, and use it in places where default is applied
* Remove calls to `convention()`; this is not required because the input is optional anyways.


HT154 (approved):
LGTM

PR #1601, opened 2026-05-21
pkl-core: Migrate nullness to JSpecify

odenix:
Replace pkl-core's local nullness annotations with JSpecify annotations.
Enable NullAway checking for pkl-core packages except org.pkl.core.ast and org.pkl.core.stdlib.

Notable code changes:
- Add a dedicated late-init constructor to VmTyped
- Move VmExceptionBuilder's fallback message derivation from withCause() to build()
- Split VmException rendering between builder-provided messages and string-backed messages
- Initialize MessageTransport handlers with default throwing handlers
- Update JSON helper collection types to allow nullable values JSON arrays and objects can contain JSON null,
   so the Java Map/List element types need to model nullable elements explicitly
- Make public command transform APIs accept nullable transformed values 
   Command transforms can produce null for optional/default handling,
   so the BiFunction and options-map element types now model that explicitly
- Make ExecutorSpiException accept nullable message and cause 
   Existing call sites can pass nullable causes from Throwable.getCause()
- Remove JSR-305 semantics from `@LateInit`
   JSpecify does not support the same type-qualifier-nickname pattern,
   so `@LateInit` is now documentation plus a NullAway constructor-initialization exemption

Out of scope:
- NullAway checking of org.pkl.core.ast and org.pkl.core.stdlib
- IntelliJ warnings related to `@LateInit` fields
- Removing the JSR-305 dependency, since concurrency annotations are still in use


HT154 (approved):
LGTM!

bioball (approved):
LGTM, this is great, thank you!

PR #1606, opened 2026-05-21
Configure IntelliJ to respect `@LateInit` annotations

bioball:
IntelliJ can understand that some annotations on fields mean that they are implicitly initialized, which means we don't get the "field XXX is not initialized" warning for `@LateInit` fields.

This setting, unfortunately, is recorded into `.idea/misc.xml`, which contains a bunch of arbitrary stuff that we don't want to check into source control

This adds some logic to touch up that file to mark `@LateInit` as implicitly initialized fields, so we don't get any editor warnings.

Also, suppress some warnings.

CC @odenix


HT154 (approved):
Nice!

PR #1608, opened 2026-05-22
Remove JSR-305 dependency

odenix:
- Remove single usage of `@Immutable` without replacement
- Remove HttpClient's usages of `@ThreadSafe` without replacement
- Replace javax.annotation.concurrent.GuardedBy with org.pkl.core.util.GuardedBy

Also:
- Remove redundant final modifiers from members of a final class


HT154 (approved):
Very nice!!

pkl-core/src/main/java/org/pkl/core/util/GuardedBy.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Indicates that accessing the annotated field requires holding the lock object stored in the
+ * specified field.
+ *
+ * <p>This annotation is for documentation purposes only.
+ */
+@Target(ElementType.FIELD)
+@Retention(RetentionPolicy.SOURCE)
+public @interface GuardedBy {
```
bioball, line 31: I don't think it's an improvement to replace this with our own annotation.

If we want to replace this, we should use another annotation that is understood by errorprone (they recommend just using `com.google.errorprone.annotations.concurrent.GuardedBy`).

odenix: It's an improvement because it removes the JSR-305 dependency.
I can look into `com.google.errorprone.annotations.concurrent.GuardedBy` as part of turning on Error Prone checks, which is next on my list.

bioball: I get that JSR-305 is a dormant library that we don't want, but, replacing it with an annotation that we'll need to replace immediately afterwards feels incorrect.

I'd be open to just replacing this with the errorprone annotation, though.

odenix: > replacing it with an annotation that we'll need to replace immediately afterwards feels incorrect.

To me, this is an open question. Depending on errorprone-annotations will cause some of the same headaches that depending on jsr-305 did:
* We probably don't want to make it an `api` or `implementation` dependency.
* If we make it a `compileOnly` dependency, 
   * consumers of Pkl libraries may get the same "unknown enum constant" etc. javac warnings as with jsr-305, which are quite annoying.
   * Consumers compiling their own code with `-Werror` might even be forced to add errorprone-annotations to their own compile classpath.
   * Annotation usages will show as errors in IDEs of consuming projects.

If Pkl was an application, none of this would be a concern.

bioball: Mm, I don't think `compileOnly` is too big of a concern here:

1. `javac` ignores unknown annotations not on the classpath
2. `GuardedBy` annotations are all on private fields, not public APIs

Pkl's users _shouldn't_ be seeing any compiler warnings here.

odenix: ad 1. At least in some cases, javac warns about them, and `-Werror` turns warnings into errors. This is definitely the case for jsr-305.
ad 2. This _may_ save us as long as we only use Error Prone's `GuardedBy` and not `ThreadSafe` etc.

`compileOnly`'s IDE issue is unavoidable, and will leave consumers with a "something is broken here" feeling.

For `GuardedBy` specifically, I think that adding the dependency isn't worth it. I'd rather forgo the checks.
If the goal is to double down on Error Prone annotations, they may be worth the consumer pain.

bioball: Yeah, I think we should also enable the error prone check. I just tried running that check locally, and found a bunch of issues around unsafe access!

odenix: OK, I'll leave this PR and Error Prone follow-ups to you.

PR #1611, opened 2026-05-22
Remove jsr305; switch GuardedBy to com.google.errorprone

bioball:
Closes #1608 

Motivation: jsr305 is now a dormant package and will never be adopted by Java (and contributes to the `javax` package, which is supposed to be reserved). OTOH, com.google.errorprone is actively maintained.


HT154 (approved):

odenix:
@bioball We should try to get the most out of errorprone-annotations by keeping `Immutable` and `ThreadSafe`.

PR #1615, opened 2026-05-25
Fix incorrect Facts in pkl:base doc comments

adityasingh2400:
Several `Facts:` examples in the `pkl:base` standard library docs assert statements that are false when evaluated. Since these examples are not run by any test, the mistakes went unnoticed and are rendered verbatim into the generated API docs, where they mislead readers.

The corrected examples:

- `String.isNotBlank`: `"\t\n\r".isNotBlank` was listed as holding, but a string of only whitespace is blank, so it is false. Negated it to match the neighboring `!"".isNotBlank` and `!"   ".isNotBlank` examples.
- `DataSize.toBinaryUnit` / `toDecimalUnit`: the `mb`/`mib` lines mirrored the `kb`/`kib` lines, but the identity only holds for adjacent units. `1024.kb == 1000.kib` (both 1,024,000 b), whereas `1024.mb` is 1,024,000,000 b and `1000.mib` is 1,048,576,000 b, so they are not equal. There is no clean round-number equivalent at this magnitude, so I removed the two false lines; the remaining examples still demonstrate the conversion.
- `Collection.any`: `!List(1, 2, 3).any((n) -> n.isEven)` is false because 2 is even. Changed the list to `List(1, 3, 5)` so the negation holds.
- `IntSeq.end`: the example read `IntSeq(2, 5).start == 5`, which documents the wrong property and is false (`start` is 2). Corrected it to `IntSeq(2, 5).end == 5`.
- `List.isDistinctBy` / `distinctBy`: `List("a", "b", "abc")` is not distinct by length, since `"a"` and `"b"` both have length 1. Switched to `List("a", "bb", "ccc")` so the distinctness examples hold.
- `Map`: `Map(...).values` returns a `List`, not a `Set`. Corrected the expected type.

I verified that each corrected example evaluates to `true`, and that the neighboring examples I kept still pass, using the released Pkl 0.31.1 binary. The changes are confined to doc-comment text, so formatting is unaffected.

### DIFF
--- a/stdlib/base.pkl
+++ b/stdlib/base.pkl
@@ -1314,7 +1314,7 @@ external class String extends Any {
   /// ```
   /// !"".isNotBlank
   /// !"   ".isNotBlank
-  /// "\t\n\r".isNotBlank
+  /// !"\t\n\r".isNotBlank
   /// "abc".isNotBlank
   /// ```
   @Since { version = "0.31.0" }
@@ -1904,7 +1904,6 @@ external class DataSize extends Any {
   /// Facts:
   /// ```
   /// 1024.kb.toBinaryUnit() == 1000.kib
-  /// 1024.mb.toBinaryUnit() == 1000.mib
   ///
   /// 1000.kib.toBinaryUnit() == 1000.kib
   /// 1000.b.toBinaryUnit() == 1000.b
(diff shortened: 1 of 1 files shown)


adityasingh2400:
Withdrawing this one: these are correct but doc-only Facts corrections in the stdlib, which are better handled as a single maintainer-side cleanup than a one-off external PR.

bioball:
@adityasingh2400 this is a good PR, why don't you re-open this?

adityasingh2400:
Thanks @bioball, happy to. I had withdrawn this one thinking it was better as a maintainer-side cleanup, but if you would like it as a PR I am glad to carry it. The fork that backed this branch was deleted in the meantime, so GitHub would not let me reopen #1615 directly. I have restored the exact same change in #1669. No changes beyond what you reviewed here.

PR #1618, opened 2026-05-26
Fixes #1617 : Use Locale.ROOT when lowercasing rewrite URIs

### LINKED ISSUE #1617: HTTP rewrite fails with PklBugException under tr-TR locale when host contains 'I'
RequestRewritingClient will throw a PklBugException (Unreachable code) when an URL being rewritten has a host containing the letter `I` and the JVM default locale is Turkish, or actually also Azerbaijani (AZ).

**How to reproduce**

1. Configure a rewrite rule on the HTTP client with a URI with an `i` (example: `https://github.com`)
2. Evaluate a module that imports or reads a URL whose host contains a capital `I`, (example: `import "https://GITHUB.COM/shared.pkl"`).
3. Run the JVM with `-Duser.language=tr -Duser.country=TR`

**Actual**
```
org.pkl.core.PklBugException: Unreachable code.
      at org.pkl.core.http.RequestRewritingClient.normalizeRewrite(RequestRewritingClient.java:217)
      at org.pkl.core.http.RequestRewritingClient.rewriteUri(...)
```

**Expected**

URL is rewritten and fetched.

mirkoalicastro:
Use Locale.ROOT to apply the lowecase format. For URI scheme and host locale-neutral casing is the semantically the correct choice. Added a unit test that sets the default locale to tr-TR and that would fail without the fix.

Fixes #1617 .

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
+++ b/pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
@@ -26,6 +26,7 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


bioball (approved):
LGTM, thank you!

PR #1620, opened 2026-05-26
Fix calls to string case api

bioball:
* Enable IntelliJ inspection for calls to `String.toLowerCase()` and `String.toUpperCase()`
* Enable error prone check
* Fix all issues

### DIFF
--- a/build-logic/src/main/kotlin/pklJSpecify.gradle.kts
+++ b/build-logic/src/main/kotlin/pklJSpecify.gradle.kts
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import net.ltgt.gradle.errorprone.CheckSeverity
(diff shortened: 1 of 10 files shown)


HT154 (approved):
Good call! It'd be great if we had a way to block this sort of thing in CI as well, but that'd likely require adopting additional build tooling.

bioball:
We can add this as an error prone check!

I'll go ahead and add it.

build-logic/src/main/kotlin/pklJSpecify.gradle.kts
```diff
@@ -36,6 +37,7 @@ nullaway { onlyNullMarked = true }
 
 tasks.withType<JavaCompile>().configureEach {
   options.errorprone.disableAllChecks = true
+  options.errorprone.check("StringCaseLocaleUsage", CheckSeverity.ERROR)
```
HT154, line 40: Excellent!

PR #1621, opened 2026-05-26
Enable error-prone check for GuardedBy, fix errors

bioball:


pkl-core/src/main/java/org/pkl/core/module/ModulePathResolver.java
```diff
@@ -139,30 +140,29 @@ public void close() {
...
+        assert fileCache != null;
+        fileCache.putIfAbsent(IoUtils.toNormalizedPathString(relativized), path);
+        assert cachedPathElementRoot != null;
+        var element = cachedPathElementRoot;
+        for (var i = 0; i < relativized.getNameCount(); i++) {
+          var name = relativized.getName(i).toString();
+          var isDirectory = i < (relativized.getNameCount() - 1);
+          element = element.putIfAbsent(name, new TreePathElement(name, isDirectory));
+        }
+      }
```
bioball, line 165: Switch to using simple for loop so that error-prone knows that `fileCache` is accessed while the lock is still held.

pkl-core/src/main/java/org/pkl/core/externalreader/ExternalReaderProcessImpl.java
```diff
@@ -96,43 +93,48 @@ private MessageTransport getTransport() throws ExternalReaderProcessException {
...
-    return transport;
   }
 
   /**
    * Runs the underlying message transport so it can receive responses from the child process.
    *
    * <p>Blocks until the underlying transport is closed.
    */
-  private void runTransport() {
+  private void runTransport(MessageTransport transport) {
```
bioball, line 137: This method shouldn't try to obtain a lock because it's called from inside a thread while another thread is holding the lock. So, this is moved to accept a reference to the transport instead.

pkl-core/src/main/java/org/pkl/core/externalreader/ExternalReaderProcessImpl.java
```diff
@@ -51,13 +50,11 @@ final class ExternalReaderProcessImpl implements ExternalReaderProcess {
...
 
-  @LateInit
   @GuardedBy("lock")
-  private Process process;
+  private @Nullable Process process;
 
-  @LateInit
   @GuardedBy("lock")
-  private MessageTransport transport;
+  private @Nullable MessageTransport transport;
```
bioball, line 57: Unrelated improvement: the `@LateInit` annotation was causing plenty of superfluous warnings around null assertions. Switching to `@Nullable` makes everything a little simpler.

HT154 (approved):

PR #1622, opened 2026-05-26
Optimize let expressions as frame slot vars

bioball:
This introduces an optimization to write let expressions to frame slots of the current frame, rather than transforming them into lambdas. As a result, let expressions are _much_ faster to execute.

## Benchmarks

I'm seeing that jpkl is about 12x faster, and native Pkl is about 4000x faster:

| variant | results (mean) |
| --- | --- |
| jpkl | 1208.06.ns -> 97.76.ns |
| pkl | 1275.96.ns -> 0.32.ns |

Here is the code used by this benchmark:

```pkl
amends "pkl:Benchmark"

microbenchmarks {
  ["let"] {
    expression =
      let (a = 1)
      let (b = 2)
      let (c = 3)
      let (d = 4)
      let (e = 1)
      let (f = 2)
      let (g = 3)
      let (h = 4)
      let (i = 1)
      let (j = 2)
      let (k = 3)
      let (l = 4)
      let (m = 1)
      let (n = 2)
      let (o = 3)
      let (p = 4)
      let (q = 1)
      let (r = 2)
      let (s = 3)
      let (t = 4)
        a + b + c + d + e + f + g + h + i + j + k + l + m + n + o + p + q + r + s + t
  }
}

output {
  renderer {
    converters {
      [Duration] = (it: Duration) -> super[Duration].apply(it.toUnit("ns"))
    }
  }
}
```

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
@@ -890,7 +890,6 @@ public class AstBuilder extends AbstractAstBuilder<Object> {
     Node child = expr;
     var parent = expr.parent();
     var scope = symbolTable.getCurrentScope();
-    var levelsUp = 0;
 
     while (parent instanceof IfExpr
         || parent instanceof TraceExpr
@@ -925,8 +924,7 @@ public class AstBuilder extends AbstractAstBuilder<Object> {
               : new InferParentWithinMethodNode(
                   createSourceSection(expr.newSpan()), language, scopeName, new GetOwnerNode());
     } else if (parent instanceof LetExpr letExpr && letExpr.getBindingExpr() == child) {
-      // TODO (unclear how to infer type now that let-expression is implemented as lambda
-      // invocation)
+      // TODO correctly infer parent, e.g. `let (x: Person = new {}) ...`
       throw exceptionBuilder()
           .evalError("cannotInferParent")
           .withSourceSection(createSourceSection(expr.newSpan()))
--- a/pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
@@ -54,7 +54,7 @@ import org.pkl.core.ast.builder.SymbolTable.AnnotationScope;
 import org.pkl.core.ast.builder.SymbolTable.ClassScope;
 import org.pkl.core.ast.builder.SymbolTable.ModuleScope;
 import org.pkl.core.ast.builder.SymbolTable.ObjectScope;
-import org.pkl.core.ast.builder.VariableResolution.ForGeneratorVariable;
+import org.pkl.core.ast.builder.VariableResolution.ForGeneratorOrLetVariable;
 import org.pkl.core.ast.builder.VariableResolution.ImplicitBaseProperty;
 import org.pkl.core.ast.builder.VariableResolution.ImplicitThisProperty;
 import org.pkl.core.ast.builder.VariableResolution.LexicalProperty;
@@ -67,7 +67,7 @@ import org.pkl.core.ast.expression.binary.GreaterThanNodeGen;
 import org.pkl.core.ast.expression.binary.GreaterThanOrEqualNodeGen;
 import org.pkl.core.ast.expression.binary.LessThanNodeGen;
 import org.pkl.core.ast.expression.binary.LessThanOrEqualNodeGen;
-import org.pkl.core.ast.expression.binary.LetExprNode;
+import org.pkl.core.ast.expression.binary.LetExprNodeGen;
 import org.pkl.core.ast.expression.binary.LogicalAndNodeGen;
 import org.pkl.core.ast.expression.binary.LogicalOrNodeGen;
 import org.pkl.core.ast.expression.binary.MultiplicationNodeGen;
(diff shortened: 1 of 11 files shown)


pkl-core/src/test/files/LanguageSnippetTests/output/errors/letExpressionError2.err
```diff
@@ -3,11 +3,11 @@ ouch
 
 x | let (y = throw("ouch"))
              ^^^^^^^^^^^^^
-at letExpressionError2#res1.<function#1> (file:///$snippetsDir/input/errors/letExpressionError2.pkl)
+at letExpressionError2#res1 (file:///$snippetsDir/input/errors/letExpressionError2.pkl)
 
 x | let (x = 1)
     ^^^^^^^^^^^
-at letExpressionError2#res1 (file:///$snippetsDir/input/errors/letExpressionError2.pkl)
+at letExpressionError2#res1.<let expr> (file:///$snippetsDir/input/errors/letExpressionError2.pkl)
```
bioball, line 10: Since these are no longer function calls, the stack trace in error messages shows up a little differently.

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -931,15 +921,9 @@ private ExpressionNode doVisitNewExprWithInferredParent(NewExpr expr) {
...
+                  createSourceSection(expr.newSpan()), language, scopeName, new GetOwnerNode())
               : new InferParentWithinMethodNode(
-                  createSourceSection(expr.newSpan()),
-                  language,
-                  scopeName,
-                  levelsUp == 0 ? new GetOwnerNode() : new GetEnclosingOwnerNode(levelsUp));
+                  createSourceSection(expr.newSpan()), language, scopeName, new GetOwnerNode());
     } else if (parent instanceof LetExpr letExpr && letExpr.getBindingExpr() == child) {
       // TODO (unclear how to infer type now that let-expression is implemented as lambda
       // invocation)
```
HT154: This TODO is no longer relevant. Can we actually infer the type now?

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
HT154: `levelsUp` is now unused

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -1113,38 +1097,28 @@ public ExpressionNode visitIfExpr(IfExpr expr) {
...
-        sourceSection, functionNode, visitExpr(letExpr.getBindingExpr()), isCustomThisScope);
+      typeNode = visitTypeAnnotation(par.getTypeAnnotation());
+      slot =
+          frameDescriptorBuilder.addSlot(
+              FrameSlotKind.Illegal, toIdentifier(par.getIdentifier().getValue()), null);
+      binding = par.getIdentifier().getValue();
+    }
+    var bindingExpr = visitExpr(letExpr.getBindingExpr());
+    var t = typeNode;
+    var s = slot;
```
HT154: Why alias these variables?

pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
```diff
@@ -961,12 +961,16 @@ public static Object evaluateExpression(
...
     var language = VmLanguage.get(null);
-    var builder = new AstBuilder(source, language, moduleInfo, moduleResolver);
     var parsedExpression = parseExpressionNode(expression, source);
+    var builder = new AstBuilder(source, language, moduleInfo, moduleResolver);
     var exprNode = builder.visitExpr(parsedExpression);
     var rootNode =
         new SimpleRootNode(
-            language, new FrameDescriptor(), exprNode.getSourceSection(), "", exprNode);
+            language,
+            builder.buildModuleFrameDescriptor(),
```
HT154, line 970: This builds the current scope's frame descriptor, which is presumed to be the module scope (after `builder.visit*` completes). When would this not be the same as the empty descriptor?

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
stackoverflow: I imagine it's because Java only accept final and "effectively final" variables inside lambdas. And `slot` and `typeNode` are neither.

stackoverflow (approved):
Looks good!

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
bioball: This is this node:

```
foo = let (x = new {})
               ^^^
````

I think it's actually correct that we cannot tell what to amend here. I can update and just erase this branch altogether.

File: pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java (hunk 3 above)
Comment (bioball) [reply pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java]: Yup!

File: pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java (hunk 4 above)
Comment (bioball) [reply pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java:970]: Let expressions add frame slots to the descriptor, for example:

```pkl
foo = let (x = something) x + 1
```

This contributes slots to the module node's frame descriptor.

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
HT154: What about when an explicit type is provided for the binding?
```pkl
foo = let (x: Bar = new {})
```

bioball: Ah, that's interesting. Yeah, we should be able to infer this correctly.

HT154 (approved):
LGTM!

PR #1629, opened 2026-05-29
Fix forcing of VmValue during error rendering

stackoverflow:


pkl-core/src/main/java/org/pkl/core/util/ErrorMessages.java
```diff
@@ -33,7 +35,17 @@ public static String create(String messageName, @Nullable Object... args) {
     if (args.length == 0) return errorMessage;
 
     var formatter = new MessageFormat(errorMessage, locale);
-    return formatter.format(args);
+    var renderer = VmValueRenderer.singleLine(Integer.MAX_VALUE);
+    @Nullable Object[] actualArgs = new @Nullable Object[args.length];
```
HT154, line 42: Isn't this type incorrect?
```suggestion
    Object @Nullable [] actualArgs = new Object[args.length];
```

HT154 (changes requested):
Can we add a test to prevent regressions?

pkl-core/src/main/java/org/pkl/core/util/ErrorMessages.java
stackoverflow: No. I want an array of nullable objects (same type as `args`), not a nullable array of objects.

stackoverflow:
> Can we add a test to prevent regressions?

Added a test to make sure VmValues are not forced while rendering errors.

HT154 (approved):
LGTM

bioball (approved):

pkl-core/src/main/java/org/pkl/core/util/ErrorMessages.java
```diff
@@ -35,6 +35,8 @@ public static String create(String messageName, @Nullable Object... args) {
     if (args.length == 0) return errorMessage;
 
     var formatter = new MessageFormat(errorMessage, locale);
+    // TODO: we render VmValues here with VmValueRenderer, but that's not enough to properly
+    //       render all kinds of values, like Pkl Strings, for example
     var renderer = VmValueRenderer.singleLine(Integer.MAX_VALUE);
```
bioball: [nit] Move `renderer` to a static field

PR #1632, opened 2026-06-02
Correctly count regular members of `Dynamic`

### LINKED ISSUE #1631: pkl-binary serialization produces an incomplete msgpack value in the presence of local properties
Consider this example:
```pkl
import "pkl:pklbinary"

class Test {
}

y = new Dynamic {
  foo {
    local bar = new Test {
    }
    z = bar
  }
}

serialized = new pklbinary.Renderer {}.renderDocument(y).base64
```
Running this script as
```
pkl eval test.pkl -x serialized | base64 -d | msgpack2json -d
```
shows the following output:
```
[
    1,
    "Dynamic",
    "pkl:base",
    [
        [
            16,
            "foo",
            [
                1,
                "Dynamic",
                "pkl:base",
                [
                    [
                        16,
                        "z",
                        [
                            1,
                            "test#Test",
                            "file:///private/tmp/test.pkl",
                            []
                        ]
                    ]
msgpack2json: parse error: mpack_error_eof (9)
```

That is, the produced msgpack stream is cut off right after writing the value of the `z` field.

Removing the `local` modifier before `bar`, removing the `z` property which refers to the local property, or moving `bar` to the top level all fix the issue, so it seems to be related to the fact that `bar` is a local property. This issue also apparently needs `bar` to be a class instance, as using a primitive value instead of a class doesn't seem to trigger it.

This happens on the latest Pkl version, 0.31.1.

HT154:
Previously, `VmDynamic.isHiddenOrLocalProperty` didn't correctly identify locals whose cache key is an `ObjectMember` instead of an `Identifier`, causing `VmDynamic. getRegularMemberCount` to return an incorrect value. This caused some renderers to produce incorrect output.

Resolves #1631

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmDynamic.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmDynamic.java
@@ -158,7 +158,8 @@ public final class VmDynamic extends VmObject {
   }
 
   private boolean isHiddenOrLocalProperty(Object key) {
-    return key instanceof Identifier
-        && (key == Identifier.DEFAULT || ((Identifier) key).isLocalProp());
+    return key instanceof ObjectMember member && member.isLocal()
+        || key instanceof Identifier identifier
+            && (key == Identifier.DEFAULT || identifier.isLocalProp());
   }
 }
--- a/pkl-core/src/test/kotlin/org/pkl/core/EvaluatorTest.kt
+++ b/pkl-core/src/test/kotlin/org/pkl/core/EvaluatorTest.kt
@@ -635,7 +635,7 @@ class EvaluatorTest {
   }
 
   @Test
-  fun `nested pkl-binary rendiring produces correct results`() {
+  fun `nested pkl-binary rendering produces correct results`() {
     val evaluator =
       with(EvaluatorBuilder.preconfigured()) {
         allowedResources.add(Pattern.compile("b64:"))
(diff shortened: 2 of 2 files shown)


pkl-core/src/main/java/org/pkl/core/runtime/VmDynamic.java
```diff
@@ -158,7 +158,8 @@ public int getRegularMemberCount() {
   }
 
   private boolean isHiddenOrLocalProperty(Object key) {
-    return key instanceof Identifier
-        && (key == Identifier.DEFAULT || ((Identifier) key).isLocalProp());
+    return key instanceof ObjectMember member && member.isLocal()
+        || key instanceof Identifier identifier
+            && (key == Identifier.DEFAULT || identifier.isLocalProp());
```
bioball, line 163:
```suggestion
    // only local members have the entire `ObjectMember` stored as a key in cachedValues
    if (key instanceof ObjectMember member) {
      assert member.isLocal();
      return true;
    }
    return key instanceof Identifier identifier
        && (key == Identifier.DEFAULT || identifier.isLocalProp());
```

pkl-core/src/test/kotlin/org/pkl/core/EvaluatorTest.kt
```diff
@@ -696,6 +696,48 @@ class EvaluatorTest {
...
+
+          class MyClass {
+            bar: Test
+          }
+
+          class Test
+
+          output {
+            renderer = new pklbinary.Renderer {}
+          }
```
bioball, line 732: Maybe put this in `pkl-core/src/test/files/LanguageSnippetTests/input/pklbinary` instead

bioball (approved):
Two nits; approving to unblock

PR #1634, opened 2026-06-02
Improve handling of frame slots

bioball:
This makes various improvements to the handling of frame slot vars, and includes some bug fixes introduced by https://github.com/apple/pkl/pull/1622

* Refactor SymbolTable to track for-generator and parameter slots in each scope
* Execute let expressions in their own root node in some places
* Unify how frame slots are managed; they are all represented as `FrameSlotVariable`, created in `AstBuilder`, and passed into `SymbolTable`.
* Fix how let expressions are executed in custom this scopes (introduce a new root node when needed)


pkl-core/src/test/files/LanguageSnippetTests/input/generators/predicateMembersListing.pkl
```diff
@@ -43,7 +43,8 @@ res5 = new {
 
 // nested predicate
 res6 = (people) {
-  [[(people) { [[name == "Barn Owl"]] { age = 99 } }.toList().find((it) -> it.age == 99).name == name]] {
+  [[(people) { [[name == "Barn Owl"]] { age = 99 } }.toList().find((it) -> it.age == 99).name
+    == name]] {
```
HT154, line 47: Hmm odd formatting here

pkl-core/src/main/java/org/pkl/core/ast/builder/SymbolTable.java
```diff
@@ -505,21 +547,45 @@ public interface LexicalScope {
   }
 
   public static class ObjectScope extends Scope implements LexicalScope {
-    private final Map<String, Integer> params;
+    private final FrameSlotVariable[] bindings;
 
+    /**
+     * NOTE: object body params desguar to wrapping this object with a lambda call.
```
HT154:
```suggestion
     * NOTE: object body params desugar to wrapping this object with a lambda call.
```

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -1281,7 +1308,15 @@ public GeneratorMemberNode visitObjectMethod(ObjectMethod memberNode) {
...
         symbolTable.enterEagerGenerator(
-            (scp) -> symbolTable.enterCustomThisScope(scope -> visitExpr(ctx.getPred())));
+            (scp) ->
+                symbolTable.enterCustomThisScope(
+                    scope -> {
+                      var currentFrameDescriptorSize = scope.frameDescriptorBuilder.getSize();
+                      var expr = visitExpr(ctx.getPred());
+                      var writesFrameSlotVars =
+                          scope.frameDescriptorBuilder.getSize() > currentFrameDescriptorSize;
+                      return getExprWithinCustomThis(scope, expr, writesFrameSlotVars);
```
HT154: This pattern gets repeated a few times. Might be worth pulling into a method of its own or subsuming into `getExprWithinCustomThis` (since it looks like that matches 1:1).

pkl-core/src/main/java/org/pkl/core/util/ArrayUtils.java
```diff
@@ -0,0 +1,50 @@
...
+package org.pkl.core.util;
+
+import java.util.Arrays;
+
+public final class ArrayUtils {
+  private ArrayUtils() {}
+
+  public static final int[] EMPTY_INT_ARRAY = new int[0];
+
+  public static int[] plus(int[] array, int elem) {
```
HT154: Hmm don't love the method name. Maybe `append`? That's what go calls the analogous operation on slices at least.

HT154 (commented):
Some initial feedback. Haven't finished looking at the whole thing yet.

pkl-core/src/main/java/org/pkl/core/ast/builder/SymbolTable.java
```diff
@@ -224,17 +234,35 @@ public abstract static class Scope {
     protected final FrameDescriptorBuilder frameDescriptorBuilder;
     private final ConstLevel constLevel;
     protected boolean isBaseModule;
+    // all for-generator slots in this scope (excludes args and let bindings).
+    protected final int[] forGeneratorSlots;
+    // all paramter slots in this scope; includes let bindings, function params, method params,
```
HT154:
```suggestion
    // all parameter slots in this scope; includes let bindings, function params, method params,
```

HT154 (approved):
No blockers from me on this, just a few minor things. Approving to unblock.

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -417,14 +420,35 @@ public UnresolvedTypeNode visitConstrainedType(ConstrainedType type) {
...
-          for (int i = 0; i < constraints.length; i++) {
+          for (var i = 0; i < constraints.length; i++) {
+            var currentFrameDescriptorSize = scope.frameDescriptorBuilder.getSize();
             var expr = visitExpr(exprs.get(i));
+            var writesFrameSlotVars =
+                scope.frameDescriptorBuilder.getSize() > currentFrameDescriptorSize;
+            // if a constraint expression writes to frame slots (only known case: is a `let` expr),
+            // create a new root node and execute the constraint within this root node
+            // e.g. String(let (x = this) x.length > 5)
+            expr = getExprWithinCustomThis(scope, expr, writesFrameSlotVars);
```
stackoverflow:
```suggestion
            expr = writesFrameSlotVars ? getExprWithinCustomThis(scope, expr) : expr;
```
nit: I think this is more readable. Or rename the function to `maybeGetExprWithinCustomThis`, as the name is misleading right now.

pkl-core/src/main/java/org/pkl/core/ast/builder/SymbolTable.java
```diff
@@ -266,16 +296,21 @@ public final String getQualifiedName() {
       return qualifiedName;
     }
 
-    public FrameDescriptor buildFrameDescriptor() {
-      return frameDescriptorBuilder.build();
+    public int[] getForGeneratorSlots() {
```
stackoverflow, line 304: As we added doc comments for `getParameterSlots` we should also add to `getForGeneratorSlots`.

pkl-core/src/main/java/org/pkl/core/ast/expression/primary/WithCustomThisExpression.java
```diff
@@ -0,0 +1,55 @@
...
+package org.pkl.core.ast.expression.primary;
+
+import com.oracle.truffle.api.CompilerDirectives;
+import com.oracle.truffle.api.CompilerDirectives.CompilationFinal;
+import com.oracle.truffle.api.frame.VirtualFrame;
+import org.pkl.core.ast.ExpressionNode;
+import org.pkl.core.runtime.VmUtils;
+import org.pkl.core.util.ArrayUtils;
+
+public final class WithCustomThisExpression extends ExpressionNode {
```
stackoverflow, line 26: nit: I'd add doc comments to give some context on why this node is needed.

pkl-core/src/main/java/org/pkl/core/util/ArrayUtils.java
stackoverflow: Agree with Jen that `append` is a better name.

stackoverflow (approved):
Great work!

PR #1636, opened 2026-06-02
Fix native build

bioball:
[native-pkl-cli]


HT154 (approved):
LGTM

It'd be really slick if we could somehow detect git diffs in methods above the truffle boundary and run native builds on PRs automatically. Probably difficult to achieve in practice, but a gal can dream!

bioball:
It _should_ be possible to statically check if we are missing truffle boundaries without requiring native builds too. Would be a nice improvement to our dev workflow.

Issue #1163, opened 2025-07-30
Package download error reporting hides true cause when redirects are involved


HT154:
I'm working on adopting Pkl 0.29 and HTTP rewrites and ran into an error condition with a misleading message.

Background:
* I'm trying to pull packages via pkg.pkl-lang.org (which redirects to github.com) without connecting to the public internet.
* I have a rewrite from https://github.com/ to https://github.internal/github-releases/
  * This endpoint uses a cert from a publicly trusted CA.
* I have a rewrite from https://pkl.pkl-lang.org to https://pkl.internal/mirror/pkg.pkl-lang.org/
  * This endpoint is using a self-signed cert.
  * My client system has this cert in ~/.pkl/cacerts.
  * This endpoint redirects to the appropriate location under https://github.internal/github-releases/ just like pkl.pkl-lang.org does.

I made the mistake of not explicitly trusting my normal public CA bundle via ~/.pkl/cacerts so attempting to access https://github.internal failed, but the error message was surprising!

```console
bash-5.1$ pkl eval package://pkg.pkl-lang.org/pkl-go/pkl.golang@0.11.0#/go.pkl \
  --http-rewrite https://pkg.pkl-lang.org/=https://pkl.internal/mirror/pkg.pkl-lang.org/ \
  --http-rewrite https://github.com/=https://github.internal/github-releases/
–– Pkl Error ––
Exception when making request `GET https://pkg.pkl-lang.org/pkl-go/pkl.golang@0.11.0`:
Error during SSL handshake with host `pkl.internal`:
unable to find valid certification path to requested target (request was rewritten: https://pkg.pkl-lang.org/pkl-go/pkl.golang@0.11.0 -> https://pkl.internal/mirror/pkg.pkl-lang.org/pkl-go/pkl.golang@0.11.0)
```

I would expect this error to look something like this:
```
Exception when making request `GET https://pkg.pkl-lang.org/pkl-go/pkl.golang@0.11.0`:
Error during SSL handshake with host `pkl.internal`:
unable to find valid certification path to requested target
* request was rewritten: https://pkg.pkl-lang.org/pkl-go/pkl.golang@0.11.0 -> https://pkl.internal/mirror/pkg.pkl-lang.org/pkl-go/pkl.golang@0.11.0
[...]

PR #1637, opened 2026-06-02
Improve HTTP redirect following

bioball:
This implements HTTP redirect following ourselves.

The goal is:

1. All I/O is checked against `--allowed-resources` and `--allowed-modules`, including HTTP redirects
2. HTTP rewrite rules can affect redirect following
3. HTTP headers can affect redirect following

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
+++ b/pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
@@ -77,7 +77,7 @@ final class RequestRewritingClient implements HttpClient {
     var maxRedirects = 20;
     if (maxRedirectProp != null) {
       try {
-        maxRedirects = Integer.parseInt(maxRedirectProp);
+        maxRedirects = Math.max(0, Integer.parseInt(maxRedirectProp));
       } catch (NumberFormatException ignored) {
       }
     }
--- a/pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
+++ b/pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
@@ -137,7 +137,7 @@ final class RequestRewritingClient implements HttpClient {
               ErrorMessages.create("httpRedirectCannotDowngrade", currentRequestUri, redirectUri));
         }
         currentRequestUri = rewriteUri(redirectUri);
-        currentRequest = rewriteRequest(currentRequest, currentRequestUri);
+        currentRequest = rewriteRequest(request, currentRequestUri);
         redirectCount++;
       } else {
         return response;
--- a/pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
+++ b/pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
@@ -134,7 +134,7 @@ final class RequestRewritingClient implements HttpClient {
         throw new HttpClientException(
             ErrorMessages.create("httpRedirectInvalidUri", currentRequestUri, location.get()));
       }
-      if (currentRequestUri.getScheme().equals("https") && redirectUri.getScheme().equals("http")) {
+      if (currentRequestUri.getScheme().equalsIgnoreCase("https") && redirectUri.getScheme().equalsIgnoreCase("http")) {
         throw new HttpClientException(
             ErrorMessages.create("httpRedirectCannotDowngrade", currentRequestUri, redirectUri));
       }
--- a/pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
+++ b/pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
@@ -134,7 +134,8 @@ final class RequestRewritingClient implements HttpClient {
         throw new HttpClientException(
             ErrorMessages.create("httpRedirectInvalidUri", currentRequestUri, location.get()));
       }
-      if (currentRequestUri.getScheme().equalsIgnoreCase("https") && redirectUri.getScheme().equalsIgnoreCase("http")) {
+      if (currentRequestUri.getScheme().equalsIgnoreCase("https")
+          && redirectUri.getScheme().equalsIgnoreCase("http")) {
         throw new HttpClientException(
             ErrorMessages.create("httpRedirectCannotDowngrade", currentRequestUri, redirectUri));
       }
--- a/pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
+++ b/pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
@@ -34,6 +34,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Pattern;
 import org.jspecify.annotations.Nullable;
 import org.pkl.core.PklBugException;
+import org.pkl.core.SecurityManagerException;
+import org.pkl.core.util.ErrorMessages;
 import org.pkl.core.util.HttpUtils;
 import org.pkl.core.util.IoUtils;
 import org.pkl.core.util.Pair;
@@ -42,13 +44,15 @@ import org.pkl.core.util.Pair;
  * An {@code HttpClient} decorator that
  *
  * <ul>
- *   <li>overrides the {@code User-Agent} header of {@code HttpRequest}s
+ *   <li>overrides the headers of {@code HttpRequest}s
  *   <li>sets a request timeout if none is present
  *   <li>ensures that {@link #close()} is idempotent.
  *   <li>rewrites outbound URI prefixes with another prefix.
+ *   <li>handles redirects, rewriting URLs after each redirect and checking against {@link
+ *       org.pkl.core.http.HttpClient.HttpRequestChecker}.
  * </ul>
  *
- * <p>Both {@code User-Agent} header and default request timeout are configurable through {@link
+ * <p>The headers, default request timeout, and URI rewrites are configurable through {@link
  * HttpClient.Builder}.
  */
 @ThreadSafe
@@ -63,6 +67,15 @@ final class RequestRewritingClient implements HttpClient {
   private final List<Entry<URI, URI>> rewrites;
 
   private final AtomicBoolean closed = new AtomicBoolean();
+
+  private static final int MAX_HTTP_REDIRECTS;
+
+  static {
+    // allow Java users to configure max redirects the same way they would Java's default HTTP
+    // client.
+    var maxRedirectProp = System.getProperty("http.maxRedirects");
+    MAX_HTTP_REDIRECTS = maxRedirectProp != null ? Integer.parseInt(maxRedirectProp) : 20;
+  }
 
   RequestRewritingClient(
       String userAgent,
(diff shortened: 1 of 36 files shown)


pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
```diff
@@ -84,12 +97,56 @@ final class RequestRewritingClient implements HttpClient {
...
+              ErrorMessages.create("httpRedirectCannotDowngrade", currentRequestUri, redirectUri));
+        }
+        currentRequestUri = redirectUri;
+        currentRequest = rewriteRequest(currentRequest, rewriteUri(redirectUri));
+        redirectCount++;
+      } else {
+        return response;
+      }
+    }
+  }
```
bioball, line 146: This class (`RequestRewritingClient`) maybe should be renamed. Alternatively, we can implement the redirect following outside of `HttpClient` (like in `HttpUtils`).

I don't feel strongly either way, but, food for thought for the future.

pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
```diff
@@ -64,6 +68,15 @@ final class RequestRewritingClient implements HttpClient {
...
   private final AtomicBoolean closed = new AtomicBoolean();
 
+  private static final int MAX_HTTP_REDIRECTS;
+
+  static {
+    // allow Java users to configure max redirects the same way they would Java's default HTTP
+    // client.
+    var maxRedirectProp = System.getProperty("http.maxRedirects");
+    MAX_HTTP_REDIRECTS = maxRedirectProp != null ? Integer.parseInt(maxRedirectProp) : 20;
+  }
```
HT154, line 85: The intent here is that this will always AOT compile to 20 in native builds, right?

pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
```diff
@@ -84,12 +97,56 @@ final class RequestRewritingClient implements HttpClient {
...
+  private <T> HttpResponse<T> doSend(
+      HttpRequest request,
+      BodyHandler<T> responseBodyHandler,
+      HttpRequestChecker httpRequestChecker)
+      throws SecurityManagerException, IOException {
+    var currentRequestUri = request.uri();
+    var redirectCount = 0;
+    var currentRequest = rewriteRequest(request, rewriteUri(request.uri()));
+    while (true) {
+      httpRequestChecker.check(currentRequestUri);
```
HT154, line 116: This makes the assumption that the initial request URI (pre-rewrite) is already checked. If a rewrite occurs, this procedure never checks the original URI, though it does if no rewrite occurs.

This is likely a safe assumption today, but maybe it would make sense for the HTTP client to check that again to ensure the "every URI (original, rewritten, redirected, final) is checked against the policy" is maintained within its own API surface. Barring a change, we should at least document the assumption.

pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
```diff
@@ -84,12 +97,56 @@ final class RequestRewritingClient implements HttpClient {
...
+      BodyHandler<T> responseBodyHandler,
+      HttpRequestChecker httpRequestChecker)
+      throws SecurityManagerException, IOException {
+    var redirectCount = 0;
+    var currentRequestUri = rewriteUri(request.uri());
+    var currentRequest = rewriteRequest(request, currentRequestUri);
+    while (true) {
+      httpRequestChecker.check(currentRequestUri);
+      var response = delegate.send(currentRequest, responseBodyHandler, httpRequestChecker);
+      if (HttpUtils.isRedirectStatusCode(response.statusCode())) {
```
HT154: Nit: swap this condition around for an early return and to reduce nesting.

pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
```diff
@@ -84,12 +97,56 @@ final class RequestRewritingClient implements HttpClient {
...
+          redirectUri = currentRequestUri.resolve(location);
+        } catch (IllegalArgumentException e) {
+          throw new HttpClientException(ErrorMessages.create("httpRedirectInvalidUri", location));
+        }
+        if (currentRequestUri.getScheme().equals("https")
+            && redirectUri.getScheme().equals("http")) {
+          throw new HttpClientException(
+              ErrorMessages.create("httpRedirectCannotDowngrade", currentRequestUri, redirectUri));
+        }
+        currentRequestUri = rewriteUri(redirectUri);
```
HT154: Similar to above, we should likely be checking both the redirected `Location` and—if it's rewritten—the rewrite target here. Currently only the rewrite target is checked if a rewrite matches.

pkl-core/src/main/java/org/pkl/core/util/HttpUtils.java
```diff
@@ -25,6 +25,17 @@
 public final class HttpUtils {
   private HttpUtils() {}
 
+  public static boolean isRedirectStatusCode(int statusCode) {
+    return switch (statusCode) {
+      // We can handle each of these status codes exactly the same because:
+      //
+      // 1. We don't implement any cacheing for HTTPS requests.
+      // 2. Pkl only makes GET requests.
+      case 301, 302, 303, 307, 308 -> true;
```
HT154, line 34: I've never seen a 303 in the wild, but no reason not to handle it here I guess!

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1151,3 +1148,20 @@ HTTP header value `{0}` has invalid syntax.
 invalidHttpHeaderValueTooLong=\
 HTTP Header value is invalid because it is longer than 4096 characters. \
 Value: `{0}`
+
+httpTooManyRedirects=\
+Too many redirects. Possible redirect loop?
```
HT154: Should this include the threshold value?

Stylistically, this message also seems incongruous in tone from our usual. Maybe the second sentence should be the hint?

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1151,3 +1148,20 @@ HTTP header value `{0}` has invalid syntax.
...
+
+httpRedirectNoLocation=\
+Cannot follow HTTP redirect because no 'Location' header was found
```
HT154:
```suggestion
Cannot follow HTTP redirect because no 'Location' header was found.
```

It also might be good to include the redirected/rewritten URI(s) in this message in case it happens after one or more redirects. Same goes for other error messages (including allowed module/resource denials).

pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
bioball: Yup

pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
bioball: Open to discussing this further; the rationale here is that the allowed modules/resources flags are I/O controls, and so it should check on what the actual I/O is.

This way you can say: "disallow requests to public.website, requests to internal.website are okay, and rewrite public.website to internal.website"

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
bioball: Added to every error message

pkl-core/src/main/java/org/pkl/core/util/HttpUtils.java
bioball: Me neither!

pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
HT154: That makes sense to me, but that's not actually what happens in practice for the initial URI since that's checked by the caller. I'm amenable to this route, but we should probably then remove the caller-side pre- and post-request checks.

bioball: The module key doesn't check the request URI before calling into `HttpClient#send` (or am I missing something)?

But, yeah, it looks like the resource read side should be refactored here; `ResourceManager` checks the URI before calling into the reader, and `read()` doesn't expect the resource reader to perform this check.

We should maybe unify these, but I think that refactor is out-of-scope. I'll insert a hack for now to skip the resource check just for `ResourceReaders.HttpResource`.

HT154: > The module key doesn't check the request URI before calling into `HttpClient#send` (or am I missing something)?

That happens here, no? https://github.com/apple/pkl/blob/68f7d06eedc27251d670252e668681d604db1203/pkl-core/src/main/java/org/pkl/core/module/ModuleKeys.java#L522

This also checks the URI of the response, which should also no longer be needed.

pkl-core/src/main/java/org/pkl/core/runtime/ResourceManager.java
```diff
@@ -98,12 +99,16 @@ public Optional<Object> read(URI resourceUri, @Nullable Node readNode) {
...
         (uri) -> {
-          try {
-            securityManager.checkReadResource(uri);
-          } catch (SecurityManagerException e) {
-            throw new VmExceptionBuilder().withCause(e).withOptionalLocation(readNode).build();
-          }
           var reader = getResourceReader(uri);
+          // hack: we don't want to call `checkReadResource` here for these resources because those
+          // readers defer to HttpClient to do the actual checks.
+          if (!(reader instanceof ResourceReaders.HttpResource) && !(reader instanceof ResourceReaders.HttpsResource)) {
```
HT154: Not in love with the `instanceof` check. Maybe a new method `ResourceReader.getUsesHttpClient()` (default impl returns false) could be used to gate this instead? Not a blocker.

pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
bioball: That code has been changed in this PR; this is the current method impl:

https://github.com/bioball/pkl/blob/613c6d3b12f62f4116cc6da89a1dbfe5518b6536/pkl-core/src/main/java/org/pkl/core/module/ModuleKeys.java#L520-L534

pkl-core/src/main/java/org/pkl/core/runtime/ResourceManager.java
bioball: I think both are hacks; this feels more tolerable of a hack for now.

pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
HT154: 🤦‍♀️ had the wrong branch checked out, my bad!

HT154 (approved):
LGTM. We definitely need to document this as a breaking change as previously-valid "inner" redirects may have worked but will now fail. We should update the documentation to reflect the precise behavior here.

HT154:
Does this PR address #1163?

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1151,3 +1148,23 @@ HTTP header value `{0}` has invalid syntax.
 invalidHttpHeaderValueTooLong=\
 HTTP Header value is invalid because it is longer than 4096 characters. \
 Value: `{0}`
+
+httpTooManyRedirects=\
+Too many redirects, exceeded the redirect threshold ({0}).
+
+httpRedirectNoLocation=\
+Cannot follow HTTP redirect because no 'Location' header was found.\n\
```
stackoverflow:
```suggestion
Cannot follow HTTP redirect because no ''Location'' header was found.\n\
```

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1151,3 +1148,23 @@ HTTP header value `{0}` has invalid syntax.
...
+\n\
+HTTP Request: `GET {0}`
+
+httpRedirectInvalidUri=\
+Cannot follow HTTP redirect because the response 'Location' header has a malformed URI.\n\
```
stackoverflow:
```suggestion
Cannot follow HTTP redirect because the response ''Location'' header has a malformed URI.\n\
```

pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
```diff
@@ -84,12 +104,55 @@ final class RequestRewritingClient implements HttpClient {
...
+            ErrorMessages.create("httpRedirectNoLocation", currentRequestUri));
+      }
+      URI redirectUri;
+      try {
+        redirectUri = currentRequestUri.resolve(location.get());
+      } catch (IllegalArgumentException e) {
+        throw new HttpClientException(
+            ErrorMessages.create("httpRedirectInvalidUri", currentRequestUri, location.get()));
+      }
+      if (currentRequestUri.getScheme().equals("https") && redirectUri.getScheme().equals("http")) {
```
stackoverflow:
```suggestion
      if (currentRequestUri.getScheme().equalsIgnoreCase("https") && redirectUri.getScheme().equalsIgnoreCase("http")) {
```
I don't think `URI.getScheme` normalizes URIs.

pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
```diff
@@ -64,6 +68,22 @@ final class RequestRewritingClient implements HttpClient {
...
+  private static final int MAX_HTTP_REDIRECTS;
+
+  static {
+    // allow Java users to configure max redirects the same way they would Java's default HTTP
+    // client.
+    var maxRedirectProp = System.getProperty("http.maxRedirects");
+    var maxRedirects = 20;
+    if (maxRedirectProp != null) {
+      try {
+        maxRedirects = Integer.parseInt(maxRedirectProp);
```
stackoverflow:
```suggestion
        maxRedirects = Math.max(0, Integer.parseInt(maxRedirectProp));
```
nit: as 0 and negative numbers behave the same, I'd rather normalize this value. Makes for better errors also.

pkl-core/src/main/java/org/pkl/core/runtime/ResourceManager.java
```diff
@@ -98,12 +99,17 @@ public Optional<Object> read(URI resourceUri, @Nullable Node readNode) {
...
         resourceUri.normalize(),
         (uri) -> {
-          try {
-            securityManager.checkReadResource(uri);
-          } catch (SecurityManagerException e) {
-            throw new VmExceptionBuilder().withCause(e).withOptionalLocation(readNode).build();
-          }
           var reader = getResourceReader(uri);
+          // hack: we don't want to call `checkReadResource` here for these resources because those
+          // readers defer to HttpClient to do the actual checks.
```
stackoverflow, line 104: Not against the hack per se. But we could remove this and make the classes private again if we add some method to `ResourceReader` like `default boolean defersSecurityCheck() {
  return false; }`

pkl-core/src/main/java/org/pkl/core/util/HttpUtils.java
```diff
@@ -25,6 +25,17 @@
 public final class HttpUtils {
   private HttpUtils() {}
 
+  public static boolean isRedirectStatusCode(int statusCode) {
+    return switch (statusCode) {
+      // We can handle each of these status codes exactly the same because:
+      //
+      // 1. We don't implement any cacheing for HTTPS requests.
```
stackoverflow:
```suggestion
      // 1. We don't implement any caching for HTTPS requests.
```

stackoverflow (approved):
Looks good. Some nits.

pkl-core/src/main/java/org/pkl/core/runtime/ResourceManager.java
bioball: I'd rather not do this because class `ResourceReader` is a public API.

The more principled approach here feels like: Pkl should pass in the security manager to `read()`, which it already does for the module reader case. However, I think that's out-of-scope for this PR.

bioball:
> Does this PR address https://github.com/apple/pkl/issues/1163?

Not directly, but now we can fix it!

Issue #1262, opened 2025-10-28
Extending an abstract class **without** overriding an abstract property crashes with ambiguous message


StefMa:
Given:
```pkl
abstract class Bird {
  abstract name: String
  function greet(bird: Bird): String = "Hello, \(bird.name)!"
}

class Pigeon extends Bird {
  // Here, I forgot to overried `name`
}

pigeon: Pigeon = new {}
```

Leads to:
<details>

```
pkl failed with error Error: Command failed: /var/task/node_modules/@pkl-community/pkl-linux-x64/bin/pkl eval - --no-project --format json --allowed-modules pkl:,repl: --allowed-modules package: --cache-dir /tmp/pkl
An unexpected error has occurred. Would you mind filing a bug report?
Cmd+Double-click the link below to open an issue.
Please copy and paste the entire error output into the issue's description, provided you can share it.

https://github.com/apple/pkl/issues/newjava.lang.IllegalArgumentException%3A%20Cannot%20convert%20VM%20value%20with%20unexpected%20type%3A%20null

–– Pkl Error ––
Cannot convert VM value with unexpected type: null

394 | external function renderDocument(value: Any): String
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
at pkl.base#JsonRenderer.renderDocument (https://github.com/apple/pkl/blob/0.25.1/stdlib/base.pkl#L394)

106 | text = renderer.renderDocument(value)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
at pkl.base#Module.output.text (https://github.com/apple/pkl/blob/0.25.1/stdlib/base.pkl#L106)

Pkl 0.25.1 (Linux 5.15.0-1050-aws, native)

java.lang.IllegalArgumentException: Cannot convert VM value with unexpected type: null
	at org.pkl.core.runtime.VmValueConverter.convert(VmValueConverter.java:100)
	at org.pkl.core.stdlib.AbstractRenderer.doVisitProperty(AbstractRenderer.java:193)
	at org.pkl.core.stdlib.AbstractRenderer.lambda$visitTyped$0(AbstractRenderer.java:256)
	at org.pkl.core.runtime.VmObject.lambda$iterateAlreadyForcedMemberValues$1(VmObject.java:132)
	at org.pkl.core.runtime.VmObject.iterateMembers(VmObject.java:148)
[...]

PR #1647, opened 2026-06-05
Reject abstract members in non-abstract classes

### LINKED ISSUE #1614: Non-abstract class can have abstract members
I'd expect the below code to fail, but it succeeds in the same way as if `abstract` wasn't present:

```
class Foo {
  abstract bar: Int
}

foo = new Foo {
  bar = 5
}
```

The same problem exists for methods:

```
class Foo {
  abstract function bar(): Int
}

foo = new Foo {}
```

vinayakj592:
Fixes #1614.

## Context

A non-abstract `class` (or `module`) was allowed to declare `abstract` properties and methods.
Because such an enclosing type is instantiable, an `abstract` member there can never be guaranteed
an implementation — so the contradiction surfaced only as a runtime error when the member was
accessed (`Cannot invoke abstract method`), or not at all.

This makes it a compile-time error to declare an `abstract` member unless its enclosing class or
module is also `abstract`. This is consistent with how Pkl already rejects instantiating an abstract
class, and mirrors how Java and Kotlin treat abstract members.

## Before

```pkl
class Foo {
  abstract bar: Int
}
res = new Foo { bar = 5 }   // evaluated successfully (should fail)
```

```pkl
class Foo {
  abstract function bar(): Int
}
res = new Foo {}            // evaluated successfully; res.bar() failed only at runtime
```

## After

```
–– Pkl Error ––
Cannot define an abstract member in a non-abstract class.

2 | abstract bar: Int
    ^^^^^^^^
at Foo

A member can only be `abstract` if its enclosing class is also `abstract`.
```

## Implementation

- `AstBuilder` now validates, while building the AST, that a non-abstract class/module declares no
  `abstract` members. The check runs in both `visitClass` and `visitModule`, and the error points at
  the `abstract` keyword.
- Adds the `abstractMemberInNonAbstractClass` error message.

## Scope: classes and modules

The issue describes classes; I applied the same rule to modules as well, since a module is a class
in Pkl and a non-abstract module is likewise directly evaluatable. Happy to narrow this to classes
only if you'd prefer — it's a one-line change either way.

The `moduleMethodModifiers` pkl-doc test fixture declared an abstract method at non-abstract module
level (relying on the old behavior); it's updated to an `abstract module`, and its expected
documentation output is regenerated.

## Tests

- New `LanguageSnippetTests` error cases: abstract property in a class, abstract method in a class,
  and abstract member in a module.
- `./gradlew build` passes (`pkl-core` and `pkl-doc` included).

### DIFF
--- a/docs/modules/release-notes/pages/changelog.adoc
+++ b/docs/modules/release-notes/pages/changelog.adoc
@@ -4,6 +4,10 @@ include::ROOT:partial$component-attributes.adoc[]
 [[release-0.32.0]]
 == 0.32.0 (UNRELEASED)
 
+=== Fixes
+
+* Fixes an issue where a non-abstract class or module could declare `abstract` members (https://github.com/apple/pkl/issues/1614[#1614]).
+
 [[release-0.31.1]]
 == 0.31.1 (2026-03-26)
 
--- a/docs/modules/release-notes/pages/changelog.adoc
+++ b/docs/modules/release-notes/pages/changelog.adoc
@@ -4,10 +4,6 @@ include::ROOT:partial$component-attributes.adoc[]
 [[release-0.32.0]]
 == 0.32.0 (UNRELEASED)
 
-=== Fixes
-
-* Fixes an issue where a non-abstract class or module could declare `abstract` members (https://github.com/apple/pkl/issues/1614[#1614]).
-
 [[release-0.31.1]]
 == 0.31.1 (2026-03-26)
 
--- a/pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
@@ -1441,6 +1441,8 @@ public class AstBuilder extends AbstractAstBuilder<Object> {
     var scope = (ModuleScope) symbolTable.getCurrentScope();
     scope.setModifiers(modifiers);
 
+    checkAbstractMembersAllowed(modifiers, mod.getProperties(), mod.getMethods());
+
     // visit imports first so that we already have the object member name available
     var imports = mod.getImports();
     var importMembers = new ObjectMember[imports.size()];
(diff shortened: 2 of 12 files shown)


docs/modules/release-notes/pages/changelog.adoc
```diff
@@ -4,6 +4,10 @@ include::ROOT:partial$component-attributes.adoc[]
 [[release-0.32.0]]
 == 0.32.0 (UNRELEASED)
 
+=== Fixes
+
+* Fixes an issue where a non-abstract class or module could declare `abstract` members (https://github.com/apple/pkl/issues/1614[#1614]).
+
```
bioball: Move this to `0.32.adoc`, or feel free to just leave this blank (we'll fill in release notes before we ship).

```suggestion
```

bioball (approved):
Thanks! I think this solution makes sense. Another approach here could have been to defer the abstract check when the member is assigned to or accessed, but this feels more principled and catches obvious bugs earlier.

Note that this doesn't fully address #1614; this still works:

```pkl
abstract class Foo {
  abstract bar: Int
}

class Qux extends Foo {}

qux: Qux = new {
  bar = 5
}
```

Waiting for eyes from @HT154 and @stackoverflow before merging this.

HT154 (approved):
LGTM. Like @bioball said, this narrowly addresses #1614 but doesn't cover the related case from #1262, but that's fine to address as follow-up.

vinayakj592:
Could you help me understand this? I understand your point that it doesn’t cover issue #1262, but I’m not clear on why it also doesn’t fully address issue #1614. Could you explain which parts of issue #1614 remain unresolved?
@HT154 @bioball

bioball:
@vinayakj592 Class `Qux` still has an abstract member, because it did not override `bar`

vinayakj592:
> @vinayakj592 Class `Qux` still has an abstract member, because it did not override `bar`

@bioball I see what you're saying, I will cover that while working on #1262 , I think this can be merged irrespective of that.

bioball:
Yes, agree!

vinayakj592:
By the way, what does the typical review and merge cycle look like here? How long does it usually take for a PR to get reviewed and merged?

I’m planning to start contributing more actively, so I was also wondering if there are any discussion channels or community spaces apart from GitHub Discussions where contributors collaborate and ask questions.

@bioball

bioball:
Feel free to join the Pkl community discord (see https://pkl.community/); us maintainers also hang out on there!

This PR is blocked on a review from @stackoverflow.

vinayakj592:
Thanks @bioball

stackoverflow (approved):
Look good. Thanks for the contribution.

bioball:
@vinayakj592 FYI: I explored a solution for #1262 in https://github.com/apple/pkl/pull/1690. I think there's a couple interesting open questions to cover, though (see that PR's description)

PR #1648, opened 2026-06-05
Fix error rendering `@ConvertProperty` annotations during error reporting

HT154:
The added snippet test originally produced error "A value of type `Function2` cannot be exported."

This PR actually fixes the bug twice:
* By marking `ConvertProperty.render` as `hidden` so that it is skipped when the enclosing object is exported. This broke any attempts to obtain the module schema because this requires exporting all annotations on all class properties.
* By changing the way that `VmUndefinedValueException.fillInHint()` obtains the module URI to avoid obtaining the module schema (and triggering the more expensive module schema generation process).

"Non-hidden property with non-exportable type found in annotation subclass" might be a good candidate for a future linter rule or even an evaluator error!

### DIFF
--- a/stdlib/base.pkl
+++ b/stdlib/base.pkl
@@ -414,7 +414,7 @@ abstract class BaseValueRenderer {
 open class ConvertProperty extends Annotation {
   /// Function called by [BaseValueRenderer] types during rendering to transform property
   /// names and values.
-  render: (Pair<String, Any>, BaseValueRenderer) -> Pair<String, Any>
+  hidden render: (Pair<String, Any>, BaseValueRenderer) -> Pair<String, Any>
 }
 
 /// Base class for rendering Pkl values in some textual output format.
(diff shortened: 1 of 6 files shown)


stdlib/base.pkl
```diff
@@ -414,7 +414,7 @@ abstract class BaseValueRenderer {
 open class ConvertProperty extends Annotation {
   /// Function called by [BaseValueRenderer] types during rendering to transform property
   /// names and values.
-  render: (Pair<String, Any>, BaseValueRenderer) -> Pair<String, Any>
+  hidden render: (Pair<String, Any>, BaseValueRenderer) -> Pair<String, Any>
```
bioball, line 417: I removed the `hidden` modifier and the `convertProperty1.pkl` test is still passing. Is this actually a problem? If so, what error case exercises it?

HT154: You have to revert both fixes to observe the error.

If you pass the snippet test to `Evaluator.evaluateSchema()` without marking this as `hidden`, you can observe the same error unaffected by the change in `mUndefinedValueException`.

bioball: Ah, gotcha!

Can we add a test around `evaluateSchema` specifically?

bioball (approved):
Approving to unblock, one comment about adding a test case for this

HT154:
Added a new test case here that fails if the `hidden` is removed (unrelated to the other fix for the reported issue). Also updated the function properties in annotation classes in `pkl:Command` to be hidden.

PR #1650, opened 2026-06-05
Check if receiver is a module object in `relativePathTo`

### LINKED ISSUE #1649: `relativePathTo` does not check if receiver is a module
This snippet fails with an unexpected exception:

```pkl
local objectThatAmendsModule = (module) {
  // add a local property to force this into a new object amends;
  // empty object amends are optimized away.
  local qux = 1
}

res = objectThatAmendsModule.relativePathTo(module)
```

bioball:
Closes #1649

### DIFF
--- a/pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
+++ b/pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
@@ -203,7 +203,10 @@ Class `{0}` does not have type parameters.
 notASubclassOfTyped=\
 Class `{0}` is not a subtype of `Typed`.
 
-expectedModule=\
+expectedModuleAsReceiver=\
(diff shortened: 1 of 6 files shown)


pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -203,7 +203,10 @@ Class `{0}` does not have type parameters.
 notASubclassOfTyped=\
 Class `{0}` is not a subtype of `Typed`.
 
-expectedModule=\
```
stackoverflow, line 206: ReflectNodes still uses `expectedModule` so you need to change it there to the new message.

stackoverflow (approved):
Approving, but there's a blocker to fix.

PR #1659, opened 2026-06-05
Revert configuration cache

bioball:
This reverts the commits that enabled Gradle's configuration cache feature.

IMO: this feature is too hard to use. We don't know if a task is valid for the configuration cache until it runs, and it's very hard to tell if something is safe when authoring Gradle code.
For example, our publish tasks are currently failing; I don't know how I would fix this without running the publish task again on my dev machine.

Also, some of our build scripts become more brittle because of this; for example, see https://github.com/apple/pkl/blob/bb07589eae0b3195a589559a3245cbc12c29b394/build-logic/src/main/kotlin/BuildInfo.kt#L291-L296


HT154 (approved):
Bummer. Hopefully this feature improves enough for us to be able to adopt it in the future.

PR #1663, opened 2026-06-06
Handle modulo by 0

### LINKED ISSUE #1660: Modulo operator with `0` produces crash
Example:

```pkl
res = 5 % 0
```

bioball:
Treat modulo by zero as division by zero

NOTE: This copies the behavior of int div, but not division.

```
pkl0> 5 / 0
Infinity
pkl1> 5 ~/ 0
–– Pkl Error ––
Division by zero.

1 | 5 ~/ 0
    ^^^^^^
at  (repl:pkl1)
```

Should we copy division or integer division?

My sense is that throw is the better behavior here. I'm not sure when `5 % 0` resulting in `Infinity` is meaningful.

Perhaps we should fix the division case too.

Fixes #1660


HT154 (approved):

bioball:
According to IEEE 754, floating point remainder by zero should return NaN. However, it also behaves differently than how mod works today.

```
pkl0> 11.0 % 4.0
3.0
```

IEEE 754 would compute this to `-1.0`.

PR #1666, opened 2026-06-08
Fix VmIntSeq.length

bioball:
This fixes an internal optimization.

No language-observable change results from this fix, so no tests are introduced.


stackoverflow (approved):

bioball:
Ah, I was mistake, this does affect language behavior; tests added here: https://github.com/apple/pkl/pull/1739

PR #1672, opened 2026-06-09
Fix String.padStart/padEnd

### LINKED ISSUE #1661: `String.padStart` and `String.padEnd` return incorrect strings sometimes
For example:

```pkl
res = "🙃".padStart(5, "h")
```

Produces:

```pkl
"hhh🙃"
```

bioball:
Fixes the following two bugs:

* Pads to incorrect string length
* Does not append full code point

Closes #1661


stackoverflow (approved):
Looks good!

PR #1673, opened 2026-06-09
Fix errors around strings with lone high or low surrogates

### LINKED ISSUE #1662: `String.getOrNull` crashes with certain strings
For example, if a string contains a single high surrogate not followed by a low surrogate:

```pkl
"\u{D800}".getOrNull(0)
```

bioball:
This fixes some errors around treatment of strings with lone high/low surrogates.

Closes #1662

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
@@ -675,13 +675,14 @@ public final class VmUtils {
     var charOffset = startIndex;
 
     while (charOffset < length && codePointOffset > 0) {
-      if (Character.isHighSurrogate(string.charAt(charOffset++))
+      var ch = string.charAt(charOffset);
+      charOffset++;
+      if (Character.isHighSurrogate(ch)
           && charOffset < length
-          && !Character.isLowSurrogate(string.charAt(charOffset++))) {
-        codePointOffset -= 2;
-      } else {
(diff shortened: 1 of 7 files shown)


pkl-core/src/test/files/LanguageSnippetTests/output/api/string.pcf
```diff
@@ -372,6 +372,11 @@ examples {
     "d"
     "g"
     null
+    "🏀"
+    null
+    "?"
+    "?"
```
bioball, line 380: Not scoped to this PR, but we should improve this: PcfRenderer should render unprintable characters as a unicode escape sequence.

pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
```diff
@@ -675,13 +675,14 @@ public static int codePointOffsetToCharOffset(
     var charOffset = startIndex;
 
     while (charOffset < length && codePointOffset > 0) {
-      if (Character.isHighSurrogate(string.charAt(charOffset++))
+      var ch = string.charAt(charOffset);
+      charOffset++;
+      if (Character.isHighSurrogate(ch)
           && charOffset < length
-          && !Character.isLowSurrogate(string.charAt(charOffset++))) {
```
bioball, line 680: We shouldn't advance `charOffset` until we know the next char is a low surrogate

stackoverflow (approved):
LGTM

PR #1674, opened 2026-06-09
Default to XDG-style file locations

florinungur:
Default the CLI's file locations to XDG-style paths instead of placing everything under `~/.pkl`.

With this PR, a default Pkl setup writes nothing to `$HOME/.pkl`:

- Package cache: `~/.cache/pkl` (was `~/.pkl/cache`)
- Settings file: `~/.config/pkl/settings.pkl` (was `~/.pkl/settings.pkl`)
- CA certificates: `~/.config/pkl/cacerts` (was `~/.pkl/cacerts`)
- REPL history: `~/.local/state/pkl/repl-history` (was `~/.pkl/repl-history`; XDG state home, where history files belong)

These are fixed defaults; no environment variables (`XDG_CACHE_HOME`, `XDG_CONFIG_HOME`, `XDG_STATE_HOME`) are read, matching the design principle that Pkl doesn't configure itself from the environment.

Backward compatible, no migration needed: each legacy `~/.pkl` location is still used when it already exists. In particular, a pre-existing `~/.pkl/cache` is still used, so cached packages aren't re-downloaded. The one-time JLine2 `~/.pkl/repl-history.bin` cleanup is left pointing at the legacy path. An empty `~/.config/pkl/cacerts` directory counts as absent, so it doesn't shadow a populated legacy `~/.pkl/cacerts`.

Public Java-API change: `PklSettings.loadFromPklHomeDir()` is renamed to `loadFromDefaultLocation()` since it no longer loads from the pkl home dir; an embedder that relied on `~/.pkl/settings.pkl` will now prefer `~/.config/pkl/settings.pkl` when both exist. Happy to pick a different name if you'd prefer.

`pkl-executor` can't depend on `pkl-core`, so its `defaultModuleCacheDir()` replicates the same fallback (with a keep-in-sync comment).

Verified `./gradlew spotlessApply` and the JVM suites plus `pkl-cli`/`pkl-executor` compile. I also built the macOS arm64 native image and ran it. The fresh-home XDG branch isn't directly exercisable on the native binary (macOS resolves `user.home` from the OS, not `$HOME`/`-D`), but it's the same code path the unit tests cover and GraalVM recomputes `user.home` at runtime.

### DIFF
--- a/docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
+++ b/docs/modules/language-tutorial/pages/02_filling_out_a_template.adoc
@@ -360,7 +360,7 @@ at pkl.base#Module.output.text (https://github.com/apple/pkl/blob/e4d8c882d/stdl
 <6> What Pkl evaluated to discover the error.
 
 When Pkl prints source locations, it also prints clickable links for easy access.
-For local files, it generates a link for your development environment (https://pkl-lang.org/main/current/pkl-cli/index.html#settings-file[configurable in `+~/.pkl/settings.pkl+`]).
+For local files, it generates a link for your development environment (https://pkl-lang.org/main/current/pkl-cli/index.html#settings-file[configurable in `+~/.config/pkl/settings.pkl+`]).
 For packages imported from elsewhere, if available, Pkl produces `https://` links to their repository.
 
 Pkl complains about a _type constraint_.
--- a/docs/modules/pkl-cli/pages/index.adoc
+++ b/docs/modules/pkl-cli/pages/index.adoc
@@ -1414,10 +1414,10 @@ When making TLS requests, Pkl comes with its own set of {uri-certificates}[CA ce
 These certificates can be overridden via either of the two options:
 
 - Set them directly via the CLI option `--ca-certificates <path>`.
-- Add them to a directory at path `~/.pkl/cacerts/`.
+- Add them to a directory at path `~/.config/pkl/cacerts/` (or the legacy `~/.pkl/cacerts/`).
 
 Both these options will *replace* the default CA certificates bundled with Pkl. +
-The CLI option takes precedence over the certificates in `~/.pkl/cacerts/`. +
+The CLI option takes precedence over the certificates in `~/.config/pkl/cacerts/`. +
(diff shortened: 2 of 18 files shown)


bioball:
I think I'm open to this, but some considerations:

* We need to update pkl-intellij and pkl-lsp to look for sources in these new directories
* Like you said: as a matter of principle, Pkl doesn't configure itself using environment variables. So, this doesn't follow the XDG specification. However, it does feel like an improvement that we write to XDG-style paths.
* We use `~/.pkl` for other stuff. for example, `pkl-lsp` writes to `~/.pkl/editor-support`
* These folders are standard on Unix (macOS/linux), but not on Windows. Windows has stuff like `%APPDATA%`. I think it'd be weird to write to XDG-style paths on Windows, although I'm not a Windows user so I defer to others (maybe @odenix has an opinion?)

florinungur:
I made https://github.com/apple/pkl-intellij/pull/216 and https://github.com/apple/pkl-lsp/pull/208 to continue this effort. Both also migrate `editor-support` to the XDG data dir (`~/.local/share/pkl/editor-support`) with a legacy `~/.pkl/editor-support` fallback.

Also took a shot at the Windows paths (`%APPDATA%` for config, `%LOCALAPPDATA%` for cache/data/state), but since I can't test it and don't have a strong opinion, I welcome feedback (cc @odenix).

HT154:
A little bit of thought needs to be put toward how this change interacts with pkl-go and pkl-swift, as their APIs involve configuration of the cache directory.

bioball:
We chatted about this a bit: we're on board with following the XDG base directory specification fully on macOS/Linux:

* If the env var is set, use the env var.
* Otherwise default to `~/.config`, `~/.local/share`, etc

On Windows, using `%APPDATA%` and `%LOCALAPPDATA%` sounds great, and we're thinking we should _also_ allow this to be configured using XDG environment variables.

The rationale here is: unlike most environment variables, XDG actually has a specification that is well defined.

florinungur:
Closing this – I don't have the bandwidth to see it through. @bioball is picking the change up for 0.33, and the commits stay fetchable from `refs/pull/1674/head`.

PR #1683, opened 2026-06-15
Reject relative package URIs

stackoverflow:


pkl-core/src/main/java/org/pkl/core/packages/PackageResolvers.java
```diff
@@ -456,10 +457,18 @@ private String getEffectivePackageUriPath(PackageUri packageUri) {
...
+      var relativePath =
+          Path.of(
+              CACHE_DIR_PREFIX,
+              IoUtils.encodePath(uri.getUri().getAuthority()),
+              getEffectivePackageUriPath(uri));
+      // ensure the derived path cannot escape the cache directory
+      var resolved = cacheDir.resolve(relativePath).normalize();
+      if (!resolved.startsWith(cacheDir.normalize())) {
+        throw new SecurityException(
+            ErrorMessages.create("packageUriEscapesCacheDir", uri.getUri()));
```
bioball: Can this ever happen? It's not possible to construct a `PackageUri` with path traversal segments now, right?

If not, we can just change this to:

```java
assert resolved.startsWith(cacheDir.normalize());
```

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -861,6 +861,12 @@ Package URIs must have an authority component.\n\
 \n\
 For example, `example.com` in URI `project://example.com/my/package@1.0.0`.
 
+invalidRelativePathInPackageUri=\
+Package URI `{0}` must not contain `..` path segments.
```
bioball: Don't need to repeat the package URI here; our error will surface the original URI to users.

Also, we should specifically call out the path segment, because they're fine in the fragment; e.g. `package://example.com/foo@1.0.0#/foo/../bar.pkl`

```suggestion
Package URIs cannot contain `..` segments in the URI's "path" component.
```

bioball (changes requested):
Some nits, but also: can you add a test in pkl-core/src/test/files/LanguageSnippetTests/input/packages?

pkl-core/src/main/java/org/pkl/core/packages/PackageResolvers.java
stackoverflow: No, this shouldn't happen. But it's a second line of defense, so I think it's worth leaving it there.

bioball (approved):

pkl-core/src/main/java/org/pkl/core/packages/PackageResolvers.java
bioball: Gotcha, makes sense. I think, in this case, we should turn this into a `PklBugException`, because this is a bug on our part.

PR #1687, opened 2026-06-20
Fix Windows GraalVM installation

fallintoplace:
## What changed

This updates the GraalVM install task so Windows installs use ZIP extraction instead of the Unix tar path. Windows now publishes the extracted directory by moving it into place, while macOS and Linux keep the existing tar extraction and symlink-based install flow.

## Why

BuildInfo.GraalVm downloads .zip archives on Windows, but InstallGraalVm always ran tar --strip-components=1 -xzf and then tried to install by creating a symbolic link. That combination is fragile on Windows: the archive format does not match the extraction command, and symlink creation can require special privileges.

## Validation

- ./gradlew -p build-logic compileKotlin
- ./gradlew -p build-logic check
- ./gradlew help

[windows]

### DIFF
--- a/build-logic/src/main/kotlin/InstallGraalVm.kt
+++ b/build-logic/src/main/kotlin/InstallGraalVm.kt
@@ -120,10 +120,7 @@ constructor(
     return normalizedPath.substring(separatorIndex + 1).takeIf { it.isNotEmpty() }
   }
 
-  private fun publishInstallDir(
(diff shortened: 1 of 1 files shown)


bioball:
Thanks for the PR! Run `./gradlew spotlessApply` to fix formatting errors.

bioball (approved):

PR #1691, opened 2026-06-23
Add equals and hashcode to PType

stackoverflow:


pkl-core/src/test/files/LanguageSnippetTests/input/api/referenceEquals.pkl
```diff
@@ -0,0 +1,50 @@
+import "pkl:ref"
```
bioball: Move this file to `input/basic/reference.pkl` and have it `amend "../snippetTest.pkl"`; follow same structure as `input/basic/set.pkl`, `input/basic/string.pkl`, etc.

Should test both equality and inequality.

pkl-core/src/main/java/org/pkl/core/PType.java
```diff
@@ -80,6 +80,17 @@ public String getLiteral() {
     public String toString() {
       return ValueFormatter.basic().formatStringValue(literal, "");
     }
+
+    @Override
+    public boolean equals(Object obj) {
```
bioball: For every implementation of `equals`: make the argument `@Nullable`

bioball (approved):
Approving to unblock; some minor comments

PR #1695, opened 2026-06-24
Improve error message for aliased references

stackoverflow:


pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2173,26 +2181,68 @@ private Object doEval(VmReference value, VmTyped module) {
...
-              if (aliasSourceSection != null) {
-                err.withSourceSection(aliasSourceSection);
-              }
-              throw err.build();
+              found[0] = typeNode.getSourceSection();
+              return false;
             }
             return true;
           });
+      return found[0];
```
bioball: [nit] prefer using `MutableReference` for this

```suggestion
      var found = new MutableReference<@Nullable SourceSection>(null);
      referentTypeNode.acceptTypeNode(
          true,
          (typeNode) -> {
            if (typeNode instanceof ConstrainedTypeNode) {
              found.set(typeNode.getSourceSection());
              return false;
            }
            return true;
          });
      return found.get();
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2138,7 +2140,13 @@ public ReferenceTypeNode(
...
       this.referentTypeNode = referentTypeNode;
       this.getModuleNode = new GetModuleNode(sourceSection);
-      validateTypeArguments(sourceSection);
+      // A constraint written directly in this annotation's referent is reported against this
+      // annotation. Constraints introduced through a generic type alias's type argument are caught
+      // later.
+      if (findReferentConstraint() != null) {
+        CompilerDirectives.transferToInterpreter();
+        throw exceptionBuilder().evalError("invalidReferenceTypeAnnotationWithConstraint").build();
+      }
```
bioball, line 2156: It's kind of strange that that this throws an eval error directly, instead of `ReferentConstraintException`.

Also, this doesn't capture the chain if given a type that indirectly is a constrained type:

```pkl
import "pkl:ref"

typealias Bar = String(isEmpty)
typealias Foo = Bar

foo: ref.Reference<ref.Domain, Foo>
```

Also, even more indirectly:

```pkl
import "pkl:ref"

typealias Bar = Mapping<String, String(isEmpty)>
typealias Foo = Bar

foo: ref.Reference<ref.Domain, Foo>
```

pkl-core/src/main/java/org/pkl/core/runtime/VmTypeAlias.java
```diff
@@ -203,8 +202,12 @@ public TypeNode instantiate(
         });
     clone.accept(
         node -> {
-          if (node instanceof ReferenceTypeNode referenceTypeNode) {
-            referenceTypeNode.validateTypeArguments(typeAliasTypeNodeSourceSection);
```
bioball, line 207: [nit] I think it's better to keep method `validateTypeArguments`, and move all this logic inside there

pkl-core/src/test/files/LanguageSnippetTests/input/errors/reference21.pkl
```diff
@@ -0,0 +1,7 @@
+import "pkl:ref"
+
+typealias Ref<T> = ref.Reference<ref.Domain, T>
+
+typealias Ref2<T> = Ref<T>
+
+foo: Ref2<String(isEmpty)>
```
bioball, line 7: Add these tests:

```pkl
import "pkl:ref"

typealias Ref<T> = ref.Reference<ref.Domain, T>
typealias Ref2<T> = Ref<T(isEmpty)>

foo: Ref2<String>
```

And:

```pkl
import "pkl:ref"

typealias Foo1 = Foo2
typealias Foo2 = String(isEmpty)

foo: ref.Reference<ref.Domain, Foo1>
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2173,26 +2181,68 @@ private Object doEval(VmReference value, VmTyped module) {
...
+                aliasLayers.get(i).getSourceSection(), owner.getQualifiedName()));
+      }
+      return frames;
+    }
+
+    /**
+     * Signals that a {@code Reference} reached through one or more generic type aliases has a
+     * constrained referent.
+     */
+    public static final class ReferentConstraintException extends RuntimeException {
```
bioball: Hmm... do we need this class? Can we just throw `VmException` instead with the leading stack frames already set?

pkl-core/src/main/java/org/pkl/core/ast/type/UnresolvedTypeNode.java
```diff
@@ -306,7 +306,15 @@ public TypeNode execute(VirtualFrame frame) {
...
           resolvedTypeArgumentNodes[i] = typeArgumentNodes[i].execute(frame);
         }
-        return new TypeAliasTypeNode(sourceSection, typeAlias, resolvedTypeArgumentNodes);
+        try {
+          return new TypeAliasTypeNode(sourceSection, typeAlias, resolvedTypeArgumentNodes);
+        } catch (ReferenceTypeNode.ReferentConstraintException e) {
+          // a constraint reached a `Reference` referent through this generic alias.
+          var exception =
+              exceptionBuilder().evalError("invalidReferenceTypeAnnotationWithConstraint").build();
+          exception.setLeadingStackFrames(e.getLeadingStackFrames());
```
bioball: [nit] Move `setLeadingStackFrames()` to the exception builder; `withLeadingStackFrames(...).build()`

bioball (changes requested):

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2173,26 +2187,67 @@ private Object doEval(VmReference value, VmTyped module) {
...
+        @Nullable VmTypeAlias outermostAlias) {
+      var frames = new ArrayList<StackFrame>();
+      for (Node node = constraintNode; node != null; node = node.getParent()) {
+        if (!(node instanceof ConstrainedTypeNode
+            || node instanceof TypeAliasTypeNode
+            || node instanceof ReferenceTypeNode)) {
+          continue;
+        }
+        var section = node.getSourceSection();
+        if (section == null || !section.isAvailable() || isWithin(usageSection, section)) {
```
bioball, line 2224: For some reason, my IntelliJ thinks that `section == null` is always false

```suggestion
        //noinspection ConstantValue
        if (section == null || !section.isAvailable() || isWithin(usageSection, section)) {
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2173,26 +2187,67 @@ private Object doEval(VmReference value, VmTyped module) {
...
+     */
+    private static @Nullable VmTypeAlias ownerAlias(
+        Node node, @Nullable VmTypeAlias outermostAlias) {
+      for (var parent = node.getParent(); parent != null; parent = parent.getParent()) {
+        if (parent instanceof TypeAliasTypeNode aliasNode) {
+          return aliasNode.getTypeAlias();
+        }
+      }
+      return outermostAlias;
+    }
```
bioball, line 2248: [nit] Truffle has a util method `NodeUtil.findParent` for this

Also, not sure if it's just me, but I'm seeing a lot of incorrect analysis of 3rd party libs (IJ incorrectly thinks they return `NonNull` values

```suggestion
    @SuppressWarnings("DataFlowIssue")
    private static @Nullable VmTypeAlias ownerAlias(
        Node node, @Nullable VmTypeAlias outermostAlias) {
      var parent = NodeUtil.findParent(node, TypeAliasTypeNode.class);
      //noinspection ConstantValue
      if (parent != null) {
        return parent.typeAlias;
      }
      return outermostAlias;
    }
```

pkl-core/src/main/java/org/pkl/core/runtime/VmExceptionBuilder.java
```diff
@@ -374,64 +384,67 @@ public VmException build() {
...
+                programValues,
+                location,
+                sourceSection,
+                memberName,
+                hintBuilder,
+                effectiveInsertedStackFrames,
+                wrappedException);
+          }
+        };
+    exception.setLeadingStackFrames(leadingStackFrames);
```
bioball: [nit] add leadingStackFrames to a constructor arg to each of these and pass it in; and remove `setLeadingStackFrames` on `VmException`

bioball (approved):
LGTM! Left some nits, but approving to unblock

HT154:
I explored taking this a little further in #1698. It's based on the code here, but extracts common validation stuff into its own class for reuse. It also avoids leaking details of `ReferenceTypeNode` into `TypeAliasTypeNode` like this PR does.

stackoverflow:
> I explored taking this a little further in #1698. It's based on the code here, but extracts common validation stuff into its own class for reuse. It also avoids leaking details of `ReferenceTypeNode` into `TypeAliasTypeNode` like this PR does.

Nice! I merged this, so you can rebase from main now.

PR #1698, opened 2026-06-24
Correctly type check `Class<T>`

### LINKED ISSUE #222: Constant expression analysis through `Class<T>` doesn't match evaluator
Code and diags:
```pkl
open class A
open class B
class C extends A

foo = C is Class<A>
//    ^^^^^^^^^^^^^
//    Expression is always 'true'

bar = C is Class<B>
//    ^^^^^^^^^^^^^
//    Expression is always 'false'
```

Actual evaluation result:
```pkl
foo = true
bar = true
```

Pkl type-erases `Class`'s argument during eval, but this isn't the case in-editor. pkl-lsp likely has the same behavior.

In this case, it might actually be preferable to have Pkl match the editor experience here instead of conforming the editor to the language.

### LINKED ISSUE #1729: `Class<T>` is always type-erased to `Class`
Currently there's no way to test that a `Class` value is the class (or a subclass of) a particular class.

```pkl
open class A
open class B
class C extends A

foo = C is Class<A>
bar = C is Class<B>
```
Results in:
```pkl
foo = true
bar = true
```
Even though `C` is not [a subclass of] `B`.

Pkl should instead _not_ erase the type parameter:
* Type checks against `Class<T>` should check that the value is a `Class` and that the class is [a subclass of] `T`
* `Class<T>` type annotations should forbid non-class types for `T`, including union, nullable, constrained, parameterized, string literal, and `nothing` types.

See related: https://github.com/apple/pkl-intellij/issues/222

HT154:
This PR removes the unconditional type-erasure of `Class<T>` to `Class`.

This is a breaking change when `T` is a union type, nullable type, string literal type, parameterized type, or `nothing`; these will never type-check.

For convenience, erasure still occurs when `T` is `unknown` or `Any`. Similarly, the subclass check is skipped when `T` is an un-replaced type variable, e.g. in generic classes/methods (type aliases replace the type arg during instantiation).

Resolves #1729
Closes https://github.com/apple/pkl-intellij/issues/222


bioball:
General comment: given this:

```pkl
typealias MyNumber = Number
foo: Class<MyNumber> = Number
```

This seems like an incorrect error, because `MyNumber` is just an alias for `Number`.

```
–– Pkl Error ––
`Class` type arguments may only be class types, module types, `unknown`, or `module`.

9 | foo: Class<MyNumber> = Number
         ^^^^^^^^^^^^^^^
at test (/Users/danielchao/code/apple/pkl/.dan-scripts/test.pkl:9)
```

Might also be okay to _not_ throw for type arguments which don't point to classes. In these cases, the type would just never typecheck, and it would behave just like the `nothing` type. For example:

```pkl
a: Class<Foo?>
b: Class<"foo" | "bar>
```

HT154:
> General comment: given this:
> 
> ```pkl
> typealias MyNumber = Number
> foo: Class<MyNumber> = Number
> ```
> 
> This seems like an incorrect error, because `MyNumber` is just an alias for `Number`.

Can you say more about how this would be useful? The only case I can think of is the compatibility shim typealias when a class is moved out into its own module.

If this was supported, it would only work for aliases to bare class types, which is of limited utility. An alias may hide that type is not an ordinary class and result in an error (as-is) or a `nothing` type (as you suggest below), especially if a library author later refactors an alias to include a constraint where one did not exist previously.

> Might also be okay to _not_ throw for type arguments which don't point to classes. In these cases, the type would just never typecheck, and it would behave just like the `nothing` type. For example:
> 
> ```pkl
> a: Class<Foo?>
> b: Class<"foo" | "bar>
> ```

Again, I think it's more surprising that users could inadvertently build a `nothing` type without realizing and have all their type checks silently fail. This would still be a breaking change compared to the existing behavior (where these are essentially erased to `unknown`) and I'd prefer to break in the direction of providing helpful user feedback than silent and potentially-surprising behavior.

bioball:
> Can you say more about how this would be useful? The only case I can think of is the compatibility shim typealias when a class is moved out into its own module.

An invariant of a typealias is that: on the type level, it just stands for its aliased type; therefore `Class<MyNumber>` should be the same as `Class<Number>`.

> Again, I think it's more surprising that users could inadvertently build a nothing type without realizing and have all their type checks silently fail

I'd expect the failures to be loud, rather than silent! If you try to assign anything to `Class<Foo?>`, Pkl throws an exception. Also, I'd expect that any usage would also be caught statically (e.g. red squiggly lines in the editor during assignment). We can also add a warning diagnostic if you declare `Class<Foo?>` (or similar).

HT154:
Updated this to unwrap typealiases during class calculation, so aliases (to aliases)+ to classes now behave as if the underlying class was used directly.

Also updated this to error at runtime rather than during `ClassClassTypeNode` init (or alias instantiation) with a hint that such type checks can never succeed. I think the errors are less helpful in cases where the `Class` type node or the type arg are behind aliases (since previously it relied on the same machinery as `ReferenceTypeNode`'s constraint-in-referent checking). The only real difference other than error message content is that "bad" type annotations that never execute never raise errors. I can live with either approach, but I think the runtime errors aren't as good.

RushikeshGandhmal:
While comparing #1763 with this PR, I noticed one case that may still need coverage: exported schemas. `ClassClassTypeNode` exposes the type argument through mirrors but does not override `doExport()`, so the inherited implementation appears to export `Class<A>` as raw `Class`.

In #1763 I added a `doExport()` implementation returning `PType.Class(Class, typeArgument)` and a regression test using `Evaluator.evaluateSchema()` to verify that `Class<A>` preserves `A`. I’d be happy to contribute that focused test/change here if useful.

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -200,6 +200,9 @@ Duplicate type parameter `{0}`.
 notAParameterizableClass=\
 Class `{0}` does not have type parameters.
 
+notAClassType=\
+`Class` type arguments may only be class types, module types, `unknown`, or `module`.
+
```
stackoverflow:
```suggestion
```
This doesn't seem to be used.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -3219,6 +3235,101 @@ protected boolean acceptTypeNode(boolean visitTypeArguments, TypeNodeConsumer co
...
+    }
+
+    @Override
+    protected boolean isIncludedInTrace(Node node) {
+      // getViolatingNode() always returns null, it's only used to re-calculate clazz
+      throw PklBugException.unreachableCode();
+    }
+
+    @Override
+    public @Nullable Node getViolatingNode() {
```
stackoverflow: The way you are using `getViolatingNode` just for side effects instead of the intended purpose is a bit hacky. I'd at least add a comment stating the reasons here.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -3219,6 +3235,101 @@ protected boolean acceptTypeNode(boolean visitTypeArguments, TypeNodeConsumer co
     }
   }
 
+  public abstract static class ClassClassTypeNode extends ValidatingObjectSlotTypeNode {
+
+    @Child private TypeNode typeNode;
+    private @Nullable VmClass clazz;
```
stackoverflow: I think this could be `@CompilationFinal`.

stackoverflow (approved):
LGTM.
Worth adding the breaking change to the changelog already.

pkl-core/src/main/java/org/pkl/core/ast/type/VmTypeMismatchException.java
```diff
@@ -169,6 +169,69 @@ protected Boolean hasHint() {
...
+    @Override
+    protected Boolean hasHint() {
+      return expectedClass == null;
+    }
+
+    @Override
+    public void buildHint(AnsiStringBuilder builder, String indent, boolean withPowerAssertions) {
+      if (expectedClass != null) return;
+      builder.append(
+          "A `Class` type check can only succeed when its type argument is an un-parameterized class, a module, `unknown`, `module`, or an alias to one of those types.");
```
bioball: [nit] Move this to `errorMessages.properties`

Also, what does "a module" mean? This error message already contains `module`; duplicated by mistake?

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -61,10 +62,18 @@
...
+   * Type node that corresponds to a simple, unparameterized {@link VmClass}.
+   *
+   * <p>This includes generic classes written without any type arguments like {@code List}.
+   */
+  public interface SimpleClassTypeNode {
     VmClass getVmClass();
   }
 
+  /** Type node that corresponds to a user-defined class (or module class). */
+  public interface UserClassTypeNode extends SimpleClassTypeNode {}
```
bioball: I'm not sure if the `SimpleClassTypeNode` here is doing too much; `TypeNode` already has `VmClass getVmClass()`.

We can keep the existing class hierarchy (only `ClassTypeNode`), and `getViolatingNode()`'s check can just be:

```java
    @Override
    public @Nullable Node getViolatingNode() {
      // use the validation hook to recalculate clazz after typealias instantiation

      var node = typeNode;
      while (node instanceof TypeAliasTypeNode typeAliasTypeNode) {
        node = typeAliasTypeNode.getAliasedTypeNode();
      }

      if (node instanceof UnknownTypeNode || node instanceof TypeVariableNode) {
        clazz = BaseModule.getAnyClass();
      } else {
        clazz = node.getVmClass();
      }
      return null;
    }
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -3164,6 +3180,104 @@ protected boolean acceptTypeNode(boolean visitTypeArguments, TypeNodeConsumer co
...
+    }
+
+    @Fallback
+    protected Object fallback(Object value) {
+      throw typeMismatch(value, BaseModule.getClassClass());
+    }
+
+    @Override
+    protected boolean acceptTypeNode(boolean visitTypeArguments, TypeNodeConsumer consumer) {
+      return consumer.accept(this);
```
bioball, line 3236: `Class<T>` is parameterized, so this should be:

```suggestion
      if (visitTypeArguments) {
        return consumer.accept(this) && typeNode.acceptTypeNode(true, consumer);
      }
      return consumer.accept(this);
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -3164,6 +3180,104 @@ protected boolean acceptTypeNode(boolean visitTypeArguments, TypeNodeConsumer co
     }
   }
 
+  public abstract static class ClassClassTypeNode extends ValidatingObjectSlotTypeNode {
```
bioball: Shouldn't need to extend `ValidatingObjectSlotTypeNode`; this can be:

```java
  public abstract static class ClassClassTypeNode extends ObjectSlotTypeNode {

    @Child private TypeNode typeNode;
    @CompilationFinal private @Nullable VmClass clazz;

    public ClassClassTypeNode(SourceSection sourceSection, TypeNode typeNode) {
      super(sourceSection);
      this.typeNode = typeNode;
    }

    private void initVmClass() {
      if (clazz != null) {
        return;
      }
      var node = typeNode;
      while (node instanceof TypeAliasTypeNode typeAliasTypeNode) {
        node = typeAliasTypeNode.getAliasedTypeNode();
      }

      if (node instanceof UnknownTypeNode || node instanceof TypeVariableNode) {
        clazz = BaseModule.getAnyClass();
      } else {
        clazz = node.getVmClass();
      }
    }

    @Specialization
    protected Object eval(VmClass value) {
      initVmClass();
      // rest of the method
    }
}
```

bioball: Then, we can get rid of all the overrides that don't do anything.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -3164,6 +3180,104 @@ protected boolean acceptTypeNode(boolean visitTypeArguments, TypeNodeConsumer co
...
+
+    @Override
+    public VmList getTypeArgumentMirrors() {
+      return VmList.of(typeNode.getMirror());
+    }
+
+    @Override
+    protected PType doExport() {
+      return new PType.Class(BaseModule.getClassClass().export(), typeNode.doExport());
+    }
```
bioball, line 3256: Since these are also changing other public API surface areas, can we add some tests around:

* reflect API giving type arguments for `Class<T>`
* Java schema evaluator giving metadata about type arguments

pkl-core/src/test/files/LanguageSnippetTests/input/types/classType.pkl
```diff
@@ -0,0 +1,26 @@
...
+
+res3 = C is Class
+res4 = C is Class<unknown>
+res5 = C is Class<Any>
+
+res15 = D is Class<module>
+res16 = D is Class<Object>
+res17 = D is Class<Typed>
+res18 = D is Class<Dynamic>
+res19 = D is Class<Int>
```
bioball, line 26: Let's also add a test for parenthesized types:

```pkl
C is Class<(C)>
C is Class<((C))>
```

bioball (changes requested):
Overall, looks good! Some comments.

pkl-core/src/main/java/org/pkl/core/ast/type/VmTypeMismatchException.java
HT154: "a module" here is an imported module used as a type, distinguished from "a class" being something declared with the `class` keyword. If you don't think it's worth distinguishing these, I'm happy to remove it.

bioball: Yeah, I think it's less confusing to just remove.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
HT154: This is not correct behavior because it implicitly erases `Class<Mapping<String, String>>` to `Class<Mapping>` which could result in behavior that is surprising to users not intimately familiar with Pkl's generics:
```pkl
foo: Mapping<Int, Int> = new { ... }
bar = foo.getClass() is Class<Mapping<String, String>> // true!
```

Marking non-parameterized `TypeNode` subclasses as implementing `SimpleClassTypeNode` is an easy way to distinguish these instead of manually checking for one of the 12 acceptable `TypeNode` subclasses.

bioball: Ah, I see! But, how about using `isParametric()` to guard against that?

```java
    @Override
    public @Nullable Node getViolatingNode() {
      // use the validation hook to recalculate clazz after typealias instantiation

      var node = typeNode;
      while (node instanceof TypeAliasTypeNode typeAliasTypeNode) {
        node = typeAliasTypeNode.getAliasedTypeNode();
      }

      if (node instanceof UnknownTypeNode || node instanceof TypeVariableNode) {
        clazz = BaseModule.getAnyClass();
      } else if (!node.isParametric()) {
        clazz = node.getVmClass();
      }
      return null;
    }
```

HT154: Ah, good call!

bioball (approved):
LGTM!

pkl-core/src/main/java/org/pkl/core/ast/type/VmTypeMismatchException.java
```diff
@@ -169,6 +169,68 @@ protected Boolean hasHint() {
...
+              "typeMismatch", indent, renderedExpected, renderedActualClass));
+    }
+
+    @Override
+    protected Boolean hasHint() {
+      return expectedClass == null;
+    }
+
+    @Override
+    public void buildHint(AnsiStringBuilder builder, String indent, boolean withPowerAssertions) {
```
stackoverflow, line 229: nit: all the other `buildHint` here have a `@TruffleBoundary`.

pkl-core/src/main/java/org/pkl/core/ast/type/VmTypeMismatchException.java
```diff
@@ -169,6 +169,68 @@ protected Boolean hasHint() {
     }
   }
 
+  public static final class Class extends VmTypeMismatchException {
```
stackoverflow: nit: `Class` shadows `java.lang.Class`. Would rather have this renamed. Maybe `ClassType` or `ParameterizedClass`.

stackoverflow (approved):
LGTM

PR #1699, opened 2026-06-24
Add 0.32 release notes

bioball:


stackoverflow (approved):
Looks good, but some missing stuff:

#1637: didn't see this anywhere in the changelog

Also #1714 was merged and can be added.

PR #1707, opened 2026-06-26
Fix many editor diagnostics

bioball:
This addresses many IDE warnings resulting from the switch to JSpecify.

In some cases, this moves `@LateInit` to `@Nullable`, because the latter requires less runtime assertions.

Also, this changes the behavior of exporting VmObject; there's no place in our code that does not force a VmObject prior to export, so the existing logic around handling nullable values has been removed.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/PcfRenderer.java
+++ b/pkl-core/src/main/java/org/pkl/core/PcfRenderer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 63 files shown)


pkl-core/src/main/java/org/pkl/core/PcfRenderer.java
```diff
@@ -227,9 +227,7 @@ private void doVisitProperties(Map<String, Object> properties) {
             write(currIndent);
             writeIdentifier(name);
 
-            if (value == null) { // unevaluated property
```
stackoverflow, line 230: This is a public API, so it's worth adding this breaking change to the changelog.

stackoverflow (approved):
Looks good

pkl-core/src/main/java/org/pkl/core/PcfRenderer.java
bioball: It's an impossible condition (exported values can never be `null`)

stackoverflow: Hmmm, yeah. In theory one can construct unforced objects and pass them to a pcf renderer. But I don't think it's worth pointing this out for such niche (and buggy) use case.

PR #1708, opened 2026-06-26
Add `this` as a self type

### LINKED ISSUE #1612: Pkl is missing a self-type
The lack of a self-type makes it hard/impossible to flow type in some scenarios.

For example, `Any.ifNonNull`:

We can probably re-use the `this` keyword to represent the self type.

Note: Pkl already has a self-type for the module, called `module`.

### LINKED ISSUE #1712: `module` types as typelias type args are evaluated in the context of the alias's module, not where `module` is used
```pkl

import "pkl:test"

x: Int = 1

res1 = fun(this)
res2 = fun1(this)
res3 = fun2(this)
```

```
–– Pkl Error ––

                                        ^^^^^^

xx | res2 = fun1(this)
            ^^^^^^^^^^

                                                                         ^^^^
```

HT154:
Depends on #1796

This PR:
* Adds type `this`, a fully featured self type for use in modules or classes.
  * Forbids use of `this` in typealias bodies.
* Deprecates usage of the `module` type in const scopes (class and typealias bodies, annotations, const properties/methods) with a warning, to be an error in a future release.
  * In some cases this induces incorrect or undefined behavior (see #1712).
  * Usages should either be replaced by `this` or a self-import and use of the current module.
* Adopts `this` in the standard library:
  * `pkl.base#Any.getClass()` now returns `Class<this>`.
  * `pkl.base#Any.ifNonNull()` now accepts an argument of type `(this) -> Result`.
  * `pkl.ref#Domain.renderReference()` now accepts a `Reference<this, Any>`.
* This also a corner case with `ReferenceTypeNode`, `module` types, and type aliases that caused `module` types in the referent position to always be resolved to the module where the `Reference` type annotation is instead of where `module` is used.

For ease of review, this PR consists of two commits; the first is all implementation changes, the second includes regenerated pkl-doc test output.

Resolves #1612
Resolves #1712


pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -410,44 +416,61 @@ protected boolean acceptTypeNode(boolean visitTypeArguments, TypeNodeConsumer co
...
+      this.pType = pType;
+      this.mirrorFactory = mirrorFactory;
+    }
+
+    public static FinalSelfTypeNode module(SourceSection sourceSection, VmClass clazz) {
+      return new FinalSelfTypeNode(
+          sourceSection, clazz, PType.MODULE, MirrorFactories.moduleTypeFactory);
+    }
+
+    public static FinalSelfTypeNode thiz(SourceSection sourceSection, VmClass clazz) {
```
HT154: Any recs for a better name here? 😅

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -185,7 +187,11 @@ public static VmList getMirrors(TypeNode[] nodes) {
     return builder.build();
   }
 
-  protected PType doExport() {
+  /**
+   * If {@code frame} is provided then self types should be resolved to real types, otherwise return
+   * the self PType
+   */
+  protected PType doExport(@Nullable VirtualFrame frame) {
```
HT154: `ReferenceTypeNode` needs to export the real class here, so when a frame is passed `NonFinalSelfTypeNode` uses it to get the target's class instead of exporting `PType.MODULE`/`PType.THIS`.

pkl-core/src/main/java/org/pkl/core/ast/type/UnresolvedTypeNode.java
```diff
@@ -99,12 +110,128 @@ public Module(SourceSection sourceSection) {
...
+
+    private void checkValidUsage() {
+      UnresolvedTypeNode.walkParents(
+          this,
+          (parent, prevParent) -> {
+            if (parent instanceof TypeAliasNode typeAliasNode) {
+              if (typeAliasNode.isBodyChild(prevParent)) {
+                throw exceptionBuilder()
+                    .evalError("invalidSelfTypeUsage", "this", "type alias")
+                    .build();
```
bioball: I think we should handle this error at parse time in AstBuilder, rather than here. We should walk the parser syntax tree, rather than Truffle

pkl-doc/src/main/kotlin/org/pkl/doc/PageGenerator.kt
```diff
@@ -347,6 +347,9 @@ internal abstract class PageGenerator<out S>(
       PType.MODULE -> {
         +"module"
       }
+      PType.THIS -> {
+        +"this"
+      }
```
bioball, line 352: Need to also handle `PType.THIS` in `SearchIndexGenerator` and `PackageDataGenerator`

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
bioball: Suggestion: `thisType`

pkl-core/src/main/java/org/pkl/core/ast/type/UnresolvedTypeNode.java
```diff
@@ -99,12 +110,128 @@ public Module(SourceSection sourceSection) {
...
+      if (parentProperty != null) {
+        var parentClass = NodeUtil.findParent(parentProperty, ClassNode.class);
+        if (parentClass != null) {
+          return parentClass.executeGeneric(frame);
+        }
+      }
+
+      return VmUtils.getClass(getReceiverNode.executeGeneric(frame));
+    }
+  }
```
bioball: Same comment as above: I think this should be done at parse time (within AstBuilder), not here.

This truffle node can receive the (nullable) enclosing class name, and use a combination of `GetModuleNode` (in the case of no enclosing module) and `ResolveSimpleDeclaredTypeNode` (in the case of an enclosing module) to get the correct `VmClass`

bioball: Within `AstBuilder`, we can walk up the symbol table scope until we find a Class scope to find the class identifier.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -461,25 +484,54 @@ protected boolean acceptTypeNode(boolean visitTypeArguments, TypeNodeConsumer co
...
+    }
+
+    public static NonFinalSelfTypeNode thiz(SourceSection sourceSection, VmClass clazz) {
+      return new NonFinalSelfTypeNode(
+          sourceSection, clazz, new GetReceiverNode(), PType.THIS, MirrorFactories.thisTypeFactory);
+    }
+
+    private VirtualFrame getEffectiveFrame(VirtualFrame frame) {
+      var realFrame = VmLanguage.get(this).localContext.get().getRealTypeAliasFrame();
+      return realFrame != null ? realFrame : frame;
```
bioball: I'm trying to understand what this code is doing; is this to address the case of `typealias Foo = module`?'

If so, I think probably a better solution here is: when we execute a `TypeAliasNode`, we can replace the `module` type when we execute the type alias itself (e.g. inside `TypeAliasNode#executeGeneric`)

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -461,25 +484,54 @@ protected boolean acceptTypeNode(boolean visitTypeArguments, TypeNodeConsumer co
...
     }
 
     @Override
     protected Object executeLazily(VirtualFrame frame, Object value) {
-      var moduleClass = ((VmTyped) getModuleNode.executeGeneric(frame)).getVmClass();
+      var moduleClass =
```
```suggestion
      var clazz =
```

pkl-core/src/test/files/LanguageSnippetTests/input/types/thisType3.pkl
```diff
@@ -0,0 +1,21 @@
...
+
+res2 = (baz) {
+  [[this is this]] {
+    res2 = "blah"
+  }
+  when (module is this) {
+    123
+  }
+  module is this
+}
```
bioball, line 21: [nit] Can you name these files as `thisTypeX`? e.g. `thisType1`, `thisType2`, etc

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -491,25 +543,28 @@ protected Object executeLazily(VirtualFrame frame, Object value) {
...
     }
 
     @Override
-    protected PType doExport() {
-      return PType.MODULE;
+    protected PType doExport(@Nullable VirtualFrame frame) {
+      if (frame == null) return pType;
+      var moduleClass =
+          ((VmObjectLike) getTargetNode.executeGeneric(getEffectiveFrame(frame))).getVmClass();
+      return new PType.Class(moduleClass.export());
```
bioball: This is a breaking change (`PType` is a user-facing API)

This affects the schema evaluator; affects the shape of declared members of `PClass`

I think it might also affect the doc generator?

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -128,35 +123,34 @@ private static void normalizeTypes(PType type, PClass moduleClass, Set<PType> re
...
+        normalizeTypes(t, result);
       }
-    } else if (type == PType.MODULE) {
-      result.add(new PType.Class(moduleClass));
     } else {
       // remaining types: PType.Function, PType.TypeVariable. no normalizing needed; TypeVariable
       // gets replaced upon instantiation, and Function can bubble up to users as a reference error
       // if accessed.
       result.add(type);
+      // PType.MODULE and PType.THIS can never be encountered here; caller must deref self types
```
bioball, line 176: I don't think this is actually happening here? This class is just calling `clazz.export()` in the constructor, which won't pass in a frame, which won't defer self types.

bioball: Also: I don't know if it's worth investing too much into this code path; I think we should introduce a `VmType` value. If we had that, the `TypeNode.export()` call wouldn't be used.

pkl-core/src/main/java/org/pkl/core/PType.java
```diff
@@ -57,6 +57,17 @@ public String toString() {
         }
       };
 
+  /** The type of the enclosing owner. */
```
bioball:
```suggestion
  /** The {@code this} type. */
```

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1203,3 +1203,6 @@ Redirected to: `{1}`
 
 invalidReferenceTypeAnnotationWithConstraint=\
 `Reference` referent type argument may not include type constraints.
+
+invalidSelfTypeUsage=\
+`{0}` type annotations are not allowed in {1} bodies.
```
bioball: A "type annotation" is the thing on the right-hand side of a typed identifier; `foo: Bar` has type annotation `: Bar`

```suggestion
`{0}` types are not allowed in {1} bodies.
```

pkl-core/src/main/java/org/pkl/core/runtime/VmTypeAlias.java
```diff
@@ -231,6 +231,7 @@ public void force(boolean allowUndefinedValues) {
   }
 
   @Override
+  @TruffleBoundary
```
bioball, line 234: Leftover code? This didn't need a truffle boundary before, so I wouldn't expect this to need one now

pkl-core/src/test/files/LanguageSnippetTests/output/types/thisType2.err
```diff
@@ -0,0 +1,14 @@
+–– Pkl Error ––
+`this` type annotations are not allowed in type alias bodies.
```
bioball: For consistency with other error messages, I think let's phrase this as:

```
Cannot declare `this` types inside type alias bodies.
```

bioball (changes requested):
Left some comments!

The major ones are:

* Much of the resolution logic should happen inside `AstBuilder`; we shouldn't need to traverse the truffle tree
* Not sure about the `realTypeAliasFrame` logic added to type node execution; this feels like a workaround to a deeper issue
* Not sure if `TypeNode.export()` should be deferencing self types; this feels like the responsibility of the caller, not `TypeNode` itself

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
HT154: How so? Only ReferenceTypeNode exercises the branch where `frame != null`. Regular `TypeNode.export()` calls are not affected at all.

bioball: ooh gotcha!

pkl-core/src/main/java/org/pkl/core/ast/type/UnresolvedTypeNode.java
bioball: We can also maybe introduce a `ReadClassNode` or something like that perhaps that receives the identifier name, whether it's local, and whether it's the module class.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
HT154: Per our discussion the other day, this approach doesn't work because we need the runtime receiver to resolve `this` when it is present via a type argument to the alias, so doing anything at alias instantiation time does not work.

I've replaced the usage of `VmLocalContext` with an auxiliary frame slot (and in doing so removed assumptions around aux slots only being used for custom `this`).

pkl-core/src/main/java/org/pkl/core/runtime/VmTypeAlias.java
HT154: This is needed because the call to `referentTypeNode.doExport` has moved above an explicit truffle boundary in `ReferenceTypeNode`. There's a matching one on `VmClass.export`, but it was missing here. All prior existing call sites happened to be below truffle boundaries too.

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
HT154: Agreed on introducing `VmType`, but we do still need to correctly handle self types in references with this PR. I've revised this a bit and added tests around properties containing self types.

pkl-core/src/main/java/org/pkl/core/ast/expression/member/ReadClassNode.java
```diff
@@ -0,0 +1,75 @@
...
+import org.pkl.core.ast.member.ObjectMember;
+import org.pkl.core.runtime.Identifier;
+import org.pkl.core.runtime.VmTyped;
+
+/** Read a class by name from the current receiver's module */
+public final class ReadClassNode extends ExpressionNode {
+
+  private final Identifier className;
+  private final boolean isLocal;
+  private @Child ExpressionNode getModuleNode;
```
bioball: `@Child` should be applied to the field, rather than the type

```suggestion
  @Child protected ExpressionNode getModuleNode;
```

pkl-core/src/main/java/org/pkl/core/ast/expression/member/ReadClassNode.java
```diff
@@ -0,0 +1,75 @@
...
+  @Child private @Nullable DirectCallNode callNode;
+
+  public ReadClassNode(SourceSection sourceSection, Identifier className, boolean isLocal) {
```
bioball: The identifier already tells if the property is local or not (see `Identifier#isLocalProp()`); the differentiating logic here can be based off `isLocalProp()`

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -293,9 +299,14 @@ public class AstBuilder extends AbstractAstBuilder<Object> {
   private final boolean isMethodReturnTypeChecked;
 
   public AstBuilder(
-      Source source, VmLanguage language, ModuleInfo moduleInfo, ModuleResolver moduleResolver) {
+      Source source,
+      VmLanguage language,
+      Logger logger,
```
bioball: No need to change anything here, just noting: I actually think it would have been fine to just call `VmContext.get(null).getLogger()` from here.


1. Passing a node in (in the case of `VmContext.get(node)`) lets truffle [optimize away the context lookup](https://www.graalvm.org/truffle/javadoc/com/oracle/truffle/api/TruffleLanguage.ContextReference.html#get(com.oracle.truffle.api.nodes.Node)). However, AstBuilder is never part of compilation, so it's not concerning to call `VmContext.get(null)`
2. The lookup cost of `VmContext.get()` is negligible compared to what `AstBuilder` is doing (and we pay for it once per source file).

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -365,7 +377,59 @@ public UnresolvedTypeNode visitNothingType(NothingType type) {
...
+            VmUtils.createStackFrame(sourceSection, null));
+        break;
+      } else if (scope.isTypeAliasScope()) {
+        logger.warn(
+            ErrorMessages.create("invalidSelfTypeUsage", "module", "type alias")
+                + " This will be an error in a future release.",
+            VmUtils.createStackFrame(sourceSection, null));
+        break;
+      }
+    }
```
bioball, line 405: [nit] `Type getSuperClass()` doesn't really match symbol table's responsibility; `Type` is a parser node.

I think this can instead be:

```suggestion
  private void checkModuleType(ModuleType moduleType, SourceSection sourceSection) {
    // `extends module` is okay
    if (moduleType.parent() instanceof Class classNode && classNode.getSuperClass() == moduleType) {
      return;
    }
    for (var scope = symbolTable.getCurrentScope(); scope != null; scope = scope.getParent()) {
      if (scope.isAnnotationScope()) break;
      if (scope.isClassScope()) {
        logger.warn(
          ErrorMessages.create("invalidSelfTypeUsage", "module", "class")
            + " This will be an error in a future release.",
          VmUtils.createStackFrame(sourceSection, null));
        break;
      } else if (scope.isTypeAliasScope()) {
        logger.warn(
          ErrorMessages.create("invalidSelfTypeUsage", "module", "type alias")
            + " This will be an error in a future release.",
          VmUtils.createStackFrame(sourceSection, null));
        break;
      }
    }
  }

  @Override
  public UnresolvedTypeNode visitModuleType(ModuleType type) {
    var sourceSection = createSourceSection(type);
    checkModuleType(type, sourceSection);
    return new UnresolvedTypeNode.Module(sourceSection);
  }
```

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -365,7 +377,59 @@ public UnresolvedTypeNode visitNothingType(NothingType type) {
...
+            .build();
+      }
+    }
+
+    var getClassNode =
+        classScope == null
+            ? new GetReceiverClassNode(sourceSection)
+            : isBaseModule
+                ? new GetBaseModuleClassNode(classScope.getName())
+                : new ReadClassNode(sourceSection, classScope.getName(), classScope.isLocal());
```
bioball: `ReadClassNode` only needs the identifier; so this can be:

```suggestion
    org.pkl.core.runtime.Identifier className = null;
    for (var scope = symbolTable.getCurrentScope(); scope != null; scope = scope.getParent()) {
      if (scope.isObjectScope() || scope.isCustomThisScope()) {
        break;
      }
      if (scope instanceof ClassScope foundClassScope) {
        className = foundClassScope.getName();
        break;
      }
      // it's still safe to break on ObjectScope because this is valid:
      // typealias Foo = List(any((it) -> it == new Dynamic { it is this })) // this == Dynamic
      if (scope.isTypeAliasScope()) {
        throw exceptionBuilder()
            .withSourceSection(sourceSection)
            .evalError("invalidSelfTypeUsage", "this", "type alias")
            .build();
      }
    }

    var getClassNode =
        className == null
            ? new GetReceiverClassNode(sourceSection)
            : isBaseModule
                ? new GetBaseModuleClassNode(className)
                : new ReadClassNode(sourceSection, className);
```

pkl-core/src/test/files/LanguageSnippetTests/input/types/currentModuleType7.pkl
```diff
@@ -0,0 +1,19 @@
+// test that `module` type use in annotations is allowed
+@Foo { baz = bar is module }
```
bioball: Not sure this should be allowed either, for the same reason that it shouldn't be allowed inside a class: this is a "static" context, where extending this module shouldn't change the meaning of the annotation.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2748,26 +2820,39 @@ protected Object executeLazily(VirtualFrame frame, Object value) {
       setOwner(frame, VmUtils.getOwner(typeAlias.getEnclosingFrame()));
       setReceiver(frame, VmUtils.getReceiver(typeAlias.getEnclosingFrame()));
 
+      var realFrameSlot = getRealFrameSlot(frame);
+      var ownsRealFrame = frame.getAuxiliarySlot(realFrameSlot) == null;
+      if (ownsRealFrame)
+        frame.setAuxiliarySlot(realFrameSlot, new FakeFrame(prevReceiver, prevOwner));
```
bioball: [nit] use curly braces unless the logic is a short, single-line clause (e.g. `if (foo == null) return`).

```suggestion
      if (ownsRealFrame) {
        frame.setAuxiliarySlot(realFrameSlot, new FakeFrame(prevReceiver, prevOwner));
      }
```

Same comment applies to the rest of the changes in this file

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
bioball: Do we need the logic around passing a frame in still?

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2875,6 +2968,45 @@ private static void setReceiver(Frame frame, Object receiver) {
...
+      @Override
+      public Object[] getArguments() {
+        return args;
+      }
+
+      @Override
+      public MaterializedFrame materialize() {
+        return this;
+      }
+    }
```
bioball: I went down the rabbit hole today and I think I have a better solution to this whole thing.

The whole reason we are doing owner/receiver replacements is so that we can resolve variables within constraint expressions.

However, now that we are resolving variables at parse time (yay!), we don't need to be swapping out owner/receiver anymore. Instead, we can just produce ReadProperty/InvokeMethod nodes that are qualified off of the typealias's enclosing module. Then, we shouldn't need to be dealing with this logic around preserving the frame's _original_ owner/reciever.

Since this is all orthogonal to the `this` type, here's a PR that implements this change!

https://github.com/apple/pkl/pull/1796

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
HT154: I think so, yeah. Even with the changes from #1796, `ReferenceTypeNode` still needs to erase self types in referents to their real classes before calling into `VmReference`. With #1796 and the other changes here, I think that _could_ be done post-export, but it would mean walking/transforming the `PType` tree, which we have no machinery for. Preserving this mechanism is both more efficient (don't need to walk the type tree twice) and avoids potentially introducing new API to or helper code for `PType`.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
HT154: I've rebased this PR on #1796 and fully agree that it is the path forward, reducing a lot of the complexity around resolving the correct receiver for determining the `this` type. All of the aux frame slot stuff has been removed here now.

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -365,7 +381,89 @@ public UnresolvedTypeNode visitNothingType(NothingType type) {
...
+            .build();
+      }
+    }
+
+    var getClassNode =
+        className == null
+            ? new GetReceiverClassNode(sourceSection)
+            : isBaseModule
+                ? new GetBaseModuleClassNode(className)
+                : new ReadClassNode(sourceSection, className);
```
bioball: With the introduction of `ReadQualifiedLocalPropertyNode`, we no longer need a `ReadClassNode`:

```suggestion
    ExpressionNode getClassNode;
    if (isBaseModule && className != null) {
      getClassNode = new GetBaseModuleClassNode(className);
    } else if (className == null) {
      getClassNode = new GetReceiverClassNode(sourceSection);
    } else if (className.isLocalProp()) {
      getClassNode = new ReadQualifiedLocalPropertyNode(sourceSection, className, false, new GetModuleNode(sourceSection));
    } else {
      getClassNode = ReadPropertyNodeGen.create(sourceSection, className, false, new GetModuleNode(sourceSection));
    }
```

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -369,7 +372,99 @@ public UnresolvedTypeNode visitNothingType(NothingType type) {
...
+      }
+    }
+    if (errorMessage != null) {
+      VmContext.get(null)
+          .getLogger()
+          .warn(
+              ErrorMessages.create(errorMessage, errorArgs)
+                  + " This will be an error in a future release.",
+              VmUtils.createStackFrame(sourceSection, null));
+    }
```
bioball: I was kind of confused why we were first looking for classes/annotations/etc before looking for properties.

I think this impl is clearer

It also addresses an issue where it would blame a nested property instead of the originating property marked `const`

```suggestion
  private void checkModuleType(ModuleType type, SourceSection sourceSection) {
    // `class X extends module` is fine
    if (type.parent() instanceof Class classNode && classNode.getSuperClass() == type) {
      return;
    }

    var currentScope = symbolTable.getCurrentScope();
    if (!currentScope.getConstLevel().isConst()) {
      return;
    }

    String errorMessage = null;

    // only classes/typealiases/annotations will apply "MODULE" const level
    if (currentScope.getConstLevel() == ConstLevel.MODULE) {
      for (var scope = currentScope; scope != null; scope = scope.getParent()) {
        if (scope.isAnnotationScope()) {
          errorMessage = ErrorMessages.create("invalidModuleTypeInAnnotation");
          break;
        } else if (scope.isClassScope()) {
          errorMessage = ErrorMessages.create("invalidModuleTypeInClass");
          break;
        } else if (scope.isTypeAliasScope()) {
          errorMessage = ErrorMessages.create("invalidModuleTypeInTypeAlias");
          break;
        }
      }
    }

    // Only properties and methods will apply "ALL" const level
    else {
      for (var scope = currentScope; scope != null; scope = scope.getParent()) {
        if (scope.isPropertyScope() || scope.isMethodScope()) {
          var parentScope = scope.getParent();
          assert parentScope != null;
          // if the parent also has "ALL", we haven't found the originating const property/method
          // yet.
          if (parentScope.getConstLevel() == ConstLevel.ALL) {
            continue;
          }
          var message =
              scope.isPropertyScope() ? "invalidModuleTypeInProperty" : "invalidModuleTypeInMethod";
          errorMessage = ErrorMessages.create(message, scope.getQualifiedName());
        }
      }
    }
    assert errorMessage != null;
    VmContext.get(null)
        .getLogger()
        .warn(
            errorMessage + " This will be an error in a future release.",
            VmUtils.createStackFrame(sourceSection, null));
  }
```

To add coverage for that bug, update test to:

```diff
diff --git a/pkl-core/src/test/files/LanguageSnippetTests/input/types/currentModuleType4.pkl b/pkl-core/src/test/files/LanguageSnippetTests/input/types/currentModuleType4.pkl
index def8514c..10907e12 100644
--- a/pkl-core/src/test/files/LanguageSnippetTests/input/types/currentModuleType4.pkl
+++ b/pkl-core/src/test/files/LanguageSnippetTests/input/types/currentModuleType4.pkl
@@ -11,3 +11,9 @@ class Baz extends Annotation {
 
 const qux = true is module
 const function quux() = true is module
+
+const bar = new {
+  baz {
+    qux = true is module
+  }
+}
```

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -369,7 +372,99 @@ public UnresolvedTypeNode visitNothingType(NothingType type) {
...
+  }
+
+  @Override
+  public UnresolvedTypeNode visitThisType(ThisType type) {
+    var sourceSection = createSourceSection(type);
+    // need to pass explicit class name for property and method arg/return type annotations
```
bioball: I forgot why we needed this and had to re-think through the whole problem again; to help out readers like me in the future let's add:

```suggestion
    // need to pass explicit class name for property and method arg/return type annotations.
    // this is because type annotations on class properties/methods are initialized when the ClassNode
    // is executed, and the frame's receiver is the enclosing module rather than the class.
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
bioball: Is it okay if we leave this out for now? I suspect this is temporary code; we shouldn't be using `PType` for references. And, the behavior of "if you give me a frame I behave differently" feels like a code smell.

I'm okay with just leaving references in a broken state for now. And, if it helps, we can change the signature for `renderReference` back to `abstract function renderReference(reference: Reference<Domain, Any>): String`

bioball (changes requested):

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -369,7 +372,99 @@ public UnresolvedTypeNode visitNothingType(NothingType type) {
...
+        continue;
+      }
+      if (scope.isPropertyScope()) {
+        errorMessage = "invalidModuleTypeInProperty";
+        errorArgs = new Object[] {scope.getQualifiedName()};
+      } else if (scope.isMethodScope()) {
+        errorMessage = "invalidModuleTypeInMethod";
+        errorArgs = new Object[] {scope.getQualifiedName()};
+      } else {
+        // all possibly const scopes should be covered in one of these loops
```
stackoverflow: Are they?
```pkl
const f = (x: module) -> x
```
```bash
❱ jpkl eval src/foo.pkl 

org.pkl.core.runtime.VmBugException%3A%20Unreachable%20code.

–– Pkl Error ––
Unreachable code.

Pkl 0.33.0-dev+7e94e965 (macOS 26.5.2, Java 25.0.3)

org.pkl.core.runtime.VmBugException: Unreachable code.
```

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
stackoverflow, line 1: Both the Java code generator and the Kotlin one don't handle `PType.MODULE` or `PType.THIS`.
```pkl
  module my.mod

  class Node {
    next: this?
  }
```

```
# or pkl-gen-kotlin
pkl-gen-java foo.pkl -o out

  Java (pkl-codegen-java @ 7e94e965):

  Exception in thread "main" java.lang.AssertionError: Encountered unexpected PType subclass: this
        at org.pkl.codegen.java.JavaCodeGenerator.toJavaPoetName(JavaCodeGenerator.kt:892)
        at org.pkl.codegen.java.JavaCodeGenerator.toJavaPoetName(JavaCodeGenerator.kt:848)
        at org.pkl.codegen.java.JavaCodeGenerator.toJavaPoetName$default(JavaCodeGenerator.kt:750)
        at org.pkl.codegen.java.JavaCodeGenerator.generateTypeSpec$addCtorParameter(JavaCodeGenerator.kt:250)
        at org.pkl.codegen.java.JavaCodeGenerator.generateTypeSpec$generateConstructor(JavaCodeGenerator.kt:286)
        at org.pkl.codegen.java.JavaCodeGenerator.getJavaFile(JavaCodeGenerator.kt:205)

  Kotlin:

  Exception in thread "main" java.lang.AssertionError: Encountered unexpected PType subclass: this
        at org.pkl.codegen.kotlin.KotlinCodeGenerator.toKotlinPoetName(KotlinCodeGenerator.kt:757)
        at org.pkl.codegen.kotlin.KotlinCodeGenerator.toKotlinPoetName(KotlinCodeGenerator.kt:707)
        at org.pkl.codegen.kotlin.KotlinCodeGenerator.generateTypeSpec$generateConstructor(KotlinCodeGenerator.kt:241)
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -460,55 +483,81 @@ protected boolean acceptTypeNode(boolean visitTypeArguments, TypeNodeConsumer co
...
+  /** The `module` or `this` type for an open module or class. */
+  public static final class NonFinalSelfTypeNode extends ObjectSlotTypeNode
       implements ClassTypeNode {
-    private final VmClass moduleClass; // only used by getVmClass()
-    @Child private ExpressionNode getModuleNode;
+    private final VmClass clazz; // only used by getVmClass()
+    @Child private ExpressionNode getTargetNode;
+    private final PType pType;
+    private final VmObjectFactory<Void> mirrorFactory;
+    @CompilationFinal private int realFrameSlot = -1;
```
stackoverflow: Not used.

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
HT154: This is resolved by Dan's suggestion here: https://github.com/apple/pkl/pull/1708#discussion_r3761805743

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
HT154: Removing this and instead grabbing the `this`/`module` classes in `ReferenceTypeNode` actually does not cause a regression (for non-deprecated usage of `module`). because the self types are now only allowed in the context matching the executing frame!

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
HT154: Given that `module` is not handled today, I'm hesitant to block this PR on support in the code generators, but happy to take that on as followup.

How should these be handled given that these languages don't have their own self types? We could erase them to the class type where the property is declared, but that doesn't quite match the intention in non-final contexts.

stackoverflow: I think it's fine to not handle them. But we should show a better error stating that we don't support self types for Java/Kotlin codegen. I know this was already the case with module types, but as we're adding more now it's a good time to improve the message.

HT154: Gotcha. Added a nicer error for both `module` and `this` to both generators.

pkl-codegen-java/src/main/kotlin/org/pkl/codegen/java/JavaCodeGenerator.kt
```diff
@@ -751,6 +751,11 @@ class JavaCodeGenerator(
     when (this) {
       PType.UNKNOWN -> OBJECT.nullableIf(nullable)
       PType.NOTHING -> TypeName.VOID
+      PType.MODULE,
+      PType.THIS ->
+        throw JavaCodeGeneratorException(
+          "Pkl `${this}` types are not supported by the Java code generator."
+        )
```
bioball, line 759: Java/Kotlin can both represent self types:

```java
class Foo<T extends Foo<T>> {
  public T getMyself();
}
```

I think it's okay to not support this for now, but we should leave a TODO comment to support this in the future maybe

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -369,7 +372,104 @@ public UnresolvedTypeNode visitNothingType(NothingType type) {
...
+    for (var scope = symbolTable.getCurrentScope(); scope != null; scope = scope.getParent()) {
+      if (scope.isObjectScope() || scope.isCustomThisScope()) {
+        break;
+      }
+      if (scope instanceof ClassScope foundClassScope) {
+        className = foundClassScope.getName();
+        break;
+      }
+      // it's still safe to break on ObjectScope because this is valid:
+      // typealias Foo = List(any((it) -> it == new Dynamic { it is this })) // this == Dynamic
```
bioball, line 451: I'm not sure if we even need the `isObjectScope`; you'll still hit `isCustomThisScope()` regardless (you can't declare an object literal inside a typealias body unless it's inside a type constraint).

Also: we're missing a test here.

I think our test can just be:

```pkl
local bar = List()

typealias Foo = List(bar is this)
```

bioball (changes requested):
I think this is getting close; did another pass.

One major comment here is: I think that we're missing a bunch of tests; there's a lot of positions in our grammar that can produce type nodes, which aren't tested.

Some of the ones that I can think of are:

* Object body param (`foo { bar: this -> /* etc */ }`)
* Supertype (`class Bar extends this {}`)
* Function literal type annotation (`(arg: this) -> ...`)
* For-generator variable
* Let expr variable
* Local object property `new { local x: this = ... }`
* Object method arguments (`new { local function foo(bar: this): this = ... }`)
* Type cast expressions (`foo as this`)
* Constraint property type (`foo: this(...)`)
* Function property type (`foo: this -> this`)
* `Listing<this>` (test lazy check and default value behavior)

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
HT154: We need the `isObjectScope` check still to break before we encounter a `ClassScope` so that `className` remains null and `getClassNode` is set to a `GetReceiverClassNode`. If we remove that check, we instead get a `ReadQualifiedLocalPropertyNode` or `ReadPropertyNode` that reads the containing class itself, which isn't correct here.

bioball: Gotcha, makes sense!

pkl-core/src/test/files/LanguageSnippetTests/input/types/thisType1.pkl
```diff
@@ -0,0 +1,201 @@
...
+
+res12 = test.catch(() -> res11.fun(module)[0].x)
+
+// this type not affected by custom this scope:
+res13: List(this is List<this>) = List(module)
+
+// test generic typealias through inheritance from module
+res14 = fun2(this)
+res14a = test.catch(() -> res11.fun2(this))
+res14b = res11.fun2(res11)
```
bioball: This test doesn't show up in the results

pkl-core/src/test/files/LanguageSnippetTests/input/types/thisType1.pkl
```diff
@@ -0,0 +1,201 @@
...
+        res18 = module.res18
+        res19 = module.res19
+        res22 = module.res22.toList().map((it) -> "\(it.getClass()): \(it.x)").toListing()
+        res22a = new Good {}.res22.toList().map((it) -> "\(it.getClass()): \(it.x)").toListing()
+        res23a = module.res23a
+        res23b = module.res23b
+        res24a = module.res24a
+        res24b = module.res24b
+      }
+}
```
bioball: I was having a really hard time matching up the expectations and the source test. While debugging things, I ended up rewriting this and I think this is a little better.

This also adds back some tests that were missing in the output value.

```pkl
open module thisType1

import "pkl:test"
import "pkl:reflect"

hidden x: Int = 1

hidden res1: this = module

res1a = res1.res2.x
res1b = res1.res3.x
res1c = res1.res5[1].x
res1d = res1.res6a.x

hidden res2: this = this
hidden res3: this = (this) { x = super.x + 1 }
hidden res4: this? = null
hidden res5: List<this> = List(this, (this) { x = super.x + 2 })
hidden res6: List<this> = fun(this)
hidden res6a: this = res6[0]
hidden res6b: this = fun(new Good { x = 3 })[0]

function fun(m: this): List<this> =
  let (_x = x)
    List((m) { x = super.x + _x })

typealias Foo<T> = T

function fun2(m: Foo<this>): List<this> =
  let (_x = x)
    List((m) { x = super.x + _x + 1 })

class Bad {
  res7: this = "abc"

  res8: this = new Person {}

  res9: this? = new Person {}

  res10: List<this> = List(new Person {})
}

class Person

res7 = test.catch(() -> new Bad {}.res7)
res8 = test.catch(() -> new Bad {}.res8)
res9 = test.catch(() -> new Bad {}.res9)
res10 = test.catch(() -> new Bad {}.res10)

class Good extends module {
  x = 2
}

local res11 = new Good {}

res11a = res11.res2.x
res11b = res11.res3.x
res11c = res11.res5[1].x
res11d = res11.res6a.x

res12 = test.catch(() -> res11.fun(module)[0].x)

// this type not affected by custom this scope:
hidden res13: List(this is List<this>) = List(module)
res13a = res13[0].x

// test generic typealias through inheritance from module
// hidden res14 = fun2(this)
res14a = test.catch(() -> res11.fun2(this))
res14b = res11.fun2(res11)[0].x

// test generic typealias without inheritance from module
open class A {
  x: Int = 1
  function fun2(m: Foo<this>): List<this> =
    let (_x = x)
      List((m) { x = super.x + _x + 1 })
}
class B extends A

local a = new A {}
local b = new B {}
res15 = a.fun2(a)
res15a = test.catch(() -> b.fun2(a))
res15b = b.fun2(b)

// test function type, function literal param, object body param
local c: (this) -> Dynamic = (xx: this) -> new Dynamic { x = xx.x * 2 }
local d = (c) { xx: this ->
  x = super.x + xx.x * 3
}
res16 = d.apply(this).x
res16a = test.catch(() -> d.apply(new Dynamic { x = 1 }).x)

// let expr binding type
res17 =
  let (xx: this = new module { x = 3 })
    xx.x

// for/when generator vars skip a level
res18: Listing<Int> = new {
  for (xx: this in List(module)) {
    xx.x
  }
}

// object property/method param/method return
res19 = new Dynamic {
  local foo: this = new Dynamic { x = 1 }
  local function bar(a: this): this = new { x = foo.x + a.x }
  baz = bar(new Dynamic { x = 2 })
}

// type cast, explicit new expr
hidden res20 = new this { x = 6 } as this
res20a = res20.x

// constrained
hidden res21: this(this.x > 5) = res20
res21a = res21.x

// lazy type check and default value
hidden res22: Listing<this> = new {
  new { x = 1 }
  new Good {}
  new { x = 3 }
  new Good { x = 4 }
}

res22a = module.res22.toList().map((it) -> "\(it.getClass()): \(it.x)").toListing()
res22b = new Good {}.res22.toList().map((it) -> "\(it.getClass()): \(it.x)").toListing()

// annotation type

open class ThisAnnotation extends Annotation {
  @this { y = 1 }
  y: Int
}

open class ThatAnnotation extends ThisAnnotation

@ThisAnnotation { y = 2 }
hidden annThis: Int = 1
res23 = reflect.Class(getClass()).properties["annThis"].annotations.single
res23a = res23.getClass().toString()
res23b = reflect.Class(res23.getClass()).allProperties["y"].annotations.single.getClass().toString()

@ThatAnnotation { y = 2 }
hidden annThat: Int = 1
res24 = reflect.Class(getClass()).properties["annThat"].annotations.single
res24a = res24.getClass().toString()
res24b = reflect.Class(res24.getClass()).allProperties["y"].annotations.single.getClass().toString()

```

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -370,7 +373,104 @@ public UnresolvedTypeNode visitNothingType(NothingType type) {
...
+          var parentScope = scope.getParent();
+          assert parentScope != null;
+          // if the parent also has "ALL", we haven't found the originating const property/method
+          // yet.
+          if (parentScope.getConstLevel() == ConstLevel.ALL) {
+            continue;
+          }
+          var message =
+              scope.isPropertyScope() ? "invalidModuleTypeInProperty" : "invalidModuleTypeInMethod";
+          errorMessage = ErrorMessages.create(message, scope.getQualifiedName());
```
bioball, line 419:
```suggestion
          errorMessage = ErrorMessages.create(message, scope.getQualifiedName());
          break;
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -461,54 +482,77 @@ protected boolean acceptTypeNode(boolean visitTypeArguments, TypeNodeConsumer co
...
     }
 
     @Override
     public boolean doIsEquivalentTo(TypeNode other) {
-      if (!(other instanceof NonFinalModuleTypeNode nonFinalModuleTypeNode)) {
+      if (!(other instanceof NonFinalSelfTypeNode nonFinalModuleTypeNode)) {
         return false;
       }
-      return moduleClass.equals(nonFinalModuleTypeNode.moduleClass);
+      return clazz.equals(nonFinalModuleTypeNode.clazz);
```
```suggestion
      if (!(other instanceof NonFinalSelfTypeNode nonFinalSelfTypeNode)) {
return false;
}
      return clazz.equals(nonFinalSelfTypeNode.clazz);
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -411,43 +414,61 @@ protected boolean acceptTypeNode(boolean visitTypeArguments, TypeNodeConsumer co
...
     }
 
     @Override
     public boolean doIsEquivalentTo(TypeNode other) {
-      if (!(other instanceof FinalModuleTypeNode finalModuleTypeNode)) {
+      if (!(other instanceof FinalSelfTypeNode finalModuleTypeNode)) {
         return false;
       }
-      return moduleClass.equals(finalModuleTypeNode.moduleClass);
+      return clazz.equals(finalModuleTypeNode.clazz);
```
```suggestion
      if (!(other instanceof FinalSelfTypeNode finalSelfTypeNode)) {
return false;
}
      return clazz.equals(finalSelfTypeNode.clazz);
```

bioball (approved):
Some minor nits and a test refactor, but otherwise LGTM!

Approving to unblock

pkl-core/src/test/files/LanguageSnippetTests/input/types/thisType1.pkl
HT154: Which one specifically? All of the `res14/a/b` properties are forced by the block of `let`s in the module output.

pkl-core/src/test/files/LanguageSnippetTests/input/types/thisType1.pkl
HT154: This has a different meaning, no? Things now marked `hidden` are not forced like the original implementation.

bioball: They're forced by the derived properties; e.g. 

```pkl
hidden res1: this = module

// forces `res1`
res1a = res1.res2.x
```

HT154: Some of them are, sure, but not all. Just on a cursory glance, `res4` and `res6b` never get called. But point taken, I'll iterate on this to make it easier to work with.

stackoverflow (approved):

PR #1709, opened 2026-06-27
Fix type argument lost in constraint expressions within generic typea…

### LINKED ISSUE #1705: Type argument is lost in constraint expressions within generic typealias
This snippet:

```pkl
typealias MyList<T> = List(every((it) -> it is T))

foo: MyList<Int> = List("uhoh")
```

Produces:

```pkl
foo = List("uhoh")
```

The `it is T` should become `it is Int`, that check isn't happening

adityabagchi24:
# Description

When using a generic type alias with a constraint such as `every((it) -> it is T)`, the type argument (for example, `Int`) was not being substituted into the constraint expression. As a result, the `it is T` check silently accepted any value instead of performing an actual type check.

Fixes #1705

## Before

```text
typealias MyList<T> = List(every((it) -> it is T))
foo: MyList<Int> = List("uhoh")  // accepted — incorrect
```

## After

```text
typealias MyList<T> = List(every((it) -> it is T))
foo: MyList<Int> = List("uhoh")  // rejected — correct
```

## Root Cause

1. The lambda body inside constraint expressions (`it is T`) was stored without Truffle's `@Child` annotation. Consequently, the tree-walking logic responsible for substituting type parameters could not traverse into the lambda body.
2. The instantiation logic only replaced resolved type variable nodes. However, inside constraint expressions, the type variable still exists in its unresolved form when substitution occurs.

## Fix

* Marked the lambda body expression as `@Child` so it can be properly traversed and deep-copied during type alias instantiation.
* Extended the instantiation logic to also handle unresolved type variables and replace them with concrete type arguments.
* Added a lightweight wrapper type that carries a pre-resolved type node into constraint expressions.

## Validation

* `List("uhoh")` used as `MyList<Int>` now correctly produces a type constraint violation error.
* `List(1, 2, 3)` used as `MyList<Int>` continues to pass.
* Non-generic constrained type aliases (for example, `typealias MyList = List(every((it) -> it is Int))`) continue to work correctly.
* All existing `LanguageSnippetTests`, `pkl-codegen-java`, `pkl-codegen-kotlin`, and `pkl-cli` tests pass.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/ast/type/UnresolvedTypeNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/type/UnresolvedTypeNode.java
@@ -457,8 +457,6 @@ public abstract class UnresolvedTypeNode extends PklNode {
 
     @Override
     public TypeNode execute(VirtualFrame frame) {
-      CompilerDirectives.transferToInterpreter();
-
       return typeNode;
     }
   }
--- a/pkl-core/src/main/java/org/pkl/core/ast/type/UnresolvedTypeNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/type/UnresolvedTypeNode.java
@@ -448,7 +448,7 @@ public abstract class UnresolvedTypeNode extends PklNode {
    * with the corresponding concrete type argument.
    */
   public static final class Resolved extends UnresolvedTypeNode {
-    private final TypeNode typeNode;
+    @Child private TypeNode typeNode;
 
     public Resolved(SourceSection sourceSection, TypeNode typeNode) {
       super(sourceSection);
--- a/pkl-core/src/main/java/org/pkl/core/ast/type/UnresolvedTypeNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/type/UnresolvedTypeNode.java
@@ -442,10 +442,9 @@ public abstract class UnresolvedTypeNode extends PklNode {
   }
 
   /**
-   * An unresolved type node that is pre-resolved to a concrete type node.
-   * Used during type alias instantiation to replace type variable references
-   * inside constraint expressions (e.g., {@code every((it) -> it is T)})
-   * with the corresponding concrete type argument.
+   * An unresolved type node that is pre-resolved to a concrete type node. Used during type alias
+   * instantiation to replace type variable references inside constraint expressions (e.g., {@code
+   * every((it) -> it is T)}) with the corresponding concrete type argument.
    */
   public static final class Resolved extends UnresolvedTypeNode {
     @Child private TypeNode typeNode;
--- a/pkl-core/src/main/java/org/pkl/core/ast/type/UnresolvedTypeNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/type/UnresolvedTypeNode.java
@@ -429,11 +429,37 @@ public abstract class UnresolvedTypeNode extends PklNode {
       this.typeParameter = typeParameter;
     }
 
+    public int getTypeParameterIndex() {
+      return typeParameter.getIndex();
+    }
+
     @Override
     public TypeNode execute(VirtualFrame frame) {
       CompilerDirectives.transferToInterpreter();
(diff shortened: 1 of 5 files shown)


pkl-core/src/main/java/org/pkl/core/runtime/VmTypeAlias.java
```diff
@@ -198,6 +199,15 @@ public TypeNode instantiate(
                 typeArgumentNodes.length == 0
                     ? new UnknownTypeNode(sourceSection)
                     : typeArgumentNodes[index]);
+          } else if (node instanceof UnresolvedTypeNode.TypeVariable unresolvedTypeVar) {
+            // Type variables inside constraint expressions (e.g. `every((it) -> it is T)`)
+            // are still unresolved at instantiation time. Replace them with a resolved
+            // unresolved type node that returns the concrete type argument.
+            int index = unresolvedTypeVar.getTypeParameterIndex();
```
bioball:
```suggestion
            var index = unresolvedTypeVar.getTypeParameterIndex();
```

pkl-core/src/main/java/org/pkl/core/ast/type/UnresolvedTypeNode.java
```diff
@@ -429,11 +429,37 @@ public TypeVariable(SourceSection sourceSection, TypeParameter typeParameter) {
...
   }
+
+  /**
+   * An unresolved type node that is pre-resolved to a concrete type node.
+   * Used during type alias instantiation to replace type variable references
+   * inside constraint expressions (e.g., {@code every((it) -> it is T)})
+   * with the corresponding concrete type argument.
+   */
+  public static final class Resolved extends UnresolvedTypeNode {
+    private final TypeNode typeNode;
```
bioball:
```suggestion
    @Child private TypeNode typeNode;
```

pkl-core/src/main/java/org/pkl/core/ast/type/UnresolvedTypeNode.java
```diff
@@ -429,11 +429,37 @@ public TypeVariable(SourceSection sourceSection, TypeParameter typeParameter) {
...
+    public Resolved(SourceSection sourceSection, TypeNode typeNode) {
+      super(sourceSection);
+      this.typeNode = typeNode;
+    }
+
+    @Override
+    public TypeNode execute(VirtualFrame frame) {
+      CompilerDirectives.transferToInterpreter();
+
+      return typeNode;
```
bioball:
```suggestion
      return typeNode;
```

bioball (changes requested):
Thanks! In general, this fix looks good.

We need some tests here; add this:

```pkl
typealias MyList<T> = List(this[0] is T)

foo: MyList<Int> = List("uh oh")
```

Create file at path pkl-core/src/test/files/LanguageSnippetTests/input/types/typeAliasConstraint3.pkl

And then run class `org.pkl.core.LanguageSnippetTests` to fill in the expected result.
You can either navigate to that class in IntelliJ and click the run button in the gutter, or you can run this command:

```
./gradlew pkl-core:test --tests 'org.pkl.core.LanguageSnippetTests'
```

adityabagchi24:
@bioball Done! Created the test at `pkl-core/src/test/files/LanguageSnippetTests/input/types/typeAliasConstraint3.pkl` and ran the snippet tests to generate the expected `.err` output. Also rebased on latest `upstream/main` to resolve the merge conflict.

bioball (approved):
LGTM; looking for thumbs from @stackoverflow before merging

adityabagchi24:
@bioball There was some formatting error in the file `UnresolvedTypeNode.java`. I think, I have fixed it now. Can you please re run the workflow?

stackoverflow (approved):
Looks good. Thanks for the contribution.

adityabagchi24:
@bioball Thank you so much for your guidance, thorough reviews, and for merging my PR. I really appreciate your time and support throughout the process. Also, thank you @stackoverflow for reviewing my changes.

PR #1713, opened 2026-06-28
Correctly handle `module` types as typealias parameters

### LINKED ISSUE #1712: `module` types as typelias type args are evaluated in the context of the alias's module, not where `module` is used

HT154:
Really not happy with this solution, but I'm not sure how else to achieve this!!

Depends on #1717

Resolves #1712


HT154:
Preferable to disallow `module` types in typealias and class bodies (with a deprecation cycle) and #1708 should disallow `this` types in typealias bodies.

PR #1717, opened 2026-06-29
Invalidate cached noop typechecks during typealias instantiation

### LINKED ISSUE #1710: Incorrect `List`/`Set`/`Map`/`Listing`/`Mapping`/union type check behavior through typealiases
Correct behavior:
```pkl

import "pkl:test"

x: Int = 1

function fun(val: List<module>): Int = val.first.x

res1 = fun(List(this))
res4 = test.catch(() -> fun(List(new Dynamic { x = "nope" })))
```
Result:
```pkl
x = 1
res1 = 1
res4 = "Expected value of type `currentModuleType4`, but got type `Dynamic`. Value: new Dynamic { x = ? }"
```

Incorrect behavior (`List` type annotation behind alias):
```pkl
x = 1
res3 = 1
res6 = "Expected value of type `Int`, but got type `String`. Value: \"nope\""
```

This happens because the `ListTypeNode`'s original `elementTypeNode` is a `TypeVariableTypeNode`, which unconditionally returns `true` for `isNoopTypeCheck()`. But when the alias is instantiated, this is replaced by the `NonFinalModuleTypeNode` (which is not a noop) without resetting `skipElementTypeChecks`.

This speaks to a need for better handling of internal cached (non-child) `TypeNode` state, which #1698 takes steps to try and address.

This affects all type nodes that attempts to optimize by detect noop checks at init time.

### LINKED ISSUE #1716: Type checks through aliases are always run lazily even when they should be eager
This affects `Listing` and `Mapping` similarly:
```pkl
typealias MyListing<T> = Listing<T>
res4 = new Listing { new Dynamic {} } is MyListing<module>
```
Result:
```pkl
res4 = true // should be false
```

HT154:
This also fixes `TypeAliasTypeNode` always performing lazy type checks even when they should be eager.

Resolves #1710
Resolves #1716

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
@@ -70,16 +70,6 @@ public abstract class TypeNode extends PklNode {
     super(sourceSection);
   }
 
-  /**
-   * Invalidate any cached state and recalculate it.
-   *
-   * <p>Subclasses should override this and call invalidate on each child before invalidating its
-   * own state.
-   */
-  protected void invalidate() {
-    // noop by default
-  }
-
   public boolean isNoopTypeCheck() {
     return false;
   }
@@ -821,11 +811,6 @@ public abstract class TypeNode extends PklNode {
     }
 
     @Override
-    protected final void invalidate() {
-      elementTypeNode.invalidate();
-    }
-
-    @Override
     protected final PType doExport() {
       return new PType.Nullable(elementTypeNode.doExport());
     }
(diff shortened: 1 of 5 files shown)


GitHubismae (approved):

stackoverflow (commented):
Nested unions still suffer from the same issue:
```pkl
typealias Flat<A, B> = A | B
res_flat = new Dynamic {} is Flat<module, module>

typealias Nested<A, B> = (A | A) | (B | B)
res_nested = new Dynamic {} is Nested<module, module>
```
Results in
```
res_flat = false
res_nested = true
```

Children should be invalidated before parents. The truffle visitor does the opposite (goes from parent to children).

bioball (changes requested):
I don't think we need an `invalidate()` method. Instead, we can just call `if (elementTypeNode.isNoopTypeCheck())` inside the `execute` methods of these nodes.

`invalidate()` suggests that a type node has stale state. But, the core problem is that we don't know yet if this is a noop check yet when inside the constructor of `ListTypeNode`, `SetTypeNode`, etc.

We can probably also change the `isNoopTypeCheck` method in class `TypeVariableNode` to this:

```java
    @Override
    public boolean isNoopTypeCheck() {
      throw exceptionBuilder().bug("isNoopTypeCheck called on an unreplaced type variable").build();
    }
```

HT154:
> I don't think we need an `invalidate()` method. Instead, we can just call if (`elementTypeNode.isNoopTypeCheck()`) inside the execute methods of these nodes.

Are you suggesting removing the `skip_TypeChecks` fields from each class?

> We can probably also change the `isNoopTypeCheck` method in class `TypeVariableNode` to this:
> 
> ```java
>     @Override
>     public boolean isNoopTypeCheck() {
>       CompilerDirectives.transferToInterpreter();
>       throw exceptionBuilder().bug("isNoopTypeCheck called on an unreplaced type variable").build();
>     }
> ```

This particular change isn't safe since `TypeVariableNode` is not replaced in generic classes/methods.

bioball:
> This particular change isn't safe since TypeVariableNode is not replaced in generic classes/methods.

Ah, gotcha. Then, yeah, we should keep that impl as-is for now (it basically stands for `Unknown`).

>Are you suggesting removing the skip_TypeChecks fields from each class?

Yup! Or, possibly, defer setting that field until `execute` gets called. But, it seems fine to just remove.

stackoverflow (approved):
Looks good!

HT154:
An alternate approach here is having `TypeVariableNode.isNoopTypeCheck()` return `false` (only when it's in a type alias body?). This would mean slightly less efficient type checks when type vars are replaced with no-ops, but would preserve the optimization in other cases (like when no typealiases are used). Thoughts?

HT154:
> An alternate approach here is having `TypeVariableNode.isNoopTypeCheck()` return `false` only when it's in a type alias body.

#1722 implements this approach. I think this is probably preferable to this PR as it avoids losing the caching optimization in non-type-parameter scenarios.

bioball (approved):
LGTM! Merging this one; I'm not totally sold that https://github.com/apple/pkl/pull/1722 semantically makes sense (`TypeVariableNode` is always meant to be replaced, I think)

PR #1718, opened 2026-06-29
Enforce stricter rules on reference access

bioball:
This changes the following:

1. If given a union type, the member must exist on each member of that type
2. If accessing a member off `Reference<D, Null>`, give a `Reference<D, Null>`

Also:
* Improve error messages thrown during member access.
* Add test around accessing members off of a function type


pkl-core/src/main/java/org/pkl/core/ast/expression/binary/SubscriptNode.java
```diff
@@ -100,18 +104,39 @@ protected Object eval(
...
+                  getShortName(),
+                  reference.exportType(),
+                  VmUtils.getClass(key))
+              .withProgramValue("Left operand", reference)
+              .withProgramValue("Right operand", key);
+      var hint = getReferenceHint(reference, err, key);
+      if (hint != null) {
+        e.withHint(hint);
+      }
+      throw e.build();
```
stackoverflow:
```suggestion
              .withProgramValue("Right operand", key)
              .withHint(getReferenceHint(reference, err, key));
      throw e.build();
```
`withHint` already checks for `null`.

stackoverflow (approved):
Looks good.

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -44,6 +44,7 @@ public final class VmReference extends VmValue {
   // candidate types can only be: PType.Class, PType.Alias (only preservedAliasTypes),
   // PType.StringLiteral, PType.UNKNOWN, or PType.Union (containing only the previous; flattened)
   private final PType referentType;
+  private static final PType nullType = new PType.Class(BaseModule.getNullClass().export());
```
HT154: I think I had native compilation issues when I did this statically previously, so double check that those builds pass!

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -148,6 +152,11 @@ private static void normalizeTypes(PType type, PClass moduleClass, Set<PType> re
       }
     } else if (type == PType.MODULE) {
       result.add(new PType.Class(moduleClass));
+    } else {
+      // remaining types: PType.Function, PType.TypeVariable. no normalizing needed; TypeVariable
+      // gets replaced upon instantiation, and Function can bubble up to users as a reference error
+      // if accessed.
+      result.add(type);
```
HT154, line 159: Comment on L44 needs an update to reflect this change

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -189,33 +195,54 @@ private static void getCandidatePropertyType(PType type, String property, Set<PT
       return;
     }
     // restriction: only class types can have their properties referenced
-    if (!(type instanceof PType.Class clazz)) return;
-    // restriction: cannot reference properties of external classes
-    if (clazz.getPClass().isExternal()) return;
```
HT154, line 194: Was dropping the external class check intentional?

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -1183,3 +1203,6 @@ Redirected to: `{1}`
 
 invalidReferenceTypeAnnotationWithConstraint=\
 `Reference` referent type argument may not include type constraints.
+
+invalidReferenceTypeAnnotationWithNullableType=\
+`Reference` referent type argument may not include nullable types.
```
HT154: Pretty sure this is unused now

HT154 (changes requested):

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
bioball: Yeah, because this check is already covered by L272 (cannot reference properties defined in external classes)?

I was trying to figure out what this check is covering; removing this line doesn't introduce any test failures.

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
bioball: Ah, yeah I see! I'll just inline this, then.

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
bioball: This block would also unintentionally catch referencing members off `Reference<D, Null>` (because `Null` is an external class)

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -490,4 +526,29 @@ public String toString() {
...
   }
+
+  public enum VmReferenceAccessErrorType {
+    CANNOT_FIND_MEMBER,
+    EXTERNAL_MEMBER,
+    DEFAULT_MEMBER,
+    EXTERNAL_CLASS
+  }
+
+  public static final class VmReferenceAccessError extends RuntimeException {
```
HT154, line 537: Does this need to subclass `RuntimeException`? Seems like it might actually be useful for this to be a checked exception instead.

HT154 (approved):
LGTM! Just one non-blocker

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
bioball: I might agree if this was a generic API where we didn't know who the callers were, but this will only ever need to be caught by `ReadPropertyNode` and `SubscriptNode`, so it didn't seem too meaningful to turn into a checked exception. We'd also have to add a `throws` clause to multiple methods here if so.

PR #1721, opened 2026-07-01
Refine handling of constraints in typealiases

bioball:
Motivation: clean up handling of unresolved type nodes.

* Introduce `TypeExpressionNode`, and add `insertTypeNode` method
* Remove `UnresolvedTypeNode.Resolved`; skip unnecessary `execute()` and `CompilerDirectives.transferToInterprefer()` calls when executing TestTestNode/TypeCastNode
* Use `insertTypeNode` to inject type vars into expressions

Also, fixes an issue where nested constraints still aren't replaced.

This is a follow-up fix for #1709


stackoverflow (commented):
There's a regression in your fix:
```pkl
typealias A<T> = List((this[0] is T(true)))
fail1 = List("uh oh")
fail11 = foo as A<Int>

typealias B<T> = List(this[0] is T?)
fail2 = List("uh oh") as B<Int>
```

Both of these succeed instead of failing, while they failed before.

bioball:
Ah, good catch! Yeah, this isn't a viable solution then.

I don't think we can avoid `UnresolvedTypeNode.Resolved`

PR #1722, opened 2026-07-01
Do not treat replaceable `TypeVariableNode` instances as noops

### LINKED ISSUE #1710: Incorrect `List`/`Set`/`Map`/`Listing`/`Mapping`/union type check behavior through typealiases

### LINKED ISSUE #1716: Type checks through aliases are always run lazily even when they should be eager

HT154:
This also fixes `TypeAliasTypeNode` always performing lazy type checks even when they should be eager.

Resolves #1710
Resolves #1716
Supersedes #1717

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
@@ -2395,11 +2395,13 @@ public abstract class TypeNode extends PklNode {
 
   public static final class TypeVariableNode extends WriteFrameSlotTypeNode {
     private final TypeParameter typeParameter;
+    private final boolean isInTypeAlias;
 
-    public TypeVariableNode(SourceSection sourceSection, TypeParameter typeParameter) {
-
+    public TypeVariableNode(
+        SourceSection sourceSection, TypeParameter typeParameter, boolean isInTypeAlias) {
       super(sourceSection);
       this.typeParameter = typeParameter;
+      this.isInTypeAlias = isInTypeAlias;
     }
 
     public int getTypeParameterIndex() {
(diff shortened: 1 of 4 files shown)


pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2408,7 +2410,10 @@ public int getTypeParameterIndex() {
 
     @Override
     public boolean isNoopTypeCheck() {
-      return true;
+      // if in a type alias, this node will be replaced by another TypeNode that may not be a noop
+      // so we return false so that containing TypeNodes don't skip checks erroneously
+      // this is slightly less efficient if the replacement _is_ a noop, but only inside typealiases
+      return !isInTypeAlias;
```
bioball, line 2416: We should never hit the case where `isNoopTypeCheck()` is called when this is inside a typealias, because these nodes should be replaced before the alias can ever run.

Also: I feel that we should replace the usage of `TypeVariableNode` in other positions (e.g. `function <T>(foo: T)`) with `TypeNode.Unknown`, at least until we have support for type arguments there. Probably would be better if `TypeVariableNode` never gets fired.

PR #1749, opened 2026-07-09
Close intermediate redirect response bodies before following

KushalP:
When following redirects, `RequestRewritingClient` discarded each 3xx response without closing its body, leaking the underlying connection.

Close the body as soon as we know the status is a redirect and won't be returned, so the too-many-redirects, missing-`Location`, invalid-URI and downgrade error paths all release the connection too.

This prevents leaking one `BodyHandlers.ofInputStream()` connection per hop. Now the code closes the body before following the redirect.

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
+++ b/pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
@@ -16,6 +16,7 @@
 package org.pkl.core.http;
 
 import com.google.errorprone.annotations.ThreadSafe;
+import java.io.Closeable;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -117,6 +118,9 @@ final class RequestRewritingClient implements HttpClient {
       var response = delegate.send(currentRequest, responseBodyHandler, httpRequestChecker);
       if (!HttpUtils.isRedirectStatusCode(response.statusCode())) {
         return response;
+      }
+      if (response.body() instanceof Closeable closeable) {
+        closeable.close();
       }
       if (redirectCount >= MAX_HTTP_REDIRECTS) {
         throw new HttpClientException(
--- a/pkl-core/src/test/kotlin/org/pkl/core/http/RequestRewritingClientTest.kt
+++ b/pkl-core/src/test/kotlin/org/pkl/core/http/RequestRewritingClientTest.kt
@@ -15,6 +15,8 @@
  */
 package org.pkl.core.http
 
+import java.io.ByteArrayInputStream
+import java.io.InputStream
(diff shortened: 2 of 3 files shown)


pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
```diff
@@ -134,12 +135,16 @@ private <T> HttpResponse<T> doSend(
...
-          && redirectUri.getScheme().equalsIgnoreCase("http")) {
+          && nextRequestUri.getScheme().equalsIgnoreCase("http")) {
         throw new HttpClientException(
-            ErrorMessages.create("httpRedirectCannotDowngrade", currentRequestUri, redirectUri));
+            ErrorMessages.create("httpRedirectCannotDowngrade", currentRequestUri, nextRequestUri));
       }
-      currentRequestUri = rewriteUri(redirectUri);
+      if (response.body() instanceof Closeable closeable) {
+        closeable.close();
+      }
```
bioball: Move this to line 122; otherwise the throws here still won't close the response body.

pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
```diff
@@ -134,12 +135,16 @@ private <T> HttpResponse<T> doSend(
         throw new HttpClientException(
             ErrorMessages.create("httpRedirectInvalidUri", currentRequestUri, location.get()));
       }
+      var nextRequestUri = rewriteUri(redirectUri);
       if (currentRequestUri.getScheme().equalsIgnoreCase("https")
-          && redirectUri.getScheme().equalsIgnoreCase("http")) {
+          && nextRequestUri.getScheme().equalsIgnoreCase("http")) {
```
bioball: The current behavior is actually is intentional; if a user _rewrites_ a request URI, they're on the hook for the security implications there.

You can already downgrade when not in the context of redirect following; just because we're following redirects here shouldn't change that.

bioball (changes requested):
Thanks for the PR! Some comments below.

pkl-core/src/main/java/org/pkl/core/http/RequestRewritingClient.java
HT154: Probably prudent to stick this in a `finally` block instead

bioball: Should be okay to just move to line 122; once we know we're dealing with a redirect status code, we can already close the input stream because we know we're not going to return it.

HT154 (approved):

bioball (approved):
LGTM, thank you!

PR #1760, opened 2026-07-10
Fix `eval -x` of local members

### LINKED ISSUE #1759: `eval -x` cannot find local properties
New regression in 0.32; `eval -x` cannot find local properties.

Given code:

```pkl
local foo = 1
```

`pkl eval -x foo` gives:

```
–– Pkl Error ––
Cannot find property `foo`.
```

bioball:
This fixes a regression where local members cannot be seen by the expression evaluator.

Because variable are now parse-time resolved, this builds a truffle node by first constructing a synthetic module of all defined local members, and calls AstBuilder on this synthetic module first.

This moves some existing logic from ReplServer over to VmUtils.

Closes #1759


HT154 (approved):
LGTM

PR #1763, opened 2026-07-12
Preserve Class<T> type arguments

### LINKED ISSUE #1729: `Class<T>` is always type-erased to `Class`

RushikeshGandhmal:
Fixes #1729

Previously, `Class<T>` was explicitly erased to `Class` during type resolution, so `C is Class<A>` could not be distinguished from `C is Class<B>`.

This change introduces a `ClassValueTypeNode` that:
- preserves the type argument through type resolution and runtime checks
- validates class values by requiring the represented class to be a subclass of the type argument
- rejects invalid concrete type arguments such as unions, nullables, constrained types, parameterized types, and modules
- still allows type variables like `Class<Type>` used in stdlib generic signatures

A regression test is added covering:
- positive cases (`C is Class<A>` vs `Class<B>`)
- aliases to class types and alias-preserving generics
- exported schema preservation
- stdlib compatibility with type variables
- rejection of invalid `Class<T>` arguments

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
+++ b/pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
@@ -777,6 +777,102 @@ public abstract class TypeNode extends PklNode {
     }
   }
 
+  public static final class ClassValueTypeNode extends ValidatingObjectSlotTypeNode {
+    @Child private TypeNode typeArgumentNode;
+
+    public ClassValueTypeNode(SourceSection sourceSection, TypeNode typeArgumentNode) {
+      super(sourceSection);
+      this.typeArgumentNode = typeArgumentNode;
+      validate();
+    }
+
+    @Override
+    protected Object executeLazily(VirtualFrame frame, Object value) {
+      var typeArgumentClass = getTypeArgumentClass();
+      if (value instanceof VmClass vmClass
+          && (typeArgumentClass == null || typeArgumentClass.isSuperclassOf(vmClass))) {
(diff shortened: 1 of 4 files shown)


RushikeshGandhmal:
Hi @bioball -- this is a more substantial type-system fix for #1729. I kept the change focused on Class<T> resolution and used the existing ValidatingObjectSlotTypeNode pattern for validation. The regression test covers both the new semantics and the stdlib compatibility case. Thanks for taking a look!

HT154:
This appears to be a duplicate of #1698. If you have input on this change, I would encourage you to discuss there so we can avoid splitting the discussion.

RushikeshGandhmal:
Thanks for pointing this out. I missed #1698, and after comparing the implementations I agree that #1763 overlaps it. I’ll close this PR and move the additional schema-export observation and regression test to #1698 so the discussion stays in one place.

PR #1766, opened 2026-07-12
codegen-java: Support generating nullable annotations

### LINKED ISSUE #811: codegen-java: Support generating @Nullable annotations
Motivation:
codegen-java already supports generating `@NonNull` annotations.
However, many projects prefer `@Nullable` over `@NonNull` annotations.
JSpecify also recommends to use `@Nullable`.

Changes:
- Add a `--nullable-annotation` CLI parameter that defaults to `none`.
- Add a `nullableAnnotation` Gradle/CliJavaCodeGenerator/JavaCodeGenerator property that defaults to `null`.
- Annotate nullable types with the configured annotation (if any).

As part of this feature, I propose to also make the following changes:
- Change the `-non-null-annotation` CLI parameter and `nonNullAnnotation` properties accordingly.
- Do not generate `@NonNull` annotations by default.
- Deprecate `org.pkl.java.config.mapper.NonNull` and encourage users to switch to JSpecify.
   Note that tools won't recognize `org.pkl.java.config.mapper.NonNull` unless explicitly configured.

Open question:
`@Nullable` must be accompanied by `@NullMarked` (JSpecify) or `@NonnullByDefault` (JSR 305).
I can think of two ways to go about this:
1. Introduce `--null-marked-annotation` and annotate config classes accordingly.
    This is the most consistent solution, but it requires yet another CLI parameter and property.
    It can be made more convenient by inferring the correct annotation based on `--nullable-annotation` and `--non-null-annotation`
    for popular libraries such as JSpecify and JSR 305.
2. Leave it to users to add a `package-info.java` with a `@NullMarked` annotation.
    This is a clean and simple solution. However, it is less convenient than (1).
    It could, in theory, cause problems if users compile generated and handwritten code separately and don't have a place where to put `package-info.java`.

Final thought: The Java ecosystem is converging on JSpecify. For many users, the best option would be `--generate-jspecify-annotations`.

RushikeshGandhmal:
Fixes #811

Adds an optional `nullableAnnotation` setting across the Java codegen API, CLI, and Gradle plugin. When configured, nullable Pkl types receive the specified TYPE_USE annotation across fields, getters, constructor parameters, `withX()` methods, nested generic arguments, boxed primitives, and byte arrays. The default remains no nullable annotation, preserving existing generated output.

Tests cover direct API generation, the actual `--nullable-annotation` CLI flag, Gradle DSL/task propagation, configuration cache, and compilation of generated JSpecify-annotated Java.

### DIFF
--- a/docs/modules/java-binding/pages/codegen.adoc
+++ b/docs/modules/java-binding/pages/codegen.adoc
@@ -204,6 +204,16 @@ The specified annotation type must be annotated with `@java.lang.annotation.Targ
 or the generated code may not compile.
 ====
 
+.--nullable-annotation
+[%collapsible]
+====
+Default: `none` +
+Fully qualified name of the annotation type to use for annotating nullable types. +
+The specified annotation type must be annotated with `@java.lang.annotation.Target(ElementType.TYPE_USE)`
+or the generated code may not compile.
+If unset, nullable types are not annotated.
+====
+
 Common code generator options:
 
 include::{partialsdir}/cli-codegen-options.adoc[]
--- a/docs/modules/pkl-gradle/pages/index.adoc
+++ b/docs/modules/pkl-gradle/pages/index.adoc
@@ -430,6 +430,17 @@ The specified annotation type must be annotated with `@java.lang.annotation.Targ
 or the generated code may not compile.
 ====
 
+.nullableAnnotation: Property<String>
+[%collapsible]
+====
+Default: `null` +
+Example: `nullableAnnotation = "org.project.MyAnnotation"` +
+Fully qualified name of the annotation type to use for annotating nullable types. +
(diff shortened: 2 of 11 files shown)


RushikeshGandhmal:
Hi @bioball -- this PR implements the nullable annotation support discussed in #811. I kept the change additive (no default behavior change) and covered the API, CLI, Gradle DSL, and compilation of generated JSpecify-annotated code. Thanks for taking a look!

bioball:
Hi @RushikeshGandhmal, we definitely value contributions, but we expect them to be carefully reviewed by the author. This means that we cannot take on the review and maintenance burden of code that is only generated by AI. Your pull requests appear to be AI authored (please correct us if we made a mistake). We are going to close your PRs for now, and invite you to write your first pull request by hand before relying on AI.

PR #1768, opened 2026-07-13
Add inference for implicitly typed `new` in method arguments

HT154:


pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -951,6 +952,71 @@ private ExpressionNode doVisitNewExprWithInferredParent(NewExpr expr) {
...
+      if (access instanceof QualifiedAccessExpr qualifiedAccessExpr) {
+        var methodName = qualifiedAccessExpr.getIdentifier().getValue();
+        var methodIdentifier = org.pkl.core.runtime.Identifier.get(methodName);
+        inferredParentNode =
+            new InferParentWithinMethodArgumentNode.Virtual(
+                sourceSection,
+                language,
+                argIndex,
+                methodIdentifier,
+                visitExpr(qualifiedAccessExpr.getExpr()));
```
HT154: In qualified access, `expr` (in `expr.method(arg)`) is re-parsed and re-evaluated for each arg containing an implicitly typed `new`. I wasn't able to determine how to reuse the receiver node from the parent `InvokeMethodVirtualNode`.

bioball: How about our good ol' friend, the auxiliary frame slot? Where, the eval methods in `InvokeMethodVirtualNode` sets the receiver in an auxiliary frame slot, and it gets read inside `InferParentWithinMethodArgumentNode.Virtual`

We might be able to do this every time we call a method (including direct calls, super calls), which would mean we don't even need to do any variable resolution here?

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinMethodArgumentNode.java
```diff
@@ -0,0 +1,183 @@
...
+
+    protected org.pkl.core.ast.member.ClassMethod getClassMethod(VirtualFrame frame) {
+      var receiverClass = receiverClassNode.executeGeneric(frame);
+      var method = ((VmClass) receiverClass).getMethod(methodName);
+      assert method != null;
+      return method;
+    }
+  }
+
+  public static final class Super extends ClassMethod {
```
bioball: Style-wise, we typically prefer standalone classes over nested static classes. E.g. `InferParentWithinSuperMethodArgumentNode`

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2854,25 +2854,22 @@ public TypeNode initWriteSlotNode(int slot) {
...
-        assert numberOfAuxiliarySlots == 1;
-        customThisSlot = 0;
-      }
       var ret = childNode.executeLazily(frame, value);
 
       var localContext = language.localContext.get();
       var prevShouldTypeCheck = localContext.shouldEagerTypecheck();
       localContext.shouldEagerTypecheck(true);
-      frame.setAuxiliarySlot(customThisSlot, value);
+      frame.setAuxiliarySlot(getCustomThisSlot(frame), value);
```
HT154: This change unwinds the commented assumption about custom `this` being the only use for aux slots.

pkl-core/src/main/java/org/pkl/core/ast/expression/member/AbstractInvokeMethodNode.java
```diff
@@ -15,79 +15,29 @@
...
-  @Children private ExpressionNode[] argumentNodes;
-  @Child private @Nullable DirectCallNode callNode;
-  @CompilationFinal protected boolean isConstChecked;
+  public static final Object METHOD_FRAME_SLOT_ID =
+      new Object() {
+        @Override
+        public String toString() {
+          return "method";
+        }
+      };
```
bioball: For code organization sake: move this into `VmUtils`, and also put `org.pkl.core.ast.builder.SymbolTable.CustomThisScope#FRAME_SLOT_ID` into there too

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinMethodArgumentNode.java
```diff
@@ -0,0 +1,75 @@
...
+    var parentTypeNode = method.getParameterTypeNode(argIndex);
+    var parameterTypeDefaultValue =
+        parentTypeNode.createDefaultValue(
+            frame, language, method.getHeaderSection(), method.getQualifiedName());
+    if (parameterTypeDefaultValue != null) {
+      inferredParent = parameterTypeDefaultValue;
+      return inferredParent;
+    }
+
+    throw exceptionBuilder().evalError("cannotInferParent").build();
```
bioball: [nit]

```suggestion
    if (parameterTypeDefaultValue == null) {
      throw exceptionBuilder().evalError("cannotInferParent").build();
    }
    inferredParent = parameterTypeDefaultValue;
    return inferredParent;
```

pkl-core/src/main/java/org/pkl/core/ast/expression/member/AbstractInvokeMethodNode.java
```diff
@@ -15,79 +15,29 @@
...
-      callNode = DirectCallNode.create(getCallTarget(owner));
-      insert(callNode);
-    }
-    assert callNode != null;
-    return callNode;
+  protected int getMethodSlot(VirtualFrame frame) {
+    // can't store the slot id as this node may be called from different root nodes
+    // (see constraints14 snippet)
+    CompilerDirectives.transferToInterpreter();
+    return frame.getFrameDescriptor().findOrAddAuxiliarySlot(METHOD_FRAME_SLOT_ID);
```
bioball: Calling `transferToInterpreter()` will de-optimize and will prevent Truffle from compiling basically all method calls. We only use it before throwing exceptions.

We should be using truffle boundaries instead; and note that you can't pass a VirtualFrame into a truffle boundary so you need to pass the frame descriptor instead (for reference, see `TypeNode#getCustomThisSlot`).

```suggestion
  @TruffleBoundary
  protected int getMethodSlot(FrameDescriptor frameDescriptor) {
    // can't store the slot id as this node may be called from different root nodes
    // (see constraints14 snippet)
    return frameDescriptor.findOrAddAuxiliarySlot(METHOD_FRAME_SLOT_ID);
```

pkl-core/src/main/java/org/pkl/core/ast/expression/member/AbstractInvokeLexicalOrQualifiedMethodNode.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  @ExplodeLoop
+  protected final Object invoke(VirtualFrame frame, VmObjectLike owner, Object receiver) {
+    checkConst(owner);
+    var method = getMethod(owner);
+    var args = new Object[2 + argumentNodes.length];
+    args[0] = receiver;
+    args[1] = owner;
+    frame.setAuxiliarySlot(getMethodSlot(frame), method);
```
bioball: If there's already a method slot, we need to restore the current method slot after this exits. Otherwise, this will clobber the auxiliary slot of an outer method call in the same frame:

```pkl
res = outerMethod(
  innerMethod(new {}), // sets the auxiliary slot
  new {} // now has `innerMethod` in its auxiliary slot
)
```

Looks like we're missing tests for this!

bioball (changes requested):

pkl-core/src/main/java/org/pkl/core/ast/expression/member/AbstractInvokeLexicalOrQualifiedMethodNode.java
HT154: Ahh good catch! I centralized evaluation of args into `AbstractInvokeMethodNode` since it was pretty repetitive. I don't think this impacts PE at all (the `@ExplodeLoop` is retained), but let me know if I should revert that part.

pkl-core/src/main/java/org/pkl/core/ast/expression/member/AbstractInvokeMethodNode.java
```diff
@@ -15,79 +15,52 @@
...
+  public AbstractInvokeMethodNode(SourceSection sourceSection, ExpressionNode[] argumentNodes) {
     super(sourceSection);
-    this.methodName = methodName;
     this.argumentNodes = argumentNodes;
-    this.needsConst = needsConst;
-    this.isConstChecked = false;
   }
 
+  @TruffleBoundary
+  protected int getMethodSlot(FrameDescriptor frameDescriptor) {
```
bioball:
```suggestion
  private int getMethodSlot(FrameDescriptor frameDescriptor) {
```

pkl-core/src/main/java/org/pkl/core/ast/member/ObjectMember.java
```diff
@@ -23,11 +24,13 @@
...
 import org.pkl.core.ast.MemberNode;
 import org.pkl.core.ast.VmModifier;
+import org.pkl.core.ast.type.TypeNode;
 import org.pkl.core.runtime.Identifier;
 import org.pkl.core.runtime.VmDynamic;
+import org.pkl.core.runtime.VmObjectLike;
 import org.pkl.core.runtime.VmUtils;
 
-public final class ObjectMember extends Member {
+public final class ObjectMember extends Member implements Method {
```
bioball: Doesn't feel right that this implements `Method`.

How about we make `ObjectMethodNode` implement `Method` instead?

On the invoke node side, you'd just do something like:

```java
  @Override
  protected Method getMethod(VmObjectLike owner) {
    var member = owner.getMember(methodName);
    assert member != null && member.isLocal();
    var memberNode = (ObjectMethodNode) member.getMemberNode();
    assert memberNode != null;
    return memberNode;
  }
```

pkl-core/src/main/java/org/pkl/core/ast/member/ObjectMethodNode.java
```diff
@@ -57,6 +57,11 @@ public ObjectMethodNode(
     return functionNode.getReturnTypeNode();
   }
 
+  public TypeNode getParameterTypeNode(int idx) {
+    assert functionNode != null;
```
bioball: This isn't safe; `functionNode` isn't guaranteed to exist yet.

I'd recommend a `getFunctionNode()` method that sets the function node if it doesn't exist; follow the same pattern as `org.pkl.core.ast.expression.member.AbstractReadLocalPropertyNode#getCallNode`

Here's a test that breaks right now:

```pkl
buzz {
  local function bar(prop: Person) = prop

  res = bar(new { name = "Sammy" })
}

class Person {
  name: String
}
```

pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
```diff
@@ -76,6 +75,21 @@ public final class VmUtils {
...
   public static final URI REPL_TEXT_URI = URI.create(REPL_TEXT);
 
+  public static final Object CUSTOM_THIS_FRAME_SLOT_ID =
+      new Object() {
+        @Override
+        public String toString() {
+          return "customThisSlot";
+        }
+      };
+  public static final Object METHOD_FRAME_SLOT_ID =
```
bioball, line 86:
```suggestion

  public static final Object METHOD_FRAME_SLOT_ID =
```

bioball (changes requested):

pkl-core/src/main/java/org/pkl/core/ast/member/ObjectMember.java
HT154: Ah I think I tried this and ran into some issue because `ObjectMember.getCallTarget()` runs `getMemberNode().getCallTarget().call()`, but the implementation of `Method.getCallTarget` can just do that itself!

pkl-core/src/main/java/org/pkl/core/ast/member/ObjectMethodNode.java
```diff
@@ -57,34 +58,41 @@ public ObjectMethodNode(
...
-    if (functionNode == null) {
-      CompilerDirectives.transferToInterpreter();
+    return getFunctionNode(frame).getCallTarget();
+  }
 
-      var parameterTypeNodes =
-          VmUtils.resolveParameterTypes(frame, getFrameDescriptor(), unresolvedParameterTypeNodes);
+  private FunctionNode getFunctionNode(VirtualFrame frame) {
+    if (functionNode != null) return functionNode;
+    CompilerDirectives.transferToInterpreter();
```
bioball: We need `CompilerDirectives.transferToInterpreterAndInvalidate()` here because we're setting a `@CompilationFinal` field

```suggestion
    CompilerDirectives.transferToInterpreterAndInvalidate();
```

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinMethodArgumentNode.java
```diff
@@ -0,0 +1,87 @@
...
+
+  @Override
+  public Object executeGeneric(VirtualFrame frame) {
+    if (inferredParent != null) return inferredParent;
+
+    // remaining code only runs first time this node is executed
+    // (assuming evaluation isn't continued despite errors)
+    // except when param type is a non-final self type (not cacheable)
+
+    CompilerDirectives.transferToInterpreter();
```
bioball:
```suggestion
    CompilerDirectives.transferToInterpreterAndInvalidate();
```

bioball (approved):
Two minor issues left, but approving to unblock. Good work on this!

Looking for a 👍  from @stackoverflow too

pkl-core/src/main/java/org/pkl/core/ast/member/ObjectMethodNode.java
HT154: Huh, the original in `executeImpl` didn't invalidate! This was a direct cut/paste.

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinMethodArgumentNode.java
HT154: Likewise, the other `Infer*Node`s don't invalidate here either! Updated them too.

pkl-core/src/main/java/org/pkl/core/ast/member/ObjectMethodNode.java
bioball: Ah, whoops!

pkl-core/src/main/java/org/pkl/core/ast/member/ClassMethod.java
```diff
@@ -131,4 +137,8 @@ public VmTyped getReturnTypeMirror() {
   public PClass.Method export(PClass owner) {
     return functionNode.export(owner, docComment, annotations, modifiers, typeParameters);
   }
+
+  public TypeNode getParameterTypeNode(VirtualFrame frame, int idx) {
+    return functionNode.getParameterTypeNodes()[idx];
```
stackoverflow: This seems to come from source position.
```pkl
class Bar
function f(p: Bar) = p

foo = f(new {}, new {})
```
```bash
org.pkl.core.PklBugException: An unexpected error has occurred. Would you mind filing a bug report?

java.lang.ArrayIndexOutOfBoundsException%3A%20Index%201%20out%20of%20bounds%20for%20length%201

–– Pkl Error ––
Index 1 out of bounds for length 1

4 | foo = f(new {}, new {})
    ^^^
```

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -991,6 +993,16 @@ private ExpressionNode doVisitNewExprWithInferredParent(NewExpr expr) {
           .evalError("cannotInferParent")
           .withSourceSection(createSourceSection(expr.newSpan()))
           .build();
+    } else if (parent instanceof ArgumentList argumentList) {
+      // cases we can't cover currently
+      // - FunctionN.apply: the parameter type nodes are not stored
+      // - pkl.base intrinsic constructorts: List(), Set(), Map(), Bytes()
+      // - generic methods: pkl.base#Pair(), etc.
+      // these will throw cannotInferParent at runtime
```
stackoverflow, line 1102: In your snippet `genericMethod` results in "Cannot instantiate type `First`.", not cannot infer type. Or am I reading this wrong?

pkl-core/src/main/java/org/pkl/core/ast/expression/member/AbstractInvokeMethodNode.java
```diff
@@ -15,79 +15,51 @@
...
-public abstract sealed class AbstractInvokeMethodNode extends ExpressionNode
-    permits AbstractInvokeQualifiedMethodNode, AbstractInvokeLexicalMethodNode {
+public abstract class AbstractInvokeMethodNode extends ExpressionNode {
 
-  protected final Identifier methodName;
-  private final boolean needsConst;
-  @Children private ExpressionNode[] argumentNodes;
-  @Child private @Nullable DirectCallNode callNode;
-  @CompilationFinal protected boolean isConstChecked;
+  @Children protected ExpressionNode[] argumentNodes;
```
stackoverflow: This could be `final`.

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinMethodArgumentNode.java
```diff
@@ -0,0 +1,56 @@
...
+import com.oracle.truffle.api.frame.VirtualFrame;
+import com.oracle.truffle.api.source.SourceSection;
+import org.jspecify.annotations.Nullable;
+import org.pkl.core.ast.member.Method;
+import org.pkl.core.runtime.VmLanguage;
+import org.pkl.core.runtime.VmUtils;
+
+public final class InferParentWithinMethodArgumentNode extends AbstractInferParentNode {
+  private final int argIndex;
+  @CompilationFinal private @Nullable Object inferredParent;
```
stackoverflow: Why redeclare this?

pkl-core/src/main/java/org/pkl/core/ast/expression/member/AbstractInferParentNode.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
```
stackoverflow:
```suggestion
 * Copyright © 2026 Apple Inc. and the Pkl project authors. All rights reserved.
```

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinMethodArgumentNode.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
```
stackoverflow:
```suggestion
 * Copyright © 2026 Apple Inc. and the Pkl project authors. All rights reserved.
```

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinMethodArgumentNode.java
HT154: Copy pasta

pkl-core/src/main/java/org/pkl/core/ast/member/ObjectMethodNode.java
```diff
@@ -57,29 +58,42 @@ public ObjectMethodNode(
...
 
-      var parameterTypeNodes =
-          VmUtils.resolveParameterTypes(frame, getFrameDescriptor(), unresolvedParameterTypeNodes);
+  private FunctionNode getFunctionNode(VirtualFrame frame) {
+    if (functionNode != null) return functionNode;
+    CompilerDirectives.transferToInterpreterAndInvalidate();
 
-      var returnTypeNode =
-          unresolvedReturnTypeNode != null ? unresolvedReturnTypeNode.execute(frame) : null;
+    adoptChildren();
```
HT154: Not sure why adopting children is needed here, but resolving parameter types fails without it!

pkl-core/src/main/java/org/pkl/core/ast/member/ClassMethod.java
stackoverflow: This is still happening for object methods:
```pkl
class Bar { x: Int = 1 }
obj {
  local function f(p: Bar) = p.x
  res = f(new {}, new {})
}
```
```bash

java.lang.ArrayIndexOutOfBoundsException%3A%20Index%201%20out%20of%20bounds%20for%20length%201

–– Pkl Error ––
Index 1 out of bounds for length 1

4 | res = f(new {}, new {})   // expect ArrayIndexOutOfBoundsException, not "Cannot tell which parent to amend"
    ^^^
```

pkl-core/src/main/java/org/pkl/core/ast/expression/member/AbstractInferParentNode.java
```diff
@@ -0,0 +1,91 @@
...
+      // try to produce a more specific error message than "cannotInstantiateType"
+      var clazz = typeNode.getVmClass();
+      if (clazz != null) VmUtils.checkIsInstantiable(clazz, typeNode);
+
+      throw exceptionBuilder()
+          .evalError("cannotInstantiateType", typeNode.getSourceSection().getCharacters())
+          .build();
+    }
+
+    if (typeNode.isFinalType()) {
```
stackoverflow: This is cached per node, but call sites can be polymorphic for arguments, regardless of the type being final or not:
```pkl
class Bar1 { x: Int = 1 }
class Bar2 { x: Int = 2 }
class C1 { function f(p: Bar1) = p.x }
class C2 { function f(p: Bar2) = p.x * 10 }

local function call(a) = a.f(new {}) // poly call

res1 = call(new C1 {})
res2 = call(new C2 {})
```
```bash
–– Pkl Error ––
Expected value of type `foo#Bar2`, but got type `foo#Bar1`.
Value: new Bar1 { x = 1 }

4 | class C2 { function f(p: Bar2) = p.x * 10 }
                             ^^^^
at foo#C2.f (file:///.../foo.pkl, line 4)

6 | local function call(a) = a.f(new {}) // poly call
                             ^^^^^^^^^^^
at foo#call (file:///.../foo.pkl, line 6)

9 | res2 = call(new C2 {})
           ^^^^^^^^^^^^^^^
```

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -991,6 +993,16 @@ private ExpressionNode doVisitNewExprWithInferredParent(NewExpr expr) {
           .evalError("cannotInferParent")
           .withSourceSection(createSourceSection(expr.newSpan()))
           .build();
+    } else if (parent instanceof ArgumentList argumentList) {
+      // cases we can't cover currently
+      // - FunctionN.apply: the parameter type nodes are not stored
+      // - pkl.base intrinsic constructorts: List(), Set(), Map(), Bytes()
```
stackoverflow:
```suggestion
      // - pkl.base intrinsic constructors: List(), Set(), Map(), Bytes()
```

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InvokeSuperMethodNode.java
```diff
@@ -39,13 +38,12 @@ protected InvokeSuperMethodNode(
...
+    super(sourceSection, argumentNodes);
     this.needsConst = needsConst;
 
     assert !methodName.isLocalMethod();
 
     this.methodName = methodName;
-    this.argumentNodes = argumentNodes;
   }
 
   @ExplodeLoop
```
stackoverflow:
```suggestion
```
No loop here anymore.

pkl-core/src/main/java/org/pkl/core/ast/member/FunctionNode.java
```diff
@@ -81,6 +81,10 @@ public int getParameterCount() {
     return paramCount;
   }
 
+  public TypeNode[] getParameterTypeNodes() {
```
stackoverflow: nit: a bit unsafe to handle an array to the caller, that could mutate it. Could have being a `getParameterTypeNode(int)`

pkl-core/src/test/files/LanguageSnippetTests/output/methods/inferParameterType.pcf
```diff
@@ -0,0 +1,12 @@
+qualified = 1
+unqualifiedLexical = 1
+unqualifiedThis = 8
+`super` = 16
+nestedMethodCalls = 26
+objectMethod {
+  call = 0
+}
+intrinsicConstructor = "Cannot tell which parent to amend."
+genericMethod = "Cannot instantiate type `First`."
```
stackoverflow: nit: ```"Cannot instantiate type `First`."``` is a strange message. What is type `First`? Can I `new First {}`?
Using `cannotInferParent` would be better.

pkl-core/src/main/java/org/pkl/core/ast/expression/member/AbstractInferParentNode.java
HT154: Hmm yeah I think the options here are either a) don't cache at all or b) cache based on method identity. I'm going to avoid caching for now but we can revisit this later if needed.

bioball: You can use `@Cached` to cache based on computed values (e.g. the receiver's class); for reference, see https://github.com/apple/pkl/blob/1208fa94388725a7c67de4d617d893e3a5141cbf/pkl-core/src/main/java/org/pkl/core/ast/expression/member/ReadPropertyNode.java#L104

HT154: @bioball I gave this a try in https://github.com/apple/pkl/pull/1768/changes/d51ceb1b7c4b66b334b7c5629930bfc6df412ad3. Unsure if this is on target, though.

stackoverflow (approved):
Looks good!

pkl-core/src/main/java/org/pkl/core/ast/expression/member/AbstractInferParentNode.java
```diff
@@ -0,0 +1,69 @@
...
+    // try to produce a more specific error message than "cannotInstantiateType"
+    var clazz = typeNode.getVmClass();
+    if (clazz != null) {
+      VmUtils.checkIsInstantiable(clazz, typeNode);
+    }
+
+    CompilerDirectives.transferToInterpreter();
+    throw exceptionBuilder()
+        .evalError("cannotInstantiateType", typeNode.getSourceSection().getCharacters())
+        .build();
```
bioball, line 67:
```suggestion
    CompilerDirectives.transferToInterpreter();

    if (typeNode instanceof TypeVariableNode) {
      throw exceptionBuilder().evalError("cannotInferParent").build();
    }

    // try to produce a more specific error message than "cannotInstantiateType"
    var clazz = typeNode.getVmClass();
    if (clazz != null) {
      VmUtils.checkIsInstantiable(clazz, typeNode);
    }

    throw exceptionBuilder()
        .evalError("cannotInstantiateType", typeNode.getSourceSection().getCharacters())
        .build();
```

pkl-core/src/main/java/org/pkl/core/ast/expression/member/AbstractInvokeMethodNode.java
```diff
@@ -15,79 +15,51 @@
...
   }
 
   @ExplodeLoop
-  protected final Object invoke(VirtualFrame frame, VmObjectLike owner, Object receiver) {
-    checkConst(owner);
+  protected Object[] evalArgs(
+      VirtualFrame frame, @Nullable Method method, Object owner, @Nullable Object receiver) {
+    var methodSlot = getMethodSlot(frame.getFrameDescriptor());
+    var prevMethod = frame.getAuxiliarySlot(methodSlot);
+    frame.setAuxiliarySlot(methodSlot, method);
```
bioball, line 50: Ah, one final issue: we can optimize these nodes when no argument is an implicit `new {}`.

I think what we can do here is: when constructing the argument nodes, we can keep track of whether the node is an implicit `new {}`. Then, we can pass that as a boolean argument to the constructor of this node.

If none of the arguments are `new {}`, we can skip all of the logic around setting the auxiliary slots before executing the argument nodes.

We can also handle this in a follow-up PR, too, so, not blocking for here.

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinMethodArgumentNode.java
```diff
@@ -0,0 +1,88 @@
...
+  protected @Nullable TypeNode getTypeNode(VirtualFrame frame, Method method) {
+    return method.getParameterTypeNode(frame, argIndex);
+  }
+
+  // keep specializations in sync with other AbstractInferParentFromMethodNode subclasses
+
+  @Specialization(guards = {"method == cachedMethod", "isFinalType(cachedMethod, typeNode)"})
+  protected final Object evalCached(
+      @SuppressWarnings("unused") VirtualFrame frame,
+      @Bind("getMethod(frame)") Method method,
```
bioball:
```suggestion
      @Bind("getMethod(frame)") @SuppressWarnings("unused") Method method,
```

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinMethodArgumentNode.java
```diff
@@ -0,0 +1,88 @@
...
+      @Cached("getMethod(frame)") @SuppressWarnings("unused") Method cachedMethod,
+      @Cached("getTypeNode(frame, cachedMethod)") @SuppressWarnings("unused") TypeNode typeNode,
+      @Cached(
+              "getDefaultValue(frame, typeNode, cachedMethod.getHeaderSection(), cachedMethod.getQualifiedName())")
+          Object defaultValue) {
+    return defaultValue;
+  }
+
+  @Specialization(replaces = "evalCached")
+  protected final Object eval(VirtualFrame frame, @Bind("getMethod(frame)") Method method) {
```
bioball: FYI I don't see that this `getMethod(frame)` call is shared between `evalCached` and `eval`. I also tried adding `@Idempotent` to `getMethod(Frame)`, and that didn't change the generated code.

bioball (approved):
LGTM!

One comment about optimization, but non-blocking.

pkl-core/src/main/java/org/pkl/core/ast/expression/member/AbstractInvokeMethodNode.java
HT154: Left a TODO here for later. Agreed that it would be good not to hit the slot allocation truffle boundary on every invocation. Do you have an idea how to best go about tracking this in `AstBuilder`? I'm loathe to add an instance property for this, but off the cuff I can't think of a better way to approach it.

PR #1770, opened 2026-07-14
Typo fix: `s/descendent/descendant`

HT154:


HT154:
Sources are conflicting on what's most correct here! Some recommend -ent for the adjective form used here, while others recommend -ant in all cases. TIL how mixed opinions are on this! That said, this is clear enough as-is so I'm going to close this.

PR #1778, opened 2026-07-15
Mark protobuf module as deprecated

bioball:
This moves the protobuf module from experimental to deprecated

Rationale:

* This is very seldom used, if used at all
* Not fully fleshed out (this really should be paired with a protoc-gen-pkl, which doesn't exist)
* There is a ProtoJSON wire format that we can already render to

### DIFF
--- a/docs/modules/pkl-cli/pages/index.adoc
+++ b/docs/modules/pkl-cli/pages/index.adoc
@@ -307,7 +307,7 @@ The default output renderer for a module supports the following formats:
 * `pcf`
 * `plist`
 * `properties`
-* `textproto`
+* `textproto` (deprecated in Pkl 0.33)
 * `xml`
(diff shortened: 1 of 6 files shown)


HT154 (approved):
Do we also want to remove `textproto` from the `pkl eval --format` CLI help?

StefMa:
Does it really make sense to only deprecate it instead of just removing it? It is marked as experimental and is probably not used often, if ever. I'm not sure you really want to take on the burden of going through a deprecation cycle.

bioball:
@StefMa good thought, although, we don't ever straight up APIs from the stdlib; they always go through a deprecation cycle first.

We'll likely get rid of this entirely in Pkl 0.34.

stackoverflow (approved):

PR #1781, opened 2026-07-16
Remove `abstract` properties

HT154:
It is not clear what an abstract property means as properties do not need an implementation even in non-abstract classes. Additionally, `fixed` properties in abstract classes already do not require implementations while subclasses must provide one.

This change ignores the `abstract` modifier when applied to properties. It does not make marking a property `abstract` an error at this time to avoid breaking user code.

Corresponding changes to pkl-intellij and pkl-lsp will mark use of the abstract modifier on properties as a warning+unused and provide a quick fix to remove it.

Removing abstract properties affects only this behavior:
* Removal of validation that abstract properties may not have an implementation (default value)
* Removal of tailored error for access to an abstract property
* Reflection no longer includes `abstract` modifiers for properties, affecting `pkl:reflect`, exported Java values (`PClass` and friends), and pkl-doc output


pkl-core/src/main/java/org/pkl/core/ast/builder/CannotInvokeAbstractPropertyNode.java
stackoverflow, line 1: You deleted this class but not the error message: `cannotInvokeAbstractProperty`.

stackoverflow (approved):
Looks good. One comment.

pkl-core/src/main/java/org/pkl/core/ast/builder/CannotInvokeAbstractPropertyNode.java
HT154: Ah good catch, fixed!

stdlib/base.pkl
```diff
@@ -2824,10 +2824,10 @@ abstract external class Collection<out Element> extends Any {
   /// Shorthand for `maxWith((a, b) -> a < b)`.
   ///
   /// Throws if this collection is empty, or if any two elements cannot be compared with `<`.
-  abstract max: Element
+  max: Element
 
   /// Same as [max] but returns [null] if this collection empty.
-  abstract maxOrNull: Element
+  maxOrNull: Element
```
bioball, line 2830: I wonder if we should make all of these `external`, and remove the corresponding overrides in `Int` and `Float`?

We can have a `NumberNodes` that defines specializations for `long` and `double`, e.g.

```java
  public abstract static class ns extends ExternalPropertyNode {
    @Specialization
    protected VmDuration eval(double self) {
      return new VmDuration(self, DurationUnit.NANOS);
    }

    @Specialization
    protected VmDuration eval(long self) {
      return new VmDuration(self, DurationUnit.NANOS);
    }
  }
```

I'm not really sure what's better. From a code perspective, it seems nicer without all of these overrides in `base.pkl`.

bioball (approved):
LGTM

stdlib/base.pkl
HT154: A case of six meet half-dozen, I think. The one argument I can think of against this is that pkl-doc hides superclass members by default, so it may be harder to discover the unit properties.

PR #1783, opened 2026-07-17
Add feature flagging system

HT154:
Adds:
* New CLI flag `--feature=<name>=[true|false]`
* Pkl API
  * Property `pkl.EvaluatorSettings#featureFlags: Mapping<KnownFeatureFlags | String, Boolean>`
  * Typealias `pkl.EvaluatorSettings#FeatureFlags = nothing` (no feature flags yet, but this will be a string literal union)
* Java API
  * `org.pkl.core.FeatureFlags` - enum to document feature flags and their default values.
  * `org.pkl.core.evaluatorSettings.PklEvaluatorSettings.featureFlags`
  * `org.pkl.core.Analyzer` now requires constructor argument `Map<FeatureFlag, Boolean> featureFlags`
  * `org.pkl.core.EvaluatorBuilder`
    * `addFeatureFlag`, `addFeatureFlags`, `setFeatureFlags`, `getFeatureFlags`
    * `applyFromProject(Project, Logger)` (for capturing string->`FeatureFlag` conversion warnings)
      * The overload without the `Logger` argument now passes `Loggers.noop()` to this API
  * `org.pkl.config.java.ConfigEvaluatorBuilder.applyFromProject(Project, Logger)` (for capturing string->`FeatureFlag` conversion warnings)
    * The overload without the `Logger` argument now passes `Loggers.noop()` to this API
  * `org.pkl.core.Logger` (changes to assist with warnings not from Pkl code)
    * `trace(String message, String frameUri)` (now called by `trace(String, StackFrame)`)
    * `warn(String message, String frameUri)` (now called by `warn(String, StackFrame)`)
* Gradle tasks now accept `featureFlags` (as a `MapProperty<String, Boolean>`)
* Message passing API `CreateEvaluatorRequest` now accepts a map property `featureFlags`, which is a map of string to boolean values.

Feature flag lifecycle:
1. Feature flags are added by adding a case to `org.pkl.core.FeatureFlags`.
2. Flags may change default values as the language evolves.
3. Prior to removal, enum cases should be marked `@Deprecated` for at least one release to provide Java API users an opportunity to discontinue usage.
4. Finally, a flag is retired by removing its enum case and any code accessing its value.

Notably, because Pkl's stdlib is statically initialized with a separate `VmContext`, it can set its own feature flags independently of each user-initiated evaluation!

### DIFF
--- a/docs/modules/bindings-specification/pages/message-passing-api.adoc
+++ b/docs/modules/bindings-specification/pages/message-passing-api.adoc
@@ -126,6 +126,28 @@ project: Project?
 /// Added in Pkl 0.26.0.
 http: Http?
 
+/// Register external module readers.
+///
+/// Added in Pkl 0.27.0.
+externalModuleReaders: Mapping<String, ExternalReader>?
+
+/// Register external resource readers.
+///
+/// Added in Pkl 0.27.0.
+externalResourceReaders: Mapping<String, ExternalReader>?
+
+/// Dictates the rendering of calls to the trace() method within Pkl.
+///
+/// Added in Pkl 0.30.0.
+traceMode: ("compact" | "pretty")?
+
+/// Feature flags to enable or disable.
+///
+/// Specifying a feature flag unknown to the current version of Pkl will result in a warning.
+///
+/// Added in Pkl 0.33.0.
+featureFlags: Mapping<String, Boolean>?
+
 class ClientResourceReader {
   /// The URI scheme this reader is responsible for reading.
   scheme: String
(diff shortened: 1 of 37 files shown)


docs/modules/bindings-specification/pages/message-passing-api.adoc
```diff
@@ -126,6 +126,28 @@ project: Project?
...
+
+/// Register external resource readers.
+///
+/// Added in Pkl 0.27.0.
+externalResourceReaders: Mapping<String, ExternalReader>?
+
+/// Dictates the rendering of calls to the trace() method within Pkl.
+///
+/// Added in Pkl 0.30.0.
+traceMode: ("compact" | "pretty")?
```
HT154, line 142: Whoops, we've missed updating this for a bit!

PR #1788, opened 2026-07-22
Fix variable resolution of object body params

### LINKED ISSUE #1787: Regression in variable resolution
This snippet used to work in Pkl 0.31, but is now failing:

```pkl
foo = new Mapping {
  default { key ->
    when (key is String) {
      bar = key
    }
  }
  ["x"] {}
}
```

bioball:
Fixes a regression introduced in Pkl 0.32.

Closes #1787


HT154 (approved):
LGTM

PR #1793, opened 2026-07-24
Fix negative caret width rendering multi-line expression errors

protobufel2:
If you evaluate a multi-line expression that fails, you don't get the error — you get a crash:

```java  
  Evaluator ev = Evaluator.preconfigured();
  ev.evaluateExpressionString(
      ModuleSource.text("x = 1"),
      "let (_a = 1)\n    let (_b = 2)\n        throw(\"boom\")");
  // java.lang.IllegalArgumentException: count is negative: -11
```

  A single-line expression renders fine, so this crept in with the preamble handling from #1760: the preamble length gets subtracted from endColumn even when endColumn is the already-stripped sourceLine.length(). On a multi-line frame that double-counts and the caret width goes negative, so String.repeat
  throws instead of the renderer drawing the ^^^.

  Fix is to apply the offset only to the frame-derived columns. Added a regression test; build and buildNative pass.

### DIFF
--- a/pkl-core/src/test/kotlin/org/pkl/core/EvaluateExpressionTest.kt
+++ b/pkl-core/src/test/kotlin/org/pkl/core/EvaluateExpressionTest.kt
@@ -166,6 +166,34 @@ class EvaluateExpressionTest {
   }
 
   @Test
+  fun `evaluate throwing multi-line expression renders without crashing`() {
+    val error =
+      assertThrows<PklException> {
+        evaluate("x = 1", "let (_a = 1)\n    let (_b = 2)\n        throw(\"boom\")")
+      }
+
+    // A multi-line frame used to throw `IllegalArgumentException: count is negative` while
+    // rendering.
+    assertThat(error.message)
+      .isEqualTo(
+        """
+        –– Pkl Error ––
+        boom
+
+        3 | throw("boom")
+            ^^^^^^^^^^^^^
+        at  (repl:text)
+
+        1 | let (_a = 1)
+            ^^^^^^^^^^^^
+        at repl:text.<let expr> (repl:text)
+
+        """
+          .trimIndent()
+      )
+  }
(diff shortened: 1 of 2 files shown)


docs/modules/release-notes/pages/0.33.adoc
```diff
@@ -27,7 +27,7 @@ To continue running new Pkl releases on these systems, use an appropriate Java r
 
 == Bug Fixes [small]#🐜#
 
-* XXX
+* Rendering an error raised by a multi-line expression input crashes with `IllegalArgumentException: count is negative` instead of reporting the Pkl error (https://github.com/apple/pkl/pull/XXXX[#XXXX])
```
HT154: Please back this one out for now. We may release this in a patch release to 0.32 and we don't cover patch release changes in subsequent minor release notes.

HT154 (approved):
LGTM, modulo the release notes change. Approving to unblock.

pkl-core/src/test/kotlin/org/pkl/core/EvaluateExpressionTest.kt
```diff
@@ -165,6 +165,35 @@ class EvaluateExpressionTest {
...
+  @Test
+  fun `evaluate throwing multi-line expression renders without crashing`() {
+    val error =
+      assertThrows<PklException> {
+        evaluate("x = 1", "let (_a = 1)\n    let (_b = 2)\n        throw(\"boom\")")
+      }
+
+    // Regression: a frame spanning multiple lines used to make the caret width go negative once the
+    // expression preamble was accounted for, throwing `IllegalArgumentException: count is negative`
+    // instead of rendering the Pkl error.
```
bioball: This feels overly informative

```suggestion
```

bioball (approved):
Thanks!

PR #1795, opened 2026-07-26
Add `pkl project package --install` to local cache

HT154:


pkl-core/src/main/java/org/pkl/core/packages/PackageResolvers.java
```diff
@@ -684,5 +698,20 @@ public void close() throws IOException {
...
+      var metadataCacheFile = doGetMetadataPath(packageUri);
+      Files.createDirectories(metadataCacheFile.getParent());
+      Files.copy(metadataFile, metadataCacheFile);
+      Files.setPosixFilePermissions(metadataCacheFile, FILE_PERMISSIONS);
+
+      var zipCacheFile = doGetZipFilePath(packageUri);
+      Files.createDirectories(zipCacheFile.getParent());
+      Files.createDirectories(zipCacheFile.getParent());
+      Files.copy(zipFile, zipCacheFile);
+      Files.setPosixFilePermissions(zipCacheFile, FILE_PERMISSIONS);
```
stackoverflow: Don't you need the `if (!IoUtils.isWindows())` guard here like in many other places in this file?

pkl-core/src/main/java/org/pkl/core/project/ProjectPackager.java
```diff
@@ -178,6 +186,12 @@ public PackageResult doPackage(Project project) throws IOException {
         new PackageResult(
             metadataFile, metadataChecksumFile, zipFile, zipChecksumFile, metadataFileChecksum);
     packageResults.put(pkg.uri(), result);
+    if (packageWriteResolver != null) {
+      try {
+        packageWriteResolver.writePackage(pkg.uri(), metadataFile, zipFile);
+      } catch (UnsupportedOperationException ignored) {
```
stackoverflow: Do you need to ignore the exception here? If I read the code correctly, `packageWriteResolver` is always `DiskCachedPackageResolver`, not the in memory one.

pkl-core/src/main/java/org/pkl/core/packages/PackageResolvers.java
```diff
@@ -684,5 +698,20 @@ public void close() throws IOException {
         fileSystems.clear();
       }
     }
+
+    @Override
+    public void writePackage(PackageUri packageUri, Path metadataFile, Path zipFile)
+        throws IOException {
+      var metadataCacheFile = doGetMetadataPath(packageUri);
+      Files.createDirectories(metadataCacheFile.getParent());
+      Files.copy(metadataFile, metadataCacheFile);
```
stackoverflow: What if the package was already cached? Worth adding a `StandardCopyOption.REPLACE_EXISTING`.
Same for the copy bellow.

pkl-core/src/main/java/org/pkl/core/packages/PackageResolvers.java
```diff
@@ -684,5 +698,20 @@ public void close() throws IOException {
...
+      Files.setPosixFilePermissions(metadataCacheFile, FILE_PERMISSIONS);
+
+      var zipCacheFile = doGetZipFilePath(packageUri);
+      Files.createDirectories(zipCacheFile.getParent());
+      Files.createDirectories(zipCacheFile.getParent());
```
stackoverflow:
```suggestion
      Files.createDirectories(zipCacheFile.getParent());
```

pkl-core/src/main/java/org/pkl/core/packages/PackageResolvers.java
```diff
@@ -684,5 +698,20 @@ public void close() throws IOException {
         fileSystems.clear();
       }
     }
+
+    @Override
+    public void writePackage(PackageUri packageUri, Path metadataFile, Path zipFile)
```
stackoverflow: `getMetadataPath`/`getZipFilePath` write to a tmp dir and use `Files.move(... , StandardCopyOption.ATOMIC_MOVE)` to make sure the move is atomic. Worth thinking about this case here so we don't end up with a partial write.

pkl-cli/src/main/kotlin/org/pkl/cli/commands/ProjectCommand.kt
```diff
@@ -128,13 +128,22 @@ class PackageCommand : BaseCommand(name = "package", helpLink = helpLink) {
       .single()
       .flag()
 
+  private val install: Boolean by
+    option(
+        names = arrayOf("--install"),
+        help = "Install the built package into the local package cache",
```
bioball:
```suggestion
        help = "Install the built package into the module cache dir",
```

bioball (changes requested):
Some questions:

1. Should the package command fail if `--install` and `--no-cache` are both supplied? It seems strange that this succeeds; these are conflicting flags.
2. Should there be another flag that accepts a directory to write to? Running the packager can possibly download modules, because it executes the evaluator, and the evaluator can possibly download packages to the cache dir. The act of writing dependencies to the cache is separate from the act of installing the input package.
3. Should the CLI write some additional output in the case of `--install`? Right now, it writes the package's filenames.

Also, this is missing docs for both the CLI and Gradle.

HT154:
> 1. Should the package command fail if `--install` and `--no-cache` are both supplied? It seems strange that this succeeds; these are conflicting flags.

Yup, added this. It validates it twice: one in the CLI command itself so you get a nice clikt-y error) and once in `ProjectPackager` to catch other entrypoints.

> 2. Should there be another flag that accepts a directory to write to? Running the packager can possibly download modules, because it executes the evaluator, and the evaluator can possibly download packages to the cache dir. The act of writing dependencies to the cache is separate from the act of installing the input package.

Hmm in what situation would you want to write a package to the cache but _not_ write its dependencies there as well?

> 3. Should the CLI write some additional output in the case of `--install`? Right now, it writes the package's filenames.

Added output of the (absolute) paths to the written location within the cache.

> Also, this is missing docs for both the CLI and Gradle.

Added.

pkl-cli/.out/mypackage@1.0.0/mypackage@1.0.0.zip.sha256
```diff
@@ -0,0 +1 @@
+8739c76e681f900923b900c9df0ef75cf421d39cabb54650c4b9ad19b6a76d85
```
bioball: Accidentally commited?

bioball (approved):
LGTM; just one comment about removing the stuff in pkl-cli/.out, but approving to unblock

> Hmm in what situation would you want to write a package to the cache but not write its dependencies there as well?

Yeah, not sure. Probably fine as is; we can always add another flag if needed too.

pkl-cli/.out/mypackage@1.0.0/mypackage@1.0.0.zip.sha256
HT154: Could've sworn I'd rm'd those, but that might've been before I fixed the test. Gone now.

PR #1796, opened 2026-07-28
Improve evaluation of typealiases

### LINKED ISSUE #1712: `module` types as typelias type args are evaluated in the context of the alias's module, not where `module` is used

bioball:
This simplifies the resolution of names in typealias constraints. This removes the existing logic around swapping out the frame's owner/receiver, and instead favors resolving the variable at parse time.

* During variable resolution, create read variable nodes that read off of a receiver node
* Introduce `GetTypeAliasModuleNode` that provides the typealias's enclosing module
* Inject the enclosing module during typealias instantiation

Fixes #1712
Closes #1712


pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -2735,40 +2735,13 @@ public VmList getTypeArgumentMirrors() {
...
-      setOwner(frame, VmUtils.getOwner(typeAlias.getEnclosingFrame()));
-      setReceiver(frame, VmUtils.getReceiver(typeAlias.getEnclosingFrame()));
-
-      try {
-        return aliasedTypeNode.executeLazily(frame, value);
-      } finally {
-        setOwner(frame, prevOwner);
-        setReceiver(frame, prevReceiver);
-      }
+      return aliasedTypeNode.executeLazily(frame, value);
```
HT154, line 2738: rm also in `executeAndSet`? If so, also rm `setOwner`/`setReceiver` entirely.

pkl-core/src/test/files/LanguageSnippetTests/input-helper/types/typeAliasAmendBase.pkl
```diff
@@ -0,0 +1 @@
+name: String = "base"
```
HT154: newline at eof is missing in several of these snippet test inputs

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InvokeLexicalObjectMethodNode.java
```diff
@@ -16,23 +16,37 @@
...
+    this.levelsUp = levelsUp;
   }
 
+  @Override
+  public Object executeGeneric(VirtualFrame frame) {
+    var capturedFrame = VmUtils.getFrame(frame, levelsUp);
+    var owner = VmUtils.getOwner(capturedFrame);
+    var receiver = VmUtils.getReceiver(capturedFrame);
+    return invoke(frame, owner, receiver);
+  }
```
HT154: This is identical to `InvokeLexicalClassMethodNode`'s. Worth DRYing this up in the class hierarchy further?

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InvokeQualifiedObjectMethodNode.java
```diff
@@ -0,0 +1,62 @@
...
+      ExpressionNode getReceiverNode) {
+    super(sourceSection, methodName, argumentNodes, needsConst);
+    this.getReceiverNode = getReceiverNode;
+  }
+
+  @Override
+  public Object executeGeneric(VirtualFrame frame) {
+    var receiver = (VmTyped) getReceiverNode.executeGeneric(frame);
+    return invoke(frame, receiver, receiver);
+  }
```
HT154: Same here: DRY up w.r.t. `InvokeQualifiedClassMethodNode`?

HT154 (commented):
Very slick! A few fairly minor things, but this definitely seems like the right direction to be moving in.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
bioball: Whoops, good catch

HT154 (approved):
LGTM!

pkl-core/src/main/java/org/pkl/core/ast/expression/primary/GetTypeAliasModuleNode.java
```diff
@@ -0,0 +1,53 @@
...
+    super(sourceSection);
+  }
+
+  @Override
+  public boolean isInstrumentable() {
+    return false;
+  }
+
+  public void lateInitModule(VmTyped module) {
+    // should only set this once
```
stackoverflow: nit: good idea to explain why here: that a node may be visited again by a referencing alias's walk and must not be overwritten.
Perhaps changing the should to must.

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -687,12 +691,28 @@ private ExpressionNode resolveReadVariable(UnqualifiedAccessExpr expr) {
             case MODULE -> p.isModuleScope();
             case ALL -> p.levelsUp() > constDepth;
           };
+      if (scope.isInTypeAliasScope() && p.isModuleScope()) {
```
stackoverflow: nit: worth a comment that if we ever support nested typealiases we need to expand this code (and the method one).

pkl-core/src/main/java/org/pkl/core/ast/expression/primary/GetTypeAliasModuleNode.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.oracle.truffle.api.CompilerDirectives;
+import com.oracle.truffle.api.CompilerDirectives.CompilationFinal;
+import com.oracle.truffle.api.frame.VirtualFrame;
+import com.oracle.truffle.api.source.SourceSection;
+import org.jspecify.annotations.Nullable;
+import org.pkl.core.ast.ExpressionNode;
+import org.pkl.core.runtime.VmTyped;
+
+public class GetTypeAliasModuleNode extends ExpressionNode {
```
stackoverflow:
```suggestion
public final class GetTypeAliasModuleNode extends ExpressionNode {
```
nit

pkl-core/src/main/java/org/pkl/core/ast/expression/member/ReadQualifiedLocalPropertyNode.java
```diff
@@ -0,0 +1,53 @@
...
+package org.pkl.core.ast.expression.member;
+
+import com.oracle.truffle.api.frame.VirtualFrame;
+import com.oracle.truffle.api.source.SourceSection;
+import org.pkl.core.ast.ExpressionNode;
+import org.pkl.core.runtime.Identifier;
+import org.pkl.core.runtime.VmTyped;
+
+/** Reads a local property off of the receiver node. */
+public class ReadQualifiedLocalPropertyNode extends AbstractReadLocalPropertyNode {
```
stackoverflow:
```suggestion
public final class ReadQualifiedLocalPropertyNode extends AbstractReadLocalPropertyNode {
```
nit: `ReadLexicalLocalPropertyNode` is final

pkl-core/src/main/java/org/pkl/core/ast/expression/primary/GetTypeAliasModuleNode.java
```diff
@@ -0,0 +1,53 @@
...
+
+  @CompilationFinal public @Nullable VmTyped module;
```
stackoverflow: nit: a bit dangerous to make this public mutable. Adding a getter and making it private/protected would be better.

stackoverflow (approved):
LGTM

pkl-core/src/main/java/org/pkl/core/ast/expression/primary/GetTypeAliasModuleNode.java
bioball: Good catch; this can just be private

PR #1803, opened 2026-07-29
Use EnumSet internally for Token type checks

HT154:
@bioball ran a quick benchmark on this and `EnumSet` has higher throughput:
```
Benchmark           Mode  Cnt         Score        Error  Units
EnumSet.enumSet    thrpt    5  71988081.614 ± 895845.964  ops/s
EnumSet.isKeyword  thrpt    5  28355512.127 ± 528516.078  ops/s
```


bioball (approved):
LGTM

PR #1805, opened 2026-07-30
Fix various issues in libpkl

bioball:
* Fix possible SIGSEGV from host process.
   - Require that the same `pkl_exec_t` be used in the same OS thread; returning meaningful error if this fails
* Remove `System.exitProcess(1)` logic in NativeTransport; this would kill the host process too and isn't an appropriate action for a received ProtocolException
* Allow multiple calls to pkl_init without pkl_close; this limitation doesn't really make any sense
* Prefer calling methods defined in graal_isolate.h, instead of creating the same method via CEntryPoint
* Add a CMakeLists.txt so that CLion can follow the code and provide proper diagnostics
* Improve doc comments; describe params as either in or out

### DIFF
--- a/libpkl/src/main/c/include/pkl.h
+++ b/libpkl/src/main/c/include/pkl.h
@@ -27,7 +27,7 @@ extern "C" {
 #define PKL_EXPORT __attribute__((visibility("default")))
 #endif
 
-#define PKL_ERR_LOCK      1     /* Failed to create a mutex, or acquire a lock on a mutex */
+#define PKL_ERR_THREAD    1     /* Called using the same pexec_t but from a different thread */
 #define PKL_ERR_PROTOCOL  2     /* Failed to decode a message */
 
 /** Error details that occurred during a method call */
@@ -38,16 +38,17 @@ typedef struct {
 /**
  * Pkl executor instance that manages communication with the Pkl runtime.
  *
- * Instances should be created via `pkl_init()` and destroyed via `pkl_close().`
+ * Instances should be created via `pkl_init` and destroyed via `pkl_close`.
  *
- * All operations on this struct are considered thread-safe and are synchronized via a mutex.
+ * All calls using this executor should be synchronized in the same thread.
+ * Failing to do so will possibly cause the process to be aborted with a segfault.
  */
 typedef struct __pkl_exec_t pkl_exec_t;
 
 /**
  * The callback that gets called when a message is received from Pkl.
  *
- * Messages must be deserialized to Pkl's Message Passing API: 
+ * Messages must be deserialized to Pkl's Message Passing API:
  * https://pkl-lang.org/main/current/bindings-specification/message-passing-api.html
  *
  * @param[in] length    The length of the message bytes
@@ -62,6 +63,8 @@ typedef void (*pkl_message_response_handler)(unsigned int length, char *message,
  *
  * To clean up resources allocated by the executor, use `pkl_close()`.
  *
+ * All calls using this executor should come from the same thread.
+ *
  * @param[in] handler   The callback that gets called when a message is received from Pkl.
  * @param[in] userData  User-defined data that gets passed to handler.
  * @param[out] exec      The pointer to write the created pkl_exec_t to.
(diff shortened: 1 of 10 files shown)


libpkl/src/nativeTest/c/test_pkl.c
```diff
@@ -77,13 +77,28 @@ void test_empty_message() {
...
+	pkl_error_t err = { 0 };
+	pkl_exec_t *exec = NULL;
+	assert(pkl_init(test_message_handler, NULL, &exec, &err) == 0);
+
+	char *message = (char*) (unsigned char[] ) { 0x41, 0x42, 0x43 };
+	int result = pkl_send_message(exec, 3, message, &err);
+	assert(result == PKL_ERR_PROTOCOL);
+
+	result = pkl_close(NULL, &err);
+	assert(result == -1);
```
stackoverflow:
```suggestion
	result = pkl_close(exec, &err);
	assert(result == 0);
```
You should close `exec`, no? The null close is already tested by `test_null_send_message`

stackoverflow (approved):

HT154 (approved):

libpkl/src/nativeTest/c/test_pkl.c
bioball: Ah, bad copy/paste!

libpkl/src/main/c/include/pkl.h
```diff
@@ -38,34 +38,37 @@ typedef struct {
 /**
  * Pkl executor instance that manages communication with the Pkl runtime.
  *
- * Instances should be created via `pkl_init()` and destroyed via `pkl_close().`
+ * Instances should be created via `pkl_init` and destroyed via `pkl_close`.
  *
- * All operations on this struct are considered thread-safe and are synchronized via a mutex.
+ * All calls using this executor should be synchronized in the same thread.
+ * Failing to do so will possibly cause the process to be aborted with a segfault.
```
stackoverflow: Does it segfault now? Doesn't it return `PKL_ERR_THREAD`?

libpkl/src/main/c/include/pkl.h
```diff
@@ -75,25 +78,34 @@ PKL_EXPORT int pkl_init(pkl_message_response_handler handler, void *userData,
...
  * @return 0 on success, and non-zero otherwise.
  */
-PKL_EXPORT int pkl_send_message(pkl_exec_t *pexec, unsigned int length, char *message,
+PKL_EXPORT int pkl_send_message(const pkl_exec_t *pexec, unsigned int length, char *message,
 		pkl_error_t *error);
 
 /**
  * Cleans up any resources that were created as part of the `pkl_init` process
  * for our `pkl_exec_t` instance.
+*
```
stackoverflow:
```suggestion
 *
```

pkl-server/src/main/kotlin/org/pkl/server/Server.kt
```diff
@@ -232,6 +232,8 @@ class Server(private val transport: MessageTransport) : AutoCloseable {
       }
     } catch (e: IllegalArgumentException) {
       throw ProtocolException(e.message ?: "Failed to create an evalutor. $e", e)
+    } catch (e: IllegalStateException) {
+      throw ProtocolException(e.message ?: "Failed to create an evalutor. $e", e)
```
stackoverflow:
```suggestion
      throw ProtocolException(e.message ?: "Failed to create an evaluator. $e", e)
    } catch (e: IllegalStateException) {
      throw ProtocolException(e.message ?: "Failed to create an evaluator. $e", e)
```

stackoverflow (approved):

libpkl/src/main/c/include/pkl.h
bioball: Looks like `pkl.h` wasn't being formatted by spotless; fixed

libpkl/src/main/c/include/pkl.h
bioball: Good catch!

PR #1806, opened 2026-07-30
Skip loading known-bad certs

HT154:
In some cases, macOS can produce certs (used internally by some daemons) with multiple EKU extensions, which Java cannot load. This PR detects such certs and silently skips loading them.

This also skips hidden files (`.*`) from the cacerts dir when locating cert files, which avoids issues if, for example, a `.DS_Store` file ends up there (ask me how I know).

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/http/JdkHttpClient.java
+++ b/pkl-core/src/main/java/org/pkl/core/http/JdkHttpClient.java
@@ -19,6 +19,7 @@ import com.google.errorprone.annotations.ThreadSafe;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.PushbackInputStream;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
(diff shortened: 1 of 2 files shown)


pkl-core/src/main/java/org/pkl/core/http/JdkHttpClient.java
```diff
@@ -177,13 +176,27 @@ private static void collectCertificates(
...
+            }
+            break;
+          }
+          if (Objects.equals(ioExc.getMessage(), "Duplicate extensions not allowed")) continue;
+        }
+        throw new HttpClientException(
+            ErrorMessages.create("cannotParseCertFile", source, Exceptions.getRootReason(e)));
+      } finally {
+        first = false;
+      }
```
bioball, line 213: Can we have some tests that cover this behavior?

HT154: The double extension handling? I'm not entirely sure how to produce such a cert on-demand and I'm hesitant to include the real sample from a user in the repo.

bioball: Hmm... fair. There's some things that can be tested, though, e.g. the handling of empty bytes.

bioball (approved):
Approving to unblock; would be nice to have some tests

pkl-core/src/main/java/org/pkl/core/http/JdkHttpClient.java
HT154: There are already tests that cover several cases (garbage data, empty file, otherwise invalid cert) in `CliEvaluatorTest`. Only the new duplicate extension case isn't covered.

PR #1808, opened 2026-08-03
Add hex/binary/octal support to `String.toInt()`

HT154:
This also eliminates some duplicated number handling between `AstBuilder` and `StringNodes`

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
@@ -1094,7 +1094,7 @@ public final class VmUtils {
   }
 
   /** Removes `_` from numbers to be parsed. Returns the string unmodified if it's invalid. */
-  public static String removeUnderscoresFromNumber(String number, boolean isHex) {
+  public static String removeUnderscoresFromNumber(String number, boolean allowExponents) {
     if (number.indexOf('_') < 0) return number;
 
     var builder = new StringBuilder();
@@ -1108,7 +1108,7 @@ public final class VmUtils {
         return number;
       }
 
-      numberStart = c == '.' || (!isHex && (c == 'e' || c == 'E'));
+      numberStart = c == '.' || (allowExponents && (c == 'e' || c == 'E'));
     }
 
     return builder.toString();
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
@@ -1092,4 +1092,25 @@ public final class VmUtils {
     var truffleStackTraceElements = TruffleStackTrace.getStackTrace(e);
     return truffleStackTraceElements != null && truffleStackTraceElements.size() < 100;
   }
+
+  /** Removes `_` from numbers to be parsed. Returns the string unmodified if it's invalid. */
+  public static String removeUnderscoresFromNumber(String number, boolean isHex) {
+    if (number.indexOf('_') < 0) return number;
+
+    var builder = new StringBuilder();
+    var numberStart = true;
+    for (var i = 0; i < number.length(); i++) {
+      var c = number.charAt(i);
+      if (c != '_') {
+        builder.append(c);
+      } else if (numberStart) {
+        // invalid: _ at start or after [.eE]
+        return number;
+      }
+
+      numberStart = c == '.' || (!isHex && (c == 'e' || c == 'E'));
+    }
+
+    return builder.toString();
+  }
 }
--- a/pkl-core/src/main/java/org/pkl/core/stdlib/base/StringNodes.java
+++ b/pkl-core/src/main/java/org/pkl/core/stdlib/base/StringNodes.java
@@ -33,6 +33,9 @@ import org.pkl.core.util.GlobResolver;
 import org.pkl.core.util.GlobResolver.InvalidGlobPatternException;
 import org.pkl.core.util.Pair;
 import org.pkl.core.util.StringUtils;
+import org.pkl.parser.Lexer;
+import org.pkl.parser.ParserError;
+import org.pkl.parser.Token;
 
 @SuppressWarnings("unused")
 public final class StringNodes {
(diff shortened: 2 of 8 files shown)


pkl-core/src/main/java/org/pkl/core/stdlib/base/StringNodes.java
```diff
@@ -835,7 +835,8 @@ public abstract static class toInt extends ExternalMethod0Node {
     @Specialization
     protected long eval(String self) {
       try {
-        return Long.parseLong(removeUnderlinesFromNumber(self));
+        var negate = self.charAt(0) == '-';
```
bioball: This will throw if given an empty string.

Looks like we're missing this from our tests

pkl-core/src/main/java/org/pkl/core/stdlib/base/StringNodes.java
```diff
@@ -850,7 +851,8 @@ public abstract static class toIntOrNull extends ExternalMethod0Node {
     @Specialization
     protected Object eval(String self) {
       try {
-        return Long.parseLong(removeUnderlinesFromNumber(self));
+        var negate = self.charAt(0) == '-';
+        return VmUtils.parseInteger(negate ? self.substring(1) : self, negate, Long::parseLong);
```
bioball: `Long.parseLong()` accepts `-` and `+` signs (e.g. `Long.parseLong("-5")`); so this would parse `0x-5` as `-5`, `0x+5` as `5`, etc, whereas this isn't actually accepted by Pkl.

One solution is to just reject any sequence after `0x`/`0b`/etc that isn't a digit. However, that's starting to feel like a lexer. Maybe we should just use our Lexer? That would keep this in sync with our grammar.

bioball (changes requested):

pkl-core/src/test/files/LanguageSnippetTests/input/api/string.pkl
```diff
@@ -266,15 +266,56 @@ examples {
...
     module.catch(() -> "1.2".toInt())
     module.catch(() -> "9223372036854775808".toInt())
     module.catch(() -> "-9223372036854775809".toInt())
     module.catch(() -> "abc".toInt())
     module.catch(() -> "_1_000".toInt())
+    module.catch(() -> "0p0".toInt())
+    module.catch(() -> "0x-0".toInt())
+    module.catch(() -> "0x+0".toInt())
+    module.catch(() -> "0x0 + 0x1".toInt())
+    module.catch(() -> "".toInt())
```
bioball, line 338: It got pretty hard to read through this one test. Mind if we break this up?

This should do:

```pkl
  ["toInt() decimal"] {
    "123".toInt()
    "-123".toInt()
    "+123".toInt()
    "1_2__3___".toInt()
    "-1_2__3___".toInt()
    "+1_2__3___".toInt()
    "0".toInt()
    "-0".toInt()
    "+0".toInt()
    "-9223372036854775808".toInt() // math.minInt
    "9223372036854775807".toInt() // math.maxInt
  }

  ["toInt() hex"] {
    "0x123".toInt()
    "-0x123".toInt()
    "+0x123".toInt()
    "0x1_2__3___".toInt()
    "-0x1_2__3___".toInt()
    "+0x1_2__3___".toInt()
    "0x0".toInt()
    "-0x0".toInt()
    "+0x0".toInt()
  }

  ["toInt() binary"] {
    "0b101".toInt()
    "-0b101".toInt()
    "+0b101".toInt()
    "0b1_0__1___".toInt()
    "-0b1_0__1___".toInt()
    "+0b1_0__1___".toInt()
    "0b0".toInt()
    "-0b0".toInt()
    "+0b0".toInt()
  }

  ["toInt() octal"] {
    "0o123".toInt()
    "-0o123".toInt()
    "+0o123".toInt()
    "0o1_2__3___".toInt()
    "-0o1_2__3___".toInt()
    "+0o1_2__3___".toInt()
    "0o0".toInt()
    "-0o0".toInt()
    "+0o0".toInt()
  }

  ["toInt() error cases"] {
    module.catch(() -> "1.2".toInt())
    module.catch(() -> "9223372036854775808".toInt())
    module.catch(() -> "-9223372036854775809".toInt())
    module.catch(() -> "abc".toInt())
    module.catch(() -> "_1_000".toInt())
    module.catch(() -> "0p0".toInt())
    module.catch(() -> "0x-0".toInt())
    module.catch(() -> "0x+0".toInt())
    module.catch(() -> "0x0 + 0x1".toInt())
    module.catch(() -> "".toInt())
    module.catch(() -> "0b9".toInt())
    module.catch(() -> "0xz".toInt())
    module.catch(() -> "0o9".toInt())
  }
```

bioball (approved):
Just one nit, but otherwise looks good!

pkl-core/src/test/files/LanguageSnippetTests/input/api/string.pkl
```diff
@@ -263,18 +263,67 @@ examples {
     str1.codePoints
   }
 
-  ["toInt()"] {
+  ["toInt() decimal"] {
```
stackoverflow, line 266: Now that you use the lexer, whitespace is allowed. So we should also test it: ex.: `-  123`.

pkl-core/src/main/java/org/pkl/core/stdlib/base/StringNodes.java
```diff
@@ -830,12 +833,58 @@ protected String eval(String self) {
...
+        case OCT -> {
+          radix = 8;
+          text = text.substring(2); // strip "0o"
+        }
+        case HEX -> {
+          radix = 16;
+          text = text.substring(2); // strip "0x"
+        }
+        default -> throw new NumberFormatException();
+      }
```
stackoverflow, line 867: This is inconsistent with AstBuilder (so, regular pkl numbers).
`toInt` accepts uppercase radixes (because the lexer does), while regular Pkl numbers don't.
Either we change this code or we should drop this limitation from AstBuilder (which I honestly don't remember why it was there).

pkl-core/src/main/java/org/pkl/core/stdlib/base/StringNodes.java
```diff
@@ -830,12 +833,58 @@ protected String eval(String self) {
     }
   }
 
+  /** Use the lexer to parse integer values in the same forms that Pkl itself accepts */
+  private static Long toInt(String self) throws NumberFormatException {
```
stackoverflow:
```suggestion
  private static long toInt(String self) throws NumberFormatException {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
```diff
@@ -1092,4 +1092,25 @@ public static boolean isPklBug(VmStackOverflowException e) {
     var truffleStackTraceElements = TruffleStackTrace.getStackTrace(e);
     return truffleStackTraceElements != null && truffleStackTraceElements.size() < 100;
   }
+
+  /** Removes `_` from numbers to be parsed. Returns the string unmodified if it's invalid. */
+  public static String removeUnderscoresFromNumber(String number, boolean isHex) {
```
stackoverflow:
```suggestion
  public static String removeUnderscoresFromNumber(String number, boolean allowExponent) {
```
nit

pkl-parser/src/main/java/org/pkl/parser/Lexer.java
```diff
@@ -236,7 +236,7 @@ private Token nextDefault() {
           } else {
             throw unexpectedChar("..", ".", "...", "...?");
           }
-        } else if (lookahead >= 48 && lookahead <= 57) {
+        } else if (lookahead >= 48 /* '0' */ && lookahead <= 57 /* '9' */) {
```
stackoverflow: nit: all these comments are fine, but we could go all the way and change the numbers to char directly, which works fine in Java: `lookahead >= '0'`.

pkl-core/src/test/files/LanguageSnippetTests/input/api/string.pkl
```diff
@@ -263,18 +263,67 @@ examples {
...
     "-1_2__3___".toInt()
+    "+1_2__3___".toInt()
     "0".toInt()
     "-0".toInt()
+    "+0".toInt()
+    "-9223372036854775808".toInt() // math.minInt
+    "9223372036854775807".toInt() // math.maxInt
+  }
+  
+  ["toInt() hex"] {
```
stackoverflow, line 282: Missing tests for uppercase radixes (see comment above about differences from regular Pkl code).

stackoverflow, line 282: We should also test the `isHex` flag on the remove underlines function: `"0x1E_2".toInt()`

pkl-core/src/main/java/org/pkl/core/stdlib/base/StringNodes.java
HT154: I opted to drop this in `AstBuilder`

stackoverflow (approved):
LGTM

PR #1809, opened 2026-08-03
Use XDG base directories and Known Folders on Windows

bioball:
This changes logic that previously read/wrote from `~/.pkl` to use XDG base directories (all OSes), and Known Folders locations on Windows.

For example, Pkl will look for `settings.pkl` in:

1. `$XDG_CONFIG_HOME/pkl/settings.pkl`
2. `$APPDATA/pkl/settings.pkl`
4. Path `pkl/settings.pkl` within `$XDG_CONFIG_DIRS`
5. `/etc/xdg/pkl/settings.pkl`
3. `~/.pkl/settings.pkl`

The settings and cacerts dirs will still read from the legacy `~/.pkl/` directory if not found in the new locations.

See https://specifications.freedesktop.org/basedir/latest/

SPICE here: https://github.com/apple/pkl-evolution/pull/34

### DIFF
--- a/docs/modules/pkl-cli/partials/cli-common-options.adoc
+++ b/docs/modules/pkl-cli/partials/cli-common-options.adoc
@@ -43,7 +43,7 @@ The cache directory for storing packages.
 If unset, defaults to the following locations:
 
 . `$XDG_CACHE_HOME/pkl`
-. `$LOCALAPPDATA/pkl/cache` (on Windows only)
+. `$LOCALAPPDATA/pkl/Cache` (on Windows only)
 . `~/.cache/pkl` (if `$XDG_CACHE_HOME` and `$LOCALAPPDATA` are both unset)
 ====
 
--- a/docs/modules/pkl-cli/partials/cli-common-options.adoc
+++ b/docs/modules/pkl-cli/partials/cli-common-options.adoc
@@ -36,9 +36,15 @@ Possible values:
 .--cache-dir
 [%collapsible]
 ====
-Default: `~/.pkl/cache` +
 Example: `/path/to/module/cache/` +
+
 The cache directory for storing packages.
+
+If unset, defaults to the following locations:
+
+. `$XDG_CACHE_HOME/pkl`
+. `$LOCALAPPDATA/pkl/cache` (on Windows only)
+. `~/.cache/pkl` (if `$XDG_CACHE_HOME` and `$LOCALAPPDATA` are both unset)
 ====
 
 .--no-cache
@@ -97,7 +103,7 @@ Any symlinks are resolved before this check is performed.
 Default: (none) +
 Example: `mySettings.pkl` +
 File path of the Pkl settings file to use.
-If not set, `~/.pkl/settings.pkl` or defaults specified in the `pkl.settings` standard library module are used.
+If not set, `~/.config/pkl/settings.pkl` on Unix or `%APPDATA%/pkl/settings.pkl` on Windows (or the legacy `~/.pkl/settings.pkl`), or defaults specified in the `pkl.settings` standard library module are used.
 ====
 
 .-t, --timeout
--- a/docs/modules/release-notes/pages/0.33.adoc
+++ b/docs/modules/release-notes/pages/0.33.adoc
@@ -17,7 +17,7 @@ include::partial$intro.adoc[]
 
 ==== Default file locations
 
-For new setups, the CLI no longer stores anything under `~/.pkl` (https://github.com/apple/pkl/pull/1674[#1674]).
+For new setups, the CLI no longer stores anything under `~/.pkl` (https://github.com/apple/pkl/pull/1809[#1809]).
 
 It uses XDG-style locations on Unix and Known Folder locations on Windows:
 
@@ -46,7 +46,7 @@ It uses XDG-style locations on Unix and Known Folder locations on Windows:
 | none
 |===
 
-On every OS, these locations can be overriden with XDG-style env vars.
+On every OS, these locations can be overridden with XDG-style env vars.
 For example, setting `XDG_CACHE_HOME` will configure the cache directory.
 
 Note that the existing `~/.pkl/cache` directory is ignored, so Pkl will download packages to populate its cache if configured to do so.
--- a/docs/modules/release-notes/pages/0.33.adoc
+++ b/docs/modules/release-notes/pages/0.33.adoc
@@ -13,7 +13,43 @@ include::partial$intro.adoc[]
 
 == Noteworthy [small]#🎶#
 
-=== XXX
+=== CLI Changes
+
+==== Default file locations
+
+For new setups, the CLI no longer stores anything under `~/.pkl` (https://github.com/apple/pkl/pull/1674[#1674]).
+
+It uses XDG-style locations on Unix and Known Folder locations on Windows:
+
+[cols="1,2,2,2",options="header"]
+|===
+| Concern | Unix (Linux/macOS) | Windows | Legacy fallback
+
+| Package cache
+| `~/.cache/pkl`
+| `$LOCALAPPDATA/pkl/Cache`
+| none
+
+| Settings file
+| `~/.config/pkl/settings.pkl`
+| `$APPDATA/pkl/settings.pkl`
+| `~/.pkl/settings.pkl`
+
+| CA certificates
+| `~/.config/pkl/cacerts`
+| `$APPDATA/pkl/cacerts`
+| `~/.pkl/cacerts`
+
+| REPL history
+| `~/.local/state/pkl/repl-history`
+| `$LOCALAPPDATA/pkl/repl-history`
+| none
+|===
+
+On every OS, these locations can be overriden with XDG-style env vars.
+For example, setting `XDG_CACHE_HOME` will configure the cache directory.
+
+Note that the existing `~/.pkl/cache` directory is ignored, so Pkl will download packages to populate its cache if configured to do so.
 
 == Breaking Changes [small]#💔#
 
--- a/docs/modules/release-notes/pages/changelog.adoc
+++ b/docs/modules/release-notes/pages/changelog.adoc
@@ -27,6 +27,11 @@ Thank you to all the contributors for this release!
 
 xref:0.32.adoc[Release Notes]
 
+=== Changes
+
+* Default the CLI's package cache, settings file, CA certificates directory, and REPL history to XDG-style locations instead of `~/.pkl`: `~/.cache/pkl`, `~/.config/pkl/settings.pkl`, `~/.config/pkl/cacerts`, and `~/.local/state/pkl/repl-history` (pr:https://github.com/apple/pkl/pull/1674[]).
+The legacy `~/.pkl` locations are still used when they already exist, so existing setups keep working without migration.
+
 [[release-0.31.1]]
 == 0.31.1 (2026-03-26)
 
--- a/docs/modules/release-notes/pages/changelog.adoc
+++ b/docs/modules/release-notes/pages/changelog.adoc
@@ -27,11 +27,6 @@ Thank you to all the contributors for this release!
 
 xref:0.32.adoc[Release Notes]
 
-=== Changes
-
-* Default the CLI's package cache, settings file, CA certificates directory, and REPL history to XDG-style locations instead of `~/.pkl`: `~/.cache/pkl`, `~/.config/pkl/settings.pkl`, `~/.config/pkl/cacerts`, and `~/.local/state/pkl/repl-history` (pr:https://github.com/apple/pkl/pull/1674[]).
-The legacy `~/.pkl` locations are still used when they already exist, so existing setups keep working without migration.
-
 [[release-0.31.1]]
 == 0.31.1 (2026-03-26)
 
--- a/pkl-cli/src/main/kotlin/org/pkl/cli/repl/Repl.kt
+++ b/pkl-cli/src/main/kotlin/org/pkl/cli/repl/Repl.kt
@@ -63,7 +63,10 @@ internal class Repl(workingDir: Path, private val server: ReplServer, private va
         }
         completer(AggregateCompleter(CommandCompleter, FileCompleter(workingDir)))
         option(Option.DISABLE_EVENT_EXPANSION, true)
-        variable(LineReader.HISTORY_FILE, (IoUtils.getPklHomeDir().resolve("repl-history")))
+        // will be null if `$HOME` is not set. if so, don't bother writing repl history.
+        IoUtils.getReplHistoryFile()?.let { historyFile ->
+          variable(LineReader.HISTORY_FILE, historyFile)
+        }
       }
       .build()
 
@@ -80,7 +83,7 @@ internal class Repl(workingDir: Path, private val server: ReplServer, private va
 
   fun run() {
     // JLine 2 history file is incompatible with JLine 3
-    IoUtils.getPklHomeDir().resolve("repl-history.bin").deleteIfExists()
+    IoUtils.getLegacyPklHomeDir().resolve("repl-history.bin").deleteIfExists()
 
     println(ReplMessages.welcome)
     println()
--- a/pkl-cli/src/main/kotlin/org/pkl/cli/repl/Repl.kt
+++ b/pkl-cli/src/main/kotlin/org/pkl/cli/repl/Repl.kt
@@ -63,7 +63,8 @@ internal class Repl(workingDir: Path, private val server: ReplServer, private va
         }
         completer(AggregateCompleter(CommandCompleter, FileCompleter(workingDir)))
         option(Option.DISABLE_EVENT_EXPANSION, true)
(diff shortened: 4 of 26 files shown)


docs/modules/release-notes/pages/changelog.adoc
```diff
@@ -27,6 +27,11 @@ Thank you to all the contributors for this release!
 
 xref:0.32.adoc[Release Notes]
 
+=== Changes
+
+* Default the CLI's package cache, settings file, CA certificates directory, and REPL history to XDG-style locations instead of `~/.pkl`: `~/.cache/pkl`, `~/.config/pkl/settings.pkl`, `~/.config/pkl/cacerts`, and `~/.local/state/pkl/repl-history` (pr:https://github.com/apple/pkl/pull/1674[]).
```
HT154: Nit: took me a couple scans to parse "default" as a verb

```suggestion
* Use XDG-style locations by default for the CLI's package cache, settings file, CA certificates directory, and REPL history instead of `~/.pkl`: `~/.cache/pkl`, `~/.config/pkl/settings.pkl`, `~/.config/pkl/cacerts`, and `~/.local/state/pkl/repl-history` (pr:https://github.com/apple/pkl/pull/1674[]).
```

docs/modules/release-notes/pages/0.33.adoc
```diff
@@ -13,7 +13,43 @@ include::partial$intro.adoc[]
...
+It uses XDG-style locations on Unix and Known Folder locations on Windows:
+
+[cols="1,2,2,2",options="header"]
+|===
+| Concern | Unix (Linux/macOS) | Windows | Legacy fallback
+
+| Package cache
+| `~/.cache/pkl`
+| `$LOCALAPPDATA/pkl/Cache`
+| none
```
HT154, line 31: I wonder if it might make sense to have a one-time "new cache not found, but old cache is present" copy/link migration to make some attempt to reuse existing cache dirs. Cost to re-build is low, but still non-zero, especially on slow connections.

HT154 (commented):
LGTM!

HT154 (approved):

pkl-core/src/test/kotlin/org/pkl/core/util/IoUtilsTest.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2024-2025 Apple Inc. and the Pkl project authors. All rights reserved.
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
```
stackoverflow: nit: no changes in this file.

pkl-executor/src/main/java/org/pkl/executor/ExecutorOptions.java
```diff
@@ -67,7 +68,33 @@ public final class ExecutorOptions {
...
+
+  // Package-private; injectable so tests can exercise the Windows code path on a Unix CI box.
+  static Path defaultModuleCacheDir(
+      Path home, boolean isWindows, Map<String, String> environmentVariables) {
+    // Keep in sync with org.pkl.core.util.IoUtils.getSystemModuleCacheDir (pkl-executor cannot
+    // depend on pkl-core). On Unix prefer the XDG-style `~/.cache/pkl`; on Windows prefer
+    // `%LOCALAPPDATA%/pkl/cache`. Keep using a pre-existing legacy `~/.pkl/cache` so that
+    // already-populated caches aren't orphaned.
+    var xdgConfig = environmentVariables.get("XDG_CACHE_HOME");
+    if (xdgConfig != null) {
```
stackoverflow: In `BaseDirectory` you treat empty values as unset (`if (value == null || value.isEmpty())`). Here you don't.
You also don't catch `InvalidPathException` like in `BaseDirectory`, so it'll blow.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/CliCommand.kt
```diff
@@ -215,7 +215,7 @@ abstract class CliCommand(protected val cliOptions: CliBaseOptions) {
   }
 
   private fun HttpClient.Builder.addDefaultCliCertificates() {
-    val caCertsDir = IoUtils.getPklHomeDir().resolve("cacerts")
+    val caCertsDir = IoUtils.getSystemCaCertsDir()
```
stackoverflow, line 220: The change to cacerts means that the certs can now be silently diverted via an env var. I'm not sure this is a good call security-wise.

pkl-core/src/main/java/org/pkl/core/util/BaseDirectory.java
```diff
@@ -0,0 +1,185 @@
...
+    return null;
+  }
+
+  // for testing only
+  @Nullable Path firstExistingPath(String subpath, Map<String, String> envVars, boolean isWindows) {
+    var home = getHome(envVars, isWindows);
+    Path candidate;
+    if (home != null) {
+      candidate = home.resolve(subpath);
+      if (Files.exists(candidate)) {
```
stackoverflow: `File.exists` means a file named `cacerts` will return a candidate, but later we check if it's a directory, which will fail, meaning we'll use the default bundled certs even if there's a valid one somewhere else.
Worth adding another parameter or predicate to check the correct expectation.

docs/modules/release-notes/pages/0.33.adoc
```diff
@@ -13,7 +13,43 @@ include::partial$intro.adoc[]
...
+| `$LOCALAPPDATA/pkl/repl-history`
+| none
+|===
+
+On every OS, these locations can be overriden with XDG-style env vars.
+For example, setting `XDG_CACHE_HOME` will configure the cache directory.
+
+Note that the existing `~/.pkl/cache` directory is ignored, so Pkl will download packages to populate its cache if configured to do so.
 
 == Breaking Changes [small]#💔#
```
stackoverflow, line 54: We need a breaking change announcement here. If you rely on the cache on ~/.pkl/cache for air-gapped runs, your setup might be broken.

pkl-core/src/main/java/org/pkl/core/util/BaseDirectories.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.pkl.core.util;
+
+public final class BaseDirectories {
```
stackoverflow: Do we need this class and `BaseDirectory` to be public? Can't we hide them behind IoUtils?

pkl-core/src/main/java/org/pkl/core/util/BaseDirectory.java
```diff
@@ -0,0 +1,185 @@
...
+ * <p>On Windows, follows {@code $APPDATA} and {@code $LOCALAPPDATA} conventions, but can be
+ * overridden by {@code XDG} style env vars.
+ */
+public record BaseDirectory(
+    String xdgHomeEnvVar,
+    @Nullable String xdgDirsEnvVar,
+    String windowsEnvVar,
+    @Nullable String windowsSubpath,
+    String homeDefault,
+    String @Nullable [] dirsDefault) {
```
stackoverflow: nit: BaseDirectory is a record (so, immutable) but `dirsDefault` is mutable. `List<String>` would keep the class immutable and make equals/hashCode not rely on identity equality.

pkl-core/src/main/java/org/pkl/core/util/IoUtils.java
```diff
@@ -228,13 +227,33 @@ public static Path getCurrentWorkingDir() {
   }
 
   // not stored to avoid build-time initialization by native-image
-  public static Path getPklHomeDir() {
+  public static Path getLegacyPklHomeDir() {
     return Path.of(System.getProperty("user.home"), ".pkl");
   }
 
+  public static @Nullable Path getSystemModuleCacheDir() {
+    return BaseDirectories.cache.resolveHome("");
```
stackoverflow: nit: `resolveHome("")` reads like a bug. Having a BaseDirectories.home() would be better.

pkl-cli/src/main/kotlin/org/pkl/cli/repl/Repl.kt
```diff
@@ -63,7 +63,10 @@ internal class Repl(workingDir: Path, private val server: ReplServer, private va
         }
         completer(AggregateCompleter(CommandCompleter, FileCompleter(workingDir)))
         option(Option.DISABLE_EVENT_EXPANSION, true)
-        variable(LineReader.HISTORY_FILE, (IoUtils.getPklHomeDir().resolve("repl-history")))
+        // will be null if `$HOME` is not set. if so, don't bother writing repl history.
```
stackoverflow:
```suggestion
        // will be null if `user.home` is not set. if so, don't bother writing repl history.
```
nit: they are different things and it's technically possible for them to differ.

docs/modules/release-notes/pages/0.33.adoc
```diff
@@ -13,7 +13,43 @@ include::partial$intro.adoc[]
 
 == Noteworthy [small]#🎶#
 
-=== XXX
+=== CLI Changes
+
+==== Default file locations
+
+For new setups, the CLI no longer stores anything under `~/.pkl` (https://github.com/apple/pkl/pull/1674[#1674]).
```
stackoverflow: 1674 is closed. Shouldn't it point to this PR?

docs/modules/release-notes/pages/0.33.adoc
```diff
@@ -13,7 +13,43 @@ include::partial$intro.adoc[]
...
+| `$APPDATA/pkl/cacerts`
+| `~/.pkl/cacerts`
+
+| REPL history
+| `~/.local/state/pkl/repl-history`
+| `$LOCALAPPDATA/pkl/repl-history`
+| none
+|===
+
+On every OS, these locations can be overriden with XDG-style env vars.
```
stackoverflow:
```suggestion
On every OS, these locations can be overridden with XDG-style env vars.
```

stdlib/settings.pkl
```diff
@@ -17,8 +17,16 @@
...
+///
 /// Unless CLI commands and build tool plugins are explicitly configured with a settings file,
-/// they will use `~/.pkl/settings.pkl` or the defaults specified in this module.
+/// looks in the following locations in order of precedence:
+///
+/// 1. `$XDG_CONFIG_HOME/pkl/settings.pkl`
+/// 2. `$APPDATA/pkl/settings.pkl` (on Windows only)
+/// 3. `~/.config/pkl/settings.pkl`
+/// 4. Path `pkl/settings.pkl` within the `$XDG_CONFIG_DIRS` search path
+///    (dirs separted by `:` on Unix, `;` on Windows).
```
stackoverflow:
```suggestion
///    (dirs separated by `:` on Unix, `;` on Windows).
```

stdlib/settings.pkl
```diff
@@ -17,8 +17,16 @@
 /// Configuration settings for Pkl itself.
 ///
 /// Every settings file must amend this module.
+///
 /// Unless CLI commands and build tool plugins are explicitly configured with a settings file,
-/// they will use `~/.pkl/settings.pkl` or the defaults specified in this module.
+/// looks in the following locations in order of precedence:
```
stackoverflow: Wording is strange here.
What looks in the following locations?

docs/modules/pkl-cli/partials/cli-common-options.adoc
```diff
@@ -36,9 +36,15 @@ Possible values:
...
 ====
-Default: `~/.pkl/cache` +
 Example: `/path/to/module/cache/` +
+
 The cache directory for storing packages.
+
+If unset, defaults to the following locations:
+
+. `$XDG_CACHE_HOME/pkl`
+. `$LOCALAPPDATA/pkl/cache` (on Windows only)
```
stackoverflow: nit: on `0.33.adoc` you wrote `$LOCALAPPDATA/pkl/Cache`. Here and in the gradle docs you wrote in lowercase.

pkl-commons-cli/src/main/kotlin/org/pkl/commons/cli/CliCommand.kt
bioball: Is this any different from diverting the path of the `settings.pkl` via an env var? Both have security implications.

docs/modules/release-notes/pages/0.33.adoc
bioball: IMO: this isn't a breaking change. It's worth noting in our release notes, but the system cache should be treated as non-critical to a program.

Also: we don't present the system cache as an option in our blog post [_Using packages in air-gapped environments_](https://pkl-lang.org/blog/using-packages-in-air-gapped-environments.html).

pkl-core/src/main/java/org/pkl/core/util/BaseDirectories.java
bioball: Yeah, true; change this to package-internal.

pkl-core/src/main/java/org/pkl/core/util/BaseDirectory.java
bioball: We don't really care about mutable/immutable; nor equals/hashCode for this. I think I'll just rewrite this to a class so it's not confusing.

stdlib/settings.pkl
bioball: Updated this; the subject is "CLI commands and build tool plugins"

docs/modules/pkl-cli/partials/cli-common-options.adoc
bioball: Good catch!

pkl-core/src/main/java/org/pkl/core/util/BaseDirectory.java
bioball: Good point; I changed this to accept a predicate.

pkl-core/src/main/java/org/pkl/core/util/BaseDirectory.java
```diff
@@ -0,0 +1,205 @@
...
+    var userHome = System.getProperty("user.home");
+    if (userHome != null) {
+      homeDir = Path.of(userHome);
+    } else {
+      homeDir = null;
+    }
+  }
+
+  private static Path @Nullable [] getConfiguredPaths(String envVar, Map<String, String> envVars) {
+    try {
```
stackoverflow, line 112: The `try` is over the whole for loop, so a single invalid entry in XDG_CONFIG_DIRS discards every valid one.
We could catch per entry and return only the valid ones.

pkl-core/src/main/java/org/pkl/core/settings/PklSettings.java
```diff
@@ -26,32 +26,62 @@
...
-  static PklSettings loadFromPklHomeDir(Path pklHomeDir) throws VmEvalException {
-    var path = pklHomeDir.resolve("settings.pkl");
+  static PklSettings loadFromSettingsDir(Path settingsDir) throws VmEvalException {
+    return loadFromSettingsFile(settingsDir.resolve("settings.pkl"));
+  }
+
+  private static PklSettings loadFromSettingsFile(Path path) throws VmEvalException {
     return Files.exists(path)
         ? load(ModuleSource.path(path))
         : new PklSettings(Editor.SYSTEM, null);
```
stackoverflow, line 87:
```suggestion
        : PklSettings.defaultInstance;
```

stackoverflow (approved):

madrob:
Should this go through SPICE?

bioball:
> Should this go through SPICE?

Yeah, will write one

stdlib/settings.pkl
```diff
@@ -17,8 +17,17 @@
...
-/// they will use `~/.pkl/settings.pkl` or the defaults specified in this module.
+/// they look in the following locations in order of precedence:
+///
+/// 1. `$XDG_CONFIG_HOME/pkl/settings.pkl`
+/// 2. `$APPDATA/pkl/settings.pkl` (on Windows only)
+/// 3. `~/.config/pkl/settings.pkl`
+/// 4. Path `pkl/settings.pkl` within the `$XDG_CONFIG_DIRS` search path
+///    (dirs separated by `:` on Unix, `;` on Windows).
+/// 4. `/etc/xdg/pkl/settings.pkl`
+/// 5. `~/.pkl/settings.pkl` (legacy location used by Pkl 0.32 and lower)
```
HT154, line 30:
```suggestion
/// 4. Path `pkl/settings.pkl` within the `$XDG_CONFIG_DIRS` search path
///    (dirs separated by `:` on Unix, `;` on Windows).
/// 5. `/etc/xdg/pkl/settings.pkl`
/// 6. `~/.pkl/settings.pkl` (legacy location used by Pkl 0.32 and lower)
```

PR #1811, opened 2026-08-05
Throw when an object key `default` would become a `Dynamic` property

### LINKED ISSUE #561: pkl:json.Parser should throw an error when `useMapping = false` and key "default" is encountered

adityasingh2400:
Fixes #561

`Dynamic` declares a hidden `default` property, so an object key named `default` cannot be represented as a `Dynamic` property. `json.Parser.parse()`, `yaml.Parser.parse()` and `yaml.Parser.parseAll()` silently dropped the key instead, and reading `.default` on the result then failed with a confusing type error about `Function1`. These methods now throw, and the message points at `useMapping = true`, which parses into a `Mapping` and preserves the key.

This is a redo of #700, which carried approvals from bioball, stackoverflow and holzensp before its own author closed it and deleted the head ref. I kept the error wording that was settled on that thread. One deliberate difference: SnakeYAML's `BaseConstructor` rethrows `YamlEngineException` unchanged but wraps every other `RuntimeException`, so the new `org.pkl.core.util.yaml.ParseException` is unwrapped from the cause rather than caught directly. Catching it directly is what made the rendered hint carry a `org.pkl.core.util.yaml.ParseException:` prefix in the old thread.

Verified by restoring the three changed sources to `main` and rerunning: four snippet tests fail, and both new error snippets reproduce the reported bug verbatim, with the `default` key simply absent from the output. With the change, `:pkl-core:test` is 1583 tests with 0 failures and spotless is clean. I grepped the repo for YAML and JSON fixtures containing a `default` key and for `.pkl` files calling either parser outside `pkl-core/src/test` and found none, so no other module reaches the new error path. Native-image and the multi-JDK matrix were not run locally.

One thing for maintainers to weigh: bioball flagged this as a breaking change on the original thread. I added a release note under 0.33 Breaking Changes, but if it needs to wait for a major version window that is entirely reasonable and I am happy to retarget it.

### DIFF
--- a/docs/modules/release-notes/pages/0.33.adoc
+++ b/docs/modules/release-notes/pages/0.33.adoc
@@ -23,6 +23,20 @@ Native `pkl` and `pkldoc` binaries for Intel Mac systems are no longer published
 
 To continue running new Pkl releases on these systems, use an appropriate Java runtime and the `jpkl` and `jpkldoc` Java executables.
 
+=== JSON and YAML Parsers Reject the Object Key `default`
+
+`Dynamic` declares a hidden `default` property, so an object key named `default` could not be represented as a `Dynamic` property.
+Previously, `json.Parser.parse()`, `yaml.Parser.parse()` and `yaml.Parser.parseAll()` silently produced an object that was missing the key, and reading `.default` on it failed with a confusing type error.
+
+These methods now throw, and the error message points at `useMapping = true`, which parses into a `Mapping` and preserves the key:
+
+[source,pkl]
+----
+new json.Parser { useMapping = true }.parse("""
+  {"default": "greeting"}
+  """)
+----
+
 === XXX
 
 == Bug Fixes [small]#🐜#
--- a/pkl-core/src/main/java/org/pkl/core/stdlib/json/ParserNodes.java
+++ b/pkl-core/src/main/java/org/pkl/core/stdlib/json/ParserNodes.java
@@ -28,6 +28,7 @@ import org.pkl.core.runtime.*;
 import org.pkl.core.stdlib.ExternalMethod1Node;
 import org.pkl.core.stdlib.PklConverter;
 import org.pkl.core.util.EconomicMaps;
+import org.pkl.core.util.ErrorMessages;
 import org.pkl.core.util.LateInit;
 import org.pkl.core.util.json.JsonHandler;
 import org.pkl.core.util.json.JsonParser;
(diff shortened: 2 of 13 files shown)


adityasingh2400:
Flagging a red I saw on my own fork's CI in case it shows up here once the workflows are approved to run.

The `libpkl-alpine-linux-amd64-snapshot` job fails in GraalVM native-image with `exit status 30`. The log ends with "Image generator watchdog is aborting image generation" and memory stats showing 18MB free of a 12GB heap, so the image build was killed by the deadlock watchdog after running out of headroom rather than hitting a compile error.

That job is unrelated to this change, which is a parser-level check on object keys named `default` and does not affect native image generation. Mentioning it only so it is not mistaken for a real failure of this PR.

adityasingh2400:
No checks run on this PR here, so flagging the build result from my fork since it is the only CI signal.

The full `Build` workflow is green on `150ef0e`: 16 jobs pass, 3 macOS aarch64 jobs skip, and `gradle-check` passes with the new `jsonParserDefaultProperty` and `yamlParserDefaultProperty` snippet tests.

`libpkl-alpine-linux-amd64-snapshot` was red on the first attempt and passed unchanged on a re-run, so that was transient. Every other `libpkl-*` target and every other `alpine-*` target had already passed on the first attempt, which is why I re-ran rather than treating it as a real break.

PR #1812, opened 2026-08-05
Split libpkl libraries between shared and static libs

bioball:
This adds a `libpkl-static.pc` so pkg-config users can link either statically or dynamically via `pkg-config --libs libpkl`, or `pkg-config --libs libpkl-static`.

Additionally, this fixes an issue where the replace tokens wasn't doing anything, leaving the `@version@` token intact in the archive.

[native-libpkl]

### DIFF
--- a/docs/modules/libpkl/pages/index.adoc
+++ b/docs/modules/libpkl/pages/index.adoc
@@ -1,4 +1,5 @@
 = libpkl
+include::ROOT:partial$component-attributes.adoc[]
 
 libpkl is a C library for driving Pkl evaluation.
 
@@ -32,12 +33,10 @@ The archive has the following structure:
     ├── include
     │   └── pkl.h
     ├── lib
-    |   └── shared
-    |   |   └── libpkl.<shared library extension>
-    |   └── static
-    |   |   └── libpkl.<static library extension>
+    │   ├── libpkl.<shared library extension>
+    │   ├── libpkl.<static library extension>
     │   └── pkgconfig
-    |       ├── libpkl-static.pc
+    │       ├── libpkl-static.pc
     │       └── libpkl.pc
     ├── LICENSE.txt
     ├── README.md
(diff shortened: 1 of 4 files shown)


docs/modules/libpkl/pages/index.adoc
```diff
@@ -32,9 +33,10 @@ The archive has the following structure:
     ├── include
     │   └── pkl.h
     ├── lib
+    │   ├── libpkl.<shared library extension>
     │   ├── libpkl.<static library extension>
-    │   ├── libpkl.<dynamic library extension>
```
bioball, line 36: Unrelated: standardize on "shared library" instead of "dynamic library"

docs/modules/libpkl/pages/index.adoc
```diff
@@ -58,6 +60,44 @@ The file extensions are as follows:
...
+----
+# Link against the shared library
+cc -o myprogram myprogram.c $(pkg-config --libs --cflags libpkl)
+
+# Link against the static library
+cc -o myprogram myprogram.c $(pkg-config --static --libs --cflags libpkl-static)
+----
+
+
+If libpkl is installed in a non-system directory, set `PKG_CONFIG_PATH` to the subpath `lib/pkgconfig` within directory that contains libpkl. For example, given libpkl directory `/path/to/libpkl`, the environment variable should have `/path/to/libpkl/lib/pkgconfig`.
```
stackoverflow:
```suggestion
If libpkl is installed in a non-system directory, set `PKG_CONFIG_PATH` to the subpath `lib/pkgconfig` within the directory that contains libpkl. For example, given libpkl directory `/path/to/libpkl`, the environment variable should have `/path/to/libpkl/lib/pkgconfig`.
```

docs/modules/libpkl/pages/index.adoc
```diff
@@ -58,6 +60,44 @@ The file extensions are as follows:
...
+
+NOTE: Windows can use pkg-config through the https://www.msys2.org/docs/pkgconfig/[MSYS2 toolchain]. Using pkg-config on Windows implies compiling with gcc-style compiler like https://code.visualstudio.com/docs/cpp/config-mingw[MinGW], instead of MSVC).
```
stackoverflow:
```suggestion
NOTE: Windows can use pkg-config through the https://www.msys2.org/docs/pkgconfig/[MSYS2 toolchain]. Using pkg-config on Windows implies compiling with gcc-style compiler like https://code.visualstudio.com/docs/cpp/config-mingw[MinGW], instead of MSVC.
```

docs/modules/libpkl/pages/index.adoc
stackoverflow:
```suggestion
|OS |Static Library Extension |Shared library extension
```

stackoverflow (approved):
LGTM

PR #1816, opened 2026-08-07
Fix reflect crash on modules using glob imports

### LINKED ISSUE #1815: reflect.Module(m).imports throws ClassCastException when m uses a glob import
## Context

I was using `pkl:reflect` to walk the dependency graph of a set of Pkl
modules — for each module, following `supermodule` and `imports` to build
the transitive closure of everything it is built from.

This works fine until any module in the graph declares a glob import.
Reading `imports` off such a module fails with a raw
`java.lang.ClassCastException`.

**Impact:** the failure cannot be caught or guarded from within Pkl, and it
surfaces as an internal Java error rather than a Pkl diagnostic. Any code
that reflects over modules it does not control has to defensively pre-scan
module *source text* for `import*` and skip those modules — which both
misses data and is obviously fragile.

## Steps to reproduce

`sub/x.pkl`
```pkl
x = 1
```

`globby.pkl`
```pkl
import* "sub/*.pkl" as all
count = all.length
```

`main.pkl`
```pkl
import "pkl:reflect"
import "globby.pkl" as globby
imports = reflect.Module(globby).imports
```

```console
$ pkl eval main.pkl
```

## Error message

```

java.lang.ClassCastException%3A%20org.pkl.core.ast.expression.unary.ImportGlobNode%20cannot%20be%20cast%20to%20org.pkl.core.ast.expression.unary.ImportNode

–– Pkl Error ––
org.pkl.core.ast.expression.unary.ImportGlobNode cannot be cast to org.pkl.core.ast.expression.unary.ImportNode

3 | imports = reflect.Module(globby).imports
at main#imports (file:///tmp/pklbug/main.pkl, line 3)
```

## Expected

`imports` returns the module's imports, with the glob import mapped to the
URI of its glob pattern.

## Notes

- `pkl eval globby.pkl` succeeds (`count = 1`), so the glob import itself is
  fine — only reflecting over the module fails.
- With a plain `import` instead, reflection works as expected:
  `imports = Map("one", "file:///tmp/pklbug/sub/x.pkl")`.
- Root cause: `ObjectMember.isImport()` is true for both `import` and
  `import*`, but `VmTyped.getImports()` casts the member's body node to
  `ImportNode`, while a glob import's body is an `ImportGlobNode`. Both
  extend `AbstractImportNode`, which is where `getImportUri()` is declared.
- `ModuleInfo`, which builds the equivalent map for the Java `ModuleSchema`
  API, already casts to `AbstractImportNode` — so the two paths disagree
  today.
- Possibly related but distinct: #166 (NullPointerException, glob import
  reached through a package-relative module import).

I have a fix and will open a PR shortly.

## Version

```
$ pkl --version
Pkl 0.31.1 (macOS 26.2, native)
```

Also present on `main` (c5896d76) — `VmTyped.java:118`.

sergiosalvi:
Fixes #1815.

`reflect.Module(m).imports` threw a raw `java.lang.ClassCastException`
whenever `m` declared a glob import:

```
org.pkl.core.ast.expression.unary.ImportGlobNode cannot be cast to
org.pkl.core.ast.expression.unary.ImportNode
```

`ObjectMember.isImport()` is true for both `import` and `import*`, but
`VmTyped.getImports()` cast the member's body node to `ImportNode`, and a
glob import's body is an `ImportGlobNode`. This casts to their common
supertype `AbstractImportNode` instead, which is where `getImportUri()` is
declared.

`ModuleInfo` — which builds the equivalent map for the Java `ModuleSchema`
API — already casts to `AbstractImportNode`, so this brings the two paths
back in agreement.

For a glob import the mapped value is the URI of the glob pattern itself;
`stdlib/reflect.pkl` is updated to say so.

### Testing

Added `LanguageSnippetTests` case `api/reflect6`, next to the existing
`reflect5` which guards an earlier fix to this same method. It reflects over
a module declaring a glob import and asserts both that `imports` resolves
and that the glob import still works:

```
res = Map("reflect", "pkl:reflect", "globbed", "file:///$snippetsDir/input-helper/globtest/child/*.pkl")
globbedKeys {
  "../../input-helper/globtest/child/moduleC.pkl"
}
```

Verified `./gradlew build` and `./gradlew :pkl-cli:buildNative` both succeed,
and confirmed against the reproduction from #1815 using the built native
binary:

```console
$ pkl-macos-aarch64 eval main.pkl
imports = Map("all", "file:///private/tmp/pklbug/sub/*.pkl")
```

### DIFF
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmTyped.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmTyped.java
@@ -26,7 +26,7 @@ import org.pkl.core.Composite;
 import org.pkl.core.PModule;
 import org.pkl.core.PObject;
 import org.pkl.core.ast.ExpressionNode;
-import org.pkl.core.ast.expression.unary.AbstractImportNode;
+import org.pkl.core.ast.expression.unary.ImportNode;
 import org.pkl.core.ast.member.ObjectMember;
 import org.pkl.core.util.EconomicMaps;
 
@@ -115,8 +115,11 @@ public final class VmTyped extends VmObject {
         if (bodyNode instanceof WrapperNode wrapper) {
           bodyNode = (ExpressionNode) wrapper.getDelegateNode();
         }
-        builder.add(
-            member.getName().toString(), ((AbstractImportNode) bodyNode).getImportUri().toString());
+        // Glob imports (`import*`) don't import a type; they resolve to a mapping.
+        // Exclude them here, consistent with `imports` also not covering import expressions.
+        if (bodyNode instanceof ImportNode importNode) {
+          builder.add(member.getName().toString(), importNode.getImportUri().toString());
+        }
       }
     }
     return builder.build();
--- a/pkl-core/src/main/java/org/pkl/core/runtime/VmTyped.java
+++ b/pkl-core/src/main/java/org/pkl/core/runtime/VmTyped.java
@@ -26,7 +26,7 @@ import org.pkl.core.Composite;
 import org.pkl.core.PModule;
 import org.pkl.core.PObject;
 import org.pkl.core.ast.ExpressionNode;
-import org.pkl.core.ast.expression.unary.ImportNode;
+import org.pkl.core.ast.expression.unary.AbstractImportNode;
 import org.pkl.core.ast.member.ObjectMember;
 import org.pkl.core.util.EconomicMaps;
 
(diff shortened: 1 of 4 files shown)


bioball (changes requested):
Thanks for the PR!

I think it'd be better to just exclude glob imports from the `imports` property. 

* Glob imports are quite different from regular imports (they don't import a type, they result in a mapping).
* `imports` also doesn't cover import expressions (also don't import a type)

If we want to convey glob import information, might be better to introduce a new `globImports` property in the future.

sergiosalvi:
@bioball thanks, that makes sense. I've updated the PR to exclude glob imports from `imports` entirely rather than reporting the glob pattern URI. `getImports()` now skips `import*` members (they resolve to a mapping, not a module type), consistent with `imports` also not covering import expressions. I updated the doc comment to match, and the `reflect6` regression test still verifies that reflecting a module using `import*` no longer throws `ClassCastException` (with the glob import now absent from the result). Happy to add a separate `globImports` property in a follow-up if that's useful.

HT154 (approved):
LGTM

bioball (approved):

PR #1817, opened 2026-08-08
Add withNetrc support for evaluator settings

ayashjorden:
Add native `.netrc` (and Windows `_netrc`) parsing support to Pkl evaluator settings via `withNetrc` and `withNetrcFile` standard library functions in `pkl:EvaluatorSettings` and `pkl:settings`.

Closes / References https://github.com/apple/pkl/discussions/1767
  
### Changes

* Added `org.pkl.core.util.Netrc` to parse `.netrc` files (supporting `machine`, `default`, `login`/`user`, `password`/`passwd`, `account`, comments, quotes, and `macdef`).
* Enforces strict POSIX file permissions (`0600` / `0400`) before reading credentials on POSIX filesystems.
* Generates `Authorization` headers (Basic & Bearer) for target hosts.
* Added `withNetrc` and `withNetrcFile` to `pkl:EvaluatorSettings` and `pkl:settings`.


stdlib/EvaluatorSettings.pkl
```diff
@@ -205,6 +205,60 @@ external local function resolvePath(
...
+function withNetrcFile(http: Http?, netrcFile: String?): Http =
+  let (netrcHeaders = readNetrcHeaders(netrcFile))
+    if (netrcHeaders == null)
+      http ?? new Http {}
+    else
+      new Http {
+        proxy = http?.proxy
+        rewrites = http?.rewrites
+        headers = mergeHeaders(http?.headers, netrcHeaders)
+      }
```
bioball: Thinking about this some more, I think the API should look like this:

```pkl
hidden const netrcHttp: Http

external const function netrcHttp(contents: String | Resource): Http
```

On the usage side, users can simply set it:

```pkl
// PklProject

evaluatorSettings {
  http = netrcHttp
}
```

Or amend it if they want to add more headers:

```pkl
evaluatorSettings {
  http = (netrcHttp) {
    ["**"] {
      ["User-Agent"] = "Hello"
    }
  }
}
```

This means that we don't need to write `mergeHeaders` logic.

bioball: Implementation would look something like:

```pkl
hidden const netrcHttp: Http =
  let (home = read?("env:HOME"))
    if (home == null)
      new {}
    else
      let (
        netRcContents =
          if (platform.current.operatingSystem.name == "Windows")
            read?("\(home)/_netrc") ?? read?("\(home)/.netrc")
          else
            read?("\(home)/.netrc")
      )
        if (netRcContents == null)
          new {}
        else
          netrcHttp(netRcContents)
```

And the `external const function netrcHttp()` would be implemented as a truffle node in Java.

This means that we don't have file permissions checks, but I'm not sure that it's Pkl's responsibility to check that. It's a good thing for users to set, but other tools like `curl` don't do a file permissions check either.

pkl-core/src/main/java/org/pkl/core/stdlib/evaluatorsettings/EvaluatorSettingsNodes.java
```diff
@@ -52,4 +64,88 @@ protected String eval(VmTyped ignored, String uriStr, String path, boolean forWi
...
+      if (Files.notExists(path)) {
+        if (isExplicitPath) {
+          throw exceptionBuilder()
+              .evalError("netrcFileNotFound", path.toAbsolutePath().toString())
+              .build();
+        }
+        return VmNull.withoutDefault();
+      }
+
+      Netrc.checkPermissions(path);
```
bioball: Although this is definitely the best practice for _users_ to set these permissions, I don't know if we need to enforce this on the Pkl side?

I can't figure out what wget does. Curl [does not complain either](https://curl.se/docs/manpage.html#--netrc).

pkl-core/src/main/java/org/pkl/core/stdlib/evaluatorsettings/EvaluatorSettingsNodes.java
```diff
@@ -52,4 +64,88 @@ protected String eval(VmTyped ignored, String uriStr, String path, boolean forWi
...
+      } else {
+        String envNetrc = System.getenv("NETRC");
+        if (envNetrc != null && !envNetrc.isBlank()) {
+          path = Path.of(envNetrc);
+          isExplicitPath = true;
+        } else {
+          String userHome = System.getProperty("user.home", "");
+          boolean isWindows = IoUtils.isWindows();
+          if (isWindows) {
+            String userProfile = System.getenv("USERPROFILE");
```
bioball: Do we need to look in `USERPROFILE`?

As far as I can tell, everything else that supports netrc on Windows just looks in `HOME`.

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,230 @@
...
+  public static Map<String, Map<String, List<String>>> toHeadersMap(List<Entry> entries) {
+    Map<String, Map<String, List<String>>> result = new LinkedHashMap<>();
+    for (Entry entry : entries) {
+      String authHeaderValue = computeAuthHeaderValue(entry.login(), entry.password());
+      if (authHeaderValue == null) {
+        continue;
+      }
+      Map<String, List<String>> headerMap = Map.of("Authorization", List.of(authHeaderValue));
+      if (entry.machine().equals("default") || entry.machine().equals("*")) {
+        result.put("**", headerMap);
```
bioball: Couple comments:

1. I don't know of any netrc files that accepts `machine *`. Is there prior art for this?
2. Setting `**` as the URL pattern here will add the default headers to _every request_, because Pkl combines headers. I'd consider this a credential leak (adding a default login shouldn't add that login for a more specific machine).

Pkl doesn't really have a good way to say "don't combine these headers".
To be on the safe side, it's probably better to just ignore if we find a default login.
Additionally, it's a security best practice to avoid setting defaults in a netrc file, so it's not too bad to ignore methinks.

pkl-core/src/test/kotlin/org/pkl/core/util/NetrcTest.kt
```diff
@@ -0,0 +1,92 @@
...
+import org.assertj.core.api.Assertions.assertThatThrownBy
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.io.TempDir
+import org.pkl.commons.writeString
+import org.pkl.core.runtime.VmEvalException
+
+class NetrcTest {
+
+  @Test
+  fun `parse netrc file content`() {
```
bioball, line 26: Do you mind adding some more tests for the parser code? It looks like you have a lot of this stuff covered in your parser already, but we only have this one test right now.

Some tests in mind:

* Quotes
* Trailing comments after a line 
    - `password foo # some comment`
    - `password foo#some comment`
* Hash character inside quote (e.g. `login "foo # bar"`)
* Tabs as whitespace separators
* CRLF line endings
* Entry with login but no password
* Entry with password but not login
* Duplicate `machine`s
* machine that collides with keyword (e.g. `machine machine`)
* files with only comments
* skipped macdef
* Empty quotes `e.g. login ""`
* Unterminated quote on a single line (e.g. `login "`)
* Entry that's only one line (e.g. `machine foo.com login bar password pass1`)

It'd be good to add these as language snippet tests, rather than Kotlin tests.

To do that, create a new file (e.g. `pkl-core/src/test/files/LanguageSnippetTests/input/api/evaluatorSettingsModuleNetRc.pkl`), and define things as a Pkl test.

Here's an example, assuming that there is a `netrcHttp` method:

```pkl
amends "../snippetTest.pkl"

import "pkl:EvaluatorSettings"

examples {
  ["netrcHttp - quotes"] {
    EvaluatorSettings.netrcHttp("""
      machine foo.com
        login "bar bar"
        password baz
      """)
  }
}
```

You'd then run `LanguageSnippetTests` to update and generate expected output.
It's also really useful to use the internal IntelliJ plugin when writing language snippet tests: https://github.com/apple/pkl-project-commons#internal-intellij-plugin

pkl-core/src/main/java/org/pkl/core/stdlib/evaluatorsettings/EvaluatorSettingsNodes.java
```diff
@@ -52,4 +64,88 @@ protected String eval(VmTyped ignored, String uriStr, String path, boolean forWi
...
+        var innerBuilder = VmMap.builder();
+        for (var headerEntry : innerMap.entrySet()) {
+          var headerName = headerEntry.getKey();
+          var headerValuesList = headerEntry.getValue();
+
+          var listBuilder = VmList.EMPTY.builder();
+          for (String val : headerValuesList) {
+            listBuilder.add(val);
+          }
+          innerBuilder.add(headerName, listBuilder.build().toListing());
```
bioball:
```suggestion
          innerBuilder.add(headerName, VmList.create(headerValuesList).toListing());
```

bioball (changes requested):
Thanks for the PR, this looks like a great start!

ayashjorden:
@bioball 
I have refactored the `.netrc` implementation based on the review feedback:

1. **API Redesign**: Replaced `withNetrc`/`withNetrcFile` with `EvaluatorSettings.netrcHttp(contents: String | Resource): Http`. `settings.pkl` now sets `http` to `EvaluatorSettings.netrcHttp` by default.
2. **Security & Parser Hardening**:
   - Removed POSIX permission checks to align with tools like `curl`/`git`.
   - Excluded `default` and `*` machines from `toHeadersMap` to avoid leaking credentials via wildcard matching.
   - Added support for quoted tokens, comments, CRLF endings, and skipping `macdef` blocks.
3. **Runtime & Singleton Caching**: Registered `pkl:EvaluatorSettings` as a cached static singleton module (`EvaluatorSettingsModule` in `ModuleCache`), avoiding module version conflicts when imported by `pkl:settings`, and registered the `file:` resource reader for settings evaluation.
4. **Tests**: Added tests for quoted strings, comments, duplicate machines, and language snippet verification in `evaluatorSettingsModuleNetRc.pkl`. All tests and formatting checks (`./gradlew check`) pass.

stdlib/EvaluatorSettings.pkl
ayashjorden: Done

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Removed

pkl-core/src/test/kotlin/org/pkl/core/util/NetrcTest.kt
ayashjorden: Done

pkl-core/src/main/java/org/pkl/core/stdlib/evaluatorsettings/EvaluatorSettingsNodes.java
ayashjorden: Removed

stdlib/settings.pkl
```diff
@@ -28,8 +28,10 @@ import "pkl:EvaluatorSettings"
 editor: Editor = System
 
 /// Settings for controlling how Pkl makes HTTP(S) requests.
+///
+/// By default, credentials are populated from a `.netrc` file if available.
 @Since { version = "0.26.0" }
-http: EvaluatorSettings.Http?
+http: EvaluatorSettings.Http = EvaluatorSettings.netrcHttp
```
bioball: I don't think we want this to be the default; using a netrc should be opt-in rather than opt-out.

ayashjorden:
@bioball gentle ping

bioball:
I'll get back to you on this next week!

If you'd like, feel free to join our [community call](https://github.com/apple/pkl/discussions/1164) to chat about the feature and we can do a design review then too.

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,206 @@
...
+  public record Entry(
+      String machine,
+      @Nullable String login,
+      @Nullable String password,
+      @Nullable String account) {}
+
+  /** Parses the content of a .netrc file into a list of [Entry]. */
+  public static List<Entry> parse(String content) {
+    List<String> tokens = tokenize(content);
+    List<Entry> entries = new ArrayList<>();
```
bioball: [nit] here, and everywhere else: style-wise, we prefer `var` initializers wherever possible.

```suggestion
    var tokens = tokenize(content);
    var entries = new ArrayList<Entry>();
```

It's not always possible (like `String currentMachine = null;`)

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,206 @@
...
+public final class Netrc {
+  private Netrc() {}
+
+  public record Entry(
+      String machine,
+      @Nullable String login,
+      @Nullable String password,
+      @Nullable String account) {}
+
+  /** Parses the content of a .netrc file into a list of [Entry]. */
```
bioball:
```suggestion
  /** Parses the content of a .netrc file into a list of {@link Entry}. */
```

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,206 @@
...
+      if (entry.machine().equals("default") || entry.machine().equals("*")) {
+        continue;
+      }
+      String authHeaderValue = computeAuthHeaderValue(entry.login(), entry.password());
+      if (authHeaderValue == null) {
+        continue;
+      }
+      Map<String, List<String>> headerMap = Map.of("Authorization", List.of(authHeaderValue));
+      result.putIfAbsent("https://" + entry.machine() + "/**", headerMap);
+      result.putIfAbsent("http://" + entry.machine() + "/**", headerMap);
```
bioball: Two comments:
1. We just need one pattern; `http{,s}://` will match both `http:` and `https:`.
2. We need to escape any characters that would be interpreted as glob pattern wildcards:

```suggestion
      var headerMap = Map.of("Authorization", List.of(authHeaderValue));
      result.putIfAbsent("http{,s}://" + escapeGlobPattern(entry.machine()) + "/**", headerMap);
    }
    return result;
  }

  private static String escapeGlobPattern(String value) {
    var sb = new StringBuilder();
    for (var i = 0; i < value.length(); i++) {
      var c = value.charAt(i);
      if (c == '?' || c == '*' || c == '[' || c == '{' || c == '\\') {
        sb.append('[').append(c).append(']');
      }
      sb.append(c);
    }
    return sb.toString();
  }
```

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  private static @Nullable String computeAuthHeaderValue(
+      @Nullable String login, @Nullable String password) {
+    if (password == null && login == null) {
+      return null;
+    }
+    if (login != null && login.equalsIgnoreCase("bearer")) {
+      return "Bearer " + (password == null ? "" : password);
+    }
```
bioball: This doesn't seem right; it's possible that someone's username is actually "bearer".

bioball:
```suggestion
```

This makes it impossible to use "bearer" as the username. I also can't find any other tool that has the logic of: "if login is 'bearer'; use password as bearer token"

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,206 @@
...
+
+      int len = line.length();
+      for (int i = 0; i < len; i++) {
+        char c = line.charAt(i);
+        if (inQuotes) {
+          if (c == '"') {
+            inQuotes = false;
+          } else {
+            sb.append(c);
+          }
```
bioball: I think let's follow curl here and accept escape sequences in quotes; see https://everything.curl.dev/usingcurl/netrc.html#the-netrc-file-format

I played around with it, and see the following:

| Line | Decoded auth header |
| --- | --- |
| `login "foo` | Parser error |
| `login "foo\"` | Parser error |
| `login "foo \` | Parser error |
| `login "foo bar"` | `foo bar:` |
| `login "foo \z"` | `foo z:` |
| `login "foo\"foo"` | `foo"foo:` |
| `login "foo""bar"` | `foo:` (the second quoted text is discarded) |
| `"login" foo` | `foo:` (keywords can be quoted) |

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  /**
+   * Converts a list of [Entry] into a map of host glob pattern to header map (header name to list
+   * of header values).
+   */
+  public static Map<String, Map<String, List<String>>> toHeadersMap(List<Entry> entries) {
+    Map<String, Map<String, List<String>>> result = new LinkedHashMap<>();
+    for (Entry entry : entries) {
+      if (entry.machine().equals("default") || entry.machine().equals("*")) {
```
bioball: I can't find any other tool that treats `machine *` as a wildcard. Therefore, let's just treat it like a verbatim hostname.

Also: we need to omit any machines that contain `/`. This is because it would otherwise be parsed as part of the URL path.
I think other tools out there with netrc support have logic like `if url.host == machine.name`; which would always fail if the machine's name contains `/`.
In Pkl's case, because these are glob patterns, we need to just exclude this from the entry set.

```suggestion
      if (entry.machine().equals("default") || entry.machine().contains("/")) {
```

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,206 @@
...
+            currentAccount = null;
+          }
+          currentMachine = "default";
+        }
+        case "login", "user" -> {
+          if (i < tokens.size()) {
+            currentLogin = tokens.get(i++);
+          }
+        }
+        case "password", "passwd" -> {
```
bioball: I can't find any tool that accepts `passwd`; where does this come from?

```suggestion
        case "password" -> {
```

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,206 @@
...
+              }
+              tokens.add(token);
+            }
+          } else if (c == '#') {
+            if (!sb.isEmpty() || sawQuotes) {
+              String token = sb.toString();
+              sb.setLength(0);
+              sawQuotes = false;
+              if (token.equals("macdef")) {
+                inMacdef = true;
```
bioball: This would prevent "macdef" from being a username, password, or host (e.g. `login macdef`)

I think the tokenize method shouldn't really be the one that cares about macdefs, since this is designed to just produce (grammar-free) tokens.

`tokenize` can produce "null line" as a token (two consecutive newlines as defined by [netrc(5)](https://linux.die.net/man/5/netrc)); and the parser can just skip until EOF or the next null line.

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,206 @@
...
+          }
+        }
+        case "default" -> {
+          if (currentMachine != null) {
+            entries.add(new Entry(currentMachine, currentLogin, currentPassword, currentAccount));
+            currentLogin = null;
+            currentPassword = null;
+            currentAccount = null;
+          }
+          currentMachine = "default";
```
bioball, line 85: This clashes with entry like `machine default`; which is meaningfully different (one is hostname "default", one is mean to match on all hosts).

I think you can resolve this by just having record `Entry` have field `boolean isDefault`

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,206 @@
...
+    int i = 0;
+    while (i < tokens.size()) {
+      String token = tokens.get(i++);
+      switch (token) {
+        case "machine" -> {
+          if (currentMachine != null) {
+            entries.add(new Entry(currentMachine, currentLogin, currentPassword, currentAccount));
+            currentLogin = null;
+            currentPassword = null;
+            currentAccount = null;
```
bioball: Edge case: if the last token is just `machine`; don't produce an entry for it

```suggestion
            currentAccount = null;
            currentMachine = null;
```

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,206 @@
...
+        if (token.equals("macdef")) {
+          inMacdef = true;
+        } else {
+          tokens.add(token);
+        }
+      }
+    }
+    return tokens;
+  }
+}
```
bioball, line 278: This is missing some error handling (see comment above WRT what `curl` does)

stdlib/EvaluatorSettings.pkl
```diff
@@ -205,6 +205,36 @@ external local function resolvePath(
...
+          else if (platform.current.operatingSystem.name == "Windows")
+            read?("\(home.text)/_netrc") ?? read?("\(home.text)/.netrc")
+          else
+            read?("\(home.text)/.netrc")
+  )
+    if (netRcContents == null)
+      new {}
+    else
+      netrcHttp(netRcContents)
+
```
bioball: We (the Pkl team) chatted about this a little bit, and we're leaning towards not providing a helper like this for now.

The problem is that we'd like to change how `read()` works; instead of accepting an arbitrary expression, to accept a string constant. So, we'd rather not introduce stdlib implementation that has a dynamic read.

I think let's just have a helper method that turns netrc contents into HTTP headers.

For ergonomics, we think it'd be better to put the method inside class `Http`, e.g.

```pkl
class Http {
  /// Returns parsed headers from the given netrc file contents.
  ///
  /// Example:
  /// ```
  /// headers = netRcHeaders(read("/path/to/.netrc"))
  /// ```
  @Since { version = "0.33.0" }
  external function netRcHeaders(
    contents: String | Resource,
  ): Mapping<String(isGlobPattern), HttpHeaders>
}
```

This is nice because the method is available to the scope within `http { ... }` block; no qualified access or import needed.

pkl-core/src/main/java/org/pkl/core/runtime/StdLibModule.java
```diff
@@ -44,11 +43,10 @@ private static void doLoad(URI uri, VmTyped instance) {
...
               var resourceReaders =
-                  isPklBaseModule
-                      // needed when initializing `pkl:base` because of
-                      // `read("prop:pkl.outputFormat")`
-                      ? List.of(ResourceReaders.externalProperty())
-                      : List.<ResourceReader>of();
+                  List.of(
+                      ResourceReaders.environmentVariable(),
+                      ResourceReaders.externalProperty(),
+                      ResourceReaders.file());
```
bioball: Undo this change (see comment WRT removing the netrc property)

bioball (changes requested):
Some more feedback!

ayashjorden:
> I'll get back to you on this next week!
> 
> If you'd like, feel free to join our [community call](https://github.com/apple/pkl/discussions/1164) to chat about the feature and we can do a design review then too.

I have a work conflict at that time slot. I'll try to think how can we meet online (this feature is on my own personal time).
Will post an update to the PR in a little bit.

stdlib/settings.pkl
ayashjorden: Reverted

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Done

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Done

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Fixed

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Removed

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Fixed and updated tests

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Done

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Resolved

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Done

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Done

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Done

stdlib/EvaluatorSettings.pkl
ayashjorden: Done

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Done

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,275 @@
...
+
+      if (c == '\n') {
+        consecutiveNewlines++;
+        if (consecutiveNewlines >= 2) {
+          tokens.add(NULL_LINE);
+          consecutiveNewlines = 0;
+        }
+        i++;
+        continue;
+      }
```
bioball: Should handle two consecutive CRLF's too

```suggestion
      if (c == '\n' || c == '\r' && i < len - 1 && content.charAt(i + 1) == '\n') {
        consecutiveNewlines++;
        if (consecutiveNewlines >= 2) {
          tokens.add(NULL_LINE);
          consecutiveNewlines = 0;
        }
        if (c == '\r') {
          i++;
        } else {
          i += 2;
        }
        continue;
      }
```

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,275 @@
...
+          }
+        }
+        continue;
+      }
+
+      // Unquoted token
+      var sb = new StringBuilder();
+      while (i < len) {
+        var uc = content.charAt(i);
+        if (Character.isWhitespace(uc) || uc == ',' || uc == '#') {
```
bioball: Two comments:
1.  I can't find any other netrc parser that treats `,` as a separator
2. Java's `Character.isWhitespace()` is more broad than other parser whitespace; we should probably just adhere to what's accepted in other parsers.

```suggestion
        if (isWhitespace(uc) || uc == '#') {
```

This implies this method:

```java
  private static boolean isWhitespace(char c) {
    return c == ' ' || c == '\n' || c == '\f' || c == '\t' || c == '\r' || c == '\u000B';
  }
```

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,275 @@
...
+          break;
+        }
+        sb.append(uc);
+        i++;
+      }
+      tokens.add(sb.toString());
+    }
+    return tokens;
+  }
+}
```
bioball: Couple issues with this impl right now:

1. I don't know of any other netrc parsers that treat `,` as a separator?
2. It looks like `curl` actually only considers lines that start with `#` (allowing leading whitespace) to 
3. The "consecutive quoted tokens" thing was actually incorrect (my bad); it looks like curl's behavior is actually: the parser just discards any tokens that it's not expecting. So `login foo bar baz password qux` should turn into credentials `foo:qux`. The tokenizer doesn't actually need to care about consecutive quoted tokens; the parse method will just ignore unexpected tokens anyways.
4. `Character.isWhitespace()` is much broader than what curl uses, which is just `\n`, `\t`, `\r`, ` `.
5. This method is kind of long; would be nice to split it up
6. Inside quoted strings, `\n`, `\t`, and `\r` should turn into their respective control characters
7. Currently is incorrectly handling two consecutive CRLF
8. No need to keep a counter for consecutive newlines; should be good enough to know if we're already at a newline

I played around with this locally and came up with this, to address all the issues:

```suggestion
  public static List<String> tokenize(String content) {
    var tokens = new ArrayList<String>();
    var len = content.length();
    var i = 0;
    var atLineStart = true;

    while (i < len) {
      var c = content.charAt(i);

      if (c == '\n' || c == '\r' && i < len - 1 && content.charAt(i + 1) == '\n') {
        if (atLineStart) {
          tokens.add(NULL_LINE);
        }
        atLineStart = true;
        if (c == '\r') {
          i += 2;
        } else {
          i++;
        }
        continue;
      } else if (isWhitespace(c)) {
        i++;
        continue;
      }

      // lines starting with `#` (after any leading blanks) are treated as comments
      if (c == '#' && atLineStart) {
        // Skip comment until end of line
        i = consumeLineComment(i, content, len);
        continue;
      }

      // Non-whitespace character encountered
      atLineStart = false;
      if (c == '"') {
        i = consumeQuotedToken(i, content, len, tokens);
      } else {
        i = consumeUnquotedToken(i, content, len, tokens);
      }
    }
    return tokens;
  }

  private static int consumeLineComment(int i, String content, int len) {
    while (i < len && content.charAt(i) != '\n') {
      i++;
    }
    return i;
  }

  private static int consumeQuotedToken(int i, String content, int len, List<String> tokens) {
    i++; // skip opening quote
    var escape = false;
    var sb = new StringBuilder();
    while (i < len) {
      var ch = content.charAt(i);
      if (escape) {
        var escapedChar = switch (ch) {
          case 'n' -> '\n';
          case 't' -> '\t';
          case 'r' -> '\r';
          default -> ch;
        };
        sb.append(escapedChar);
        escape = false;
        i++;
        continue;
      }
      switch (ch) {
        case '"': {
          tokens.add(sb.toString());
          return i + 1;
        }
        case '\\': {
          escape = true;
          break;
        }
        default: {
          sb.append(ch);
        }
      }
      i++;
    }
    var reason = escape ? "invalid escape" : "unclosed quote";
    throw new VmExceptionBuilder().evalError("cannotParseNetrc", reason).build();
  }

  private static int consumeUnquotedToken(int i, String content, int len, List<String> tokens) {
    var sb = new StringBuilder();
    while (i < len) {
      var ch = content.charAt(i);
      if (isWhitespace(ch)) {
        tokens.add(sb.toString());
        return i;
      }
      sb.append(ch);
      i++;
    }
    tokens.add(sb.toString());
    return i;
  }

  private static boolean isWhitespace(char c) {
    return c == ' ' || c == '\n' || c == '\t' || c == '\r';
  }
```

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,275 @@
...
+    String currentMachine = null;
+    var isDefault = false;
+    String currentLogin = null;
+    String currentPassword = null;
+    String currentAccount = null;
+
+    var i = 0;
+    while (i < tokens.size()) {
+      var token = tokens.get(i++);
+      switch (token) {
```
bioball: I'm seeing that curl does a case-insensitive match for keywords:

```suggestion
      switch (token.toLowerCase(Locale.ROOT)) {
```

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,275 @@
...
+          if (i < tokens.size()) {
+            i++; // skip macro name
+          }
+          while (i < tokens.size() && !tokens.get(i).equals(NULL_LINE)) {
+            i++;
+          }
+          if (i < tokens.size() && tokens.get(i).equals(NULL_LINE)) {
+            i++;
+          }
+        }
```
bioball, line 111:
```suggestion
        case "macdef" -> {
          while (i < tokens.size()) {
            if (tokens.get(i).equals(NULL_LINE)) {
              i++;
              break;
            }
            i++;
          }
        }
```

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,275 @@
...
+        case NULL_LINE -> {
+          // Ignore null lines outside macdef
+        }
+      }
```
bioball: Superfluous condition; the block here is empty anyways

```suggestion
```

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,275 @@
...
+            currentMachine = null;
+            isDefault = false;
+            currentLogin = null;
+            currentPassword = null;
+            currentAccount = null;
+          }
+          if (i < tokens.size()) {
+            currentMachine = tokens.get(i++);
+            isDefault = false;
+          }
```
bioball, line 69:
```suggestion
        case "machine" -> {
          if (currentMachine != null) {
            entries.add(
                new Entry(
                    currentMachine, isDefault, currentLogin, currentPassword, currentAccount));
          }
          isDefault = false;
          currentLogin = null;
          currentPassword = null;
          currentAccount = null;
          currentMachine = i < tokens.size() ? tokens.get(i++) : null;
        }
```

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,275 @@
...
+                    currentMachine, isDefault, currentLogin, currentPassword, currentAccount));
+            currentMachine = null;
+            isDefault = false;
+            currentLogin = null;
+            currentPassword = null;
+            currentAccount = null;
+          }
+          currentMachine = "default";
+          isDefault = true;
+        }
```
bioball, line 87:
```suggestion
        case "default" -> {
          if (currentMachine != null) {
            entries.add(
                new Entry(
                    currentMachine, isDefault, currentLogin, currentPassword, currentAccount));
          }
          currentLogin = null;
          currentPassword = null;
          currentAccount = null;
          currentMachine = "default";
          isDefault = true;
        }
```

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,275 @@
...
+        case "login", "user" -> {
```
bioball: I think given our stance of "we'll just match curl", let's not accept "user" as a keyword.

```suggestion
        case "login" -> {
```

bioball (changes requested):
This is pretty close!

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Fixed

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Fixed

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Fixed

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Removed

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Fixed

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Updated

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Applied your suggestion

ayashjorden:
> This is pretty close!

Applied your suggestions.
Might be free this Wednesday@9am PT (I think you have the call this week.)

pkl-core/src/main/java/org/pkl/core/runtime/ModuleCache.java
```diff
@@ -94,6 +94,8 @@ public synchronized VmTyped getOrLoad(
           return BenchmarkModule.getModule();
         case "Command":
           return CommandModule.getModule();
+        case "EvaluatorSettings":
+          return EvaluatorSettingsModule.getModule();
```
HT154: We generally try to avoid making stdlib modules singletons when possible. If this PR works without this change (and `EvaluatorSettingsModule`), then that would be preferable.

pkl-core/src/main/java/org/pkl/core/stdlib/evaluatorsettings/HttpNodes.java
```diff
@@ -0,0 +1,77 @@
...
+    @TruffleBoundary
+    protected VmMapping eval(
+        VmTyped self, VmTyped resource, @Cached("create()") IndirectCallNode callNode) {
+      var text = (String) VmUtils.readMember(resource, Identifier.TEXT, callNode);
+      return doParse(text);
+    }
+
+    private VmMapping doParse(String text) {
+      List<Netrc.Entry> entries = Netrc.parse(text);
+      Map<String, Map<String, List<String>>> headersMap = Netrc.toHeadersMap(entries);
```
HT154:
```suggestion
      var entries = Netrc.parse(text);
      var headersMap = Netrc.toHeadersMap(entries);
```

pkl-core/src/main/java/org/pkl/core/stdlib/evaluatorsettings/HttpNodes.java
```diff
@@ -0,0 +1,77 @@
...
+          var headerName = headerEntry.getKey();
+          var headerValuesList = headerEntry.getValue();
+
+          innerBuilder.add(headerName, VmList.create(headerValuesList).toListing());
+        }
+
+        outerBuilder.add(globPattern, innerBuilder.build().toMapping());
+      }
+
+      return outerBuilder.build().toMapping();
```
HT154: Since you're building a `Mapping`, it's better to use `VmObjectBuilder` to produce one directly instead of creating a `Map` then paying to convert to a `Mapping`. You can see an example of how to do this by checking the source of `VmMap.toMapping()`.

pkl-core/src/main/java/org/pkl/core/runtime/ModuleCache.java
ayashjorden: Removed

pkl-core/src/main/java/org/pkl/core/stdlib/evaluatorsettings/HttpNodes.java
ayashjorden: Done

pkl-core/src/main/java/org/pkl/core/stdlib/evaluatorsettings/HttpNodes.java
ayashjorden: Done

ayashjorden:
@HT154 Thank you for your comments, all were addressed.

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,269 @@
...
+          break;
+        }
+        default: {
+          sb.append(ch);
+        }
+      }
+      i++;
+    }
+    var reason = escape ? "invalid escape" : "unclosed quote";
+    throw new VmExceptionBuilder().evalError("cannotParseNetrc", reason).build();
```
stackoverflow: `VmExceptionBuilder` should be used from inside a truffle node (via `exceptionBuilder()`, otherwise the error will not have a source section attached.

pkl-core/src/main/java/org/pkl/core/PClassInfo.java
```diff
@@ -38,6 +38,7 @@ public final class PClassInfo<T> implements Serializable {
   public static final URI pklBaseUri = URI.create("pkl:base");
   public static final URI pklSemverUri = URI.create("pkl:semver");
   public static final URI pklSettingsUri = URI.create("pkl:settings");
+  public static final URI pklEvaluatorSettingsUri = URI.create("pkl:EvaluatorSettings");
```
stackoverflow:
```suggestion
```
This is not used anywhere.

stackoverflow (commented):
Looks good. Just some comments.

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
bioball: You can also propagate in a `Node` as an argument, for example, see https://github.com/apple/pkl/blob/7f5ae2a712481df36f9fd42dc83814f1635b0644/pkl-core/src/main/java/org/pkl/core/runtime/ResourceManager.java#L61-L70

Another pattern that we have is for this to throw its own exception type (e.g. `NetrcException` or something). You'd then catch that exception in the truffle node, and throw `exceptionBuilder().adhocEvalError(...).build()` there.
For reference, see how we do this with GlobResolver: https://github.com/apple/pkl/blob/7f5ae2a712481df36f9fd42dc83814f1635b0644/pkl-core/src/main/java/org/pkl/core/util/GlobResolver.java#L43

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
```diff
@@ -0,0 +1,269 @@
...
+      var authHeaderValue = computeAuthHeaderValue(entry.login(), entry.password());
+      if (authHeaderValue == null) {
+        continue;
+      }
+      var headerMap = Map.of("Authorization", List.of(authHeaderValue));
+      if (entry.isDefault()) {
+        result.putIfAbsent("**", headerMap);
+      } else if (entry.machine() != null && !entry.machine().contains("/")) {
+        result.putIfAbsent("http{,s}://" + escapeGlobPattern(entry.machine()) + "/**", headerMap);
+      }
```
HT154, line 138: As discussed [here](https://github.com/apple/pkl/pull/1817#discussion_r3754216847), we would prefer to discard the `default` machine to avoid unintentional credential leaks.


```suggestion
      if (entry.isDefault()) continue;
      result.putIfAbsent("http{,s}://" + escapeGlobPattern(entry.machine()) + "/**", headerMap);
```

ayashjorden: Just to make sure, I did keep the null-check.
If the test `netRcHeaders - machine with slash omitted` is invalid, I'll also remove the `!entry.machine().contains("/")`. But that test was added by request from @bioball to handle accidental "foo.bar/zaz" entries in `.netrc`.

pkl-core/src/main/java/org/pkl/core/PClassInfo.java
ayashjorden: Done

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
ayashjorden: Done

ayashjorden:
@HT154 , added a reply, to make sure I understand your review.

pkl-core/src/main/java/org/pkl/core/util/Netrc.java
HT154: Ah my bad. You got my intention 👍

HT154 (approved):

bioball (approved):

PR #1823, opened 2026-08-14
Add a CPU profiler

### LINKED ISSUE #1800: CPU/memory profiler
To help debug performance issues, there should be a way to profile CPU/memory usage when running a Pkl program.

bioball:
This introduces a way to run CPU profiling.

CLI users can enable CPU profiling via command line flags.

Examples:

```bash
# Write pprof output to `pkl-profile.pb.gz`
pkl eval --profile-cpu-output pkl.pb.gz myModule.pkl

# Write flamegraph output to `pkl-profile.html`
pkl eval --profile-cpu-output pkl-profile.html --profile-cpu-output-format flamegraph myModule.pkl
```

This uses GraalVM's `CPUSampler` underneath the hood to collect all profiling data.

The profiling API, for now, is only offered via CLI flags. This is because of the internals of how Pkl evaluation works: Profiling uses the truffle instrumentation API, which are associated with individual `Engine` instances.
However, our `Engine` is a singleton; this is because Pkl modules are also singletons, and using another engine here would introduce errors like "Invalid sharing of AST Nodes".
So, we can't really offer profiling as an API on a per-`Evaluator` level.

Closes #1800


pkl-core/src/main/java/org/pkl/core/EvaluatorBuilder.java
```diff
@@ -481,6 +483,16 @@ public boolean getPowerAssertionsEnabled() {
     return powerAssertionsEnabled;
   }
 
+  /**
+   * Configures Pkl profiling.
+   *
+   * <p>Requires the {@code org.pkl-lang:pkl-profiler} dependency to be on the classpath.
+   */
+  public EvaluatorBuilder setPklProfilerOptions(PklProfilerOptions pklProfilerOptions) {
```
HT154: Is there a possibility that we add "non-Pkl" profiler options in the future? Why clarify in the API name?

pkl-core/src/main/java/org/pkl/core/ProfilerOptions.java
```diff
@@ -13,17 +13,14 @@
...
+import org.jspecify.annotations.Nullable;
 
-public class LibPklLogger implements Logger {
-
-  @Override
-  public void log(String msg) {
-    if (Objects.equals(System.getenv("PKL_DEBUG"), "1")) {
-      System.err.println("[libpkl] " + msg);
-    }
+public record ProfilerOptions(Path outputPath, @Nullable Duration sampleDuration) {
```
HT154: Is this actually used? Or left over from an earlier iteration?

pkl-profiler/src/main/java/org/pkl/profiler/ProfilerOutput.java
```diff
@@ -0,0 +1,40 @@
...
+  private final Path outputFile;
+
+  public ProfilerOutput(PklProfilerData data, Path outputFile) {
+    this.data = data;
+    this.outputFile = outputFile;
+  }
+
+  protected abstract void doWriteOutput(PklProfilerData data, Path outputFile) throws Exception;
+
+  public void write() {
```
HT154: Should this be `final`?

pkl-profiler/src/main/java/org/pkl/profiler/PklProfiler.java
```diff
@@ -0,0 +1,134 @@
...
+  @Override
+  protected void onFinalize(Env env) {
+    var sampler = cpuSampler;
+    if (sampler == null) return;
+    var data = new PklProfilerData(startTime, periodMs, sampler.getDataList());
+    var output =
+        switch (outputFormat) {
+          case PPROF -> new PProfOutput(data, outputFile);
+          case FLAMEGRAPH -> new FlamegraphOutput(data, outputFile);
+        };
```
HT154: Thoughts on using service providers here instead of a fixed enum in case JVM clients want to build their own output?

HT154 (commented):
Awesome!! Took an initial pass on the code, still need to sit down and play around with this.

pkl-core/src/main/java/org/pkl/core/EvaluatorBuilder.java
bioball: This is now out-dated code; turns out we can't really enable profiling on a per-evaluator basis (because the polyglot engine needs to be a singleton given how our stdlib modules are singletons)

pkl-profiler/src/main/java/org/pkl/profiler/ProfilerOutput.java
bioball: Yup!

pkl-profiler/src/main/java/org/pkl/profiler/PklProfiler.java
bioball: I feel like it's probably better to keep the API footprint small for now; pprof output _should_ give you everything you need. I actually removed the flamegraph output here, too.

pkl-cli/src/test/kotlin/org/pkl/cli/ProfilerOutputTest.kt
```diff
@@ -0,0 +1,83 @@
...
+import kotlin.io.path.inputStream
+import kotlin.io.path.writeText
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.io.TempDir
+
+/**
+ * Exercises the CPU profiler end-to-end by running the packaged executable jar as a subprocess.
+ *
+ * This must run out-of-process: enabling the profiler flips global JVM state (system properties
```
HT154, line 31:
```suggestion
 * This must run out-of-process: enabling the profiler flips global JVM state (system properties)
```

pkl-cli/src/test/kotlin/org/pkl/cli/ProfilerOutputTest.kt
```diff
@@ -0,0 +1,83 @@
...
+          "1",
+          sourceFile.absolutePathString(),
+        )
+        .redirectErrorStream(true)
+        .start()
+    val output = process.inputStream.bufferedReader().readText()
+    val exitCode = process.waitFor()
+    assertThat(exitCode).withFailMessage { "process failed with output:\n$output" }.isZero()
+
+    // sanity check only
```
HT154: Nit: word choice
```suggestion
    // smoke test only
```

pkl-cli/src/main/kotlin/org/pkl/cli/commands/EvalCommand.kt
```diff
@@ -81,6 +85,24 @@ class EvalCommand : ModulesCommand(name = "eval", helpLink = helpLink) {
...
+      .path()
+
+  private val profileCpuSamplePeriod: Long by
+    option(
+        names = arrayOf("--profile-cpu-sample-period"),
+        help = "Duration, in milliseconds, for the CPU profiler to poll for samples.",
+      )
+      .single()
+      .long()
+      .default(ProfilerOptions.DEFAULT.cpu.samplePeriod)
```
HT154: Can we move these to an `OptionGroup`?

Can we also consider putting these on `run` and/or `test` as well? Seems like it would be useful there too.

pkl-core/src/main/java/org/pkl/core/ProfilerOptions.java
```diff
@@ -0,0 +1,70 @@
...
+  private static <T> void writeProperty(String name, @Nullable T prop, Function<T, String> mapper) {
+    if (prop == null) return;
+    System.setProperty(name, mapper.apply(prop));
+  }
+
+  /** Convenience method for configuring Pkl with profiler options. */
+  public void configureSystemProperties() {
+    // if profiling is not enabled, don't bother writing any system properties
+    if (cpu.outputFile == null) return;
+    writeProperty(CPU_OUTPUT_FILE, cpu.outputFile, Path::toString);
```
HT154, line 67: Something in here seems to be problematic for relative paths with no `/`:
```console
$ jpkl eval ... --profile-cpu-output=1.pb.gz
...
[pkl-profiler] WARN: failed to write profile to 1.pb.gz: Cannot invoke "java.nio.file.Path.getFileSystem()" because "path" is null
```

Using `--profile-cpu-output=./1.pb.gz` works fine.

Maybe we need to write the resolved absolute path to the property instead.

pkl-cli/src/main/kotlin/org/pkl/cli/commands/EvalCommand.kt
bioball: Yeah, fair!

PR #1824, opened 2026-08-15
Add inference for let bindings

HT154:
Depends on #1768


pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
stackoverflow, line 1: I'm seeing some issues with nested `let`s:
```pkl
class Foo { y: Int }
res = let (x: Foo = let (a = 1) new { y = a }) x.y
```
Before:
```bash
–– Pkl Error ––
Cannot tell which parent to amend.

2 | res = let (x: Foo = let (a = 1) new { y = a }) x.y
                                    ^^^
```
Now:
```bash
–– Pkl Error ––
Cannot instantiate type ``.

2 | res = let (x: Foo = let (a = 1) new { y = a }) x.y
                                    ^^^
```

Same issue here:
```
class Foo { y: Int }
class Baz { z: Int }
res = let (x: Foo = let (a: Baz = new { z = 2 }) new { y = a.z }) x.y
```
Before we got "Cannot tell which parent to amend.", now it's "Cannot find property `y` in object of type `foo#Baz`."

stackoverflow, line 1: Underscore inference might lead to an internal error:
```pkl
res = let (_ = new { y = 1 }) 1
```
Results now in
```bash

java.lang.ArrayIndexOutOfBoundsException%3A%20Index%20-1%20out%20of%20bounds%20for%20length%200

–– Pkl Error ––
Index -1 out of bounds for length 0

1 | res = let (_ = new { y = 1 }) 1
    ^^^
```

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinLetBindingNode.java
```diff
@@ -0,0 +1,73 @@
...
+  public Object executeGeneric(VirtualFrame frame) {
+    if (inferredParent != null) return inferredParent;
+
+    // remaining code only runs first time this node is executed
+    // (assuming evaluation isn't continued despite errors)
+    // except when binding type is a non-final self type (not cacheable)
+
+    CompilerDirectives.transferToInterpreterAndInvalidate();
+
+    // >> Keep in sync with GetParentForTypeNode.executeGeneric
```
stackoverflow: Worth creating a static helper so we don't duplicate this code, or forget to update both places.
Or at least adding this comment on `GetParentForTypeNode.executeGeneric` if we are not DRYing it.

pkl-core/src/test/files/LanguageSnippetTests/input/basic/letTyped.pkl
stackoverflow, line 1: We should have more tests.
- All the examples I gave above (untyped, underscore, etc.)
- Non-instantiable binding to verify the `checkIsInstantiable`:
```pkl
abstract class Bar { y: Int }
let (x: Bar = new { y = 1 }) x.y
```
- Type with no default: `let (x: Int = new {}) x`
- Maybe also a self type and constrained type

stackoverflow (changes requested):

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinLetBindingNode.java
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright © 2024-2026 Apple Inc. and the Pkl project authors. All rights reserved.
```
stackoverflow: New file should be only 2026.

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinLetBindingNode.java
HT154: I'm going to explore this in #1768

pkl-core/src/test/files/LanguageSnippetTests/input/basic/letTyped.pkl
```diff
@@ -10,26 +14,75 @@ res2 =
...
+
+// test inference with let in binding expression
+
+res10 =
+  let (x: Foo = let (z = 0) new { y = z })
+    x.y
+
+// test inference against underscore binding
+
+res11 = test.catch(() -> let (_ = new { y = 1 }) null)
```
stackoverflow: This is inconsistent with the rest of the language: `foo = new { y = 1 }` will give you a dynamic, not an error.
Though the current state is already a big improvement. We could fix it in a further PR. But it'd be better to fix it here.

pkl-core/src/test/files/LanguageSnippetTests/input/basic/letTyped.pkl
```diff
@@ -10,26 +14,75 @@ res2 =
...
+  )
+
+// test basic `new {}` inference
+
+class Foo {
+  y: Int
+}
+
+res9 =
+  let (x: Foo = new { y = 1 })
```
stackoverflow, line 56: nit: it would be good to also support parenthesized expressions:
```pkl
let (x: Foo = (new { y = 1 }))
```
```
–– Pkl Error ––
Cannot tell which parent to amend.
```

HT154: Changing this also affects other scenarios where the parent is inferred, which I think is fine.

stackoverflow (approved):

pkl-core/src/main/java/org/pkl/core/ast/expression/binary/LetExprNode.java
```diff
@@ -27,29 +25,32 @@
 import org.pkl.core.runtime.VmException;
 import org.pkl.core.runtime.VmUtils;
 
-@NodeChild(value = "bindingNode", type = ExpressionNode.class)
-public abstract class LetExprNode extends ExpressionNode {
+public final class LetExprNode extends ExpressionNode {
```
bioball: Was this turned into a regular class because you want to expose `bindingNode`? 

If so: to get a handle on the binding node, you can just add a getter for it (e.g. `public abstract ExpressionNode getBindingNode()`), and Truffle will implement this when it generates the "Gen" class.

However, we don't really need to use a `@Specialization` here, so, your call whether you want to revert this back into using the Truffle DSL or not.

pkl-core/src/main/java/org/pkl/core/ast/expression/binary/LetExprNode.java
bioball: Even in the `slot == -1` case, we should cache and invalidate (don't create a new Truffle node every time this is called; insert this new truffle node s.t. when it's executed it is part of the tree)

```suggestion
    if (typeNode == null) {
      CompilerDirectives.transferToInterpreterAndInvalidate();
      if (slot == -1) {
        typeNode = new TypeNode.UnknownTypeNode(VmUtils.unavailableSourceSection());
      } else if (unresolvedTypeNode != null) {
        typeNode = unresolvedTypeNode.execute(frame);
      } else {
        typeNode = new TypeNode.UnknownTypeNode(VmUtils.unavailableSourceSection());
      }
```
```

File: pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinLetBindingNode.java [hunk 6]
```diff
@@ -0,0 +1,70 @@
...
+import org.pkl.core.ast.expression.binary.LetExprNode;
+import org.pkl.core.ast.type.TypeNode;
+import org.pkl.core.runtime.VmLanguage;
+
+public abstract class InferParentWithinLetBindingNode extends AbstractInferParentNode {
+  public InferParentWithinLetBindingNode(SourceSection sourceSection, VmLanguage language) {
+    super(sourceSection, language);
+  }
+
+  protected LetExprNode getLetNode(VirtualFrame frame) {
```
bioball: This method doesn't need this parameter
```suggestion
  protected LetExprNode getLetNode() {
```

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinLetBindingNode.java
```diff
@@ -0,0 +1,70 @@
...
+      @Bind("letNode.getTypeNode(frame)") @SuppressWarnings("unused") TypeNode typeNode,
+      @Cached(
+              value =
+                  "getDefaultValue(frame, typeNode, letNode.getSourceSection(), letNode.getQualifiedName())",
+              neverDefault = true)
+          Object defaultValue) {
+    return defaultValue;
+  }
+
+  @Specialization(replaces = "evalCached")
```
bioball: The `replaces` clause is redundant, because the type node will always either be final or not final; it can't change between successive executions of this node.

```suggestion
  @Specialization
```

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
bioball:
```suggestion
                  sourceSection, language, scopeName, new GetOwnerNode())
```

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -1114,25 +1114,20 @@ private ExpressionNode doVisitNewExprWithInferredParent(NewExpr expr) {
               : InferParentWithinMethodNodeGen.create(
                   createSourceSection(expr.newSpan()), language, scopeName, new GetOwnerNode());
```
bioball:
```suggestion
                  sourceSection, language, scopeName, new GetOwnerNode());
```

pkl-core/src/main/java/org/pkl/core/ast/expression/binary/LetExprNode.java
```diff
@@ -65,8 +66,17 @@ private TypeNode getTypeNode(VirtualFrame frame) {
     return typeNode;
   }
```
bioball, line 67: Even in the `slot == -1` case, we should cache and insert the type node:
  - Don't create a new truffle node every time this is called
  - Ensure the type node is part of the truffle tree; no orphaned node

```suggestion
  public TypeNode getTypeNode(VirtualFrame frame) {
    if (typeNode == null) {
      CompilerDirectives.transferToInterpreterAndInvalidate();
      if (slot == -1) {
        typeNode = new TypeNode.UnknownTypeNode(VmUtils.unavailableSourceSection());
      } else if (unresolvedTypeNode != null) {
        typeNode = unresolvedTypeNode.execute(frame);
      } else {
        typeNode = new TypeNode.UnknownTypeNode(VmUtils.unavailableSourceSection());
      }
      if (slot >= 0) {
        typeNode.initWriteSlotNode(slot);
        frame.getFrameDescriptor().setSlotKind(slot, typeNode.getFrameSlotKind());
      }
      insert(typeNode);
    }
    assert typeNode != null;
    return typeNode;
  }
```

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinLetBindingNode.java
```diff
@@ -0,0 +1,70 @@
...
+        : "AstBuilder created an InferParentWithinLetBindingNode outside of a let binding";
+    return letNode;
+  }
+
+  // keep specializations in sync with other AbstractInferParentNode subclasses
+
+  @Specialization(guards = {"typeNode.isFinalType()"})
+  protected final Object evalCached(
+      @SuppressWarnings("unused") VirtualFrame frame,
+      @Bind("getLetNode(frame)") LetExprNode letNode,
```
bioball: We can cache this (and share it) to avoid paying for this lookup on the next eval

```suggestion
      @Cached(value = "getLetNode()", neverDefault = true, adopt = false)
          @Shared
          @SuppressWarnings("unused")
          LetExprNode letNode,
```

pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinLetBindingNode.java
```diff
@@ -0,0 +1,70 @@
...
+                  "getDefaultValue(frame, typeNode, letNode.getSourceSection(), letNode.getQualifiedName())",
+              neverDefault = true)
+          Object defaultValue) {
+    return defaultValue;
+  }
+
+  @Specialization(replaces = "evalCached")
+  protected final Object eval(
+      VirtualFrame frame,
+      @Bind("getLetNode(frame)") LetExprNode letNode,
```
bioball: See above WRT cacheing the let node

```suggestion
      @Cached(value = "getLetNode()", neverDefault = true, adopt = false) @Shared
          LetExprNode letNode,
```

bioball (changes requested):

pkl-core/src/main/java/org/pkl/core/ast/expression/binary/LetExprNode.java
```diff
@@ -36,7 +36,7 @@ public abstract class LetExprNode extends ExpressionNode {
   private @Child @Nullable TypeNode typeNode;
   private final int slot;
 
-  protected LetExprNode(
+  public LetExprNode(
```
bioball:
```suggestion
  protected LetExprNode(
```

bioball (approved):
Just one minor suggestion, but otherwise LGTM!

PR #1826, opened 2026-08-18
Fix name resolution of types in constraints

bioball:
This fixes an issue where type names are resolved incorrectly in typealias constraints.


pkl-core/src/main/java/org/pkl/core/ast/member/TypeAliasNode.java
```diff
@@ -87,14 +87,14 @@ public Object executeGeneric(VirtualFrame frame) {
             frame.materialize());
 
     VmUtils.evaluateAnnotations(frame, annotationNodes, annotations);
-    var bodyTypeNode = typeAnnotationNode.execute(frame);
```
bioball, line 90: Type name resolution happens when `typeAnnotationNode.execute()` is called, so we need to move the replacement logic to here.

pkl-core/src/main/java/org/pkl/core/ast/expression/primary/GetModuleOwnerNode.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+package org.pkl.core.ast.expression.primary;
+
+import com.oracle.truffle.api.CompilerDirectives;
+import com.oracle.truffle.api.frame.VirtualFrame;
+import com.oracle.truffle.api.source.SourceSection;
+import org.pkl.core.ast.ExpressionNode;
+import org.pkl.core.runtime.VmUtils;
+
+public class GetModuleOwnerNode extends ExpressionNode {
```
stackoverflow:
```suggestion
public final class GetModuleOwnerNode extends ExpressionNode {
```
nit

stackoverflow (approved):
LGTM

HT154 (approved):

HT154 (approved):

PR #1828, opened 2026-08-21
Fix block comment at EOF without trailing newline

### LINKED ISSUE #1827: Block comment at end of file without trailing newline fails with "Unexpected end of file."
### Context

A `/* ... */` block comment that is the **last content in a file and has no trailing newline** is incorrectly rejected as an unterminated comment. Adding a trailing newline makes the exact same source parse fine.

This affects any Pkl file whose final line is a block comment with no terminating newline (e.g. files produced by tools/editors that don't append one). The workaround — add a newline — is trivial once you know the cause, but the error points at the end of the file and gives no hint that a missing newline is the difference.

### `pkl --version`

Reproduces on the latest release (**0.32.1**) and on current `main` (0.33.0-SNAPSHOT). The faulty check has been present since **0.28.0**.

### Error message

```
–– Pkl Error ––
Unexpected end of file.
```

### Steps to reproduce

`printf` is used so the trailing newline is controlled precisely:

```console
# No trailing newline -> FAILS
$ printf 'hi = "Dan"\n/*\nbar\n*/' > no-newline.pkl
$ pkl eval no-newline.pkl
–– Pkl Error ––
Unexpected end of file.

# Identical source + trailing newline -> OK
$ printf 'hi = "Dan"\n/*\nbar\n*/\n' > with-newline.pkl
$ pkl eval with-newline.pkl
hi = "Dan"
```

It is not specific to having code before the comment: a file containing only `/*\nbar\n*/` (no trailing newline), or even just `/**/`, fails the same way.

### Root cause

In `Lexer.lexBlockComment()` (`pkl-parser`), after the closing `*/` is consumed via `nextChar()`, `lookahead` becomes `EOF` when the `/` is the final character of the file. The method then unconditionally runs `if (lookahead == EOF) throw unexpectedEndOfFile();`, so a correctly terminated comment is misreported as unterminated. The check cannot distinguish "found `*/` exactly at EOF" from "reached EOF without finding `*/`".

I have a fix and regression tests ready and will open a PR.

sergiosalvi:
## What

A `/* ... */` block comment that is the last content in a file **with no trailing newline** was rejected as unterminated and failed with `Unexpected end of file.` The identical source with a trailing newline parsed fine.

This is not specific to having code before the comment — a file containing only `/*\nbar\n*/` (no trailing newline), or even just `/**/`, failed the same way. It reproduces on the latest release (0.32.1) and on `main`.

## Why

In `Lexer.lexBlockComment()`, after the closing `*/` is consumed, `lookahead` becomes `EOF` when `/` is the final character of the file. The trailing `if (lookahead == EOF) throw unexpectedEndOfFile();` then fired even though the comment was correctly terminated. The check could not distinguish "found `*/` exactly at EOF" from "reached EOF without finding `*/`".

## How

Return as soon as `*/` is consumed, and report unterminated input only when the scan loop reaches EOF without finding `*/`. Added `LexerTest` regression cases for a trailing block comment (with and without preceding code, and the empty `/**/` case), plus genuinely unterminated comments (`/* bar`, `/*`).

Fixes #1827.

### DIFF
--- a/pkl-parser/src/test/kotlin/org/pkl/parser/LexerTest.kt
+++ b/pkl-parser/src/test/kotlin/org/pkl/parser/LexerTest.kt
@@ -114,34 +114,35 @@ class LexerTest {
   }
 
   @Test
-  fun blockCommentAtEndOfFileWithoutTrailingNewline() {
+  fun `block comment at EOF without trailing newline`() {
     // A block comment that is the last thing in the file, with no trailing
     // newline, must lex correctly and not be mistaken for an unterminated comment.
     val lexer = Lexer("/*\nbar\n*/")
     assertThat(lexer.next()).isEqualTo(Token.BLOCK_COMMENT)
     assertThat(lexer.next()).isEqualTo(Token.EOF)
-
-    // Preceding code followed by the trailing block comment.
-    val withCode = Lexer("foo = 123\n/*\nbar\n*/")
-    assertThat(withCode.next()).isEqualTo(Token.IDENTIFIER)
-    assertThat(withCode.next()).isEqualTo(Token.ASSIGN)
-    assertThat(withCode.next()).isEqualTo(Token.INT)
-    assertThat(withCode.next()).isEqualTo(Token.BLOCK_COMMENT)
-    assertThat(withCode.next()).isEqualTo(Token.EOF)
-
-    // Empty block comment at end of file.
-    val empty = Lexer("/**/")
-    assertThat(empty.next()).isEqualTo(Token.BLOCK_COMMENT)
-    assertThat(empty.next()).isEqualTo(Token.EOF)
   }
 
   @Test
-  fun plainCodeAtEndOfFileWithoutTrailingNewline() {
-    // Guard/boundary test — NOT a regression test for the block-comment EOF
-    // bug. Plain code never enters the block-comment path, so this passes with
-    // or without that fix. It documents that code whose final line has no
-    // trailing newline lexes cleanly and terminates with EOF rather than
-    // erroring.
+  fun `block comment after code at EOF without trailing newline`() {
+    val lexer = Lexer("foo = 123\n/*\nbar\n*/")
+    assertThat(lexer.next()).isEqualTo(Token.IDENTIFIER)
+    assertThat(lexer.next()).isEqualTo(Token.ASSIGN)
+    assertThat(lexer.next()).isEqualTo(Token.INT)
+    assertThat(lexer.next()).isEqualTo(Token.BLOCK_COMMENT)
+    assertThat(lexer.next()).isEqualTo(Token.EOF)
+  }
+
+  @Test
+  fun `empty block comment at EOF without trailing newline`() {
+    val lexer = Lexer("/**/")
+    assertThat(lexer.next()).isEqualTo(Token.BLOCK_COMMENT)
+    assertThat(lexer.next()).isEqualTo(Token.EOF)
+  }
+
+  @Test
+  fun `plain code at EOF without trailing newline`() {
+    // Documents that code whose final line has no trailing newline lexes
+    // cleanly and terminates with EOF rather than erroring.
     val lexer = Lexer("foo = 123")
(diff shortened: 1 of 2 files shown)


pkl-parser/src/test/kotlin/org/pkl/parser/LexerTest.kt
```diff
@@ -113,6 +113,51 @@ class LexerTest {
...
+    assertThat(empty.next()).isEqualTo(Token.EOF)
+  }
+
+  @Test
+  fun plainCodeAtEndOfFileWithoutTrailingNewline() {
+    // Guard/boundary test — NOT a regression test for the block-comment EOF
+    // bug. Plain code never enters the block-comment path, so this passes with
+    // or without that fix. It documents that code whose final line has no
+    // trailing newline lexes cleanly and terminates with EOF rather than
+    // erroring.
```
HT154: Comment is a little verbose here. In particular "the block-comment EOF bug" loses all context once this PR merges.

```suggestion
    // Documents that code whose final line has no
    // trailing newline lexes cleanly and terminates with EOF rather than
    // erroring.
```

pkl-parser/src/test/kotlin/org/pkl/parser/LexerTest.kt
```diff
@@ -113,6 +113,51 @@ class LexerTest {
...
+    assertThat(withCode.next()).isEqualTo(Token.ASSIGN)
+    assertThat(withCode.next()).isEqualTo(Token.INT)
+    assertThat(withCode.next()).isEqualTo(Token.BLOCK_COMMENT)
+    assertThat(withCode.next()).isEqualTo(Token.EOF)
+
+    // Empty block comment at end of file.
+    val empty = Lexer("/**/")
+    assertThat(empty.next()).isEqualTo(Token.BLOCK_COMMENT)
+    assertThat(empty.next()).isEqualTo(Token.EOF)
+  }
```
HT154, line 140: Would you mind splitting this into three separate tests?

pkl-parser/src/test/kotlin/org/pkl/parser/LexerTest.kt
```diff
@@ -113,6 +113,51 @@ class LexerTest {
     assertThat(thrown.message).contains("Whitespace")
   }
 
+  @Test
+  fun blockCommentAtEndOfFileWithoutTrailingNewline() {
```
HT154: Nit: style. Not a blocker, but we tend to prefer `` `space case` `` for test method names.

```suggestion
  fun `block comment at EOF without trailing newline`() {
```

HT154 (changes requested):
Thanks for the fix! I have a few requests around testing this.

sergiosalvi:
Aye aye. I'm putting my people (🤖) to address your comments, just a sec.

pkl-parser/src/test/kotlin/org/pkl/parser/LexerTest.kt
sergiosalvi: Done — renamed the added tests to `` `space case` ``.

pkl-parser/src/test/kotlin/org/pkl/parser/LexerTest.kt
sergiosalvi: Done — split into three focused tests: a bare block comment, a block comment after code, and the empty `/**/` case.

pkl-parser/src/test/kotlin/org/pkl/parser/LexerTest.kt
sergiosalvi: Good call — trimmed it down and dropped the reference to this bug so it still reads sensibly after the PR merges. Went with your suggested wording.

HT154 (approved):
Thanks!

pkl-parser/src/test/kotlin/org/pkl/parser/LexerTest.kt
sergiosalvi: I've opened https://github.com/apple/pkl/pull/1829 for the naming convention.

PR #1834, opened 2026-08-27
Add net stdlib module with URL support

stackoverflow:


stdlib/url.pkl
```diff
@@ -0,0 +1,314 @@
...
+  /// When [null], only absolute URLs can be parsed.
+  ///
+  /// Throws if it is not a parseable absolute URL.
+  base: String?
+
+  /// Parses [input] as a URL.
+  ///
+  /// Parsing is lenient: malformed input is coerced rather than
+  /// rejected. Returns [null] only when [input] cannot be parsed at all.
+  external function parse(input: String): Url?
```
bioball: Following our other APIs, should have `parse` and `parseOrNull` variants (one throws, the other returns nullable in case of error)

stdlib/url.pkl
```diff
@@ -0,0 +1,314 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+/// Parsing and manipulation of URLs, following the
+/// [WHATWG URL Standard](https://url.spec.whatwg.org/).
+@ModuleInfo { minPklVersion = "0.33.0" }
+module pkl.url
```
bioball:
```suggestion
@Since { version = "0.33.0" }
module pkl.url
```

stdlib/url.pkl
```diff
@@ -0,0 +1,314 @@
...
+  /// new Builder { scheme = "not a scheme"; host = "e.com" }.buildOrNull() == null
+  /// new Builder { scheme = "https"; host = "e.com/evil" }.buildOrNull() == null
+  /// ```
+  external function buildOrNull(): Url?
+}
+
+/// Percent-encodes [value] so that it can be interpolated into a URL component.
+///
+/// Every code point other than an ASCII alphanumeric or one of `!'()*-._~` is percent-encoded as
+/// its UTF-8 bytes, which is what JavaScript's `encodeURIComponent` does.
```
bioball:
```suggestion
/// Every code point other than an ASCII alphanumeric or one of `!'()*-._~` is percent-encoded
/// as its UTF-8 octets.
///
/// Follows the rules for the `encodeURI` function as described by
/// [ECMA-262](https://262.ecma-international.org/5.1/#sec-15.1.3.3).
```

stdlib/net.pkl
```diff
@@ -0,0 +1,223 @@
...
+  ///
+  /// A URL has an authority exactly when it has a [host].
+  /// This is [userInfo], [host] and [port] joined back together, encoded as in [toString()].
+  external fixed authority: String?
+
+  /// The percent-decoded segments of [path].
+  ///
+  /// Facts:
+  /// ```
+  /// net.Url("https://example.com/a/b%20c").segments == List("a", "b c")
```
bioball: Shouldn't the first segment be `""`? Since the path starts with `/`?
```suggestion
  /// net.Url("https://example.com/a/b%20c").segments == List("", "a", "b c")
```

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlFactory.java
```diff
@@ -0,0 +1,119 @@
...
+  static Parsed parseOrThrow(String input, VmExceptionBuilder exceptionBuilder) {
+    var result = UrlParser.parse(input);
+    if (result instanceof Result.Failure failure) {
+      throw exceptionBuilder.evalError("cannotParseUrl", input).withHint(failure.hint()).build();
+    }
+    return ((Result.Success) result).url();
+  }
+
+  /** Reads the components back off {@code url}. */
+  static Parsed read(VmObjectLike url) {
```
bioball:
```suggestion
  static Parsed read(VmTyped url) {
```

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlFactory.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+    var port = (Long) VmNull.unwrap(VmUtils.readMember(url, Identifier.PORT));
+    return new Parsed(
+        readNullableString(url, Identifier.SCHEME),
+        readNullableString(url, Identifier.USER_INFO),
+        readNullableString(url, Identifier.HOST),
+        port == null ? null : port.intValue(),
+        (String) VmUtils.readMember(url, Identifier.PATH),
+        readNullableString(url, Identifier.QUERY),
+        readNullableString(url, Identifier.FRAGMENT));
```
bioball: We might as well save this to the object's extra storage, so that the next `read()` call doesn't do this work again.

bioball: Second comment: each of calls into `VmUtils.readMember` should provide an `IndirectCallNode`, because otherwise it will use an `IndirectCallNode.getUncached()`

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlNodes.java
```diff
@@ -0,0 +1,97 @@
...
+      return UrlFactory.createQueryParameters(UrlFactory.readQuery(self));
+    }
+  }
+
+  public abstract static class toString extends ExternalMethod0Node {
+    @Specialization
+    @TruffleBoundary
+    protected String eval(VmTyped self) {
+      return UrlFactory.read(self).serialize();
+    }
```
bioball, line 70: Shouldn't put `UrlFactory.read()` behind a truffle boundary, because that results in more truffle nodes being executed. Instead, we should put that on `org.pkl.core.stdlib.net.UrlParser.Parsed#serialize`.

Regarding my above comment on "pass an IndirectCallNode into `VmUtils.readMember()`": it can be created via the `@Cached()` method.
We can also have two specializations, one where the extra storage already exists, and one where it's not. That way we don't pay for the cost of `insert(IndirectCallNode.create())` if given a Url that already has a "Parsed" extra storage member.

```suggestion
    @Specialization(guards = "self.hasExtraStorage()")
    protected String evalCached(VmTyped self) {
      var parsed = (Parsed) self.getExtraStorage();
      return parsed.serialize();
    }

    @Specialization
    protected String eval(VmTyped self, @Cached("create()") IndirectCallNode callNode) {
      var parsed = UrlFactory.read(self, callNode);
      return parsed.serialize();
    }
```

This same pattern can be applied to the rest of the nodes that interact with `UrlFactory.read()`.

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlNodes.java
```diff
@@ -0,0 +1,97 @@
...
+    @Specialization
+    @TruffleBoundary
+    protected Object eval(VmTyped self, String ref) {
+      return resolve(self, ref);
+    }
+
+    @Specialization
+    @TruffleBoundary
+    protected Object eval(VmTyped self, VmTyped ref) {
+      return resolve(self, UrlFactory.read(ref).serialize());
```
bioball: This will do double work, where it serializes, then re-parses.

This also means that these snippets get different results, because `serialize()` does percent-encoding:

```pkl
local base = net.Url("http://a/b/c/")

path1 = base.resolve("d e").path
// gives "/b/c/d e"

path2 = base.resolve(new net.Url { path = "d e " }).path
// gives "/b/c/d%20e"
```

pkl-core/src/main/java/org/pkl/core/stdlib/net/NetNodes.java
```diff
@@ -0,0 +1,123 @@
...
+
+/** Backing nodes for {@code pkl:net}'s module-level members. */
+public final class NetNodes {
+  private NetNodes() {}
+
+  public abstract static class Url extends ExternalMethod1Node {
+    @Specialization
+    @TruffleBoundary
+    protected Object eval(@SuppressWarnings("unused") VmTyped self, String input) {
+      return UrlFactory.create(UrlFactory.parseOrThrow(input, exceptionBuilder()));
```
bioball: This is expensive, because `exceptionBuilder()` will create an exception that fills in its stack trace. On the happy path, we'd be doing that work just to be thrown away.

Pass a `Node` in instead:

```suggestion
      return UrlFactory.create(UrlFactory.parseOrThrow(input, this));
```

And change `parseOrThrow`'s implementation to:

```java
  static Parsed parseOrThrow(String input, Node node) {
    var result = UrlParser.parse(input);
    if (result instanceof Result.Failure failure) {
          .withLocation(node)
          .evalError("cannotParseUrl", input)
          .withHint(failure.hint())
          .build();
    }
    return ((Result.Success) result).url();
  }
```

This is a pattern we use elsewhere. For example, see `org.pkl.core.runtime.ModuleResolver#resolve`

pkl-core/src/main/java/org/pkl/core/stdlib/net/PercentEncoder.java
```diff
@@ -0,0 +1,214 @@
...
+            codePoint -> {
+              if (isUnreserved(codePoint)) {
+                out.appendCodePoint(codePoint);
+              } else if (codePoint == ' ') {
+                out.append('+');
+              } else {
+                encodeUtf8(codePoint, out);
+              }
+            });
+  }
```
bioball, line 213: Why not use the JDK's `java.net.URLEncoder` and `java.net.URLDecoder`?

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlParser.java
```diff
@@ -0,0 +1,712 @@
...
+          : new Result.Failure(
+              "The IP literal `" + host + "` has trailing characters after its `]`.");
+    }
+    if (host.indexOf('[') >= 0 || host.indexOf(']') >= 0) {
+      return new Result.Failure(
+          "A registered name cannot contain `[` or `]`; percent-encode them, or enclose an IP"
+              + " address in brackets.");
+    }
+    return percentEncodingFailure(host);
+  }
```
bioball, line 580: This method permits some invalid hosts, e.g. `net.Url("http://::/foo").host` gives me `":"` as its host.

bioball (changes requested):
Did a first pass!

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlParser.java
```diff
@@ -0,0 +1,712 @@
...
+    if (input.startsWith("//", pointer)) {
+      pointer += 2;
+      var end = pointer;
+      while (end < length && !isAuthorityTerminator(input.charAt(end))) {
+        end++;
+      }
+      var authority = input.substring(pointer, end);
+      pointer = end;
+
+      // the userinfo runs to the last "@", because "@" may appear literally in a host
```
HT154:
```suggestion
      // the userinfo runs to the last "@", because "@" may not appear literally in a host
```

This is backwards, no?

stdlib/net.pkl
```diff
@@ -0,0 +1,223 @@
...
+class Url {
+  /// The scheme, without the trailing `:`, or [null] if this is a relative reference.
+  ///
+  /// [Url()] lowercases this, as schemes are case-insensitive.
+  scheme: String(isValidScheme(this))?
+
+  /// The userinfo, without the trailing `@`, or [null] if this URL has none.
+  ///
+  /// Only a URL with a [host] can have userinfo.
+  userInfo: String(host != null && isValidPercentEncoding(this))?
```
HT154:
```suggestion
  userInfo: String(host != null, isValidPercentEncoding(this))?
```

stdlib/net.pkl
```diff
@@ -0,0 +1,223 @@
...
+  fragment: String(isValidPercentEncoding(this))?
+
+  /// Returns [true] if this URL is absolute: has a non-null scheme.
+  hidden fixed isAbsolute: Boolean = scheme != null
+
+  /// The authority, without the leading `//`, or [null] if this URL has none.
+  ///
+  /// A URL has an authority exactly when it has a [host].
+  /// This is [userInfo], [host] and [port] joined back together, encoded as in [toString()].
+  external fixed authority: String?
```
HT154, line 87:
```suggestion
  external authority: String?
```
External properties already behave as if they're fixed. We don't combine these modifiers anywhere else in the stdlib.

stdlib/net.pkl
```diff
@@ -0,0 +1,223 @@
...
+
+  /// Whether [other] identifies the same resource as this URL.
+  ///
+  /// Unlike `==`, this compares Urls as the *syntax-based normalization* of RFC 3986:
+  /// [scheme] and [host] are case-insensitive and percent-encoding is normalized.
+  /// An octet standing for a reserved character is kept as written.
+  ///
+  /// Dot segments are only removed from an absolute [path].
+  ///
+  /// Nothing that would take knowing the scheme is normalized.
```
HT154:
```suggestion
  /// Normalization that would require knowing the scheme is not performed.
```

stdlib/net.pkl
```diff
@@ -0,0 +1,223 @@
...
+  /// ```
+  external function resolve(ref: String | Url): Url
+
+  /// Whether [other] identifies the same resource as this URL.
+  ///
+  /// Unlike `==`, this compares Urls as the *syntax-based normalization* of RFC 3986:
+  /// [scheme] and [host] are case-insensitive and percent-encoding is normalized.
+  /// An octet standing for a reserved character is kept as written.
+  ///
+  /// Dot segments are only removed from an absolute [path].
```
HT154: For clarity:
```suggestion
  /// Dot segments (`./` and `../`) are only removed from an absolute [path].
```

stdlib/net.pkl
```diff
@@ -0,0 +1,223 @@
...
+  /// This is [userInfo], [host] and [port] joined back together, encoded as in [toString()].
+  external fixed authority: String?
+
+  /// The percent-decoded segments of [path].
+  ///
+  /// Facts:
+  /// ```
+  /// net.Url("https://example.com/a/b%20c").segments == List("a", "b c")
+  /// ```
+  external fixed segments: List<String>
```
HT154: I think the name of this property should be explicit about it being the path's segments

```suggestion
  external pathSegments: List<String>
```

stdlib/net.pkl
stackoverflow: See answer here: https://github.com/apple/pkl-evolution/pull/36#discussion_r4035092679.

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlNodes.java
stackoverflow: Good idea.

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlNodes.java
stackoverflow: Good catch!
Added a test for this.

pkl-core/src/main/java/org/pkl/core/stdlib/net/PercentEncoder.java
stackoverflow: `URLEncoder`/`URLDecoder` implement `application/x-www-form-urlencoded` (as stated in their javadoc), not RFC 3986.
For example `URLEncoder` will encode `~` which is in unreserved in 3986. It also encodes ` ` to `+`.
It could replace our `PercentEncoder.encode/decodeForm`, but this is a trivial implementation once you already have all the machinery.

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlParser.java
stackoverflow: That was indeed a bug!

HT154 (commented):
Here's my first pass

stdlib/net.pkl
stackoverflow: The external modifier has no relation to fixed. The only contract for external properties is that they can't have a default body (it comes from Java).

```pkl
local m = (import("pkl:math")) {
  minInt = 42
}

bogusMinInt = m.minInt
```
```
bogusMinInt = 42
```

The fact we don't put external + fixed together in other places is because most external properties are in a primitive class (which can't be amended), in an external class (which also can't be amended), or in a module (which can be amended, but doesn't change the original module).

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlParser.java
stackoverflow: Indeed!

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlFactory.java
```diff
@@ -0,0 +1,133 @@
...
+import org.pkl.core.runtime.VmObjectBuilder;
+import org.pkl.core.runtime.VmObjectLike;
+import org.pkl.core.runtime.VmTyped;
+import org.pkl.core.runtime.VmUtils;
+import org.pkl.core.stdlib.VmObjectFactory;
+import org.pkl.core.stdlib.net.UrlParser.Parsed;
+import org.pkl.core.stdlib.net.UrlParser.Result;
+
+/** Converts between a parsed URL and {@code pkl:net}'s {@code Url}. */
+final class UrlFactory {
```
bioball, line 36: Move this to `org.pkl.core.runtime`, since this isn't a node

pkl-core/src/main/java/org/pkl/core/stdlib/net/PercentEncoder.java
```diff
@@ -0,0 +1,214 @@
...
+import java.util.function.IntPredicate;
+
+/**
+ * Percent-encoding, and the character classes of <a
+ * href="https://www.rfc-editor.org/rfc/rfc3986#appendix-A">RFC 3986's ABNF</a>.
+ *
+ * <p>Each {@code IntPredicate} here holds for the characters that may appear literally in a
+ * component. Anything else has to be percent-encoded to appear there at all.
+ */
+final class PercentEncoder {
```
bioball, line 29: Move this to `org.pkl.core.runtime`, since this isn't a node.

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlNodes.java
```diff
@@ -0,0 +1,156 @@
...
+      var base = UrlFactory.read(self, callNode);
+      if (base.scheme() == null) {
+        CompilerDirectives.transferToInterpreter();
+        throw exceptionBuilder()
+            .evalError("cannotResolveAgainstRelativeUrl", base.serialize())
+            .build();
+      }
+      return UrlFactory.create(UrlParser.resolve(base, ref));
+    }
+  }
```
bioball, line 103: This implementation (and others) will still create an `IndirectCallNode` even if it doesn't need it (e.g. it's already in extra storage).

I played around with this and I think an elegant solution here is to introduce a specialized truffle node for it; e.g. this:

```java
package org.pkl.core.ast.internal;

import com.oracle.truffle.api.dsl.Cached;
import com.oracle.truffle.api.dsl.Specialization;
import com.oracle.truffle.api.frame.VirtualFrame;
import com.oracle.truffle.api.nodes.IndirectCallNode;
import com.oracle.truffle.api.source.SourceSection;
import org.pkl.core.ast.PklNode;
import org.pkl.core.runtime.VmTyped;
import org.pkl.core.stdlib.net.UrlFactory;
import org.pkl.core.stdlib.net.UrlParser;

public abstract class GetParsedUrlNode extends PklNode {
  protected GetParsedUrlNode(SourceSection sourceSection) {
    super(sourceSection);
  }

  @Specialization(guards = "url.hasExtraStorage()")
  protected UrlParser.Parsed evalCached(VmTyped url) {
    return (UrlParser.Parsed) url.getExtraStorage();
  }

  @Specialization
  protected UrlParser.Parsed eval(VmTyped url, @Cached("create()") IndirectCallNode callNode) {
    return UrlFactory.read(url, callNode);
  }

  public abstract UrlParser.Parsed execute(VirtualFrame frame, VmTyped url);
}
```

Then this implementation turns into:

```java
  public abstract static class resolve extends ExternalMethod1Node {
    private @Child GetParsedUrlNode getParsedUrlNode = GetParsedUrlNodeGen.create(sourceSection);

    @Specialization
    protected Object evalString(VirtualFrame frame, VmTyped self, String ref) {
      var base = getParsedUrlNode.execute(frame, self);
      return resolve(base, UrlFactory.parseOrThrow(ref, this));
    }

    @Specialization
    protected Object eval(VirtualFrame frame, VmTyped self, VmTyped ref) {
      var base = getParsedUrlNode.execute(frame, self);
      var parsed = getParsedUrlNode.execute(frame, ref);
      return resolve(base, parsed);
    }

    @SuppressWarnings("MethodNameSameAsClassName")
    private Object resolve(Parsed base, Parsed ref) {
      if (base.scheme() == null) {
        throw exceptionBuilder()
            .evalError("cannotResolveAgainstRelativeUrl", base.serialize())
            .build();
      }
      return UrlFactory.create(UrlParser.resolve(base, ref));
    }
  }
```

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlNodes.java
```diff
@@ -0,0 +1,156 @@
...
+
+/** Backing nodes for {@code pkl:net}'s {@code Url} class. */
+public final class UrlNodes {
+  private UrlNodes() {}
+
+  public abstract static class authority extends ExternalPropertyNode {
+    @Specialization
+    protected Object eval(VmTyped self, @Cached("create()") IndirectCallNode callNode) {
+      return VmNull.lift(UrlFactory.readAuthority(self, callNode));
+    }
```
bioball, line 42: Can we introduce specializations for each of these? One who guards on `self.hasExtraStorage()`?

E.g.

```java
    @Specialization(guards = "self.hasExtraStorage()")
    protected Object evalCached(VmTyped self) {
      var parsed = (Parsed) self.getExtraStorage();
      return VmNull.lift(parsed.authority());
    }

    @Specialization
    protected Object eval(VmTyped self, @Cached("create()") IndirectCallNode callNode) {
      return VmNull.lift(UrlFactory.readAuthority(self, callNode));
    }
```

And then change each of the `read*` methods in `URLFactory` to always call into `VmUtils.readMember()`

The benefit here is: if the object already has a cached `Parsed()` object, we can skip the overhead of `insert(IndirectCallNode.create())`.

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlParser.java
```diff
@@ -0,0 +1,745 @@
...
+ * <p>Both shapes of {@code URI-reference} (section 4.1) are accepted: a {@code URI}, which states a
+ * scheme, and a {@code relative-ref}, which does not.
+ *
+ * <p>Reference resolution is not part of parsing; see {@link #resolve}.
+ *
+ * <p>An IPv6 address may carry the zone identifier of <a
+ * href="https://www.rfc-editor.org/rfc/rfc6874">RFC 6874</a>, as in {@code [fe80::1%25eth0]}..
+ */
+@SuppressWarnings("JavadocLinkAsPlainText")
+public final class UrlParser {
```
bioball, line 38: Move to `org.pkl.core.runtime`, or perhaps `org.pkl.core.util.url`, since this isn't a node

pkl-core/src/main/java/org/pkl/core/stdlib/net/NetNodes.java
```diff
@@ -0,0 +1,86 @@
...
+import org.pkl.core.runtime.VmTyped;
+import org.pkl.core.stdlib.ExternalMethod1Node;
+
+/** Backing nodes for {@code pkl:net}'s module-level members. */
+public final class NetNodes {
+  private NetNodes() {}
+
+  public abstract static class Url extends ExternalMethod1Node {
+    @Specialization
+    @TruffleBoundary
```
bioball, line 31: Remove truffle boundary here, and move them to `UrlFactory.create()` and `UrlFactory.parseOrThrow()` if they need them (truffle node method bodies should be PE-friendly).

Same goes for other methods, too. Although I think `encodeUrlComponent` and `buildQuery` are maybe okay because the string builder API is obviously not PE-friendly, and we'd just introduce a new Java method for it unnecessarily.

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlFactory.java
```diff
@@ -0,0 +1,133 @@
...
+
+  private static @Nullable String readNullableString(
+      VmObjectLike url, Identifier name, IndirectCallNode callNode) {
+    return (String) VmNull.unwrap(VmUtils.readMember(url, name, callNode));
+  }
+
+  static String readPath(VmObjectLike url) {
+    return url.hasExtraStorage()
+        ? ((Parsed) url.getExtraStorage()).path()
+        : (String) VmUtils.readMember(url, Identifier.PATH);
```
bioball, line 95: Needs a call node

pkl-config-java/src/test/java/org/pkl/config/java/mapper/PUrlToURITest.java
```diff
@@ -0,0 +1,91 @@
...
+    assertThat(map("constructed")).isEqualTo(URI.create("https://example.com/a%20b?x%20y"));
+  }
+
+  @Test
+  public void ipvFuture() {
+    // `java.net.URI` doesn't know IPvFuture addresses
+    assertThatThrownBy(() -> map("ipvFuture"))
+        .isInstanceOf(ConversionException.class)
+        .hasMessageContaining("http://[v1.fe80::a+en1]/")
+        .hasCauseInstanceOf(URISyntaxException.class);
```
bioball, line 83: It doesn't seem great that you can define valid data in Pkl and have that blow up when mapped to Java, although it's definitely an edge case.

pkl-core/src/main/java/org/pkl/core/stdlib/net/PercentEncoder.java
bioball: Yeah, I'm referring to `encodeForm` and `decodeForm` specifically.

According to [WHATWG]( https://url.spec.whatwg.org/#application-x-www-form-urlencoded-percent-encode-set), the characters to encode also includes U+0021, U+0027-U+0027, and U+007E, but this implementation isn't doing that.

But also, `URLEncoder`/`URLDecoder` have already been optimized and battle-tested, so I feel like we might as well use it.

pkl-core/src/main/java/org/pkl/core/stdlib/net/PercentEncoder.java
```diff
@@ -0,0 +1,214 @@
...
+    return Character.toUpperCase(Character.forDigit(value, 16));
+  }
+
+  /**
+   * Percent-decodes {@code input} and interprets the decoded bytes as UTF-8.
+   *
+   * <p>A {@code %} that does not begin a percent-encoded octet is kept as-is.
+   */
+  static String decode(String input) {
+    var in = input.getBytes(StandardCharsets.UTF_8);
```
bioball, line 177: Optimization: we can skip the allocation overhead if there's no `%` to begin with.

```suggestion
    if (!input.contains("%")) {
      return input;
    }
    var in = input.getBytes(StandardCharsets.UTF_8);
```

stdlib/net.pkl
```diff
@@ -0,0 +1,247 @@
...
+/// ```
+/// net.encodeUrlComponent("a b&c") == "a%20b%26c"
+/// net.encodeUrlComponent("50%") == "50%25"
+/// ```
+external const function encodeUrlComponent(value: String): String
+
+/// Percent-decodes [value], interpreting the decoded bytes as UTF-8.
+///
+/// The inverse of [encodeUrlComponent()].
+/// A `%` that does not begin a percent-encoded octet is kept as-is.
```
bioball, line 211: I strongly feel that this should throw, not kept as-is. Otherwise, a malformed percent encoding would silently be passed through as a decoded string.

bioball (changes requested):
Did another pass!

Also: I more and more feel that the `Url` class should have pairs of raw/encoded properties for each component.

Otherwise, what does this mean?

```pkl
new net.Url { path = "/foo%20bar" }
```

There's a big difference between the "this is the encoded value" versus "this is the un-encoded value", and the current API is kind of a hybrid between the two.

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlFactory.java
stackoverflow: Moved to org/pkl/core/util/url

pkl-core/src/main/java/org/pkl/core/stdlib/net/PercentEncoder.java
stackoverflow: Moved to org/pkl/core/util/url

pkl-core/src/main/java/org/pkl/core/util/url/UrlParser.java
```diff
@@ -0,0 +1,745 @@
...
+ * <p>Both shapes of {@code URI-reference} (section 4.1) are accepted: a {@code URI}, which states a
+ * scheme, and a {@code relative-ref}, which does not.
+ *
+ * <p>Reference resolution is not part of parsing; see {@link #resolve}.
+ *
+ * <p>An IPv6 address may carry the zone identifier of <a
+ * href="https://www.rfc-editor.org/rfc/rfc6874">RFC 6874</a>, as in {@code [fe80::1%25eth0]}..
+ */
+@SuppressWarnings("JavadocLinkAsPlainText")
+public final class UrlParser {
```
stackoverflow: Moved to org/pkl/core/util/url

stdlib/net.pkl
stackoverflow: Indeed. Leniency makes sense in the parser, less so here.
This throws now on raw %/wrong octets. Also changed `parseQuery` which had the same problem and added some tests.

pkl-core/src/main/java/org/pkl/core/util/url/PercentEncoder.java
```diff
@@ -0,0 +1,214 @@
...
+            codePoint -> {
+              if (isUnreserved(codePoint)) {
+                out.appendCodePoint(codePoint);
+              } else if (codePoint == ' ') {
+                out.append('+');
+              } else {
+                encodeUtf8(codePoint, out);
+              }
+            });
+  }
```
stackoverflow: `URLEncoder` is not RFC 3986 compatible so we can't use it (it would encode `~`, for example, which is in the unreserved set).
`URLDecoder` seems to be compatible, from the tests I did, but I still don't think it's a good idea to replace it. Specially because we'll use our own encoder. Also, it allocates a full input size string builder every time, where we short circuit the happy path where there's nothing to encode. So I'm not convinced it's going to be any faster.

pkl-config-java/src/test/java/org/pkl/config/java/mapper/PUrlToURITest.java
stackoverflow: It's unfortunate, but we can't make Pkl support every language URL library out there. Go won't have a problem (as it supports RFC 6874).

pkl-core/src/main/java/org/pkl/core/util/url/PercentEncoder.java
bioball: To be clear, I'm referring to `encodeForm` and `decodeForm`, which is your implementation for coding/decoding `application/x-www-form-urlencoded`.

For this, it _should_ percent-encode `~` (see https://url.spec.whatwg.org/#application-x-www-form-urlencoded-percent-encode-set)

stackoverflow: Was also talking about `encodeForm` and `decodeForm`.
RFC 3986 doesn't actually have any rule about `application/x-www-form-urlencoded`. I was using the default rules for encoding, but indeed, we can just follow whatwg here. It has a slight conflict with 3986:
```
percent-encoded octets in the ranges of ALPHA (%41-%5A and %61-%7A), DIGIT (%30-%39), hyphen (%2D), period (%2E), underscore (%5F), or tilde (%7E) should not be created by URI producers
```
It only says `should`, not `must`.
Changed these functions to use the Java encoder/decoder which (according to my tests) are whatwg compatible.

pkl-core/src/main/java/org/pkl/core/ast/internal/GetParsedUrlNode.java
```diff
@@ -0,0 +1,42 @@
...
+  public static GetParsedUrlNode create() {
+    return GetParsedUrlNodeGen.create();
+  }
+
+  @Specialization(guards = "url.hasExtraStorage()")
+  protected Parsed evalCached(VmTyped url) {
+    return (Parsed) url.getExtraStorage();
+  }
+
+  @Specialization
```
bioball:
```suggestion
  @Specialization(guards = "!url.hasExtraStorage())
```

stdlib/net.pkl
```diff
@@ -0,0 +1,304 @@
...
+  ///
+  /// An encoded `/` is decoded like any other octet, so it cannot be told apart from a `/` that
+  /// separates segments; [pathSegments] keeps the two apart.
+  ///
+  /// Facts:
+  /// ```
+  /// net.Url("https://example.com/a%2Fb/c").rawPath == "/a/b/c"
+  /// net.Url("https://example.com/a%2Fb/c").pathSegments == List("a/b", "c")
+  /// ```
+  hidden fixed rawPath: String = decodeUrlComponent(path)
```
HT154: I think the `foo` and `rawFoo` properties are reversed here. As far as I can tell, the raw versions in most URL APIs are the "encoded" versions, so this may be backwards for most users' intuition.

stackoverflow: Yeah, that's more in line with other languages. My idea to keep the non-raw versions as encoded was so that users can amend the regular properties (`(url) { path = "..." }`).
Inverted the names.

stdlib/net.pkl
```diff
@@ -0,0 +1,307 @@
...
+/// Facts:
+/// ```
+/// net.encodeUrlComponent("a b&c") == "a%20b%26c"
+/// net.encodeUrlComponent("50%") == "50%25"
+/// ```
+external const function encodeUrlComponent(value: String): String
+
+/// Percent-decodes [value], interpreting the decoded bytes as UTF-8.
+///
+/// The inverse of [encodeUrlComponent()].
```
bioball: Not exactly the inverse because you can't round-trip:

```pkl
encodeUrlComponent(decodeUrlComponent("%41")) // gives you "A"
```

stdlib/net.pkl
```diff
@@ -0,0 +1,307 @@
...
+  ///
+  /// Facts:
+  /// ```
+  /// net.Url("HTTPS://Example.com/%7Euser").normalize() == net.Url("https://example.com/~user")
+  /// net.Url("https://example.com/a/b/../c").normalize().rawPath == "/a/c"
+  /// net.Url("https://example.com").normalize().rawPath == "/"
+  /// ```
+  external function normalize(): Url
+
+  /// Whether [other] identifies the same resource as this URL.
```
bioball: We use "Tells if" for every method/property that returns a boolean.

Same comment applies everywhere else in this module.

```suggestion
  /// Tells if [other] identifies the same resource as this URL.
```

pkl-core/src/main/java/org/pkl/core/util/url/UrlFactory.java
```diff
@@ -0,0 +1,139 @@
...
+          .withLocation(node)
+          .evalError(messageKey, value)
+          .withHint(failure.hint())
+          .build();
+    }
+  }
+
+  /** Reads the components back off {@code url}. */
+  public static Parsed read(VmTyped url, IndirectCallNode callNode) {
+    if (url.hasExtraStorage()) {
```
HT154: It probably makes sense to explicitly assert that `url` is indeed a `Url`:
```suggestion
  public static Parsed read(VmTyped url, IndirectCallNode callNode) {
    assert url.getVmClass() == NetModule.getUrlClass();
    if (url.hasExtraStorage()) {
```

This at least gives us a better error if we make a programming mistake.

pkl-core/src/main/java/org/pkl/core/ast/internal/GetParsedUrlNode.java
```diff
@@ -0,0 +1,44 @@
...
+public abstract class GetParsedUrlNode extends PklNode {
+  @NeverDefault
+  public static GetParsedUrlNode create() {
+    return GetParsedUrlNodeGen.create();
+  }
+
+  @Specialization(guards = "url.hasExtraStorage()")
+  protected Parsed evalCached(VmTyped url) {
+    return (Parsed) url.getExtraStorage();
+  }
```
HT154: Same here:
```suggestion
  protected Parsed evalCached(VmTyped url) {
    assert url.getVmClass() == NetModule.getUrlClass();
    return (Parsed) url.getExtraStorage();
  }
```

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlNodes.java
```diff
@@ -0,0 +1,189 @@
...
+    @Specialization(guards = "self.hasExtraStorage()")
+    protected Object evalCached(VmTyped self) {
+      var parsed = (Parsed) self.getExtraStorage();
+      return VmNull.lift(parsed.authority());
+    }
+
+    @Specialization
+    protected Object eval(VmTyped self, @Cached("create()") IndirectCallNode callNode) {
+      return VmNull.lift(UrlFactory.readAuthority(self, callNode));
+    }
```
HT154: Now that `GetParsedUrlNode`, should it be adopted everywhere?
```suggestion
    @Child private GetParsedUrlNode getParsedUrlNode = GetParsedUrlNode.create();

    @Specialization
    protected Object eval(VmTyped self) {
      return VmNull.lift(getParsedUrlNode.execute(self));
    }
```

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlNodes.java
```diff
@@ -0,0 +1,189 @@
...
+  private UrlNodes() {}
+
+  public abstract static class authority extends ExternalPropertyNode {
+    @Specialization(guards = "self.hasExtraStorage()")
+    protected Object evalCached(VmTyped self) {
+      var parsed = (Parsed) self.getExtraStorage();
+      return VmNull.lift(parsed.authority());
+    }
+
+    @Specialization
```
bioball: Sorry we also need this:

```suggestion
    @Specialization(guards = "!self.hasExtraStorage()")
```

 (same comment applies everywhere else)

stdlib/net.pkl
```diff
@@ -0,0 +1,307 @@
...
+  /// The percent-decoded [rawUserInfo].
+  hidden fixed userInfo: String? =
+    if (rawUserInfo != null) decodeUrlComponent(rawUserInfo) else null
+
+  /// The percent-decoded [rawHost].
+  hidden fixed host: String? = if (rawHost != null) decodeUrlComponent(rawHost) else null
+
+  /// The percent-decoded [rawPath].
+  ///
+  /// An encoded `/` is decoded like any other octet, so it cannot be told apart from a `/` that
```
HT154:
```suggestion
  /// An encoded `/` is decoded like any other octet, so it cannot be distinguished from a `/` that
```

stdlib/net.pkl
```diff
@@ -0,0 +1,307 @@
...
+  /// separates segments; [pathSegments] keeps the two apart.
```
HT154:
```suggestion
  /// separates segments.
  /// [pathSegments] disambiguates between separator and literal `/` characters.
```

stdlib/net.pkl
```diff
@@ -0,0 +1,307 @@
...
+  ///
+  /// A URL has an authority exactly when it has a [rawHost].
+  /// This is [rawUserInfo], [rawHost] and [port] joined back together.
+  external fixed authority: String?
+
+  /// The percent-decoded segments of [rawPath].
+  ///
+  /// Facts:
+  /// ```
+  /// net.Url("https://example.com/a/b%20c").pathSegments == List("a", "b c")
```
HT154: This fact is mentioned above for `path`, but I think it bears repeating here too:
```suggestion
  /// net.Url("https://example.com/a/b%20c").pathSegments == List("a", "b c")
  /// net.Url("https://example.com/a/b%2Fc").pathSegments == List("a", "b/c")
```

stdlib/net.pkl
```diff
@@ -0,0 +1,307 @@
...
+  /// ```
+  external fixed pathSegments: List<String>
+
+  /// The percent-decoded parameters of [rawQuery], or an empty mapping if this URL has no query.
```
HT154: Hmm I wonder if we should distinguish between "no query" and "empty query" by making this property nullable, i.e. `https://example.com/` has no query (so `queryParameters == null`) while `https://example.com/?` has an empty query (so `queryParameters == new Mapping {}`).

pkl-core/src/main/java/org/pkl/core/util/url/UrlFactory.java
```diff
@@ -0,0 +1,139 @@
...
+      return (Parsed) url.getExtraStorage();
+    }
+    var port = (Long) VmNull.unwrap(VmUtils.readMember(url, Identifier.PORT, callNode));
+    var parsed =
+        new Parsed(
+            readNullableString(url, Identifier.SCHEME, callNode),
+            readNullableString(url, Identifier.RAW_USER_INFO, callNode),
+            readNullableString(url, Identifier.RAW_HOST, callNode),
+            port == null ? null : port.intValue(),
+            (String) VmUtils.readMember(url, Identifier.RAW_PATH),
```
bioball:
```suggestion
            (String) VmUtils.readMember(url, Identifier.RAW_PATH, callNode),
```

pkl-core/src/main/java/org/pkl/core/util/url/UrlParser.java
```diff
@@ -0,0 +1,760 @@
...
+   *
+   * <p>{@code scheme} is {@code null} for a relative reference. {@code host} is {@code null} when
+   * the reference has no authority. {@code userInfo} and {@code port} are only ever set alongside a
+   * host.
+   */
+  public record Parsed(
+      @Nullable String scheme,
+      @Nullable String userInfo,
+      @Nullable String host,
+      @Nullable Integer port,
```
bioball: Prevent unnecessary casting between `Long` and `Integer`, since many method implementations go back and forth between in-language values and `Parsed`.

(Same comment applies everywhere else that deals with port)

```suggestion
      @Nullable Long port,
```

pkl-core/src/main/java/org/pkl/core/stdlib/net/UrlNodes.java
```diff
@@ -0,0 +1,189 @@
...
+    protected Object evalString(VmTyped self, String ref) {
+      return resolve(getSelfNode.execute(self), UrlFactory.parseOrThrow(ref, this));
+    }
+
+    @Specialization
+    protected Object eval(
+        VmTyped self, VmTyped ref, @Cached("create()") GetParsedUrlNode getRefNode) {
+      return resolve(getSelfNode.execute(self), getRefNode.execute(ref));
```
bioball: Can just re-use the same `GetParsedUrlNode` on both sides

```suggestion
    @Child private GetParsedUrlNode getUrlNode = GetParsedUrlNode.create();

    @Specialization
    protected Object evalString(VmTyped self, String ref) {
      return resolve(getUrlNode.execute(self), UrlFactory.parseOrThrow(ref, this));
    }

    @Specialization
    protected Object eval(
        VmTyped self, VmTyped ref) {
      return resolve(getUrlNode.execute(self), getUrlNode.execute(ref));
    }
```

PR #1838, opened 2026-08-28
Add `VmType` intermediate representation for types

HT154:
This has some downstream impacts:
- `CommandSpecParser` now deals with `VmType` instead of `TypeNode`
- `VmReference` now deals with `VmType` instead of `PType`
- `FunctionN` types are now exported as `PType.Function` instead of `PType.Class`, which may affect consumers of module schema information. This changes pkldoc rendering of such types, which we do not consider a breaking change. `pkl:reflect` behavior is unchanged.


pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,812 @@
...
+    return null;
+  }
+
+  public @Nullable VmTypeAlias getVmTypeAlias() {
+    return null;
+  }
+
+  /** Tells if this type is the same typecheck as the other type. */
+  @Override
+  public boolean equals(Object value) {
```
bioball:
```suggestion
  public final boolean equals(Object value) {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,812 @@
...
+          return PType.NOTHING;
+        }
+
+        @Override
+        public String toString() {
+          return "nothing";
+        }
+      };
+
+  public abstract static class SelfType extends VmType {
```
bioball: I don't see `SelfType` itself used for anything; just make `ModuleType` and `ThisType` extend `VmType` directly?

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,812 @@
...
+import org.pkl.core.PType.Alias;
+import org.pkl.core.PType.Class;
+import org.pkl.core.PType.Constrained;
+import org.pkl.core.PType.StringLiteral;
+import org.pkl.core.PType.TypeVariable;
+import org.pkl.core.PType.Union;
+import org.pkl.core.TypeParameter;
+import org.pkl.core.ValueFormatter;
+
+public abstract class VmType {
```
bioball: Maybe make this `sealed`, and turn `NOTHING` and `UNKNOWN` into their own classes.

You can still make them singletons by making their constructors private, and exposing `public static final <Type> INSTANCE = new <Type>();`

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,812 @@
...
+    protected boolean doIsEquivalentTo(VmType other) {
+      // consider constrained types as always different
+      return false;
+    }
+
+    @Override
+    protected boolean doIsSupertypeOf(VmType other) {
+      // constrained types can never supertypes
+      return false;
+    }
```
bioball, line 466: I think let's add an `identityHashCode` field to this type, which is the `System.identityHashCode()` of the constrained type node that this comes from. Then, these two methods should become:

```java
    @Override
    protected boolean doIsEquivalentTo(VmType other) {
      if (other instanceof ConstrainedType constrainedType) {
        return identityHashCode == constrainedType.identityHashCode;
      }
      return false;
    }

    @Override
    protected boolean doIsSupertypeOf(VmType other) {
      if (other instanceof ConstrainedType constrainedType) {
        return identityHashCode == constrainedType.identityHashCode;
      }
      return false;
    }
```

pkl-core/src/main/java/org/pkl/core/runtime/VmReference.java
```diff
@@ -88,71 +87,66 @@ public PType getReferentType() {
...
-    if (types.contains(PType.UNKNOWN)) return PType.UNKNOWN;
+    if (types.contains(VmType.UNKNOWN)) return VmType.UNKNOWN;
     // optimization: All allows all references, erase all candidates to only All
-    if (containsClass(types, BaseModule.getAnyClass().export()))
-      return new PType.Class(BaseModule.getAnyClass().export());
-    var typesList = new ArrayList<>(types);
-    typesList.sort(Comparator.comparing(Object::toString));
-    return new PType.Union(typesList);
+    if (containsClass(types, BaseModule.getAnyClass()))
+      return new VmType.ClassType(BaseModule.getAnyClass());
```
bioball:
```suggestion
    if (containsClass(types, BaseModule.getAnyClass())) {
      return new VmType.ClassType(BaseModule.getAnyClass());
    }
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -64,16 +61,44 @@
...
+
+  public VmType getType() {
+    if (type == null) {
+      type = doGetType();
+    }
+    return type;
+  }
+
+  @Override
+  public Node deepCopy() {
```
```suggestion
  public final Node deepCopy() {
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -64,16 +61,44 @@
 import org.pkl.core.util.MutableReference;
 
 public abstract class TypeNode extends PklNode {
+  private @Nullable VmType type;
```
```suggestion
  @CompilationFinal private @Nullable VmType type;
```

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -64,16 +61,44 @@
...
 
   protected TypeNode(SourceSection sourceSection) {
     super(sourceSection);
   }
 
+  protected abstract VmType doGetType();
+
+  public VmType getType() {
+    if (type == null) {
+      type = doGetType();
```
bioball, line 81:
```suggestion
      CompilerDirectives.transferToInterpreterAndInvalidate();
      type = doGetType();
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,812 @@
...
+    var clazz = getVmClass();
+    if (clazz != null) {
+      return new Class(clazz.export());
+    }
+    CompilerDirectives.transferToInterpreter();
+    throw new VmExceptionBuilder()
+        .bug("`%s` must override method `doExport()`.", getClass().getTypeName())
+        .build();
+  }
```
bioball: This method is overridden by every subclass. Is this needed?

```suggestion
  public abstract PType export();
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,812 @@
...
+            typeArguments[ft.parameterTypes.length],
+            ft.parameterTypes,
+            ft.returnType);
+      }
+      // special case: FunctionType is a subtype of class Function if return type is contravariant
+      if (clazz.isFunctionClass() && other instanceof FunctionType ft) {
+        if (typeArguments.length == 0) return true;
+        return typeArguments[0].isSupertypeOf(ft.returnType);
+      }
+      ;
```
bioball:
```suggestion
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,812 @@
...
+    for (var i = 0; i < paramCount; i++) {
+      if (!thisParams[i].isSubtypeOf(otherParams[i])) return false;
+    }
+    // return type is covariant
+    return thisReturn.isSupertypeOf(otherReturn);
+  }
+
+  private static VmType[] nUnknowns(int len) {
+    return Collections.nCopies(len, VmType.UNKNOWN).toArray(new VmType[0]);
+  }
```
bioball, line 731:
```suggestion
  private static VmType[] nUnknowns(int len) {
    var ret = new VmType[len];
    Arrays.fill(ret, VmType.UNKNOWN);
    return ret;
  }
```

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
```diff
@@ -656,137 +654,142 @@ private OptionBehavior resolveTypeArgument(
...
-        if (multiple == null) multiple = false;
-        if (metavar == null) metavar = METAVAR_CHAR;
-        return true;
-      } else if (typeNode
-          instanceof TypeNode.UnionOfStringLiteralsTypeNode unionOfStringLiteralsTypeNode) {
-        var choices = unionOfStringLiteralsTypeNode.getStringLiterals().stream().sorted().toList();
+      } else if (type instanceof VmType.UnionType ut && ut.isUnionOfStringLiterals()) {
+        var choices =
+            Arrays.stream(ut.getElementTypes())
+                .map(it -> ((VmType.StringLiteralType) it).getLiteral())
```
bioball, line 901: This used to be sorted

```suggestion
                .map(it -> ((VmType.StringLiteralType) it).getLiteral())
                .sorted()
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,812 @@
...
+      var result = clazz.getDisplayName();
+      if (typeArguments.length > 0) {
+        result +=
+            "<"
+                + Arrays.stream(typeArguments)
+                    .map(Object::toString)
+                    .collect(Collectors.joining(", "))
+                + ">";
+      }
+      return result;
```
bioball: Since this is language runtime code, we should prefer to use a string builder, and simple for loops to append to the string builder, because it's faster than the stream API version.

```suggestion
      var sb = new StringBuilder(clazz.getDisplayName());
      if (typeArguments.length > 0) {
        sb.append('<');
        var isFirst = true;
        for (var argument : typeArguments) {
          if (isFirst) {
            isFirst = false;
          } else {
            sb.append(", ");
          }
          sb.append(argument);
        }
        sb.append(">");
      }
      return sb.toString();
```

Same comment for the rest of the `toString()` methods in this file.

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,812 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+      return 31 * typeAlias.hashCode() + Arrays.hashCode(typeArguments);
+    }
+  }
+
+  public static final class FunctionType extends VmType {
```
bioball: I'm wondering if we need this as its own VmType?

At the type level, a `Function` type can just be thought of as a regular class type. It's just that we have syntactic sugar for its `toString()` method.

I wonder if we should just get rid of this, and have differentiating logic in the `toString()` and `export()` methods of `ClassType`. For example:

```java
public String toString() {
  if (clazz.isFunctionNType()) {
    // render as function
  } else {
    // render as class
  }
}
```

The subtyping logic should be the same; the same subtyping rules follow for functions as regular classes; the variance is all described in the signature, e.g. `Function2<in Param1, in Param2, out Result>`.

pkl-core/src/main/java/org/pkl/core/ast/type/TypeNode.java
```diff
@@ -1965,66 +1697,25 @@ public final VmTyped getReturnTypeMirror() {
...
-      return new PType.Function(parameterTypes, returnTypeNode.doExport());
-    }
-
     @Override
     protected boolean acceptTypeNode(boolean visitTypeArguments, TypeNodeConsumer consumer) {
       return consumer.accept(this);
     }
 
+    protected VmClass getVmClass() {
+      return ((VmType.FunctionType) getType()).getVmClass();
```
bioball: Since this is used in a specialization guard, it'd be unnecessarily expensive to allocate a `VmType` just to get its class.

```suggestion
      return BaseModule.getFunctionNClass(parameterTypeNodes.length);
```

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
```diff
@@ -630,9 +605,32 @@ private void resolve(ClassProperty prop, TypeNode typeNode) {
...
+          .withSourceSection(prop.getHeaderSection())
+          .evalError("commandOptionUnsupportedType", prop.getName(), "", type.toString())
+          .withHint(
+              type.getVmClass().getSimpleName()
+                  + " options must provide "
+                  + switch (type.getVmClass().getTypeParameterCount()) {
+                    case 1 -> "one type argument.";
+                    case 2 -> "two type arguments.";
+                    default -> throw PklBugException.unreachableCode();
+                  })
```
bioball: Move switch expression to its own var declaration

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
```diff
@@ -656,137 +654,142 @@ private OptionBehavior resolveTypeArgument(
...
+          if (all == null) all = this::allChooseLast;
+          if (multiple == null) multiple = false;
+          if (metavar == null) metavar = METAVAR_FLOAT;
+          return true;
+        } else if (ct.getVmClass() == BaseModule.getIntClass()) {
+          if (each == null) each = eachLong(Long.MIN_VALUE, Long.MAX_VALUE, METAVAR_INT);
+          if (all == null) all = this::allChooseLast;
+          if (multiple == null) multiple = false;
+          if (metavar == null) metavar = METAVAR_INT;
+          return true;
```
bioball, line 744: I realize this isn't a _changed_ style introduced in your PR, but since these lines are getting touched: can we add braces to all of these conditionals?

E.g. this (and apply this everywhere else)

```suggestion
          if (each == null) {
            each = eachLong(Long.MIN_VALUE, Long.MAX_VALUE, METAVAR_INT);
          }
          if (all == null) {
            all = this::allChooseLast;
          }
          if (multiple == null) {
            multiple = false;
          }
          if (metavar == null) {
            metavar = METAVAR_INT;
          }
```

pkl-core/src/main/java/org/pkl/core/ast/member/ClassProperty.java
```diff
@@ -101,4 +101,9 @@ public PClass.Property export(PClass owner) {
         name.toString(),
         PropertyTypeNode.export(typeNode));
   }
+
+  public VmType getType() {
+    if (typeNode == null) return VmType.UNKNOWN;
+    return typeNode.getTypeNode().getType();
+  }
```
bioball: This is misleading; a class property can possibly be treated as an object property, whose actual type is the super:

```pkl
open class Base {
  prop: String
}

class Child extends Base {
  prop = "foo" // typeNode is null, but type-checked as `String`
}
```

Maybe okay to preserve a `getType()` method if it's marked `@Nullable`. But it's also only used in `VmReference` so maybe just get rid of this and move it into `VmReference` instead?

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,808 @@
...
+    protected final VmClass clazz;
+
+    protected SelfType(VmClass clazz) {
+      this.clazz = clazz;
+    }
+
+    @Override
+    public @Nullable VmClass getVmClass() {
+      return clazz;
+    }
```
bioball, line 121: I wonder if it's logically correct to override this; a self type is not the same as a class type.

Looking through all the usages of `getVmClass()`, it looks like it's always used to compare to known classes.

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,808 @@
...
+    public String toString() {
+      return (elementType instanceof ClassType ct && ct.isFunctionNClass())
+              || elementType instanceof UnionType
+          ? "(" + elementType + ")?"
+          : elementType + "?";
+    }
+
+    @Override
+    public int hashCode() {
+      return elementType.hashCode();
```
bioball: Shouldn't conflict with the element type's hash code

```suggestion
      return 31 * elementType.hashCode();
```

Same comment for any other `hashCode()` implementations that just return another value's hash code

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,808 @@
...
+    @Override
+    public PType export() {
+      return new Union(exportTypes(elementTypes));
+    }
+
+    @Override
+    protected boolean doIsEquivalentTo(VmType other) {
+      if (!(other instanceof UnionType t)) return false;
+      return typesEquals(elementTypes, t.elementTypes);
+    }
```
bioball, line 635: Ideally, we can define `A | B` as equivalent to `B | A`.

Maybe one solution here, since we have hash codes, is to sort both by their hash code, then compare each type.

Or, we can just leave a TODO comment here and move on.

bioball (changes requested):
This mostly looks good! Left some commentts

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
HT154: This is because the element types of the union are now a `List` from `UnionType` instead of being a `Set` from `UnionOfStringLiteralsTypeNode`. I figured it may be better to reflect the command's declared element order than forcing a sort. Open to changing this if you feel strongly.

pkl-core/src/main/java/org/pkl/core/ast/member/ClassProperty.java
HT154: I don't think this is an issue in practice because above the property is accessed `VmClass.getAllProperties()`, which in this case returns the `prop: String` because it omits properties without type annotations if they exist in a superclass.

I rm'd `ClassProperty.getType()` in favor of doing this all in VmReference and added a comment documenting this interaction.

pkl-core/src/main/java/org/pkl/core/runtime/CommandSpecParser.java
bioball: Yeah, sounds better to keep the declared order. And this feels like a minor change, maybe not worth calling out in the release notes (help text isn't an API).

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
HT154: Removing this actually breaks `Class<this>` type checks (like the one on `Any.getClass()`)

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,736 @@
...
+    return false;
+  }
+
+  public @Nullable VmClass getVmClass() {
+    return null;
+  }
+
+  public @Nullable VmTypeAlias getVmTypeAlias() {
+    return null;
+  }
```
bioball: There's only one override and only one caller of the polymorphic `TypeNode.getVmTypeAlias()` (`typeAliasTypeFactory` in `MirrorFactories`).

Maybe just remove this on the base `VmType` and change the call within `typeAliasTypeFactory` to a cast? Then we'd avoid dynamic dispatch.

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,734 @@
...
+      return PType.NOTHING;
+    }
+
+    @Override
+    public String toString() {
+      return "nothing";
+    }
+  }
+
+  public abstract static sealed class SelfType extends VmType {
```
bioball: Per our convo earlier: I think we should have a different treatment of final and non-final self types.

These two types can be thought of as equivalent typechecks; and `equals` and `isSubtypeOf` should hold:

```pkl
class MyFinalClass {
  prop1: this
  prop2: MyFinalClass
}
```

pkl-core/src/test/kotlin/org/pkl/core/runtime/VmTypeTest.kt
```diff
@@ -0,0 +1,369 @@
...
+        ),
+        SubtypeCase(
+          "class: subtype of Any (unparameterized)",
+          VmType.ClassType(BaseModule.getListClass()),
+          anyClass,
+        ),
+        SubtypeCase(
+          "class: subtype (parameterized, direct inheritance)",
+          VmType.ClassType(BaseModule.getListClass()),
+          VmType.ClassType(BaseModule.getCollectionClass()),
```
bioball: This tests un-parameterized types?

pkl-core/src/test/kotlin/org/pkl/core/runtime/VmTypeTest.kt
```diff
@@ -0,0 +1,369 @@
...
+        ),
+        SubtypeCase(
+          "class: subtype (parameterized vs. unparametetized)",
+          VmType.ClassType(BaseModule.getListClass(), stringClass),
+          VmType.ClassType(BaseModule.getListClass()),
+        ),
+        SubtypeCase(
+          "class: subtype of Any (parameterized)",
+          VmType.ClassType(BaseModule.getListClass()),
+          anyClass,
```
bioball: Same comment: `VmType.ClassType(BaseModule.getListClass())` is un-parameterized?

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,734 @@
...
+
+  public @Nullable VmTypeAlias getVmTypeAlias() {
+    return null;
+  }
+
+  /** Tells if this type is the same typecheck as the other type. */
+  @Override
+  public final boolean equals(Object value) {
+    return this == value || value instanceof VmType other && doIsEquivalentTo(other);
+  }
```
bioball: Two comments here:

1. `Object.equals` should always accept a `@Nullable Object`
2. There's an asymmetry here; we unwrap aliased types when it's `this`, but not when it's `value`. That means (`AliasType(intType).equals(intType)`, but `!intType.equals(AliasType(intType)`). We should be unwrapping `value` just like we do `this` during comparison.

We're also missing tests for this, because `VmTypeTest` only covers `isSubtypeOf` and `isSupertypeOf`, and not `equals`.

```suggestion
  public final boolean equals(@Nullable Object value) {
    if (this == value) return true;
    if (value instanceof AliasType aliasType) {
      return equals(aliasType.aliasedType);
    }
    if (value instanceof VmType vmType) {
      return doIsEquivalentTo(vmType);
    }
    return false;
  }
```

pkl-core/src/test/kotlin/org/pkl/core/runtime/VmTypeTest.kt
```diff
@@ -0,0 +1,369 @@
...
+          boolClass,
+          typeVar,
+          bidi = true,
+        ),
+      )
+    }
+  }
+
+  @ParameterizedTest
+  @MethodSource("getSubtypeCases")
```
bioball: Non-blocking comment: we tend to prefer individual `@Test` cases, because parameterized tests are a little hard to use (no "run/debug" gutter icons for each test case in IntelliJ).

pkl-core/src/test/kotlin/org/pkl/core/runtime/VmTypeTest.kt
```diff
@@ -0,0 +1,369 @@
...
+          emptyList(),
+          VmModifier.NONE,
+          PClassInfo.forModuleClass("modB", URI("test:/modB.pkl")),
+          emptyList(),
+          VmUtils.createEmptyModule(),
+        )
+      modB.initSupertype(
+        TypeNodeFactory.NonFinalClassTypeNodeGen.create(VmUtils.unavailableSourceSection(), modA),
+        modA,
+      )
```
bioball: Suggestion: use the "front door" to reify these types.

E.g. this:

```kt
    private fun makeModule(text: String): VmTyped {
      var ret: VmTyped? = null
      VmUtils.createContext {
          val vmContext = VmContext.get(null)
          vmContext.initialize(
            VmContext.Holder(
              StackFrameTransformers.defaultTransformer,
              SecurityManagers.defaultManager,
              HttpClient.dummyClient(),
              ModuleResolver(listOf(ModuleKeyFactories.standardLibrary)),
              ResourceManager(SecurityManagers.defaultManager, listOf()),
              Loggers.noop(),
              mapOf(),
              mapOf(),
              null,
              null,
              null,
              null,
              TraceMode.COMPACT,
              false,
            )
          )
          val moduleKey = ModuleKeys.synthetic(URI("repl:text"), text)
          ret = VmLanguage.get(null).loadModule(moduleKey)
        }
        .close()
      return ret!!
    }

    private fun VmTyped.getTypeForProperty(property: String): VmType {
      return vmClass.getProperty(Identifier.get(property))!!.typeNode!!.typeNode.type
    }
```

Then you can create a type via:

```kotlin
    val module =
      makeModule(
        """
        class Person

        prop: Person
        """
          .trimIndent()
      )
    val personClassType = module.getTypeForProperty("prop")
```

This approach requires knowing nothing about the internals of the various truffle nodes, so that as we make changes to the truffle node tree, we don't need to touch `VmTypeTest`.

bioball (changes requested):
Getting really close!

pkl-core/src/test/kotlin/org/pkl/core/runtime/VmTypeTest.kt
HT154: Ah, two different things: `List`/`Collection` are parameterized classes, but these types are not. I'll make sure this gets clarified.

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
HT154: Good catch. Opted to unwrap aliases in a loop instead of relying on the recursive call.

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,767 @@
...
+      super(clazz, typeArguments);
+    }
+
+    private static final ClassType ANY = new ClassType(BaseModule.getAnyClass());
+
+    @Override
+    public PType export() {
+      return clazz.isFunctionNClass()
+          ? new PType.Function(
+              exportTypes(typeArguments, 1), typeArguments[typeArguments.length - 1].export())
```
stackoverflow, line 319: This assumes args and return have at least 1 entry, but the assumption doesn't hold:
```pkl
import "pkl:ref"
class D extends ref.Domain {
  function renderReference(reference: ref.Reference<this, Any>): String = "ref"
}
local d: D = new {}
class Holder { f: Function1 | Int }
res = ref.Reference(d, Holder, null).f
```
```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 0
        at org.pkl.core.runtime.VmType$ClassType.toString(VmType.java:335)
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,767 @@
...
+      this.typeArguments = typeArguments;
+    }
+
+    public VmType[] getTypeArguments() {
+      return typeArguments;
+    }
+
+    public VmType.ClassType withTypeArguments(VmType[] typeArguments) {
+      return new VmType.ClassType(clazz, typeArguments);
+    }
```
bioball: Move this method to `VmType.ClassType`

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,767 @@
...
+  public abstract static sealed class AbstractClassType extends VmType {
+    protected final VmClass clazz;
+    protected final VmType[] typeArguments;
+
+    protected AbstractClassType(VmClass clazz, VmType[] typeArguments) {
+      this.clazz = clazz;
+      this.typeArguments = typeArguments;
+    }
+
+    public VmType[] getTypeArguments() {
```
bioball:
```suggestion
    public final VmType[] getTypeArguments() {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,767 @@
...
+      return typeArguments;
+    }
+
+    public VmType.ClassType withTypeArguments(VmType[] typeArguments) {
+      return new VmType.ClassType(clazz, typeArguments);
+    }
+
+    @Override
+    public VmClass getVmClass() {
```
bioball:
```suggestion
    public final VmClass getVmClass() {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,767 @@
...
+      return clazz;
+    }
+
+    @Override
+    public boolean isParametric() {
```
bioball:
```suggestion
    public final boolean isParametric() {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,767 @@
...
+      return typeArguments.length > 0;
+    }
+
+    public boolean isFunctionNClass() {
```
bioball:
```suggestion
    public final boolean isFunctionNClass() {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,767 @@
...
+          case INVARIANT -> state[i].equals(goalState[i]);
+          case COVARIANT -> state[i].isSubtypeOf(goalState[i]);
+          case CONTRAVARIANT -> state[i].isSupertypeOf(goalState[i]);
+        }) return false;
+      }
+      return true;
+    }
+
+    @Override
+    public int hashCode() {
```
bioball:
```suggestion
    public final int hashCode() {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,767 @@
...
+    public boolean isParametric() {
+      return typeArguments.length > 0;
+    }
+
+    public boolean isFunctionNClass() {
+      return clazz.isFunctionNClass();
+    }
+
+    @Override
+    protected boolean doIsEquivalentTo(VmType other) {
```
bioball:
```suggestion
    protected final boolean doIsEquivalentTo(VmType other) {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,767 @@
...
+      if (!(other instanceof AbstractClassType t)) return false;
+      if (clazz != t.clazz) return false;
+      return typesEquals(typeArguments, t.typeArguments);
+    }
+
+    @Override
+    protected boolean doIsSupertypeOf(VmType other) {
```
bioball:
```suggestion
    protected final boolean doIsSupertypeOf(VmType other) {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,767 @@
...
+
+  private abstract static sealed class NonFinalSelfType extends VmType {
+    private final VmClass clazz;
+
+    protected NonFinalSelfType(VmClass clazz) {
+      this.clazz = clazz;
+    }
+
+    @Override
+    public VmClass getVmClass() {
```
bioball:
```suggestion
    public final VmClass getVmClass() {
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,767 @@
...
+    @Override
+    public String toString() {
+      if (typeArguments.length == 0) return typeAlias.getDisplayName();
+      return appendTypeArguments(new StringBuilder(typeAlias.getDisplayName()), typeArguments)
+          .toString();
+    }
+
+    @Override
+    public int hashCode() {
+      return 31 * typeAlias.hashCode() + Arrays.hashCode(typeArguments);
```
bioball: If two objects are `equals()`, they should have the same `hashCode()`.

Since we've defined alias equality as "equal to the aliased type", the hash code should just be the aliased type'a hash code.

```suggestion
      return aliasedType.hashCode();
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,767 @@
...
+    }
+
+    @Override
+    public String toString() {
+      return typeParameter.getName();
+    }
+
+    @Override
+    public int hashCode() {
+      return 31 * typeParameter.hashCode();
```
bioball: Same comment WRT "if they're equals, they should have the same hash code". I think we can just pick a random constant here.

```suggestion
      return -318019;
```

pkl-core/src/main/java/org/pkl/core/runtime/VmType.java
```diff
@@ -0,0 +1,767 @@
...
+              ? "(" + baseType + ")"
+              : baseType)
+          + "("
+          + String.join(", ", constraints)
+          + ")";
+    }
+
+    @Override
+    public int hashCode() {
+      return 31 * baseType.hashCode() + Arrays.hashCode(constraints);
```
bioball:
```suggestion
      return 31 * identity;
```

pkl-core/src/test/kotlin/org/pkl/core/runtime/VmTypeTest.kt
```diff
@@ -0,0 +1,450 @@
...
+
+  /** Assert this is a subtype of [other] and [other] is a subtype of this, but not equivalent */
+  private fun VmType.bidi(other: VmType) {
+    assertThat(this.isSubtypeOf(other)).isTrue
+    assertThat(other.isSupertypeOf(this)).isTrue
+    assertThat(other.isSubtypeOf(this)).isTrue
+    assertThat(this.isSupertypeOf(other)).isTrue
+    assertThat(this.equals(other)).isFalse
+    assertThat(other.equals(this)).isFalse
+  }
```
bioball, line 135: [nit] This makes the errors more descriptive if one fails
[nit] use Kotlin `==` instead of `Object.equals`

```suggestion
  private fun VmType.sub(other: VmType) {
    assertThat(this.isSubtypeOf(other)).`as`("%s isSubtypeOf %s", this, other).isTrue
    assertThat(other.isSupertypeOf(this)).`as`("%s isSupertypeOf %s", other, this).isTrue
    assertThat(other.isSubtypeOf(this)).`as`("%s isSubtypeOf %s", other, this).isFalse
    assertThat(this.isSupertypeOf(other)).`as`("%s isSupertypeOf %s", this, other).isFalse
    assertThat(this == other).`as`("%s == %s", this, other).isFalse
    assertThat(other == this).`as`("%s == %s", other, this).isFalse
  }

  /**
   * Assert this is a subtype of [other] and [other] is a supertype of this and they are equivalent
   */
  private fun VmType.eq(other: VmType) {
    assertThat(this.isSubtypeOf(other)).`as`("%s isSubtypeOf %s", this, other).isTrue
    assertThat(other.isSupertypeOf(this)).`as`("%s isSupertypeOf %s", other, this).isTrue
    assertThat(other.isSubtypeOf(this)).`as`("%s isSubtypeOf %s", other, this).isTrue
    assertThat(this.isSupertypeOf(other)).`as`("%s isSupertypeOf %s", this, other).isTrue
    assertThat(this == other).`as`("%s == %s", this, other).isTrue
    assertThat(other == this).`as`("%s == %s", other, this).isTrue
  }

  /** Assert this is a subtype of [other] and [other] is a subtype of this, but not equivalent */
  private fun VmType.bidi(other: VmType) {
    assertThat(this.isSubtypeOf(other)).`as`("%s isSubtypeOf %s", this, other).isTrue
    assertThat(other.isSupertypeOf(this)).`as`("%s isSupertypeOf %s", other, this).isTrue
    assertThat(other.isSubtypeOf(this)).`as`("%s isSubtypeOf %s", other, this).isTrue
    assertThat(this.isSupertypeOf(other)).`as`("%s isSupertypeOf %s", this, other).isTrue
    assertThat(this == other).`as`("%s == %s", this, other).isFalse
    assertThat(other == this).`as`("%s == %s", other, this).isFalse
  }
```

bioball (approved):
Just some minor comments, but approving to unblock!

stackoverflow (approved):
LGTM

PR #1845, opened 2026-09-03
Optimize `new` inference in method args

HT154:
Only store the method in an aux slot if we know the call contains an argument that is an inferred `new` expr. This avoids taking the perf hit of walking aux slots on every method call.


bioball (approved):
LGTM!

PR #1846, opened 2026-09-03
Escape characters that YAML cannot carry literally

dylanpulver:
`YamlRenderer` emits some characters as plain scalars that a YAML reader cannot accept. On released 0.32.1:

```
$ printf 'v = "a\\u{7f}b"' > t.pkl && pkl eval -f yaml t.pkl | od -c | head -1
0000000   v   :       a 177   b  \n
```

0x7F and the C1 block are outside `c-printable` (spec 5.1), so a conforming reader must reject the document; NEL, LS and PS are line breaks under YAML 1.1, so a 1.1 reader silently splits the value instead. Scanning U+0000-U+00FF plus a few outliers, **37 of 265 codepoints fail to survive a render/parse round-trip through PyYAML**, while U+0000-U+001F all pass.

Cause: `YamlEmitter` decides `needsEscaping` at two sites, both `ch < 0x20`, so nothing else ever reaches `YamlEscaper`. That also means the escapes #1165 added for 0x85, U+2028 and U+2029 only fire when some *other* character opens the gate — `"a\u{85}b"` renders raw, `"a\u{1}\u{85}b"` renders `"a\x01\Nb"`.

The fix shares one predicate between those two sites, and gives 0x7F-0x9F the 8-bit escape the table lacks.

Verification: `:pkl-core:test` 1643 tests, 0 failures, 3 skipped. The regenerated snippet output re-parses under PyYAML, 7/7 new cases back to the original strings. Three mutants each fail the snippet test: reverting both files, widening only the predicate, and extending only the escaper — so both halves are load-bearing.

Updated after review: U+FFFE and U+FFFF now route through `IoUtils.toUnicodeEscape`, so the gap this originally left is closed. Snippet cases h14 and h15 cover both, and reverting the two sources fails exactly those assertions.


pkl-core/src/main/java/org/pkl/core/util/yaml/YamlEmitter.java
```diff
@@ -42,6 +42,17 @@ public static YamlEmitter create(StringBuilder builder, String mode, String inde
...
   }
 
+  /**
+   * Tells if {@code ch} has to go through {@link YamlEscaper} instead of being emitted literally.
+   *
+   * <p>0x7F and the C1 block are outside {@code c-printable} (YAML 1.2 spec 5.1), and NEL, LS and
+   * PS are line breaks in YAML 1.1. Note that 0x0A is handled by its own {@code case} in {@link
+   * #emit(String, StringBuilder, boolean)} and never reaches here.
+   */
+  private static boolean mustEscape(char ch) {
```
HT154: Method name is a bit wonky. Maybe `getNeedsEscaping`?
```suggestion
  private static boolean getNeedsEscaping(char ch) {
```

HT154 (commented):
Thanks for the fix!

> Gap left: U+FFFE and U+FFFF still go out literally. They need a `\u` escape, which `IoUtils.toHexEscape` cannot emit; glad to add it here.

Wouldn't `IoUtils.toUnicodeEscape` work in this scenario?

HT154 (approved):
LGTM

PR #1847, opened 2026-09-03
Error when inferring a self type in a method argument

HT154:


pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinMethodArgumentNode.java
```diff
@@ -85,4 +85,16 @@ protected final Object eval(VirtualFrame frame) {
...
   }
+
+  @Override
+  protected Object getDefaultValue(
+      VirtualFrame frame,
+      @Nullable TypeNode typeNode,
+      SourceSection headerSection,
+      String qualifiedName) {
+    if (typeNode != null && typeNode.isSelfType())
+      throw exceptionBuilder().evalError("cannotInferParent").build();
```
bioball: [nit] style-wise, only simple statements (e.g. `break`, `continue`, `return foo`) can have braces omitted.

Also, this might need a `CompilerDirectives.transferToInterpreterAndInvalidate()`; try running `./gradlew pkl-cli:buildNative` first before adding.

```suggestion
    if (typeNode != null && typeNode.isSelfType()) {
      throw exceptionBuilder().evalError("cannotInferParent").build();
    }
```

bioball (approved):

PR #1848, opened 2026-09-04
Use `ChoiceFormat` for better plurals in error messages

HT154:


pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
```diff
@@ -207,7 +207,7 @@ expectedModuleAsArgument=\
 Expected a module as argument, but got an object that amends a module.
 
 wrongTypeArgumentCount=\
-Expected {0} type argument(s) but got {1}.
+Expected {0} type argument{0,choice,0#s|1#|2<s} but got {1}.
```
stackoverflow: `2<` means "greater than 2", right? And 2 is not greater than 2, so it will wrongly show the singular version. Same for line 313.
Worth a test.

stackoverflow (approved):
Only one comment. Approving to unblock.

bioball (approved):

pkl-core/src/main/resources/org/pkl/core/errorMessages.properties
HT154: Oof good catch!

PR #1849, opened 2026-09-04
Add user generic methods

HT154:
Changes:
- Allow generic methods to be declared outside the standard library
- Allow object (`local`) methods to be generic
- Allow generic method call sites to specify type arguments explicitly
  - Unchanged: when no explicit type argument is provided, type parameter type checks are no-ops
- Remove validation that `ref.Reference` type annotations may not contain constraints in the referent position. Any constraints are now silently dropped.
- Change the `ref.Reference()` constructor signature:
```pkl
import "pkl:ref"
before = ref.Reference(myDomain, MyReferent, myData)
after = ref.Reference::<MyDomain, MyReferent>(myDomain, myData)

// this can be cleaned up behind a generic typealias and method:
typealias MyRef<T> = ref.Reference<MyDomain, T>
function MyRef<T>(data: Any): MyRef<T> = ref.Reference::<MyDomain, T>(myDomain, data)
after2 = MyRef::<MyReferent>(myData)

// and references can now be constructed directly to complex types
complex1 = MyRef::<Listing<String>>(null)
complex2 = MyRef::<String | Int>(null)
complex3 = MyRef::<this>(null)
complex4 = MyRef::<(String) -> Int>(null)
```

Depends on #1838


pkl-core/src/main/java/org/pkl/core/runtime/VmTypeArgument.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    return rootNode
+        .getCallTarget()
+        .call(
+            enclosingFrame.getArguments()[0],
+            enclosingFrame.getArguments()[1],
+            enclosingFrame.getArguments()[2],
+            value);
+  }
```
bioball, line 51: Let's introduce a new node that evals the type check, maybe called `EvalTypeArgumentNode`.

Take a look at the interaction between `VmFunction` and `ApplyVmFunction1Node`, in particular, which is a good model for how to do this.

* `VmTypeArgument` exposes method `RootCallTarget getCallTarget()`
* `EvalTypeArgumentNode` creates a child `DirectCallNode` (in the case where the call target) or `IndirectCallNode` (if the call target changes).
* `TypeVariableNode` has a `@Child EvalTypeArgumentNode`, which it uses to execute the typecheck.

The benefit of this approach is that it allows us to use a `DirectCallNode` in the common case, which Truffle knows how to inline.

BTW: we don't necessarily need `EvalTypeArgumentNode`, since `TypeVariableNode` is the only call site. However, this seems useful to have (we might want to eval type arguments elsewhere, e.g. in stdlib methods).

bioball (commented):
Some issues that I've found while playing around with this:

This currently blows up:

```pkl
function foo<T>(it): Listing<T> = new Listing { it }

res = foo::<String>(5)
```

Fails with:

```
–– Pkl Error ––
class java.lang.Long cannot be cast to class [Lorg.pkl.core.runtime.VmTypeArgument; (java.lang.Long is in module java.base of loader 'bootstrap'; [Lorg.pkl.core.runtime.VmTypeArgument; is in unnamed module of loader 'app')

1 | function foo<T>(it): Listing<T> = new Listing { it }
                                 ^
at test#foo (/Users/danielchao/code/apple/pkl/.dan-scripts/test.pkl:1)

1 | function foo<T>(it): Listing<T> = new Listing { it }
                                                    ^^
at test#foo[#1] (/Users/danielchao/code/apple/pkl/.dan-scripts/test.pkl:1)

132 | renderer.renderDocument(value)
at pkl.base#Module.output.text (https://github.com/apple/pkl/blob/63604966/stdlib/base.pkl#L132)

136 | if (renderer is BytesRenderer) renderer.renderDocument(value) else text.encodeToBytes("UTF-8")
                                                                         ^^^^
at pkl.base#Module.output.bytes (https://github.com/apple/pkl/blob/63604966/stdlib/base.pkl#L136)

Pkl 0.33.0-dev+63604966 (macOS 27.0, Java 21.0.6)

java.lang.ClassCastException: class java.lang.Long cannot be cast to class [Lorg.pkl.core.runtime.VmTypeArgument; (java.lang.Long is in module java.base of loader 'bootstrap'; [Lorg.pkl.core.runtime.VmTypeArgument; is in unnamed module of loader 'app')
	at org.pkl.core.ast.type.TypeNode$TypeVariableNode.executeLazily(TypeNode.java:2029)
	at org.pkl.core.ast.type.TypeNode.execute(TypeNode.java:128)
	at org.pkl.core.ast.member.ListingOrMappingTypeCastNode.executeImpl(ListingOrMappingTypeCastNode.java:54)
	at org.pkl.core.ast.PklRootNode.execute(PklRootNode.java:56)
```

---

Type node defaults don't seem to work; this 

```pkl
function foo<T>(): Listing<T> = new { 
  new { name = "bob" }
}

class Person {
  name: String
}

res = foo::<Person>()
```

Fails with:

```
–– Pkl Error ––
Tried to read property `default` but its value is undefined.

1 | function foo<T>(): Listing<T> = new {
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
at test#foo.default (/Users/danielchao/code/apple/pkl/.dan-scripts/test.pkl:1)

The above error occurred when rendering path `res.default` of module `file:///Users/danielchao/code/apple/pkl/.dan-scripts/test.pkl`.

2 | new { name = "bob" }
    ^^^
at test#foo[#1] (/Users/danielchao/code/apple/pkl/.dan-scripts/test.pkl:2)

2 | new { name = "bob" }
    ^^^^^^^^^^^^^^^^^^^^
```

---

This doesn't check the type argument; this evals just fine but I would expect to see a type error.

I think this is because function type checks are just no-ops right now.
If we fix function checking (return a new function that wrap the old one and adds type checks), I'd expect this to work.

```pkl
res = List(1, 2, 3).map::<String>((it) -> it + 1)
```

pkl-core/src/main/java/org/pkl/core/runtime/VmTypeArgument.java
HT154: When would it be possible to use a `DirectCallNode`? If `TypeVariableNode` creates a single `EvalTypeArgumentNode` then wouldn't it need to accept the call target from any supplied `VmTypeArgument`? I don't see a way to know what that could be statically.

HT154:
> Some issues that I've found while playing around with this

All of these are resolved now! This touches a bunch more code now (and changes the arg layout of a ton more calls) to support `Listing`/`Mapping` lazy type checks. Added tests for all these cases and a few more to boot.

PR #1850, opened 2026-09-08
Avoid using `Object.toString()` for in-language API

bioball:
This changes code so that the in-language `toString()` doesn't use Java's `Object.toString()`.

This helps improve ergonomics when working on the language (can enable automatically calling `toString()` on debug values).

Additionally, external tools might call `toString()`; for example, Truffle's compiler sometimes will stringify node fields (e.g. when `-Dengine.CompilationFailureAction=Throw`) is set.


pkl-core/src/main/java/org/pkl/core/ast/internal/ToStringNode.java
```diff
@@ -75,11 +75,10 @@ protected String evalReference(
     return (String) callNode.call(value, value.getVmClass().getPrototype());
   }
 
-  @Fallback
-  @Override
+  @Specialization
   @TruffleBoundary
-  protected Object fallback(Object value) {
-    return value.toString();
+  protected String evalVmValue(VmValue value) {
```
bioball, line 79: All other types of values are already handled in their own specializations; no need to use a fallback here.

HT154 (approved):

pkl-core/src/main/java/org/pkl/core/runtime/VmObject.java
```diff
@@ -204,11 +204,21 @@ public final void force(boolean allowUndefinedValues) {
...
 
-  public final String toString() {
+  @TruffleBoundary
+  public final String toPklString() {
     force(true, true);
     return VmValueRenderer.singleLine(Integer.MAX_VALUE).render(this);
   }
 
+  /**
+   * Override default implementation because it calls {@link #hashCode()}, which will do object
```
stackoverflow: What about `VmList/VmSet/VmMap/VmPair`? Won't they trigger `hashCode`?

stackoverflow (approved):
Looks good. But what about `protobuf/RendererNodes.java` (visitIntSeq/visitPair/visitRegex/visitNull/etc.)?

pkl-core/src/main/java/org/pkl/core/runtime/VmObject.java
bioball: Good point; moving the `toString()` impl to `VmValue` itself.

I'm also changing the impl to use `VmValueRenderer`, which doesn't force and produces much nicer views for debugging.

PR #1853, opened 2026-09-09
Optimize super property and super subscript nodes

bioball:
This adds cacheing to ReadSuper* nodes so that the same node being executed can possibly re-use a previously cached value.


pkl-core/src/main/java/org/pkl/core/ast/expression/member/ReadSuperPropertyNode.java
```diff
@@ -36,15 +43,43 @@ public ReadSuperPropertyNode(
...
+    var constantValue = property.getConstantValue();
+    if (constantValue != null) return constantValue; // TODO: type check
 
-    // start from the parent of the owner of the `super.<propertyName>` expression
-    // skip any function object owners (same as when resolving `this`)
-    // `receiver` must be passed on unchanged to make sure that overridden properties still take
-    // effect
+    return callNode.call(
+        property.getCallTarget(),
+        // TODO: should the marker only turn off constraint checking, not overall type checking?
```
bioball, line 76: These TODOs just moved lines; they're not new TODOs introduced by this PR.

HT154 (approved):
LGTM! Very nice 🎉

stackoverflow (approved):

PR #1856, opened 2026-09-10
Fix reading custom this

bioball:
CustomThisNodes can possibly be executed with different frames. Because of this, it's not safe to cache the result of the auxiliary slot lookup.

Note: we can still cache the slot in `WithCustomThisExpression`, because it's the body node of `ExecuteCustomThisWithRootNode`, which is given a new frame with fresh auxiliary slots.

Also, change `VmUtils.findCustomThisSlot` to use the more efficient `findOrAddAuxiliarySlot` method underneath the hood.


pkl-core/src/main/java/org/pkl/core/runtime/VmUtils.java
```diff
@@ -1080,10 +1080,12 @@ public static Object evaluateExpression(
...
   }
 
-  public static int findCustomThisSlot(VirtualFrame frame) {
-    var result = frame.getFrameDescriptor().getAuxiliarySlots().get(CUSTOM_THIS_FRAME_SLOT_ID);
-    assert result != null;
-    return result;
+  @TruffleBoundary
+  public static int findCustomThisSlot(FrameDescriptor descriptor) {
+    var prevSlotCount = descriptor.getNumberOfAuxiliarySlots();
+    var idx = descriptor.findOrAddAuxiliarySlot(CUSTOM_THIS_FRAME_SLOT_ID);
```
bioball, line 1086: For some reason, there's no `descriptor.getAuxiliarySlotName(int index)` method.

As far as I can tell, this is the fastest way to look up an auxiliary slot index given its name.

HT154 (approved):
Approving to unblock. Can you add a test that exercises this?

bioball:
Closing this; one of the side-effects of https://github.com/apple/pkl/pull/1857 is that this fix is no longer needed.

PR #1857, opened 2026-09-11
Switch inferred `new {}` expressions in method args to using frame slots

bioball:
This changes the logic for inferring `new {}` expressions in method arguments to using regular frame slots, instead of auxiliary slots.

Overall, this is a performance improvement: the nodes are statically initialized with the frame slot index already, which makes this code much more PE friendly.

* At parse time, get or create a frame slot that holds the method
* Pass method slot directly into truffle nodes
* Introduce `ArgInfo` record (preserve primitive int) over using `Pair`
* Remove unneeded `isImplicitNewExpr` method; this is not needed because we can just keep track of the method slot on the frame descriptor builder.


pkl-core/src/main/java/org/pkl/core/ast/expression/member/InferParentWithinMethodArgumentNode.java
```diff
@@ -16,41 +16,30 @@
...
+      SourceSection sourceSection, VmLanguage language, int argIndex, int methodSlot) {
     super(sourceSection, language);
     this.argIndex = argIndex;
-  }
-
-  @TruffleBoundary
-  private int getMethodSlot(FrameDescriptor frameDescriptor) {
-    var methodSlot = frameDescriptor.getAuxiliarySlots().get(VmUtils.METHOD_FRAME_SLOT_ID);
-    if (methodSlot == null) {
-      // used in intrinsic constructor e.g. pkl.base#List()
```
HT154, line 45: Where is this case handled now?

HT154 (approved):
LGTM!

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -2334,32 +2327,15 @@ public List<TypeParameter> visitTypeParameterList(@Nullable TypeParameterList ct
   }
 
   @Override
-  public Pair<ExpressionNode[], Boolean> visitArgumentList(ArgumentList argumentList) {
+  public ArgInfo visitArgumentList(ArgumentList argumentList) {
```
stackoverflow, line 2330: Now this method returns a *per frame descriptor* flag. Before it returned a *per call site* flag. It's a semantic change.
Once an argument is inferred it allocates the slot. Subsequent calls might do inference, even when not needed.
```pkl
class Foo { bar: Int = 1 }
const function chk2(i: Int): Boolean = i > 0
const function mk(f: Foo): Listing<Int> = new Listing { f.bar }
class C {
  xs: Listing<Int(chk2(this))> = mk(new { bar = 2 })
}
c: C = new {}
r = c.xs.toList()
```
```
–– Pkl Error ––
Index 0 out of bounds for length 0

...

java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
```

This also fails with the same error:
```pkl
class Foo { bar: Int = 1 }
const function chk(f: Foo): Boolean = f.bar > 0
const function mk(f: Foo): Listing<Int> = new Listing { f.bar }
class C {
  xs: Listing<Int(chk(new { bar = 2 }))> = mk(new { bar = 1 })
}
c: C = new {}
r = c.xs.toList()
```
Though I think, this is related to the removed comment: `can't store the slot id as this node may be called from different root nodes`.

bioball: Good catch!

The problem here is `ListingOrMappingTypeCastNode`; it's given an empty frame descriptor, but should be using the enclosing type node's frame descriptor.

WRT the semantic change: that's intentional. One of the side-effects here is that sibling method calls get affected (e.g. `foo(new {}) + bar(1)`). However, reading/writing values off the frame are extremely cheap; this feels like enough of an edge case that perf differences should not be noticeable.

stackoverflow: > WRT the semantic change: that's intentional.

Yes. The semantic change is fine by itself. I was pointing out that it may lead to unexpected side-effects.

There's still some cases that throw now but worked before:
```pkl
class Foo { bar: Int }
const function isOk(f: Foo): Boolean = f.bar > 0
const function isPos(i: Int): Boolean = i > 0

typealias T = Int(isOk(new { bar = 1 }), isPos(this))

x: T = 5
```
```bash

java.lang.ArrayIndexOutOfBoundsException%3A%20Index%200%20out%20of%20bounds%20for%20length%200

–– Pkl Error ––
Index 0 out of bounds for length 0

7 | x: T = 5
       ^
```

```pkl
class Foo { bar: Int }
const function isOk(f: Foo): Boolean = f.bar > 0

function m(p: Int(isOk(new { bar = 1 }))): Int = p

x = m(3)
```
```bash

com.oracle.truffle.api.frame.FrameSlotTypeException%3A%20Frame%20slot%20kind%20Long%20expected%2C%20but%20got%20Object%20at%20frame%20slot%20index%200.

–– Pkl Error ––
Frame slot kind Long expected, but got Object at frame slot index 0.

┌─ 2 repetitions of:
│ 4 | function m(p: Int(isOk(new { bar = 1 }))): Int = p
│                       ^^^^^^^^^^^^^^^^^^^^^
│ at foo#m (file:///Users/islonscherer/dev/apple/pkl-test/src/foo.pkl, line 4)
└─

6 | x = m(3)
        ^^^^
```

bioball: I ended up rewriting some of this code:

* Keep track of whether the method call itself requires a method slot by using a "state" field on the frame descriptor builder. This eliminates the edge case where sibling method calls also handle "methodSlot" type logic (note: this still affects nested calls, e.g. `foo(bar(new {}))`, but I think this is okay).
* Don't use a "ExecuteWithCustomThisNode" _unless_ we are inside a typealias body, because it's not actually needed elsewhere. This also affects how type nodes with let bindings work (they just work as-is!).

HT154 (commented):
It looks like #1853 got rolled into this PR. Was that intentional?

bioball:
> It looks like #1853 got rolled into this PR. Was that intentional?

Woah, not sure how that happened. Fixed.

HT154 (approved):

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
stackoverflow: Looks better, but there's still some issues. Your change now requires the constraint to be visited before the `scope.buildFrameDescriptor()` runs. You fixed that in AstBuilder line 2171 (`var parameterTypes = doVisitParameterTypes(paramListCtx);`) but I think there are more places:
```pkl
foo {
  local x: Int(let (q = 0) this > q) = 5
  z = x
}
```
and
```pkl
function f(): Int(let (q = 0) this > q) = 5
res1 = f()
```
both error with
```
–– Pkl Error ––
Index 0 out of bounds for length 0
```

But there's a more damning issue:
```pkl
local f = (a, b) -> new Dynamic {
  x = a
  y = b
}
res = (f) { p, q: Int(let (z = 0) this > z) ->
  x = p
  y = q
}.apply("hi", 9)
```
This returns
```
res {
  x = 0
  y = 9
}
```
in your branch instead of `x = "hi"` like in 0.32.
Not sure exactly what's happening, but it seems you are reading values from the wrong frame.

bioball: Ah, great catches! I fixed all of these, and also found a couple more places in AstBuilder where the built frame descriptor isn't respecting frame slots added by the type node, and introduced tests for these.

stackoverflow (approved):
Still found some stuff, but these seem to be existing bugs that were just unmasked by your changes:
```pkl
class Foo { bar: Int }
const function isOk2(f: Foo, n: Int): Boolean = f.bar > 0 && n > 0
local f = (a, b) -> new Dynamic { y = a; w = b }

res = (f) { p, q ->
  when (isOk2(new { bar = 1 }, p)) { z = p }
}.apply(1, 2)
```
In 0.32:
```
–– Pkl Error ––
Cannot tell which parent to amend.
```
In your branch:
```

org.pkl.core.runtime.VmBugException: Unknown Pkl data type `org.pkl.core.ast.member.ClassMethod@2954b5ea`
```

Another variant:
```pkl
class Foo { bar: Int }
const function isOk(f: Foo): Boolean = f.bar > 0
local f = (a) -> new Dynamic { y = a }

res = let (a = 1) let (b = 2) (f) { x ->
  when (isOk(new { bar = 1 })) { z = x + a + b }
}.apply(2)
```
In 0.32:
```
–– Pkl Error ––
Cannot tell which parent to amend.
```
In your branch:
```

java.lang.ArrayIndexOutOfBoundsException%3A%20Index%202%20out%20of%20bounds%20for%20length%202

–– Pkl Error ––
Index 2 out of bounds for length 2
```

This next one I also think is an old bug, now reachable via your changes:
```pkl
class Foo { bar: Int }
const function isOk(f: Foo) = f.bar > 0
typealias T<X> = Int(isOk(new { bar = 1 }) && this is X)
res2 = 3 is T<String>
```
In 0.32:
```
–– Pkl Error ––
Cannot tell which parent to amend.
```
In your branch:
```
res2 = true
```

I'm fine with merging this PR and creating issues for the leftovers.

bioball:
Great catches again! The first one was a bug in my code (method slot set in the wrong frame descriptor), and is fixed. The second one I think is an existing bug that's unrelated to this code.

PR #1860, opened 2026-09-11
Fix `Duration.isoString` for sub-nanosecond values

Dev-next-gen:
This is the follow-up I mentioned in #1859. While looking at `Duration.isoString` I noticed that a nonzero duration below nanosecond precision renders as a bare `"PT"`:

```pkl
a = 0.4.ns.isoString     // "PT"
b = (-0.4).ns.isoString  // "-PT"
c = 0.9.ns.isoString     // "PT"
```

`"PT"` has no component at all, so it is not a valid ISO 8601 duration, and `java.time.Duration.parse("PT")` throws `DateTimeParseException`. It also gets past the declared return type, since every group in `-?PT(\d+H)?(\d+M)?(\d+([.]\d+)?S)?` is optional.

The cause is in `DurationUtils.toIsoString`: hours, minutes, seconds and nanos all come out as 0, but the seconds part is only forced when `totalSeconds == 0`. I replaced that check with "every component is zero" and skip the sign in that case, so these durations render as `"PT0S"`, like `0.ms` and `(-0.0).ms` already do. Nothing changes for a duration that has at least one nonzero component, including the documented examples.

It doesn't depend on #1859, which changes how the fractional part is rounded in the same function. The two branches merge without conflict and the tests pass on the merged result. With #1859 the affected range shrinks from below 1 ns to below 0.5 ns, and the cases I added (`0.4.ns`, `-0.4.ns`) give `"PT0S"` either way.

I added cases to `DurationTest.toIsoString()` and to the `api/duration.pkl` snippet test. `DurationTest` fails on main with `expected: "PT0S" but was: "PT"` and passes with the fix, and `./gradlew :pkl-core:spotlessCheck :pkl-core:test` passes. I also checked the three expressions above with `jpkl eval` before and after. I did not run the full `./gradlew build` or `buildNative`.

AI tools used


HT154 (approved):
Looks good to me! Thanks for the fix

PR #1862, opened 2026-09-11
Fix invalid prefix when generating spring boot config

bioball:
In Spring Boot, the prefix passed to `@ConfigurationProperties` must be in "canonical form".

The presence of this annotation allows spring boot to directly inject properties into a class, but is not actually needed.

The correct behavior here is to omit this annotation if the property name does not match Spring Boot's canonical form.


pkl-commons/src/main/kotlin/org/pkl/commons/Strings.kt
```diff
@@ -86,3 +86,24 @@ fun shlex(input: String): List<String> {
...
+    [a-z0-9]* # followed by zero or more lowercase letters or digits
+    (?:-[a-z0-9]+)* # followed by possibly kebab-cased
+    (?:\.[a-z][a-z0-9]*(?:-[a-z0-9]+)*)* # followed by dot-separated nested prefixes
+    $
+    """
+      .trimIndent(),
+    setOf(RegexOption.COMMENTS, RegexOption.MULTILINE),
+  )
+
+/** Tells if */
```
HT154: Tells if what?

HT154 (approved):
Approving to unblock

PR #1869, opened 2026-09-21
Elide capturing of closures in functions and objects

bioball:
This implements an optimization where functions and objects don't retain a closure if they don't need it.

During parse time (inside AstBuilder), we determine whether a function or an object needs its enclosing scope.

This then propagates to the truffle node that creates the runtime objects.
If an object does not need to capture its closure, its enclosing frame does not get materialized.


pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -1350,7 +1370,8 @@ public ExpressionNode visitFunctionLiteralExpr(FunctionLiteralExpr expr) {
                   null,
                   exprNode);
 
-          return new FunctionLiteralNode(sourceSection, functionNode, isCustomThisScope);
+          return new FunctionLiteralNode(
+              sourceSection, functionNode, isCustomThisScope, scope.needsCapture());
```
HT154, line 1374: There's some inconsistency in how `needsCapture` gets accessed.

```suggestion
              sourceSection, functionNode, isCustomThisScope, scope.needsCapture);
```

HT154 (commented):
Very slick! Any idea what, if any, real world perf impact this has?

pkl-core/src/main/java/org/pkl/core/ast/builder/AstBuilder.java
```diff
@@ -2366,6 +2387,23 @@ public FrameDescriptor buildModuleFrameDescriptor() {
...
+    var isBaseModuleType =
+        !symbolTable.getModuleScope().properties.containsKey(name)
+            && BaseModuleMembers.hasProperty(name);
+    if (!isBaseModuleType) {
+      scope.markCapture(symbolTable.getModuleScope());
+    }
+    var identifier = org.pkl.core.runtime.Identifier.get(name);
+    return isBaseModuleType
+        ? new GetBaseModuleTypeNode(sourceSection, identifier)
+        : new ResolveSimpleDeclaredTypeNode(sourceSection, identifier, getModuleNode);
```
bioball: When resolving simple type names, we only care about knowing if this resolves to a base module type or not (.e.g. `new Listing {}`).

If not, the type needs to be resolved off the enclosing module, therefore, we need to capture.

pkl-core/src/test/files/LanguageSnippetTests/input/objects/outer3.pkl
```diff
@@ -0,0 +1 @@
+foo = outer
```
bioball: Surprisingly, we don't have an existing test for this. We need this code covered because `OuterNode` needs a closure.

bioball:
> Very slick! Any idea what, if any, real world perf impact this has?

Right now, I'm not seeing _any_ performance impact when testing against existing codebases, which I'm really surprised by. I ran a test against a project that gave me about an 83% capture rate (17% of objects skipped capturing their closure).

Not really sure why, this change should help the truffle inline much more code. But, I still have a lot of PE fixes incoming, so maybe this change won't affect anything until that's applied.

